PR #222, opened 2020-07-09
TripVC UX Improvements

### LINKED ISSUE #220: After reloading TripVC, selecting a TripStop in trip details does not work
## Steps
1. Open TripVC
2. Tap a stop in the trip details. The map selects that stop.
3. Tap reload on the top right toolbar
4. Tap a stop in the trip details. The map doesn't select the stop and outputs an error:
```
ERROR: Trying to select an annotation which has not been added
```

I suspect that because `mapView.updateAnnotations` does diff changes, it causes the new details to unsync from what the map is displaying.

ualch9:
- [UI] Dynamic Type
   - StopArrivalView will adjust its contents depending on the drawer's position to make better use of available space during large text sizes.
- [UX] There is a progress view indicating trip detail loading progress
   - The drawer will be unavailable until trip details have been loaded.
   - You won't notice this on a fast LTE connection, it is intended for poor/slow connections, such as in a tunnel or 2G speeds.
- [UX] Fixes #220
- [UX] Fixes map zooming out to the whole world on reload when TripStatus.coordinates is 0,0
- [UI] Show activity indicator during network operations in StopVC and TripVC (in place of reload button)
- [DEV] Adds `isAccessibility` property to UITraitEnvironment (equivalent to `self.traitCollection.preferredContentSizeCategory.isAccessibilityCategory`)

No voiceover in this PR. I'm researching how to better handle the drawer in voiceover.

### DIFF
--- a/OBAKitCore/Network/Operations/NetworkOperation.swift
+++ b/OBAKitCore/Network/Operations/NetworkOperation.swift
@@ -44,6 +44,9 @@ public class NetworkOperation: AsyncOperation, Requestable {
     public let request: URLRequest
     public private(set) var response: HTTPURLResponse?
     public private(set) var data: Data?
+
+    public let progress: Progress = Progress(totalUnitCount: 1)
+
     private var dataTask: URLSessionDataTask?
     private let dataLoader: URLDataLoader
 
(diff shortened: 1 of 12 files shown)


OBAKitCore/Network/Operations/NetworkOperation.swift
```diff
@@ -74,7 +77,7 @@ public class NetworkOperation: AsyncOperation, Requestable {
 
             self.finish()
         }
-
+        progress.addChild(task.progress, withPendingUnitCount: 1)
```
aaronbrethorst, line 80: 👍

PR #246, opened 2020-07-17
MapVC Dynamic Type

ualch9:
Most of MapVC is already dynamic type'd.

- [UI] Dynamic type no stops found message and StatusOverlayView
- [UI] Show activity indicator on toolbar in MapVC when loading operations takes more than a second
- [DEV] Add `UILabel.obaLabel` helper for making accessibility labels


aaronbrethorst:
cool! I'll review tonight or tomorrow. thanks!

PR #254, opened 2020-07-21
Adds a badge to bookmarked stops on the map

### LINKED ISSUE #175: Differentiate bookmarked stops from non-bookmarked stops on the map
It may make sense to follow the lead I introduced on the trip controller and use a small badge with a star icon.

aaronbrethorst:
Fixes #175 - Differentiate bookmarked stops from non-bookmarked stops on the map


barbeau:
Could you share a screenshot for bookmarked and non-bookmarked stops? Curious for a possible Android solution.

aaronbrethorst:
@barbeau sure!
Simulator Screen Shot - iPhone SE (2nd generation) - 2020-07-21 at 09 19 54

ualch9:
Does the annotation show the bookmark name instead of the actual stop name?

PR #266, opened 2020-07-26
Fixes #264

### LINKED ISSUE #264: Agency Alerts from More has bad performance
What I've encountered (both simulator and on device):
- Slow launch (CPU spikes to 100% for about 10 seconds, app frozen during that time)
- Slow scrolling (CPU frequently reaches high usage, also freezes)

Screen Shot 2020-07-24 at 1 01 20 AM

It's possible that the unusual number of services alerts is also contributing to poor performance.

ualch9:
Instruments reveals that UILabel typesetting is taking a long time.
Screen Shot 2020-07-25 at 10 00 31 PM

Turns out this KCM alert had an essay as its body, which is put verbatim into the view model. Trying to fit 7000+ unicode characters, then truncating it to fit into two lines turns out to be extremely expensive.
Screen Shot 2020-07-25 at 9 53 26 PM

Because agencies may provide whatever they want in the service alert, the solution is to truncate the string before applying it to the UILabel. The number I came up with was 128. There is about 53 characters in the first line on an iPhone X in standard content size, so I rounded it up to 64, then added another 64 for iPad. It's kinda arbitrary but it works. I'm not trying to be exact so UILabel can still add the ellipsis if it needs to be truncated further.

The 14 second operation now only takes a second<sub>ish</sub>.

### DIFF
--- a/OBAKit/Controls/ListKit/MessageSectionController.swift
+++ b/OBAKit/Controls/ListKit/MessageSectionController.swift
@@ -119,7 +119,7 @@ final class MessageCell: BaseSelfSizingTableCell {
     override init(frame: CGRect) {
         super.init(frame: frame)
 
-        self.topStack = UIStackView.horizontalStack(arrangedSubviews: [authorLabel, dateLabel])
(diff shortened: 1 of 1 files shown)


ualch9:
Fixes #264

PR #275, opened 2020-08-23
Service Alerts UX

### LINKED ISSUE #263: MessageCell hard to read in accessibility
When in a larger text size, the MessageCell decides to be 

s
i
n
g
l
e

c
o
l
u
m
n

w
i
d
e

\>
:
(

ualch9:
~WORK IN PROGRESS~
## UX
- Perform HTML generation in background queue in ServiceAlertVC
- Keep activity indicator visible when WebView is loading in ServiceAlertVC
     - Loading is about 40% HTML generation & 60% WebView rendering
- Clean up service alerts in StopVC (note service alert != agency alert)
     - Remove summary. Most of the time, the summary provides unhelpful information and makes the view more messy.
     - Remove date from service alert. Agencies appear to keep it up to date, so the time doesn't mean anything.
     - Remove "Service Alert". Replace with an icon that is also a read/unread indicator.
     - Reduce ambiguity by specifying agency name. For example, ST has an alert of just "Reduced Service", which may not be applicable to KCM.

| Before | After |
|------|------|
|||

<strike>
- Planned: Make agency alerts look like service alerts
     - Keep date
     - Read/unread for agency alerts
- Planned: ServiceAlertVC needs help
     - Dates are incorrect (issue pending)
     - Summary is not formatted correctly (issue pending)
</strike>

## DEV: Using delegates instead of closures
There is a new `ServiceAlertsSectionController`, which is a departure from our standard section controller implementation. It is different because it is responsible for multiple service alerts (rather than one controller per alert) and tapping action is handled by a delegate instead of closure. The delegate-behavior is a result of making the section controller serve >1 cell. @aaronbrethorst thoughts?

closes #263


OBAKit/Controls/ListKit/MessageSectionController.swift
```diff
@@ -52,48 +52,178 @@ final class MessageSectionData: ListViewModel, ListDiffable {
...
+        self.serviceAlerts = serviceAlertData
+        self.isCollapsed = isCollapsed
+    }
+
+    func diffIdentifier() -> NSObjectProtocol {
+        return "ServiceAlertsSectionIdentifier" as NSString
+    }
+
+    func isEqual(toDiffableObject object: ListDiffable?) -> Bool {
+        return false // TODO: me
```
aaronbrethorst, line 85: I assume this is going to be worked out

OBAKit/ServiceAlerts/ServiceAlertViewController.swift
```diff
@@ -12,8 +12,23 @@ import OBAKitCore
 import WebKit
 import SafariServices
 
-// swiftlint:disable cyclomatic_complexity
-
+// swiftlint:disable function_body_length
+
+// Page Content Generation Flow
+// Generating HTML may take a couple of seconds, so do it in the background.
```
aaronbrethorst, line 18: 👍

OBAKitCore/Extensions/Collections.swift
```diff
@@ -26,4 +26,34 @@ extension Sequence {
...
+    }
+}
+
+extension Array {
+    /// Performs a binary search.
+    /// - important: The array must be sorted.
+    /// - parameter sortedBy: Which key the array is sorted by.
+    /// - parameter element: The element to look for.
+    /// - returns: The element and its index, if found. `nil` if not found.
+    func binarySearch<T: Comparable>(sortedBy keyPath: KeyPath<Element, T>, element: T) -> (index: Int, element: Element)? {
```
aaronbrethorst, line 44: nothing comparable exists in the stdlib?

aaronbrethorst:
The new UI looks good! Delegates vs. closures: I think your perspective is sensible and I have no feedback here.

OBAKitCore/Extensions/Collections.swift
ualch9: since i'm using generics and keypath here, it'll probably be more straightforward like this

ualch9:
|Service Alerts | Agency Alerts |
|---|---|
|||
| Service alert header collapsed state is persisted | Agency alert header collapsed state is not persisted |

OBAKit/Stops/StopViewController.swift
```diff
@@ -404,15 +405,18 @@ public class StopViewController: UIViewController,
         sections.append(stopHeaderSection)
 
         // Service Alerts
-        let serviceAlerts = unreadServiceAlertsSection
-        sections.append(contentsOf: serviceAlerts)
+//        let serviceAlerts = unreadServiceAlertsSection
```
aaronbrethorst, line 408: delete me?

OBAKit/ServiceAlerts/ServiceAlertViewController.swift
```diff
@@ -73,80 +92,120 @@ final class ServiceAlertViewController: UIViewController, WKNavigationDelegate {
...
+                    b.append(.li, value: $0)
                 }
             }
         }
 
-        webView.setPageContent(builder.HTML)
+        let body = builder.HTML
+
+        // Insert dark mode css.
+        let document = "<html><head><style>body{background-color:#000;color:#fff}@media screen and (prefers-color-scheme:light){body{background-color:#fff;color:#000}}</style></head><body>\(body)</body></html>"
```
aaronbrethorst, line 206: love to see this turned into a multiline string—easier to read that way.

OBAKit/Stops/StopViewController.swift
```diff
@@ -691,7 +690,17 @@ public class StopViewController: UIViewController,
     }
 
     public func listAdapter(_ listAdapter: ListAdapter, sectionControllerFor object: Any) -> ListSectionController {
-        return defaultSectionController(for: object)
+        let controller = defaultSectionController(for: object)
+
+        // Set delegates
+        switch controller {
```
aaronbrethorst, line 696: Is there a reason you didn't express this as:

```
if let serviceAlertsController as? ServiceAlertsSectionController {
    serviceAlertsController.delegate = self
}
```

The switch statement feels like it obscures your intent.

aaronbrethorst (changes requested):
Couple minor nits. Otherwise looks good :+1:

OBAKit/Stops/StopViewController.swift
ualch9: i kinda forgot, but i think the intention was to future proof it in case we have more delegates to set in the future.

PR #278, opened 2020-09-06
StopVC diff and actions

ualch9:
StopVC:
- Proper IGListKit diff implementation for StopVC
- Show all actions within StopVC's cells (instead of a "MORE" button) - see screenshot
- Move StopVC cell actions to delegate to accommodate inline actions

Misc:
- Fixes highlight labels highlighting when they shouldn't be highlighting
- Fixes region picker duplicate tag assertion
- Fixes panel assertion when previewing TripVC
- Fixes tapping a stop on the map should reveal the floating panel in TripVC
- Fixes New York MTA missing `predictedArrival` and `predictedDeparture` values, causing decoding error
- Fixes New York MTA missing a vehicle status, causing a force unwrap


aaronbrethorst (approved):
👍

aaronbrethorst:
merged at #279 - thanks @ualch9!

PR #281, opened 2020-09-14
Change minimum supported version to iOS 13.7

aaronbrethorst:
I had a change of heart on supporting iOS 12. This is for two primary reasons:

1. I want to be able to use XCFrameworks (consume and publish).
2. I want to set us up to move off of IGListKit and onto the built-in diffable data sources feature at some point.

Furthermore, per Apple, 92% of devices released in the last four years are running iOS 13. https://developer.apple.com/support/app-store/

Thoughts, @ualch9?

### DIFF
--- a/Apps/Shared/app_shared.yml
+++ b/Apps/Shared/app_shared.yml
@@ -9,11 +9,11 @@ settings:
     CLANG_ANALYZER_LOCALIZABILITY_NONLOCALIZED: true
 
 options:
-  minimumXcodeGenVersion: 2.10
+  minimumXcodeGenVersion: 2.17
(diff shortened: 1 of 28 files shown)


ualch9:
13.7 seems aggressive, how many users are currently on 13.7 for old oba?

aaronbrethorst:
@ualch9 Fair. I'd be willing to go down to 13.6, but I expect that it can be bumped up by the time this app ships.

image

aaronbrethorst:
OK, I updated the minimum version to iOS 13.6. LMK if you have any reservations, @ualch9

ualch9:
👀 👍

PR #284, opened 2020-09-14
Include Carthage frameworks for Xcode previews

ualch9:
hotfix for missing carthage frameworks when xcode previewing App scheme.
I don't think its necessary to include this in RELEASE since everything builds and runs OK.


aaronbrethorst:
:+1

PR #309, opened 2020-10-06
Trip Controller Improvements

aaronbrethorst:


ualch9:
+1 on the compiler error fix!!

PR #313, opened 2020-10-25
Snap SwipeCellKit and include locally

aaronbrethorst:
I continue to get strange issues with building OBAKit when including SwipeCellKit as a Swift Package Manager dependency. I am finally at the point of just snapping the source code for the project, especially since it seems to have been abandoned, and puling in the code locally.


aaronbrethorst:
cc @ualch9 - thoughts?

ualch9:
:thumbsup

PR #328, opened 2020-11-24
Fixes Xcode automatic preview updating not working

ualch9:
Automatic preview updating doesn’t work when a script causes a major change, so now the “major change” only occurs on Install only.


aaronbrethorst (approved):
👍

PR #347, opened 2021-01-09
Port TodayView from old-OBA

ualch9:
Remove IGListKit from Today extension.
Old OBA uses stack views instead of collection view.


aaronbrethorst (approved):
clever! looks good.

PR #349, opened 2021-01-18
TripFloatingPanelController → OBAListView

ualch9:
- TripFloatingPanelController → OBAListView
- Services alerts section is collapsible, if there are any alerts
- In the header, if there is more than one alert, than the number of alerts is given as well (i.e. "Service Alerts (3)")
- Adds "show on map" and "walking directions" to context menu for a more thorough iPad experience
- Fixes improper map layout margins with floating panels on iPad
- Adds a "blink" animation to OBAListViewCell, to focus user attention to a specific item
- Adds Separated to OBAListViewCell, since OBAListViewCell primarily displays information in a list that should show separation

## Tap on stop to focus on list
ezgif com-gif-maker

note: i forgot to set the time for the tripstops in the above screenshots. this is fixed.


aaronbrethorst (approved):
nice work!

Issue #359, opened 2021-03-30
ServiceAlerts failing identity checks in OBAListView, causing crash


ualch9:
It appears this occurs when many different agencies have service alerts (such as Puget Sound, which has several agencies).
This has been occurring for some time now, and I'm starting to notice TestFlight crashes.

Repro: More → Alerts, using Puget Sound region

## Fix
As a catch-all, the `hash` method in `TransitAlertViewModel` should hash all values (as opposed to hashing only `id`). It will degrade animations, but that's better than a full-on crash.

## Log
```
2021-03-30 13:31:29.411525-0700 App[5250:103251] Enqueuing URL: https://onebusaway.co/api/v1/regions/1/alerts.pb?version=2&app_uid=51193C33-1197-4D4D-BC9E-46CA24A17505&key=org.onebusaway.iphone&app_ver=21.1.0
2021-03-30 13:31:29.597716-0700 App[5250:92277] [UIDiffableDataSource] Warning: 20 inserted identifier(s) already present; existing items will be moved into place for this current insertion. Please note this will impact performance if items are not unique when inserted.
2021-03-30 13:31:29.599841-0700 App[5250:92277] [UIDiffableDataSource] Warning: 20 inserted identifier(s) already present; existing items will be moved into place for this current insertion. Please note this will impact performance if items are not unique when inserted.
2021-03-30 13:31:29.654876-0700 App[5250:92277] *** Terminating app due to uncaught exception 'NSRangeException', reason: '*** -[NSMutableOrderedSet insertObjects:atIndexes:]: index 24 in index set beyond bounds [0 .. 22]'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff20421af6 __exceptionPreprocess + 242
	1   libobjc.A.dylib                     0x00007fff20177e78 objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff2049e77f _CFThrowFormattedException + 194
	3   CoreFoundation                      0x00007fff2033c953 -[NSMutableOrderedSet insertObjects:atIndexes:] + 738
	4   UIKitCore                           0x00007fff24b57d1e ___UIDiffableDataSourceApplyInsertUpdate_block_invoke + 643
	5   UIKitCore                           0x00007fff24b56da2 _UIDiffableDataSourceApplyInsertUpdate + 1103
	6   UIKitCore                           0x00007fff24127874 -[__UIDiffableDataSourceSnapshot _commitUpdateAtomic:] + 339
[...]

PR #383, opened 2021-06-11
Clarify identity protocol usage in OBAListView

ualch9:
I've been struggling to understand the true purpose of Hashable and Equatable, but I think I have it figured out now.
For the record, AFAIK `NSDiffableDataSource` only cares about `Hashable`. Hashable is supposed to be the object's "identity", but I have been misinterpreting this as "stable identity". As such, I believe this was the root cause of #359.

This is an overview of my findings, and the new roles in OBAListView:
- The `Hashable` protocol requires a `hash` method. This is needed for `NSDiffableDataSource`, which is the "identity" of the model. It should be a combination of *all values, including the identifier*.
- The `Identifiable` protocol requires an `id` property. It is currently not in use, reserved for future item identification functionality. This value is the "stable identity" (e.g. `stopID`) of the model.
- The `Equatable` protocol requires a `==` method to compare equality. It is currently not in use, reserved for future item diffing functionality. This should compare the values between both subjects, excluding its identifier.

## Changes
- Added documentation guidance on Hashable, Equatable, and Identifiable purposes in OBAListViewItem
- Updated view models to reflect new identity guidance

Related: https://developer.apple.com/forums/thread/126742

### DIFF
--- a/OBAKit/Bookmarks/BookmarksViewModel.swift
+++ b/OBAKit/Bookmarks/BookmarksViewModel.swift
@@ -29,6 +29,10 @@ struct BookmarkArrivalViewModel: OBAListViewItem {
     let regionIdentifier: Int
     let stopID: StopID
 
+    var id: String {
+        return "bookmark=\(bookmarkID),region=\(regionIdentifier),stopID=\(stopID)"
+    }
+
     let isFavorite: Bool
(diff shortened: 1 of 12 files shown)


aaronbrethorst (approved):
👍

PR #387, opened 2021-06-11
iOS 14.5

ualch9:
Bump CocoaLumberjack and Eureka's minor version.
Update SwiftProtobuf to next major version.

### DIFF
--- a/Apps/Shared/app_shared.yml
+++ b/Apps/Shared/app_shared.yml
@@ -14,7 +14,7 @@ options:
   generateEmptyDirectories: true
   groupSortPosition: top
   deploymentTarget:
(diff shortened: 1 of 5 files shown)


ualch9:
@aaronbrethorst 

Build Warning: `OBAKitCore/Location/Location/LocationManagerProtocol.swift:50:34`: 'authorizationStatus()' was deprecated in iOS 14.0

Fixing this deprecation causes `LocationServiceTests.test_authorization_granted` to fail. Please take a look when you get the chance.

aaronbrethorst (approved):
I don't think the test failure is a big deal. I think it'll just require a little CI fiddling. I'll take care of that.

Issue #394, opened 2021-06-21
Retain cycle of unknown cause


ualch9:
Discovered while working on #391, reproducible on main branch 45d2fe6. Unsure when this was bug introduced.
Needs additional investigation, reference cycles appears to be closures associated with OBAKitCore.DecodableOperation, but I'm wondering if it's related to OBAListViewItem closures.

Screenshot attached.

![image](https://user-images.githubusercontent.com/22162410/122812074-141fa180-d286-11eb-94db-4091bcafefd9.png)

aaronbrethorst:
I don't think this blocks release.

ualch9:
agreed.

— alan
________________________________

PR #395, opened 2021-06-22
Fixes view controller retain cycles

ualch9:
Fixes view controller related retain cycles, but does not fix OBAKitCore models' cycles.
Partially `fixes` #394.


aaronbrethorst (approved):
nice work :+1:

PR #391, opened 2021-06-15
OBAListView iOS 14 list

ualch9:
## Changelog
- Replaces custom recreation of UITableView cells and supplementary views in favor of UIListContentConfiguration -- #377
- Removes SwipeCellKit in favor of UIListContentConfiguration.
- Uses `insetGrouped` visual style for most lists (not bookmarks, bookmark cells look too space-constrained in the new style).
- OBAListViewCollapsibleSectionsDelegate is broken now -- #392

## Shiny new visual look
| ooooo | ahhhhh |
| -- | -- |
| IMG_3448 |  IMG_3447 |

## References
Fixes #377

Opened:
#392
#393
#398 
#399

PR #397, opened 2021-06-23
Fixes CI

aaronbrethorst:


aaronbrethorst:
FYI @ualch9

ualch9:
@aaronbrethorst im using https://developer.apple.com/documentation/uikit/uicollectionlayoutlistconfiguration/3727738-separatorconfiguration in #391. Not a logic feature, only for visuals. I believe this is the only 14.5-specific feature I'm currently using, so I can remove it from 14.4 if necessary?

aaronbrethorst:
@ualch9 That sounds like a good approach. Hopefully, GitHub will add support for Xcode 12.5 soon and this'll all be moot? 🤞

PR #414, opened 2021-07-11
More tweaks

### LINKED ISSUE #356: WalkTimeView is misaligned on iPad
The triangle should be directly below the Person icon, pointing to the `20m`.
Screenshot:
IMG_625F21BA803F-1

ualch9:
- Fixes #356 
- Removes accessory view and use plain list style for TripFloatingPanelController
- [Show Stop On Map] action moves the panel out of the way so you can see the stop on the map
- Some M1 stuff


aaronbrethorst (approved):
nice work, looks great! 👍

Issue #421, opened 2021-07-13
Agency Alerts: OBAListView DiffableDataSource can't find item index of headers


ualch9:
Additionally, DiffableDataSource is flooding stdout with item index errors of header items (but not crashing):
```
2021-06-24 09:58:09.850758-0700 App[13091:4345657] [DiffableDataSource] Failed to find index of item OBAKit.AnyOBAListViewItem(_anyEquatable: OBAKit.(unknown context at $1080a4320).AnyEquatable(value: OBAKit.OBAListViewHeader(id: "agency_alerts_Washington State Ferries", title: Optional("Washington State Ferries"), isCollapsible: true), isEqual: (Function)), _id: (Function), _configuration: (Function), _separatorConfiguration: (Function), _onSelectAction: (Function), _onDeleteAction: (Function), _leadingContextualActions: (Function), _trailingContextualActions: (Function), _type: AnyHashable(OBAKit.OBAListViewHeader(id: "agency_alerts_Washington State Ferries", title: Optional("Washington State Ferries"), isCollapsible: true)))
```

_Originally posted by @ualch9 in https://github.com/OneBusAway/OBAKit/issues/389#issuecomment-867803156_

![image](https://user-images.githubusercontent.com/2254/126917770-9e06b8ad-ece7-4cf1-990c-1386995fc306.png)

aaronbrethorst:
I'm moving this to 'post launch'. I don't think we need to gate release on fixing this.

aaronbrethorst:
This is breaking the bookmarks tab for users who have a lot of bookmarks.

ualch9:
@aaronbrethorst any screenshots?

aaronbrethorst:
@ualch9 emailed you

ualch9:
thanks, i see it. i’ll get on it.

— alan
________________________________

ualch9:
@aaronbrethorst is this causing a crash? or is it just messing up the view?

aaronbrethorst:
@ualch9 it’s causing the app to lock up when it’s opened to the bookmarks tab

aaronbrethorst:
Any luck so far, @ualch9?

ualch9:
sorry, been busy. i took a look and reproduced the problem, but i can’t think of whats wrong off the bat.

could be something wrong with too many bookmarks causing the the list to frequently update.

i’ll have time sometime next week if this can wait, otherwise feel free to take a look.

— alan
________________________________

[...]

PR #424, opened 2021-07-14
Treat header item and contents as equals if header is non-collapsing

ualch9:
This was an attempt to `fix` #421.
It is a partial fix because it appears to have fixed this for everywhere except BookmarksViewController, which confuses me but ok.


aaronbrethorst (approved):
lgtm :shipit:

PR #439, opened 2021-08-05
Updates some text related to data migrations

aaronbrethorst:
Fixes https://github.com/OneBusAway/OBAKit/issues/433 - Remind non-migrated users that bookmarks may already exist
Fixes https://github.com/OneBusAway/OBAKit/issues/432 - Add "needs internet" message to migration bulletin

Also extracts a few new strings that need to be localized.


aaronbrethorst:
Tests pass locally.

PR #482, opened 2021-08-14
Hides departure time badge when nil configuration applied

### LINKED ISSUE #469: Little funky green dots appearing on right side of stop-level bookmarks
The big black box is obscuring an OBA user's information. What I want to call your attention to are the little green dots on the right side of some but not all of the stop-level bookmark cells. 

Untitled-1

ualch9:
I'm only able to randomly reproduce #469, so I think this might be a fix for it.

Fixes #469


aaronbrethorst:
Unfortunately, I'm still seeing the problem occurring after applying this change, but I was able to get it fixed. PR up momentarily.

ualch9:
OK. feel free to close this PR if unnecessary.

— alan
________________________________

aaronbrethorst:
Thanks for taking a stab at fixing this!

Issue #398, opened 2021-06-23
List view separators always appears on iOS 14.4


ualch9:
CI downgraded to 14.4 for compatibility, #397.

> @aaronbrethorst im using https://developer.apple.com/documentation/uikit/uicollectionlayoutlistconfiguration/3727738-separatorconfiguration in #391. Not a logic feature, only for visuals. I believe this is the only 14.5-specific feature I'm currently using, so I can remove it from 14.4 if necessary? -- https://github.com/OneBusAway/OBAKit/pull/397#issuecomment-866493581

PR #486, opened 2021-08-14
Update CI to use macOS 11/Xcode 12.5

aaronbrethorst:
Fixes https://github.com/OneBusAway/onebusaway-ios/issues/398 - List view separators always appears on iOS 14.4

This is awaiting us being granted access to GitHub's macOS 11 image on their CI system. I submitted a request on Aug 14, and it looks like they're processing requests every couple of business days.

### DIFF
--- a/.github/workflows/obakittests.yml
+++ b/.github/workflows/obakittests.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
-    runs-on: macos-latest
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
fyi @ualch9 - we've been approved to use 14.5 in CI, so we can drop support for 14.4 🎉

PR #500, opened 2021-08-24
Adds separators between buttons in hover bar

### LINKED ISSUE #471: Improve Homepage FAB Accessibility
Issue: 
Not clear whether it’s one touch target or two 
Contrast Ratio too low between icon/number and semi-transparent surface in some cases.

Recommendation: 
Add a gray separation line between the two touch targets (Reference: Apple Map)
Increase Contrast Ratio by matching FAB background color to the botton nav bar surface color with no transparency applied.

image

ualch9:
- Changes hoverbar's background to a thick material for better visual contrast.
- Fixes #471


aaronbrethorst (approved):
looks good! I have a small tweak to suggest for the separator height, which I will land separately from this PR.

PR #507, opened 2021-09-01
Adjust isAccessibility to be anything equal to or larger than extraLarge

### LINKED ISSUE #506: Departure times are not adjusting with Dynamic Type
> The estimated time of arrival on the new OneBusA app is green on white and much smaller text size so makes it harder for users with astigmatism and farsightedness like me to read.
> 
> I’ve included another app screenshot which uses black on white and slightly bolder text so easy to read.

image1

image2

ualch9:
Fixes #506


aaronbrethorst (approved):
👍

PR #509, opened 2021-09-04
Add base map type toggle to MapViewController and TripViewController

### LINKED ISSUE #493: Let users switch between different map modes
Classic OBA let users switch between the default vector-based map style and satellite. We should add that back.

ualch9:
Fixes #493 
image


aaronbrethorst (approved):
nice work, @ualch9. I think this is a good approach for toggling between satellite and vector.

PR #510, opened 2021-09-04
Improve standard visibility of StopArrivalView's minutes

### LINKED ISSUE #506: Departure times are not adjusting with Dynamic Type

### LINKED ISSUE #508: Stop departure times are harder to read than they used to be
User feedback:

> My phone seem to have just updated to a newer version about two weeks ago. Unfortunately the numbers showing the minutes until the next bus are now much smaller and a thinner font that is very hard to read in the daylight. If there is any way to change it back to a bold and/or larger font it would be super helpful.  Thanks for making a great app

ualch9:
"Standard" means regardless of iOS accessibility settings.
Fixes #508 and #506

| Before (no `.headline` font) | After (yes `.headline` font) |
| -- | -- |
|  |  |

### DIFF
--- a/OBAKitCore/Theme/Theme.swift
+++ b/OBAKitCore/Theme/Theme.swift
@@ -130,10 +130,18 @@ public class ThemeColors: NSObject {
         departureEarly = .systemRed
         departureEarlyBackground = .systemRed
 
-        // Hex #129900 -- UIColor.systemGreen is too bright for small text, which StopViewController uses.
+        // Hex #129900 is better visibility for small text in light mode.
+        // UIColor.systemGreen is better visibility for small text in dark mode.
         // See #506 and #508 for user feedback.
-        departureOnTime = UIColor(red: 0.07, green: 0.60, blue: 0.00, alpha: 1.00)
-        departureOnTimeBackground = UIColor(red: 0.07, green: 0.60, blue: 0.00, alpha: 1.00)
+        let departureOnTimeColor = UIColor { traitCollection in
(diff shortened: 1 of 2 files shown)


OBAKitCore/Theme/Theme.swift
```diff
@@ -130,8 +130,10 @@ public class ThemeColors: NSObject {
         departureEarly = .systemRed
         departureEarlyBackground = .systemRed
 
-        departureOnTime = .systemGreen
-        departureOnTimeBackground = .systemGreen
+        // Hex #129900 -- UIColor.systemGreen is too bright for small text, which StopViewController uses.
+        // See #506 and #508 for user feedback.
+        departureOnTime = UIColor(red: 0.07, green: 0.60, blue: 0.00, alpha: 1.00)
```
aaronbrethorst, line 135: How’s this look in dark mode?

aaronbrethorst (commented):
One Q for you before merging.

OBAKitCore/Theme/Theme.swift
ualch9: not good 👎

ualch9:
| Light mode | Dark mode |
| -- | -- |
| light mode | dark mode |

aaronbrethorst (approved):
🎉

PR #512, opened 2021-09-04
Consolidate LoadMore button in StopViewController

### LINKED ISSUE #475: Improve "Load More" Button
**Issue:** 

“Load More” button is large and disconnected from the bus schedule surface

Date and time below “Load More” are confusing to users (from usability test)

Recommendation: 
Consolidate “Load More” into the same surface as the bus schedule, and add down icon for better affordance

~Get rid of date and time associated with “Load More” to avoid user confusion~

Aaron notes: The date/time range was added in a few years back specifically in response to user confusion about routes that run infrequently. Perhaps we can find a way to improve the way it is displayed, but we should not remove it.

image

ualch9:
Fixes #475

Moves data load time to the data attribution footer (iOS "Increase Contrast" improves the visibility of data data range text).

| Sort by Time | Sort by Route |
| -- | -- |
| sort by time | sort by route |

| Sort by Time with Error | Sort by Route with Error |
| -- | -- | 
| sort by time with error | sort by error with error |

### DIFF
--- a/OBAKit/Controls/ListView/OBAListView.swift
+++ b/OBAKit/Controls/ListView/OBAListView.swift
@@ -365,7 +365,9 @@ public class OBAListView: UICollectionView, UICollectionViewDelegate {
     }
 
     public func scrollTo(section: OBAListViewSection, at position: UICollectionView.ScrollPosition, animated: Bool) {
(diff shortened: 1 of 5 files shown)


aaronbrethorst (approved):
👍

PR #528, opened 2021-09-19
Skip map annotation callout when voiceover is enabled

### LINKED ISSUE #526: Voiceover unable to select stop from Map
See attached.

image

-- image from direct user feedback

ualch9:
Fixes #526


aaronbrethorst (approved):
great work 👍

PR #536, opened 2021-09-28
Various tweaks & fixes

### LINKED ISSUE #505: `StopViewController` not deallocating, causing random haptic feedback
From a user:

> When I have the app open, the haptic feedback seems to just be making random ticks. More so when moving than when still. Other than that, I can’t determine what’s causing it, but it’s sort of annoying.

ualch9:
- Use static tab bar appearance in MapViewController since the dynamically changing background is interfering with floating panel's scrolling (iOS 15).
- Fix #505 & Undo #518
- Adds a setting toggle for data load haptic


aaronbrethorst (approved):
lgtm 👍

PR #541, opened 2021-10-04
Fixes various retain cycles

### LINKED ISSUE #394: Retain cycle of unknown cause

ualch9:
Fixes #394


aaronbrethorst (approved):
looks good, nice work. I'll get this into TestFlight later today.

Issue #557, opened 2021-10-13
Editing a bookmark prompts "Duplicate bookmark"


ualch9:
See attached video:

https://user-images.githubusercontent.com/22162410/137218446-b4ecb7e7-cd2c-4a56-9271-cff69e562931.MP4

cathy810218:
Hi Alan! 
If we remove this duplicate bookmark check, then it would not prompt the alert nor duplicate the bookmark.
Though I am not sure if this was intended to be added as an option for duplicating the bookmark.

https://github.com/OneBusAway/onebusaway-ios/blob/9f576f31fe8a273f5418e3c8708e1e1755ee5e42/OBAKit/Bookmarks/EditBookmarkViewController.swift#L219-L228

ualch9:
@cathy810218 yes, that makes sense, i dont believe the intention was to give the user an option to duplicate bookmark. when editing the bookmark, we should not be checking if it is a duplicate. 

if you'd like to tackle this, i can assign you to this issue.

cathy810218:
Okay! I will submit a PR! Thank you.

PR #565, opened 2021-11-12
ISSUES-557: Editing a bookmark prompts "Duplicate bookmark" on Save

cathy810218:
Issues: https://github.com/OneBusAway/onebusaway-ios/issues/557

Description:
Remove the duplicated bookmark check so it does not prompt the alert asking if the intention is to duplicate bookmark. If the user did not change anything in the bookmark, on Save, it should just save the unedited bookmark.


cathy810218:
Addressed the issue in [PR#566](https://github.com/OneBusAway/onebusaway-ios/pull/566) with the correct email address for signing CLA.

Issue #480, opened 2021-08-13
Prevent Bookmark Duplication


roverman:
**Issue:** 

Stops can be bookmarked multiple times without system warning.

**Recommendation:** 

Show an alert and prevent the user from bookmarking the route or stop again in order to prevent duplicates.

Warning alert text should be one of two strings depending on whether it's a route or a stop:

`You have already bookmarked this route.`, or `You have already bookmarked this stop.`

![image](https://user-images.githubusercontent.com/68605140/129315331-395a4651-8fba-4fbe-847d-62decf69a65d.png)

PR #566, opened 2021-11-12
ISSUES-557: Editing a bookmark prompts "Duplicate bookmark" on Save

cathy810218:
Issue:
https://github.com/OneBusAway/onebusaway-ios/issues/557

Description:
Remove the duplicated bookmark check so it does not prompt the alert asking if the intention is to duplicate bookmark. If the user did not change anything in the bookmark, on Save, it should just save the unedited bookmark.

### DIFF
--- a/OBAKit/Bookmarks/EditBookmarkViewController.swift
+++ b/OBAKit/Bookmarks/EditBookmarkViewController.swift
@@ -216,7 +216,8 @@ class EditBookmarkViewController: FormViewController, AddGroupAlertDelegate {
         let faveVal = form.values()[showInTodayViewTag]
         bookmark.isFavorite = faveVal as! Bool // swiftlint:disable:this force_cast
 
-        if application.userDataStore.checkForDuplicates(bookmark: bookmark) {
+        // If user is creating a bookmark, we need to check if this is a dup
+        if addMode && application.userDataStore.checkForDuplicates(bookmark: bookmark) {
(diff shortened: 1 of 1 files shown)


ualch9:
i believe editbookmarkviewcontroller is also used when creating a new bookmark, so we'd still need to check for dupes (#480). this pr would reverse d889a08, which originally fixed #480. any insights or feedback, @aaronbrethorst?

cathy810218:
Ah I see! That makes sense!
I guess I can then add this in the `if` check to see if the user is adding a new bookmark / editing it.
Will wait for Aaron's feedback :)

https://github.com/OneBusAway/onebusaway-ios/blob/master/OBAKit/Bookmarks/EditBookmarkViewController.swift#L193

aaronbrethorst:
Hey @cathy810218 - good to see you on here! Check to see if `addMode == true` and skip the warning if it is.

cathy810218:
Hi @aaronbrethorst haha! yes it's been awhile and I am trying to re-learn ios development.
Will update the PR with the change suggested above! Thanks!

cathy810218:
https://user-images.githubusercontent.com/12047881/141601651-95535b81-1693-43a6-a8e2-9e34b3bd8ec3.mp4

ualch9 (approved):
👍

ualch9:
thanks, @cathy810218!

PR #575, opened 2021-11-22
Default to full sheet on voiceover

ualch9:
- Add preferredLoadDataRegionFudgeFactor to MapRegionManager
- Add alert informing voiceover users of the change
- Defaults floatingPanel ("sheet") to full position in MapViewController and TripViewController.

I'm using "sheet" because of `UISheetPresentationController`.

Fixes #431
Fixes #572

PR #576, opened 2021-11-30
Improve map sheet UX for VoiceOver users

ualch9:
Oops, didn't mean to merge #575 yet, shouldve marked it as a draft.

Beta feedback on #575, summarized:
When the sheet is fully expanded, the prompt is confusing for Voiceover users, since the map is now inaccessible for Voiceover (see screenshot). Additionally, the map should not be disabled for all voiceover users, since low-vision users are still able to interact with the map visually.

Stuff to do:
- [x] When the sheet is fully expanded, disable mapview interactions. When the sheet is partially collapsed, enable mapview.
- [x] When the user centers the map on their location, partially collapse the sheet to enable mapview interaction.
- ~[ ] Add a button to recent stops list to "expand" the search radius if no stops were loaded (via `perferredLoadDataRegionFudgeFactor` from commit ae73f03)~ removed from this PR, see #578 

image

### DIFF
--- a/OBAKit/Mapping/MapRegionManager.swift
+++ b/OBAKit/Mapping/MapRegionManager.swift
@@ -76,10 +76,6 @@ public class MapRegionManager: NSObject,
         mapView.showsUserLocation = true
         mapView.isRotateEnabled = false
 
-        // Disables voiceover interacting with map elements (such as streets and POIs).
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
LGTM 👍

PR #618, opened 2022-12-24
Disallow the app to “Always” get user location

### LINKED ISSUE #616: location arrow disappears when location access set to "always"
If the permission for location access is set to "While Using the App", the map shows an arrow that if clicked centers your current location, as in this screenshot:

when-using

But if you change the permission to "Always Allow" the arrow disappears:
always-allow

This doesn't seem right.

ualch9:
We don't want the user's location if they don't have the app open.

Fixes #616


ualch9:
hm, CI is on holiday break?

PR #620, opened 2022-12-29
Speculative fix for “DUPLICATE_ITEM_IDENTIFIERS” and “EQUAL_IDENTIFIERS_HAVE_DIFFERENT_HASH_VALUES” crasher

ualch9:
View models now has a guaranteed unique identifier (by UUID during initialization). This is fine for now, since we are directly consuming API results, but at the cost of potentially causing the entire table view to update when one item changes.

Ultimately, `OBAListView` implementation is flawed and requires a bit of fixing, so this is a stop gap.

### DIFF
--- a/OBAKit/Controls/ListView/Helpers/OBA/AgencyAlertListViewConverters.swift
+++ b/OBAKit/Controls/ListView/Helpers/OBA/AgencyAlertListViewConverters.swift
@@ -33,7 +33,7 @@ extension AgencyAlertListViewConverters where Self: UIViewController {
                 return TransitAlertDataListViewModel(alert, isUnread: isUnread, forLocale: Locale.current, onSelectAction: presentAlertAction)
             }
 
-            let alerts = viewModels.uniqued.sorted(by: \.title) // remove duplicates
+            let alerts = viewModels.sorted(by: \.title) // remove duplicates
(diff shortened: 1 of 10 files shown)


OBAKit/Controls/ListView/Helpers/OBA/AgencyAlertListViewConverters.swift
```diff
@@ -33,7 +33,7 @@ extension AgencyAlertListViewConverters where Self: UIViewController {
                 return TransitAlertDataListViewModel(alert, isUnread: isUnread, forLocale: Locale.current, onSelectAction: presentAlertAction)
             }
 
-            let alerts = viewModels.uniqued.sorted(by: \.title) // remove duplicates
+            let alerts = viewModels.sorted(by: \.title) // remove duplicates
```
aaronbrethorst, line 36: 👍

aaronbrethorst:
looks like a good step forward. nice work 👍

ualch9:
ah-- unit tests succeeded, but CI failing on the "Update xcresult bundle" step. Looks like forks can't access CI results here, for security. I'll move work onto this repo instead, but we should find a way to mark that step as optional for other contributors.

PR #622, opened 2022-12-29
Replace RESTAPIService Operations with Swift Concurrency implementation

ualch9:
The goal is to make the codebase more approachable to first-time contributors by adopting Swift Structured Concurrency. The result from this refactor will be more readable business logic and a net-decrease in LOCs.

# OBAKit API change
## Before
```
// Keep track of operation
var operation: DecodableOperation<RESTAPIResponse<StopArrivals>>?

deinit {
    operation?.cancel()
}

func loadData() {
    let op = apiService.getArrivalsAndDeparturesForStop(
        id: stopID, 
        minutesBefore: minutesBefore, 
        minutesAfter: minutesAfter
    )

    op.complete { [weak self] result in
        switch (result) {
        case .failure(let error):
            // Show error in UI
        case .success(let response):
            // Show arrivals in UI
    }
    self.operation = op
}
```

## After
```
func loadData() async {
    do {
        let stopArrivals: RESTAPIResponse<StopArrivals> = try await apiService
            .getArrivalsAndDeparturesForStop(
                id: stopID, 
                minutesBefore: minutesBefore, 
                minutesAfter: minutesAfter
            )
        // Handle success
    } catch {
        // Handle failure
    }
}
```

---

This is part-one of Concurrency adoption. Lots of bandaids and deprecation warnings are present, but I want to keep PRs reasonably sized.

- Deprecate `RESTAPIService` (renamed to `_RESTAPIService`), which uses Operations and replaces it with `RESTAPIService`, which uses Swift Structured Concurrency.


OBAKitCore/Network/RESTAPIService/RESTAPIService+GetAgencyAlerts.swift
```diff
@@ -0,0 +1,42 @@
+//
+//  RESTAPIService+GetAgencyAlerts.swift
+//  OBAKitCore
+//
+//  Created by Alan Chu on 12/29/22.
+//
+
+extension RESTAPIService {
+    // MARK: - Regional Alerts
+    public nonisolated func getAlerts(agencies: [AgencyWithCoverage]) async throws -> [AgencyAlert] {
```
ualch9, line 10: This used to be in `MultiAgencyAlertsOperation.swift`, which was mostly boilerplate code for Operation.

OBAKitTests/Helpers/OBATestCase.swift
```diff
@@ -90,10 +92,16 @@ open class OBATestCase: XCTestCase {
 
     var baseURL: URL { URL(string: "https://\(host)")! }
 
-    var restService: RESTAPIService!
+    var restService: _RESTAPIService!
```
ualch9, line 95: bandaid

OBAKit/Orchestration/Application.swift
```diff
@@ -110,7 +110,9 @@ public class Application: CoreApplication, PushServiceDelegate {
 
                 switch result {
                 case .failure(let error):
-                    self.displayError(error)
+                    Task { @MainActor in
```
ualch9, line 113: bandaid

OBAKitCore/Network/Operations/NetworkOperation.swift
```diff
@@ -20,6 +20,9 @@ public enum APIError: Error, LocalizedError {
     case noResponseBody
     case requestFailure(HTTPURLResponse)
 
+    /// A `404` error.
+    case requestNotFound(HTTPURLResponse)
```
ualch9, line 24: I can't think of a reason why we'd differentiate a `real 404` and an `effective 404`, so I'm treating it as equals here. This removes the  `statusCodeIsEffectively404` flag.

aaronbrethorst: You might be able to find the reason I added this if you look at the git blame for this commit. There's some OBA REST API that returns a nonsensical 200 response to some request that should really be a 404. Perhaps the assumptions in the rest of the system have changed such that a misleading 200 response will no longer cause issues, but that needs to be run down before this gets removed.

OBAKitCore/DataMigration/DataMigrator_.swift
```diff
@@ -0,0 +1,276 @@
+//
```
ualch9, line 1: self: remove this file

OBAKitCore/Network/RESTAPIService/RESTAPIService+Processing.swift
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        // The REST API doesn't do a good job of surfacing what should be 404 errors.
+        // If you request a valid endpoint, but provide it with a bogus piece of
+        // data (e.g. a non-existent Stop ID), it should return a 404 error to you.
+        // Instead, it gives a 200 and a blank body.
+        if httpResponse.expectedContentLength == 0 && httpResponse.statusCode == 200 {
+            logger.error("Failed network request for \(url, privacy: .public): 404 not found.")
+            throw APIError.requestNotFound(httpResponse)
+        }
```
ualch9, line 51: Continuing previous discussion: I kept the handling of `empty 200` around. The difference now is both `statusCode == 404` or `response.isEmpty AND statusCode == 200` returns `APIError.requestNotFound`

OBAKitCore/Network/Operations/NetworkOperation.swift
ualch9: Oops, I wasn't clear. I didn't remove the handling of empty 200. See https://github.com/OneBusAway/onebusaway-ios/pull/622/files#r1060340430

PR #647, opened 2023-02-21
Upgrade Eureka and FloatingPanel

ualch9:


aaronbrethorst:
@ualch9 does this need a review?

ualch9:
I don't think it necessarily requires a review, most of the diff is boilerplate API changes with FloatingPanel 2.x.

PR #648, opened 2023-02-21
StopViewController SwiftUI

ualch9:
Started as an experiment, but might actually build it out now. Plus, the code can be reused for Live Activities/watchOS/etc.


ualch9:
Demo
- Loading arrdeps while maintaining scroll position
- Flashing of updated `minutes` label. The main view is not concerned with displaying the *flash*, each individual label view tracks its own contents.

https://user-images.githubusercontent.com/22162410/220456469-bc36cff9-07b9-42b9-8ae0-a13654876d71.mp4

---

Each arrdep uses some set of fields as its identifier. I've tried to account for it, but if there are duplicate identifiers, SwiftUI will handle it gracefully. Previously, DiffableDataSource throw an exception and crash.

The screenshot below is showing a workaround for SwiftUI's unideal behavior for list selection. In iOS 15, creating a selectable list required displaying a checkmark in the cell, which would look unusual. The correct usage was to use a `NavigationLink` in each cell, but that doesn't interoperate well with UIKit (which is required for TripViewController). So, I manually created selection behavior into the cell, and will use a delegate to communicate to the UINavigationController to push a TripViewController. 

simulator_screenshot_82653627-46AF-4BA1-B384-FD43D0A0BA74

aaronbrethorst:
Superseded by #1182 thanks @ualch9! 🫡

PR #656, opened 2023-03-24
needsToOnboard checks shouldPerformMigration instead of hasDataToMigrate

ualch9:
User reports that the onboarding screen is appearing every time during launch.

I think this is caused by `needsToOnboard` checking the wrong state, instead of checking only the existence of classic-OBA data, we should be checking the existence AND the UserDefault value of whether the user has already upgraded their data.


aaronbrethorst (approved):
Good call @ualch9 I think you're right, and this looks good.

Issue #660, opened 2023-03-31
OBAListView is stuck in its update/layout loop, when scrolling map


ualch9:
This is a reopen of #540. I thought #652 fixed #540, but it wasn't it.

ualch9:
This appears to be fixed by #661, as crash reports for this have decreased 🤞

aaronbrethorst:
👏 👏 👏

PR #661, opened 2023-04-05
Split contents of MapFloatingPanelController into separate view controllers

ualch9:
I suspect the layout system in OBAListView is partially responsible for #660. As a hotfix, I am replacing the Recent Stops list view with a standard UICollectionView. The current search view controller was more complicated to replace, so that will stay as an OBAListView for now. To facilitate the UICollectionView and OBAListView, the two states are now separate view controllers, and MapFloatingPanelController will facilitate the view controller's lifecycles. 

MapFloatingPanelController is now spaghetti, but I want to collect crash data on whether the problem is really with OBAListView's layout before cleaning up the logic.

Also, some opportunistic fixes too.


aaronbrethorst (approved):
Looks good to me and I think your plan is smart and sensible.

PR #670, opened 2023-05-21
Reload TripConvertible from remote in TripViewController

### LINKED ISSUE #668: Time estimate does not update when in detailed route view
If you go to the detailed view of a particular route (select a bus & then select the next available time, the view that shows all the stops and real-time tracking of the bus). The time estimate, circled in red, does not seem to update if you stay on that screen even if refresh is clicked. Going back and then returning to that screen updates it. 

This is on the iOS app, 23.2.3, iOS 16.4.1. 

IMG_6942

ualch9:
We are not updating the ArrivalDeparture when reloading TripViewController. ArrivalDeparture show information to a specific stop, compared to TripDetails, which is details on the entire trip.

Fixes #668

### DIFF
--- a/OBAKit/Trip/TripFloatingPanelController.swift
+++ b/OBAKit/Trip/TripFloatingPanelController.swift
@@ -30,7 +30,7 @@ class TripFloatingPanelController: UIViewController,
 
     var tripConvertible: TripConvertible? {
         didSet {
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
Nice work. Looks good.

PR #684, opened 2023-10-26
experimental Codable macros + GRDB

ualch9:
- Added new OBAKitCoreTests target for testing against `OBAKitCore.framework` only.
    - OBAKitTests depends on `OBAKit.framework` and `App.app`.
    - OBAKitCoreTests links against only `OBAKitCore.framework`, so I can focus on Codable/GRDB without fighting the compiler or working without tests. Where appropriate, I moved some tests were moved from OBAKitTests → OBAKitCoreTests.

- GRDB and REST implementations are contained in their own files.
    - REST models are in `OBAKitCore/Models/REST`. 
    - GRDB implementations are Swift extensions in `OBAKitCore/Persistence/Models`. In general, GRDB models (such as many-to-many relations) are not exposed outside of OBAKitCore.
    - With few exceptions, REST and GRDB use the same model.

Each model defines their own tables, via `protocol DatabaseTableCreator.createTable(in:)` that is called by `PersistenceManager` when initially creating the database.

When storing on-disk, each database is stored per-region: `{app_container}/Application Support/regions/{region_id}/onebusaway.sqlite`

Screenshot 2023-10-26 at 12 12 26 PM

Undecided:
- What should be persisted (stops only? or everything?). As a first release, the database could operate only in-memory.
- Need to include some sort of expiration date. AFAIK, API responses do not include an `expires` header.
- How should bookmarks be stored so they can sync via iCloud.
- Objective-C support? Is this important enough that we should write wrappers for the new models?


aaronbrethorst:
hey @ualch9 - welcome back! Tag me on Slack when you're ready for me to look this over.

ualch9:
Continued in #686 .

PR #688, opened 2023-11-11
Drop support for iOS 15

aaronbrethorst:
* Make iOS 16.0 the minimum required version
* Bump app version to 23.3
* Fix a few deprecations from upgrading


aaronbrethorst:
@ualch9 fyi. iOS 16 and 17 represent over 90% of our user base. time to upgrade!

ualch9:
i like

PR #692, opened 2023-11-19
Disable debug mode in more view

### LINKED ISSUE #613: Disable debug mode?
When clicking on the "header" image in "More..." section, debug mode is enabled. Clicking again, debug mode is disabled.
For most users this might be confusing and disencouraging - debug mode should be rather turned on in Settings section of "more", as a button.
Simulator Screen Shot - iPhone 14 Pro - 2022-10-13 at 10 43 45

Castaneda-Angel:
Fixes #613

Previously, tapping on the header image in More would toggle Debug Mode. Now, Debug Mode is toggled by a switch in Settings.

- Toggling the `Debug Mode` setting will show/hide the other settings tied to `Debug Mode` (as shown in the images).
- If `Always Refresh` is toggled on, and then you toggle off `Debug Mode`, `Always Refresh` will follow suit and also toggle off.
- `Migrate data` is either always shown if there is data to migrate, or it is hidden/shown by the `Debug Mode` status.
- Tapping on `Push ID` now displays "Copied to clipboard" message where the push ID is shown, as a way to communicate to users that something occurred after tapping.
- All new copy added should be localized.


aaronbrethorst (approved):
looks great! thanks for tackling this!

PR #701, opened 2023-12-15
Service alert features

### LINKED ISSUE #700: Show an Alert Controller when tapping on a service alert
Compare the iOS and Android screenshots of the service alert below. In the iOS case, you only get the collection view cell text. Tapping on the collection view cell will bring you to the service alert's target URL. In the Android case, there's an interstitial alert that can show you more text.

I think the proper way of resolving this is to display an alert style UIAlertController when the service alert is tapped. The alert will have two buttons: "Dismiss" and "Read More". Tapping dismiss closes the alert. Tapping "Read More" pushes an SFSafariViewController onto the stack with the service alert's URL target.

original-FFAEFF63-BF77-4321-B63E-600874B39FC5

Screenshot_20231208-084702

aaronbrethorst:
Fixes #700 

We now show the full content of the transit alert's body in a popup view controller, and optionally include a "Learn More" button at the bottom of it if the transit alert includes a URL.

# San Diego Ux

SD

# Seattle Ux

SEA

### DIFF
--- a/OBAKit/Alerts/TransitAlertDetailViewController.swift
+++ b/OBAKit/Alerts/TransitAlertDetailViewController.swift
@@ -7,13 +7,18 @@
 
 import OBAKitCore
 import UIKit
+import WebKit
+import SafariServices
 
-class TransitAlertDetailViewController: UIViewController {
-    private let transitAlert: TransitAlertViewModel
-    private let webView = DocumentWebView()
+/// Renders a full page version of a `TransitAlertViewModel`
+///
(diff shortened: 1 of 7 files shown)


aaronbrethorst:
FYI @devinbraun

aaronbrethorst:
Also, @ualch9 love your thoughts if you have time.

devinbraun:
@aaronbrethorst Wow, looks good!  Thanks for the quick turn around!

Issue #625, opened 2023-01-06
Support "Full Page" screenshots


ualch9:
https://developer.apple.com/documentation/uikit/uiscreenshotservice

ualch9:
This issue number was misquoted, it is not completed yet.

RV784:
Hi, I would love to work on this task to understand the codebase and eventually contribute to OneBusAway under GSOC.

I build the code and added this full page screenshot functionality. I'm attaching the screen recording for the same.


https://github.com/OneBusAway/onebusaway-ios/assets/58266095/ecac0a53-69ab-4bc1-a432-e87da3ca61a2

RV784:
Before making a PR, I would like a brief discussion about how I did it and what would be the best practice to implement it in our code.

PR #716, opened 2024-03-02
Fix: issue 625

RV784:
Issue Link -> https://github.com/OneBusAway/onebusaway-ios/issues/625

## Code Changes 
* Added UIScreenshotService delegate to AppDelegate
* Adding OBAListView instance from `loadView()` instead of adding it as a subview in `viewDidLoad()`


aaronbrethorst:
@RV784 the rounded cells have black backgrounds, rendering the text illegible. Compare to the product UI where the text on this screen is clearly visible.

PR #723, opened 2024-03-28
Issue 719: Add Custom Region URL

hilmyveradin:
PR for Issue #719 

Here's a video of working add custom region via URL:

https://github.com/OneBusAway/onebusaway-ios/assets/74521244/382d027c-654f-4666-a7ed-14478a878282

Here's a video of manage edge cases, say, when there's no name on the params:
https://github.com/OneBusAway/onebusaway-ios/assets/74521244/ca22596c-a439-4b46-8007-77e9712da02b


Apps/OneBusAway/AppDelegate.m
```diff
@@ -85,7 +85,8 @@ - (void)applicationWillResignActive:(UIApplication *)application {
 #pragma mark - OBAApplicationDelegate
 
 - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDictionary<UIApplicationOpenURLOptionsKey,id> *)options {
-    return [self.app application:app open:url options:options];
+    [CustomRegionDeepLink parseURL:url application:self.app];
```
aaronbrethorst, line 88: this will break every other deep link in the app.

aaronbrethorst (changes requested):
Go look at `Application.application(_:, open url:, options:)` to see how this should be plumbed into the app. You're going to want to extend `URLSchemeRouter` and add some tests.

OBAKit/Orchestration/Application.swift
```diff
@@ -363,6 +364,18 @@ public class Application: CoreApplication, PushServiceDelegate {
             presentDonationUIOnActive = false
             donationPromptID = nil
         }
+
+        if presentAddRegionAlertOnActive, let topViewController {
```
aaronbrethorst, line 368: these strings all need to be localized.

OBAKit/Orchestration/Application.swift
```diff
@@ -363,6 +364,18 @@ public class Application: CoreApplication, PushServiceDelegate {
...
+            // Show alert for nil addRegion data
+            let alertController = UIAlertController(
+                title: "Error",
```
aaronbrethorst, line 371: `Strings.error`

OBAKit/Orchestration/Application.swift
```diff
@@ -363,6 +364,18 @@ public class Application: CoreApplication, PushServiceDelegate {
...
+                message: "The provided region URL is invalid or does not point to a functional OBA server.",
+                preferredStyle: .alert
+            )
+            alertController.addAction(UIAlertAction(title: "OK", style: .default))
```
aaronbrethorst, line 375: `Strings.ok`

OBAKit/Orchestration/Application.swift
```diff
@@ -429,15 +442,48 @@ public class Application: CoreApplication, PushServiceDelegate {
...
+        switch urlType {
+        case .viewStop(let stopData):
+            guard let topViewController = self.topViewController else { return false }
+            viewRouter.navigateTo(stopID: stopData.stopID, from: topViewController)
+            return true
+        case .addRegion(let regionData):
+            viewRouter.rootNavigateTo(page: .map)
+            Task { @MainActor in
+                do {
+                    if let regionData = regionData {
```
aaronbrethorst, line 459: let's get rid of the pyramid of doom here:

```
guard let regionData else {
    presentAddRegionAlertOnActive = true
    return
}

guard let regionCoordinate = try await self.apiService?.getAgenciesWithCoverage().list.first?.region else {
    return
}
```

OBAKitCore/DeepLinks/URLSchemeRouter.swift
```diff
@@ -15,6 +15,17 @@ public struct StopURLData {
     public let regionID: Int
 }
 
+public struct AddRegionURLData {
```
aaronbrethorst, line 18: Please add a doc comment to this struct

OBAKitCore/DeepLinks/URLSchemeRouter.swift
```diff
@@ -15,6 +15,17 @@ public struct StopURLData {
...
+    public let name: String
+    public let obaURL: URL
+    public let otpURL: URL?
+}
+
+// Enum to distinguish between URL types
+public enum URLType {
```
aaronbrethorst, line 25: Please add a doc comment to this enum

aaronbrethorst (changes requested):
this looks like great progress! There are a couple small changes I'd like to see, but I think we'll be ready to go after that. thanks again for tackling this!

hilmyveradin:
> this looks like great progress! There are a couple small changes I'd like to see, but I think we'll be ready to go after that. thanks again for tackling this!

Done, kindly check it out again!

aaronbrethorst (approved):
nice work @hilmyveradin!

PR #751, opened 2024-10-23
iOS 18 Widget for OBA

manu-r12:
A WidgetKit Widget to show transit arrivals and departures at a glance, optimized for iOS 18 compatibility


manu-r12:
@aaronbrethorst

Issue #754, opened 2024-10-28
Weather Button Tap Not Responding


manu-r12:
**I don't know if this is happening in my CASE only**, 
but when tapping the  `-` button on the map page the expected action does not occur. The button appears to register the touch (i checked it in the code), but it does not trigger the intended action.

https://github.com/user-attachments/assets/ef45f584-fa94-46dd-aaeb-d62bc3985336

PR #798, opened 2025-03-07
Fix the JSON decoder extension for Weather Button(Issue#754)

adityaanand176:
Refers to [Issue#754](https://github.com/OneBusAway/onebusaway-ios/issues/754)

The format of the date was not being changed correctly to match Swift's ISO8601 format, hence throwing the error "Expected date string to be ISO8601-formatted."

The weather button functions in a way that if the forecast is nil, then the button would not load on the screen(instead of being there as a hyphen and unresponsive).

Attached a video of me changing regions to see the temperature changes. In my opinion, there is still room for improvement if we can incorporate a more detailed UI of the weather function or show more than just a summary of the weather. 

Let me know if you want any additional changes.

https://github.com/user-attachments/assets/2bbd3d7b-f59d-463c-a71d-06d41abf329c

### DIFF
--- a/OBAKit/Mapping/MapViewController.swift
+++ b/OBAKit/Mapping/MapViewController.swift
@@ -273,6 +273,7 @@ class MapViewController: UIViewController,
                     self.forecast = forecast
                 }
             } catch {
+                weatherButton.isHidden = true
                 Logger.error(error.localizedDescription)
             }
         }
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
nice work @adityaanand176

Issue #419, opened 2021-07-13
Improve the errors message displayed for a valid vehicle not currently running 


aaronbrethorst:
Repro steps:

1. Find a valid vehicle ID (e.g. maybe `4313` in Puget Sound) and make sure it isn't currently on any trips (i.e. test around 11 PM)
2. Do a search for that vehicle ID and tap on the search result

Expected:

A useful error message that perhaps says "This vehicle is not currently on any trips."

Actual:

![image](https://user-images.githubusercontent.com/2254/125393338-63309200-e36d-11eb-9964-189c39438d35.png)

adityasharma2308:
Hi @aaronbrethorst,

I’m interested in working on this issue and have looked through the code, but I’m having trouble finding a valid vehicle ID to reproduce the error. Do you have any tips on locating an ID for a vehicle that’s valid but not currently running? Additionally, any guidance on how the error messaging is currently handled would be really helpful. I’d love to get started as soon as possible thank you!

PR #799, opened 2025-03-08
Improve error message for vehicles not currently running

ShreyasSahoo:
In the case where a valid vehicle is not on any trips, the getData func returns an empty array which produces a decoding error of the type keyNotFound, I have selectively added a custom error in that case.

- Added `SearchError.noTripsAvailable` case for vehicles with no trips.
- Localized error message in English, Spanish, Italian, Polish, and Simplified Chinese.
- Updated `Localizable.strings` files with new keys and translations.

Fixes: #419

https://github.com/user-attachments/assets/2637bc47-0049-44b9-9293-5c1a032ba458


ShreyasSahoo:
Hi @aaronbrethorst,

I’d appreciate it if you could review my PR and suggest any improvements.

I also noticed a few other places where we might need more user-friendly error messages. 
For example, when searching for a valid vehicle from the Map tab, the error bulletin currently displays "Data couldn't be read..." instead of a more descriptive message.

Here's a video demonstrating the issue.

Thanks!

https://github.com/user-attachments/assets/d35f6ca6-0d81-4004-bd83-a460ee62cda0

aaronbrethorst (approved):
nice work!

PR #806, opened 2025-03-21
Manual Regions Refresh on RegionPickerView

### LINKED ISSUE #605: Add a way to manually force a regions refresh
From the regions page in settings, there should be a button that forces regions to reload.

ShreyasSahoo:
Fixes #605

- Add a Button to manually refresh regions 
- Add a instanceId to the Region Model for proper Hashable Conformance and preventing fatal errors 

https://github.com/user-attachments/assets/67899c65-56e9-4409-8a3e-32c2704f291e

### DIFF
--- a/OBAKit/Onboarding/RegionPicker/RegionPickerView.swift
+++ b/OBAKit/Onboarding/RegionPicker/RegionPickerView.swift
@@ -61,9 +61,22 @@ public struct RegionPickerView<Provider: RegionProvider>: View, OnboardingView {
     public var body: some View {
         NavigationStack {
             List {
-                Toggle(
-                    "Automatically select region",
-                    isOn: $regionProvider.automaticallySelectRegion)
+                HStack {
+                    Toggle(
+                        "Automatically select region",
+                        isOn: $regionProvider.automaticallySelectRegion)
+
+                    Button {
+                        Task {
+                            await doRefreshRegions()
+                        }
+                    } label: {
+                        Image(systemName: "arrow.clockwise")
+                    }
+                    .disabled(disableInteractions || regionProvider.automaticallySelectRegion)
+                    .buttonStyle(.plain)
+                }
+
                 Picker("", selection: $selectedRegion) {
                     ForEach(filteredRegions, id: \.self) { region in
                         cell(for: region)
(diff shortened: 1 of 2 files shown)


aaronbrethorst (changes requested):
Thanks for pulling this together! A couple changes before I can merge:

1. Unit tests are failing; please investigate :)
2. Can you move the Refresh button onto its own line and give it a title? (`Strings.refresh`)
3. Can you say a bit more about the UUID you're generating in the Region model? That shouldn't be necessary because all regions should already have a unique ID.

ShreyasSahoo:
Hi @aaronbrethorst,

Apologies for the delayed response.

- The UUID I propose in the solution acts as an ID for the instance of that region, not the region itself (which is handled by the regionIdentifier). 
- This UUID is crucial for our Region Model because the List in RegionPickerView is not displaying quickly enough, causing hash values of regions to get mixed up. When two instances with the same properties generate the same hash value, it leads to a hash collision, resulting in a fatal error like the one shown below:

Screenshot 2025-03-23 at 2 18 46 AM

This StackOverflow thread discusses a similar issue:
[Link to StackOverflow thread](https://stackoverflow.com/questions/65466863/duplicate-keys-of-type-were-found-in-a-dictionary-when-there-is-no-dicti)

ShreyasSahoo:
**Regarding the failing tests**: I've identified why adding the instanceId causes test failures.

-  Tests like test_regionUpdated_notifications & test_init_loadsCurrentRegion_autoSelectEnabled do direct equality comparisons between Region objects:

XCTAssertEqual(regionsService.currentRegion, newRegion)

- With the UUID solution, these fail because we're comparing different object instances that have different UUIDs.

I think we can have a modified equality implementation that's stricter in production but more flexible in tests to fix this. 
I would be glad to get your further guidance on this:)

aaronbrethorst:
@ShreyasSahoo sorry for the delayed response.

> This UUID is crucial for our Region Model because the List in RegionPickerView is not displaying quickly enough, causing hash values of regions to get mixed up. When two instances with the same properties generate the same hash value, it leads to a hash collision, resulting in a fatal error like the one shown below:

Why and how is the UUID fixing a race condition? Specifically, when and why are there two instances with the same properties? What continues to make me really nervous about that UUID is that I expect that changing equality testing for regions will lead to duplicate regions appearing after the app refreshes its regions list.

Issue #802, opened 2025-03-19
Tab bar is in strange color on iOS 17.7.1


YellowColr:
I guess that the tab bar should be white when Dark Mode is off. But it seems that when I choose the Map or More tab, the tab bar appears kinda gray, and when I choose Recently Explored or Bookmarks tab, the tab bar appears black wrongly. 

![Image](https://github.com/user-attachments/assets/f4e3ff94-e587-471f-8dba-974aab433642)
![Image](https://github.com/user-attachments/assets/465f11d1-e7b6-405a-9650-de28e5b57584)
![Image](https://github.com/user-attachments/assets/ceebdf39-4617-40b9-9dca-402132d4c585)
![Image](https://github.com/user-attachments/assets/996ff2ac-a8d6-45ab-854d-99bd97570b64)

adityaanand176:
I had the same issue on my simulator iOS 18.0, but I was unable to reproduce it. Looks like it only occurs when it's in light mode.

YellowColr:
I have tested it many times on the iOS 18 simulator, and everything seems to be working fine. However this issue can be reproduced 100% of the time on a real device running iOS 17, which is very strange. 😳😳

aaronbrethorst:
no longer relevant; we only support iOS 18+

PR #808, opened 2025-03-24
Fix black tabbar issue

hfadaly99:
Fix https://github.com/OneBusAway/onebusaway-ios/issues/802

### The problem:
Tab Bar background turns black when we force `tabbar.isTranslucent = false` without any background (the background can be color, image or view) as it treated as opaque without any background

### The solution:
reset `isTranslucent` to the default value

### How to reproduce the issue
- The issue happens for `Recent` and `Bookmarks` tabs when they are **empty**, or in `more` tab when we **scroll down** (most probably because the viewcontroller doesn't extend below the tabbar when it is empty)
- It also happens for light mode only

### DIFF
--- a/OBAKit/ClassicUI/ClassicApplicationRootController.swift
+++ b/OBAKit/ClassicUI/ClassicApplicationRootController.swift
@@ -31,8 +31,6 @@ public class ClassicApplicationRootController: UITabBarController {
 
         super.init(nibName: nil, bundle: nil)
 
-        self.tabBar.isTranslucent = false
-
         self.application.viewRouter.rootController = self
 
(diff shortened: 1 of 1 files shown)


hfadaly99:
@aaronbrethorst can you check this fix please?

hfadaly99:
@aaronbrethorst, could you elaborate on what **bleed-through** means in the [previous issue](https://github.com/OneBusAway/onebusaway-ios/commit/59f9bde7dad6693ecc68a44d7031c4f86ae784e9)? I'd like to understand it better to avoid regressions. I also tried to google the problem but didn't find resources.

aaronbrethorst:
Sure. Check out the bottom of the screen in this screenshot:

IMG_F9160869AA7F-1

aaronbrethorst:
resolved in the latest code changes.

PR #809, opened 2025-03-26
Fix Filtering for Single-Route Stops and All-Routes-Hidden Case

### LINKED ISSUE #790: bus stop filter usability
### Problem

I use a bus stop that, for the last few months, showed no arrival info,, even though the transit agency's website reliably reported bus arrivals. Turns out the the stop was filtered, so I must have accidentally tapped the filter button at some point, but not selected the one bus to show. And since the stop only serves one bus route, there was nothing to show.

### Solution

- The filter option should be disabled for but stops that serve only one route.
- The filter should not be turned on unless at least one route is selected.

I think maybe there should also be some notice, below warning, to inform the user that the stop has been filtered, but maybe screen space is too limited.

flash1729:
Fixes #790 

**Description:**
Addresses the issue where filtering could hide all data at a stop, confusing users. Two main fixes:
1. **Single-Route Stops:** Modified `filterMenu()` to only show "Filtered Routes" if `stop.routes.count > 1`, preventing filtering on stops with one route.
2. **All Routes Filtered:** Updated `stopPreferences.didSet` and `stopUpdated(_:)` to detect when all routes are hidden, disable filtering (`isListFiltered = false`), and reload the UI with all routes visible.

**Changes:**
- `stopUpdated(_:)`: Resets filtering if all routes are hidden and refreshes UI.
- `filterMenu()`: Conditionally adds "Filtered Routes" for multi-route stops only.
- `stopPreferences.didSet`: Calls `stopUpdated(_:)` when filtering is active after preferences change.

**Testing:**
- Single-route stop: Confirmed "Filtered Routes" is absent, arrivals always show.
- Multi-route stop: Filtered all routes → UI reverts to showing all routes, no continuous requests.

### DIFF
--- a/OBAKit/Stops/StopViewController.swift
+++ b/OBAKit/Stops/StopViewController.swift
@@ -301,7 +301,6 @@ public class StopViewController: UIViewController,
 
         let showAll = UIAction(title: allRoutesTitle) { [unowned self] _ in
             if self.isListFiltered {
-                // Only change value if it's different to avoid unnecessary data loading.
                 self.isListFiltered = false
             }
         }
--- a/OBAKit/Stops/StopViewController.swift
+++ b/OBAKit/Stops/StopViewController.swift
@@ -102,6 +102,15 @@ public class StopViewController: UIViewController,
             application.userDataStore.addRecentStop(stop, region: region)
         }
         application.analytics?.reportStopViewed(name: stop.name, id: stop.id, stopDistance: analyticsDistanceToStop)
+
+        // Disable filtering if all routes are hidden to ensure data visibility
+        if isListFiltered {
+            let allRoutesHidden = stop.routes.allSatisfy { stopPreferences.hiddenRoutes.contains($0.id) }
+            if allRoutesHidden {
+                isListFiltered = false
+                dataDidReload() // Refresh UI to reflect the change
+            }
+        }
     }
 
     /// Arrival/Departure data for this stop.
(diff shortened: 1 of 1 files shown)


flash1729:
@aaronbrethorst please do check this pr 👍

flash1729:
Hey @aaronbrethorst  there is some feedback from Copilot available in the chats that might be helpful. I think it's worth checking out!

---

The changes made in the `StopViewController.swift` file are primarily focused on improving the handling of routes and filtering in the user interface. Here are the key changes:

1. **Ensuring Data Visibility**:
    - Added a check to disable filtering if all routes are hidden. This prevents the user from ending up with no visible data and refreshes the UI to reflect the change.
    ```swift
    if isListFiltered {
        let allRoutesHidden = stop.routes.allSatisfy { stopPreferences.hiddenRoutes.contains($0.id) }
        if allRoutesHidden {
            isListFiltered = false
            dataDidReload() // Refresh UI to reflect the change
        }
    }
    ```

2. **Streamlining UI Actions**:
    - Removed an unnecessary comment and check in the `showAll` action to simplify the code.
    ```swift
    let showAll = UIAction(title: allRoutesTitle) { [unowned self] _ in
        if self.isListFiltered {
            self.isListFiltered = false
        }
    }
    ```

3. **Guard Clause and Dynamic Menu Generation**:
    - Added a guard clause to handle the case when the `stop` variable is nil.
    - Dynamically generated the menu items based on the number of routes, ensuring that the `showFiltered` option is only added if there are multiple routes.
    ```swift
    guard let stop = stop else {
        return UIMenu(children: [showAll, showFiltered])
    }

    var children = [showAll]

    if stop.routes.count > 1 {
        if isListFiltered && stopPreferences.hasHiddenRoutes {
            showFiltered.image = UIImage(systemName: "checkmark")
        } else {
            showAll.image = UIImage(systemName: "checkmark")
        }
        children.append(showFiltered)
    }

    return UIMenu(children: children)
    ```

4. **Updating Stop Preferences**:
    - Added a call to `stopUpdated(stop)` when `stopPreferences` is set and the list is filtered, ensuring that the stop data is updated.
    ```swift
    private var stopPreferences: StopPreferences {
        didSet {
            dataDidReload()
            if let stop = stop, isListFiltered {
                stopUpdated(stop)
            }
        }
    }
    ```

Overall, the changes aim to enhance the user experience by ensuring that data is always visible, simplifying the code, and dynamically updating the UI based on the current state. These changes should improve the maintainability and functionality of the `StopViewController.swift` class.

aaronbrethorst (approved):
great work @flash1729!

PR #810, opened 2025-03-28
Fix: duplicated bookmarks in separate groups crash

hfadaly99:


hfadaly99:
@aaronbrethorst can you check this fix?

PR #813, opened 2025-04-04
Fix Fatal Error on setting Alert for 1min departure time 

### LINKED ISSUE #811: Setting Departure Alert for a bus departing in a minute crashes the app
When trying to set an alert for a bus that's about to depart in a minute, the app shows an empty time picker and then crashes if you tap it.

**Steps to reproduce:**

- In Puget Sound Region, find a stop with a bus departing in 1 minute
- Open the Slide Menu and try to set an alert for that departure
- The Reminder Time Picker will be empty
- Tapping on the empty picker crashes the app

**Expected behaviour:**
Don't show the "Add Alert" option in the slide menu for buses leaving in a minute - it's too late for an alert anyway.
**Fix suggestion:**
Hide the alert button when departure is ≤ 1 minute away

ShreyasSahoo:
Fixes #811 
Add check to show alert only for departure time >1min

### DIFF
--- a/OBAKit/Stops/StopViewController.swift
+++ b/OBAKit/Stops/StopViewController.swift
@@ -1022,7 +1022,7 @@ public class StopViewController: UIViewController,
             return false
         }
 
-        return arrivalDeparture.temporalState == .future
+        return arrivalDeparture.arrivalDepartureMinutes > 1
     }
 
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
nice work

PR #887, opened 2025-11-23
Fix: Constrain MKLocalSearch results to current region bounds

### LINKED ISSUE #858: Constrain search results to the user's area
I _thought_ that our MKLocalSearch results were being constrained to the area of the current OBA region, but this doesn't seem to be the case. Figure out why this is happening and fix it.

Image

mayukhbaidya18:
## Description
Fixes #858 

MKLocalSearch was not respecting the current OBA region bounds, returning search results from anywhere in the world regardless of the selected transit region.

## Solution
- Use `currentRegion.serviceRect` instead of visible `mapRect` to provide search region hint to MKLocalSearch
- Filter search results after retrieval to only include locations within the current region's bounds using `currentRegion.contains(location:)`
- This ensures users only see places within their transit agency's service area

## Testing
Tested with Washington D.C. region:
-  "union station" - correctly shows Union Station in D.C.
- "dupont" - correctly shows Dupont Circle in D.C.
- "pacific pla" - now shows "No results" instead of Pacific

### Before
image
image
image

### After  
image
image
image

## Changes
- Modified `OBAKit/Search/SearchInteractor.swift`:
  - Updated `placemarkSearch()` to use region's service bounds
  - Added result filtering in search completion handler

### DIFF
--- a/Apps/OneBusAway/Package.resolved
+++ b/Apps/OneBusAway/Package.resolved
@@ -78,8 +78,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/firebase/firebase-ios-sdk.git",
       "state" : {
-        "revision" : "793b67f4652e1a39d03fab6650033768afe6d15e",
-        "version" : "12.5.0"
+        "revision" : "087bb95235f676c1a37e928769a5b6645dcbd325",
+        "version" : "12.6.0"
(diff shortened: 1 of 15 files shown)


aaronbrethorst (approved):
nice work! 🔥

Issue #880, opened 2025-11-07
Stop header: don't allow it to take up available space on first display


aaronbrethorst:
there's a bug in the auto layout code in some code in/around StopViewController that causes the header to sometimes render like this:

<img width="450" alt="Image" src="https://github.com/user-attachments/assets/f7357a6c-b0af-49b5-a0d3-5972288775f3" />

instead of like this:

<img width="450" alt="Image" src="https://github.com/user-attachments/assets/d342345e-98a4-4911-b803-1500a66b92ff" />

# Acceptance criteria 

- [ ] Find and fix the bug that causes the stop header to grow on first display.

PR #889, opened 2025-11-24
fix: prevent stop header from expanding on first display (#880)

flash1729:
This PR fixes an issue where the stop header sometimes rendered with extra height on first load. The initial layout pass didn’t have accurate label sizes yet, so the header’s height calculation overshot and added unnecessary padding.

Calling `layoutIfNeeded()` at the end of `configureView()` forces the view to lay out its subviews before height computation kicks in. With this simple one line addition, the header now renders at the correct height immediately.


aaronbrethorst (approved):
🔥 nice work!

PR #913, opened 2025-12-12
Fix Sound Transit 2 Line schedule trip ordering(#909)

### LINKED ISSUE #909: 2 Line schedule is showing out of order
Sound Transit’s 2 Line schedule has trips out of order. The trip that leaves Spring District station at 4:54am should be at the start of the table.

Cf. 2 Line schedule at https://soundtransit.org/2.

Image

mayukhbaidya18:
Fixes #909
Trips in the schedule view were being sorted incorrectly because the sorting logic only looked at the first stop's departure time. This caused trips that don't serve the first stop (like the 4:54am trip from Spring District) to appear in the middle of the schedule instead of at the beginning.
Updates (Dec 17, 2025):
Fixed additional sorting issues identified by @johnhelmuth-st
Now sorts by actual trip start time (minimum departure across all stops) instead of first visible column
Fixed stop column ordering to preserve geographic sequence from API
Changes:
Sort trips by minimum departure time across all stops (not just first visible stop)
Fix AM/PM grouping to use actual trip start time
Fix References.stopsWithIDs() to preserve stop order from API
Further details given below!!

### DIFF
--- a/OBAKit/Schedules/ScheduleForRouteViewModel.swift
+++ b/OBAKit/Schedules/ScheduleForRouteViewModel.swift
@@ -97,48 +97,76 @@ class ScheduleForRouteViewModel: ObservableObject {
     }
 
     /// Returns a sorted list of departure times for display
-    /// Each row represents a trip, sorted by the earliest non-nil departure time
+    /// Each row represents a trip, sorted by the trip's actual start time (earliest departure across all stops)
     var sortedDepartureTimes: [[Date?]] {
-        let times = departureTimes
-        return times.sorted { row1, row2 in
-            // Find the first non-nil departure time for each row
-            let date1 = row1.first { $0 != nil } ?? nil
-            let date2 = row2.first { $0 != nil } ?? nil
-
-            // If both have dates, sort by date
-            switch (date1, date2) {
-            case (let d1?, let d2?):
-                return d1 < d2
+        guard let direction = currentDirection,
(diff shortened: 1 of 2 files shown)


mayukhbaidya18:
Thanks a lot @johnhelmuth-st for pointing this out!
So I've added a commit to address the additional sorting issues.
Problem with the original code:
The original implementation sorted trips by looking at the first column's departure time (row.first). This caused a couple of issues, such as:
Incorrect trip ordering: Trips that don't serve the first stop had nil in the first column, which caused them to sort incorrectly and appear in the middle or end of the schedule instead of chronologically.
Incorrect stop column ordering: The References.stopsWithIDs() function used .filter, which doesn't preserve order. This caused stop columns to appear in an arbitrary order rather than the correct geographic sequence along the route.

Here's how this new commit fixes it:
Trip sorting: Now it sorts by the actual trip start time, which finds the minimum departureTime across all stops in each trip's raw data, not just the first visible column. This ensures that all trips are ordered chronologically by when they actually begin service, regardless of which stops they serve.
Stop ordering: Changed stopsWithIDs() to use compactMap { stopWithID($0) } instead of .filter, which preserves the exact order of stop IDs provided by the API.
Testing:
Verified with Sound Transit S Line schedule
The 7:20am Tacoma Dome trip now appears in correct chronological position
image
image
image
image

PR #916, opened 2025-12-12
Fix:  stop annotation overlapping text after bookmark deletion (#915)

### LINKED ISSUE #915: [Bug] Stop annotation label renders with overlapping text after bookmark deletion
> App Version = **25.4.1**

### Description

I have observed a UI rendering issue where the stop annotation label appears garbled, distorted, or with overlapping text immediately after deleting a bookmark for that stop.
The issue occurs because the map view fails to remove the "Bookmarked" version of the stop annotation before adding the "Standard" stop annotation. This results in two labels being drawn directly on top of each other (z-fighting), creating a visual artifact where the text looks significantly jagged and unclear compared to other stops.
This issue persists until the map is refreshed or the region is changed significantly.

### 📹 Video

<video src="https://github.com/user-attachments/assets/5365fd7d-76c1-4481-9f91-65bbd244e8a3" width="220" controls></video>

### 📸 Photo

The video and screenshot demonstrates the issue: A bookmark is deleted, and upon returning to the map, the specific stop label is rendered incorrectly.

### Steps to Reproduce

1. Open the app and navigate to the map.
2. Select a stop and Add a Bookmark.
3. Go to the "Bookmarks" list.
4. Delete the Bookmark.
5. Navigate back to the Map View.

Observation: The label for the stop that was just un-bookmarked appears garbled, with overlapping text, or distorted compared to surrounding stops.

diveshpatil9104:
### Description
Fixes #915
Fixes the issue where stop annotation labels appear garbled with overlapping text immediately after deleting a bookmark. The map now properly cleans up and refreshes annotation views when bookmark state changes, eliminating the visual ghosting effect.

### Root Cause
The` MapRegionManager.displayUniqueStopAnnotations()` method had two critical issues:

1. Stale Bookmark Annotations: When a bookmark was deleted, the bookmarked annotation remained on the map while a new regular stop annotation was added at the same location, causing both to render simultaneously.

2. No View Cleanup: MapKit reuses annotation views for performance, but the method didn't call `prepareForReuse()` or reset view state when swapping between bookmark and stop annotations. This caused the old visual state to persist and overlap with the new rendering.

### Problem
When users deleted a bookmark and returned to the map, the stop label for that location appeared distorted with overlapping or doubled text. This occurred because:

- The old "bookmarked" annotation view wasn't removed from the map
- The annotation view was reused without clearing its previous styling
- Both the bookmark and stop renderings were drawn on top of each other (z-fighting)

The issue persisted until the user manually refresh map.

### Solution

1. Remove Stale Bookmarks: Added logic to detect and remove bookmark annotations that no longer exist in the bookmarks list, preventing duplicate annotations at the same location.
2. Proper Deselection: Deselect annotations before removal to prevent MapKit from maintaining corrupted selection state on views that are being recycled.
3. Force View Refresh: Created a new refreshAnnotationViews() method that:
- Tracks all stops affected by bookmark state changes
- Calls `prepareForReuse()` on affected annotation views to clear old rendering state
- Reassigns the annotation property to trigger proper redraw
- Reconfigures delegate and customization callbacks

### video
Before (Bug) | After (Fixed)
-- | --
<video src="https://github.com/user-attachments/assets/52ddf8ea-9238-46db-b298-c169aec772b2" width="400"/> | <video src="https://github.com/user-attachments/assets/d8af8ef2-40fc-4fab-94db-50eb2c3df2a1" width="400"/>
Stop label appears garbled with overlapping text after deleting bookmark | Stop label renders cleanly and clearly after deleting bookmark
### screenshot :
(Bug) | (Fixed)
-- | --
<image src="https://github.com/user-attachments/assets/0f0c84b2-2293-4bea-87a9-157407785b83" width="400"/> | <image src="https://github.com/user-attachments/assets/3097846f-28c8-44ad-9cb3-3b0ec6ad52d8" width="400"/>

### Modified Files
`OBAKit/MapRegionManager.swift`

### DIFF
--- a/OBAKit/Mapping/MapRegionManager.swift
+++ b/OBAKit/Mapping/MapRegionManager.swift
@@ -382,18 +382,13 @@ public class MapRegionManager: NSObject,
                 bookmarksHash[bookmark.stopID] == nil
             else {
                 return nil
-                }
+            }
             affectedStopIDs.insert(bookmark.stopID)
             return bookmark
         }
         let allAnnotationsToRemove = stopAnnotationsToRemove + bookmarkAnnotationsToRemove
-        for annotation in allAnnotationsToRemove {
-            if mapView.selectedAnnotations.contains(where: {
-                ($0 as? Stop)?.id == (annotation as? Stop)?.id ||
-                ($0 as? Bookmark)?.stopID == (annotation as? Bookmark)?.stopID
-            }) {
-                mapView.deselectAnnotation(annotation, animated: false)
-            }
+        for annotation in allAnnotationsToRemove where mapView.selectedAnnotations.contains(where: { $0 === annotation }) {
+            mapView.deselectAnnotation(annotation, animated: false)
         }
         mapView.removeAnnotations(allAnnotationsToRemove)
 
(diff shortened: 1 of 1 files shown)


aaronbrethorst (changes requested):
Hi Divesh,

Thank you for taking the time to investigate and fix this annotation rendering bug! Your root cause analysis is spot-on — the z-fighting issue from stale bookmark annotations not being removed is exactly what's causing the garbled text. I appreciate the thorough PR description explaining the problem and your approach.

Before I can merge this, I'd like to see a few changes:

### Main Concern: Async Dispatch in `refreshAnnotationViews()`

The `DispatchQueue.main.async` dispatch here concerns me:

```swift
private func refreshAnnotationViews(for affectedStopIDs: [StopID], bookmarksHash: [StopID: Bookmark]) {
    DispatchQueue.main.async { [weak self] in
        // ...
    }
}
```

Since displayUniqueStopAnnotations() modifies annotations synchronously but the view refresh happens on the next run loop iteration, there's a timing gap where:
- MapKit may render the view incorrectly before the refresh executes
- Other code could modify annotations in between
- Rapid bookmark deletions could cause overlapping refresh calls

Could you either:
1. Remove the async dispatch if we're already on the main thread (add an assertion to verify), or
2. Explain why the async dispatch is specifically needed here?

Minor Items

1. Unused parameter: The bookmarksHash parameter in refreshAnnotationViews() is passed but never used — please remove it.
2. Formatting: Small indentation issue on line 385:

```
    return nil
        }  // <-- should align with the guard/else
```

3. Selection check simplification (optional): The deselection logic could be simplified to:

```
if mapView.selectedAnnotations.contains(where: { $0 === annotation }) {
    mapView.deselectAnnotation(annotation, animated: false)
}
```

Thanks again for your contribution to OneBusAway! The core logic of your fix is solid — I just want to make sure we don't introduce potential timing issues. Let me know if you have any questions about the feedback.

diveshpatil9104:
Hi @aaronbrethorst,
Thank you for the thorough review! You're right about the async dispatch . I've removed it since we're already on the main thread. I've addressed all your feedback - 
Changes made:

- [x] 1. Removed` DispatchQueue.main.async` and added `assert(Thread.isMainThread)` for safety

 - The async was unnecessary as `displayUniqueStopAnnotations()` is always called on the main thread
 - The assertion ensures we catch any future misuse during development .

- [x] 2. Removed unused `bookmarksHash` parameter from ` refreshAnnotationViews() `
- Updated method signature: `private func refreshAnnotationViews(for affectedStopIDs: [StopID])`
- Updated call site: `refreshAnnotationViews(for: Array(affectedStopIDs)) `

- [x] 3. Fixed indentation on guard/else blocks 

- All closing braces now properly align with their guard/else statements


- [x] 4. Simplified selection check to use identity comparison 

- Changed to: `for annotation in allAnnotationsToRemove where mapView.selectedAnnotations.contains(where: { $0 === annotation })`
- More efficient and cleaner than property comparison

You were correct that the async was unnecessary. I originally added it thinking MapKit might need a run loop iteration to complete its internal updates, but testing confirmed the synchronous approach works perfectly and eliminates the timing concerns you mentioned. The fix now executes atomically without any gaps where state could be inconsistent.

Issue #614, opened 2022-10-13
Customize "more" page


wkulesza:
Another idea that comes to my mind is ability to customize the "More..." page with following things:
- link to OneBusAway (or other) page (where detailed tutorials/user manual are located) 
- link to phone number of agency - ability to click and call the number to get passenger informaton help directly from agency
- link to text service - click to text to obtain text information.

adityaanand176:
Hi! I've added the option to open the Tutorial page in the "About This App" section. I've attached the demonstration with this comment. 

https://github.com/user-attachments/assets/07e01266-ff05-493d-af4a-66d0178272dd

Borisserz:
Taking this — adding first-class More tab config keys for tutorial, phone, and text/SMS links (CustomLinks already covers generic URLs).

PR #917, opened 2025-12-13
Add call and text actions for agencies (issue #614)

Vedd-Patel:
This PR is made in reference of issue [#614](https://github.com/OneBusAway/onebusaway-ios/issues/614)
This change adds call and text actions for each agency in the Agencies screen. When a rider taps an agency, an action sheet now appears with options to open the agency website, call the agency using its published phone number, or start a text message to that same number, making it easier to contact specific agencies directly from the app.

Simulator Screenshot - iPhone 17 Pro - 2025-12-13 at 18 33 44


Vedd-Patel:
Okay done!!

Issue #655, opened 2023-03-24
Edit bookmarks not working


ualch9:
Multiple users are reporting that the standard edit bookmark workflow is silently failing. When users click on edit, change a value, then click done, the changes are lost.

Issue #922, opened 2025-12-16
Index Out of Bounds When Reordering Bookmarks in ManageBookmarksViewController


Vedd-Patel:
###  Problem Description

Attempting to reorder bookmarks in the **Edit Bookmarks** screen (`ManageBookmarksViewController`) using the drag handle causes the application to terminate immediately with a fatal error. This makes the reordering feature unusable.

###  Steps to Reproduce

1.  Navigate to the **Bookmarks** tab.
2.  Tap the **Edit** button.
3.  Drag any bookmark row using the reorder handle ($\equiv$) to a new position (either within the same group or to a different group).
4.  Drop the row to complete the move.
5.  **Expected Result:** The row moves successfully, and the table updates smoothly.
6.  **Actual Result:** The application terminates immediately.

###  Observed Errors and Behavior

The crash is traceable in the debugger to an **`Index out of bounds`** error within the Eureka framework's internal collection management.

A secondary symptom observed is that when the form reload logic is initially present (before the index crash occurs), the table view exhibits severe visual glitches and row duplication during the drag animation.

###  Attached Media

* **Video:** Includes a screen recording demonstrating the fatal error/crash when attempting to reorder a bookmark row. 
*  https://github.com/user-attachments/assets/969cd37d-a01c-4314-b493-247afd6cd565

* after this clip it redirects to the screenshot attached below

* **Screenshot:** Includes a screenshot showing the `Thread 1: Fatal error: Index out of bounds` traceback in Xcode.
<img width="1470" height="956" alt="Image" src="https://github.com/user-attachments/assets/64f6113a-8f70-4582-a6a2-5079d966dc73" />

###  Solution I've thought of

The fatal crash and visual glitches stemmed from a synchronization issue between the persistent data store and the Eureka form model during the `UITableView` move animation. The final fix utilizes a delayed, complete form refresh to prevent conflicts.

The following logic was implemented in `tableView(_:moveRowAt:to:)`:

1.  **Persistent Data Update:** The application's data store is updated immediately.
2.  **Delayed Form Reload:** A `DispatchQueue.main.asyncAfter(deadline: .now() + 0.5)` block is used to completely reset and rebuild the Eureka form from the corrected data store state.

This delay ensures that the `UITableView` and Eureka finish their internal move animations before the model is rebuilt, successfully resolving the `Index out of bounds` crash and eliminating the row duplication glitch.

* **Video (Demonstrating Fix):** 

[...]

PR #923, opened 2025-12-17
Fix: safely reorder bookmarks between groups (#922)

Vedd-Patel:
Fixes [#922](https://github.com/OneBusAway/onebusaway-ios/issues/922)

This PR fixes a crash in the **Edit Bookmarks** screen where reordering bookmarks caused an “Index out of bounds” error inside Eureka. The change updates the persistent data store when a row is moved, then refreshes the form after a short delay so `UITableView` and Eureka can safely finish their internal animations before the form is rebuilt. A screen recording and screenshot demonstrating the original issue and the corrected behavior are attached in the linked issue for reference.

### DIFF
--- a/OBAKit/Bookmarks/ManageBookmarksViewController.swift
+++ b/OBAKit/Bookmarks/ManageBookmarksViewController.swift
@@ -10,7 +10,6 @@
 import UIKit
 import Eureka
 import OBAKitCore
-import QuartzCore
 
 class ManageBookmarksViewController: FormViewController {
     private let application: Application
@@ -48,8 +47,6 @@ class ManageBookmarksViewController: FormViewController {
 
     // MARK: - TableView Delegate Overrides
 
-    private var pendingReloadWork: DispatchWorkItem?
-
     override func tableView(_ tableView: UITableView, moveRowAt sourceIndexPath: IndexPath, to destinationIndexPath: IndexPath) {
         guard let bookmark = bookmarkForBookmarkIndexPath(sourceIndexPath) else {
             return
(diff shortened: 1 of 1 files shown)


aaronbrethorst (changes requested):
Thanks for tackling this crash - the reordering feature is definitely broken and this is a good start. I have some feedback before we merge.

### Critical Issues

**1. Code Duplication**

The new code duplicates the existing `loadForm()` method. Replace lines 60-64 with a single call:

```swift
DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) { [weak self] in
    self?.loadForm()
}
```

2. Race Condition with Rapid Reordering

If a user drags multiple bookmarks quickly (before the 0.5s delay elapses), multiple delayed blocks will queue up. This can cause data inconsistency or crashes. Add a cancellation mechanism:

```swift
private var pendingReloadWork: DispatchWorkItem?

override func tableView(_ tableView: UITableView, moveRowAt sourceIndexPath: IndexPath, to destinationIndexPath: IndexPath) {
    guard let bookmark = bookmarkForBookmarkIndexPath(sourceIndexPath) else {
        return
    }

    let destinationGroup = groupForBookmarkIndexPath(destinationIndexPath)
    application.userDataStore.add(bookmark, to: destinationGroup, index: destinationIndexPath.row)

    // Cancel any pending reload to handle rapid reordering
    pendingReloadWork?.cancel()

    // Delay form refresh to allow UITableView move animation to complete.
    // Refreshing immediately causes index-out-of-bounds in Eureka.
    // See: https://github.com/OneBusAway/onebusaway-ios/issues/922
    let workItem = DispatchWorkItem { [weak self] in
        self?.loadForm()
    }
    pendingReloadWork = workItem
    DispatchQueue.main.asyncAfter(deadline: .now() + 0.3, execute: workItem)
}
```

3. Indentation

The new code uses inconsistent indentation (~16-20 spaces inside the closure). The rest of the file uses 4-space indentation. Please fix to match.

Issue #655 Concern

The PR description says this fixes #655 ("Edit bookmarks not working"), but that issue describes a different problem: bookmark name edits being silently lost when the user taps Done.

This PR only addresses the reordering crash from #922. Looking at the code, there's no onChange handler on the NameRow to persist edits, and no viewWillDisappear save logic like ManageGroupsViewController has.

Please either:

- Remove #655 from the PR description, or
- Add a separate commit that actually addresses saving bookmark name edits

Minor Suggestion

The 0.5s delay is a magic number. Consider using CATransaction.setCompletionBlock to properly wait for the animation, or at minimum add the comment explaining why the delay exists (as shown in the code block above).

Let me know if you have questions about any of this feedback.

aaronbrethorst (approved):
nice work! this is ready to merge. thanks for seeing this through!

PR #926, opened 2025-12-20
Add manual refresh button to Choose Region screen

Vedd-Patel:
FIXES [605](https://github.com/OneBusAway/onebusaway-ios/issues/605)

This PR adds a manual Refresh button to the `Choose Region` screen to allow users to update the regions list on demand. It also resolves a critical crash (`Fatal error: Duplicate keys of type 'ID'`) by updating the `RegionPickerView` to use `id: \.regionIdentifier` instead of `id: \.self`. This change ensures that SwiftUI identifies each region by its stable primary key rather than its mutable data properties. By separating stable identity from the "nuances" of equality (like changing names or URLs), the UI remains robust during data reloads while preserving the model's ability to compare full data sets for logic and caching.

Attaching the screenshot for the UI reference:
mark3


aaronbrethorst:
This looks great. The only change I would have suggested was to use the `TaskButton` component for the Refresh button to give the user feedback a task is running after they tap on it, but that's a nice to have.

PR #928, opened 2025-12-20
Fix :hidden arrival times when stop names are too long (#920)

### LINKED ISSUE #920: Arrival time truncated by long stop names in Trip Details view
### Description
I found a UI layout bug in the Trip Details view (the list of stops for a specific trip). When a stop name is very long , it expands too far to the right, causing the arrival time to be truncated or out of the visible area.
The arrival time is critical information and should take priority over the full display of the street name.

### Screenshots
The following screenshots demonstrate the issue where the arrival time is truncated  or out of the visible area :
| Screenshot 1 | Screenshot 2 |
|-------------|--------------|
|  |  |

### Steps to Reproduce
1. Tap on any stop on the map to view the list of arriving buses .
2. Tap on a specific arrival row to open the Trip Details view .
3. Scroll through the list of stops for that trip .
4. Locate a stop with a long name .
5. Observe that the time on the right is cut off .

### Expected Behavior 
The arrival time (e.g., "11:13 PM") should always be fully visible and pinned to the right. The stop name label should truncate with an ellipsis (`...`) if there is not enough horizontal space.

### Actual Behavior 
The stop name label expands to fill the width, pushing the time label out of the visible area or truncating it.

@aaronbrethorst, I believe I can fix this issue. I would like to work on this. Please let me know if I should proceed.

diveshpatil9104:
### Description
Fixes #920
Fixes the UI layout bug in the Trip Details view where arrival times are truncated or pushed off-screen when stop names are very long. The time label now always remains fully visible while stop names truncate gracefully with ellipsis.

### Root Cause
In `OBAKit/Trip/TripStopListItem.swift` the  `TripStopCell` class had incorrect Auto Layout priority configuration:
Both labels had .required compression resistance:
`titleLabel` (stop name): .`setContentCompressionResistancePriority(.required, for: .horizontal)`
`timeLabel` (arrival time): `.setContentCompressionResistancePriority(.required, for: .horizontal)`

When both labels refuse to compress, UIStackView cannot resolve the layout conflict. Since the stop name appears first in the stack and expands to accommodate long text, it pushes the time label beyond the visible bounds of the cell.

### Problem
When viewing the Trip Details screen (list of stops for a specific trip), stops with long names caused the arrival time to be truncated or completely invisible. This occurred because:
The stop name label was configured to never compress (`.required `priority)
The time label was also configured to never compress (`.required` priority)
The horizontal stack view had no way to resolve this conflict
The stop name, appearing first, would expand and push the time off-screen
Critical information (arrival time) was hidden while less critical information (full stop name) was prioritized.

### Solution
Fixed the Auto Layout priority hierarchy to ensure arrival times are always visible:
1. Title Label (Stop Name) - Allow Compression:
Changed compression resistance from `.required` to `.defaultLow`
Added `lineBreakMode = .byTruncatingTail` to show ellipsis (...) when text is too long
Added `numberOfLines = 0 `to allow wrapping to 2 lines when space is available

2. Time Label - Maintain Priority:
Kept compression resistance at `.required` to ensure it never truncates
Kept content hugging at `.required `to prevent unnecessary expansion
This ensures UIStackView always prioritizes the time label, allowing the stop name to truncate gracefully.

### Screenshots

| **Before (Bug)** | **After (Fixed)** |
|------------------|-------------------|
|  |  |
|  |  |

**Before:** Arrival times are cut off or completely hidden when stop names are long  
**After:** Arrival times are always fully visible; stop names truncate with an ellipsis

Modified Files : 
`OBAKit/Trip/TripStopListItem.swift`

### DIFF
--- a/OBAKit/Trip/TripStopListItem.swift
+++ b/OBAKit/Trip/TripStopListItem.swift
@@ -134,8 +134,10 @@ final class TripStopCell: OBAListViewCell {
 
     let titleLabel: UILabel = {
         let label = UILabel.obaLabel(textColor: ThemeColors.shared.label)
+        label.numberOfLines = 0
+        label.lineBreakMode = .byTruncatingTail
         label.setContentHuggingPriority(.defaultLow, for: .horizontal)
-        label.setContentCompressionResistancePriority(.required, for: .horizontal)
+        label.setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
         return label
     }()
 


aaronbrethorst (approved):
Good work, this will definitely fix the problem. However I am confused about why you're setting `label.lineBreakMode = .byTruncatingTail`, as that's the default value of the property as described here: https://developer.apple.com/documentation/uikit/uilabel/linebreakmode

PR #933, opened 2025-12-24
Fix : duplicate alerts in Alerts view (#932 )

### LINKED ISSUE #932: Duplicate alerts appearing in the "More" > "Alerts" list
### Description
An issue was found in the "Alerts" list (under the "More" tab). The exact same alert appears many times in a row instead of just once.

### Screenshots 
The following screenshots show the problem where the same alert repeats over and over:

| Screenshot 1 | Screenshot 2 |
|--------------|--------------|
|  |  |

| Screenshot 3 | Screenshot 4 |
|--------------|--------------|
|  |  |

### Steps to Reproduce

1. Open the app.
2. Tap on the More tab at the bottom navigation bar.
3. Tap on Alerts.
4. Scroll through the list and notice that the same alerts appears multiple times.

### Expected Behavior 
Each alert should only appear once in the list.

### Actual Behavior 
The same alert is listed again and again.

@aaronbrethorst, I would like to work on this.

diveshpatil9104:
### Description

> fixes #932

Fixes the duplicate alerts bug in the Alerts screen, where the same alert appeared multiple times in the list. Alerts are now deduplicated by their unique ID, ensuring each alert appears only once while preserving the original order.

### Problem
When viewing the Alerts screen (More tab → Alerts), users would see the same alert repeated multiple times. This occurred because:

- The alerts data source (`alertsStore.agencyAlerts`) contained duplicate entries with identical IDs
- No deduplication logic existed between the data source and UI rendering
- The list view displayed every alert in the array, including duplicates
- Users had to scroll through repeated copies of the same alert

### Root Cause
In `OBAKit/AgencyAlertsViewController.swift`, the `items(for listView:)` method passed  `alertsStore.agencyAlerts` directly to the list rendering logic without removing duplicates. When the `AgencyAlertsStore `received updates or loaded data from the Protobuf feed, it could accumulate duplicate alert entries with the same ID. Without filtering, all duplicates were rendered in the UI, creating a confusing user experience where identical alerts appeared multiple times.
### Screenshots 

| **Before (Bug)** | **After (Fixed)** |
|------------------|-------------------|
|  |  |
|  |  |

**Before** : The same alert appeared multiple times in the Alerts list.
**After**: Each alert now appears only once. Duplicate entries are removed , resulting in a clean, readable Alerts list with no repeated information.

### Solution
- Tracks alert IDs using a `Set<String>`
- Filters the alerts array using `Set.insert(_:).inserted` to ensure each alert appears only once
- Passes only unique alerts to the list rendering function

Modified Files

`OBAKit/Alerts/AgencyAlertsViewController.swift`

### DIFF
--- a/OBAKit/Alerts/AgencyAlertsViewController.swift
+++ b/OBAKit/Alerts/AgencyAlertsViewController.swift
@@ -148,7 +148,11 @@ class AgencyAlertsViewController: UIViewController,
     // MARK: - List data
 
     func items(for listView: OBAListView) -> [OBAListViewSection] {
-        return listSections(agencyAlerts: alertsStore.agencyAlerts)
+        var seenIDs = Set<String>()
+        let uniqueAlerts = alertsStore.agencyAlerts.filter { alert in
+            seenIDs.insert(alert.id).inserted
+        }
+        return listSections(agencyAlerts: uniqueAlerts)
     }
 
     func emptyData(for listView: OBAListView) -> OBAListView.EmptyData? {


aaronbrethorst (approved):
looks good, nice work Divesh.

PR #935, opened 2025-12-29
add schedule toggle and stop view (fixes #929)

### LINKED ISSUE #929: Schedule for stop
The access to schedule is a great thing. Thanks for adding this @aaronbrethorst 
Since, this is accessible through arrivals-and-departure for stop, i would have two suggestions:
1) make this more prominent - i.e. accessible directly from the view of live departures as a switch between "live departures" and "schedules for stop"
2) current view is a schedule for full route for all stops of that route. it would be great to list only schedules for currently selected stop ie, in our webapp, for a given stop, we're showing schedules like this:
https://pksonline.pl/stop/399-9/route/172/direction/0
- you can see all departures for current day (with hours and minutes)
- you can click on any departure and see full trip
- you can switch between schedules-for-stop for different routes visiting this stop.

Let me know if that makes sense.

Vedd-Patel:
**FIXES [#929](https://github.com/OneBusAway/onebusaway-ios/issues/929)**

I have implemented a two-level toggle system in the Live Departures view for better schedule access.

The first toggle provides single-click access to schedules directly from Live Departures. Within the schedule view, I've added a dedicated stop schedule that shows only departures for the current stop, plus a second toggle to switch between "Stop Schedule" and "Full Route Schedule" based on user preference.

## Level 1: Live Departures ↔ Schedules
**Live Departures** (existing) 
ScheduleMark1

## Level 2: Stop Schedule ↔ Full Route Schedule
<table>
<tr>
<td align="center">
<h4>Stop Schedule</h4>
<em>Only current stop departures</em><br><br>
Stop Schedule
</td>
<td align="center" style="padding-left:30px">
<h4>Full Route Schedule</h4>
<em>All stops on route</em><br><br>
Full Route
</td>
</tr>
</table>

### DIFF
--- a/OBAKit/Schedules/ScheduleForStopView.swift
+++ b/OBAKit/Schedules/ScheduleForStopView.swift
@@ -57,6 +57,9 @@ struct ScheduleForStopView: View {
                     await stopViewModel.fetchSchedule()
                 }
                 .onChange(of: stopViewModel.selectedRouteID) { _, newRouteID in
+                    // Reset the toggle so the user starts fresh on the new route
+                    isShowingFullRouteSchedule = false
+
                     updateRouteViewModel(for: newRouteID)
                 }
                 .onChange(of: stopViewModel.selectedDate) { _, newDate in
@@ -158,7 +161,7 @@ struct ScheduleForStopView: View {
 
             // Segmented control: Stop Focused Schedules vs Full Route Schedules
             if stopViewModel.selectedRouteID != nil {
-                Picker("", selection: $isShowingFullRouteSchedule) {
+                Picker(StopScheduleStrings.chooseScheduleType, selection: $isShowingFullRouteSchedule) {
                     Text(StopScheduleStrings.stopSchedule).tag(false)
                     Text(StopScheduleStrings.fullRouteSchedule).tag(true)
                 }
--- a/OBAKit/Stops/StopViewController.swift
+++ b/OBAKit/Stops/StopViewController.swift
@@ -221,7 +221,6 @@ public class StopViewController: UIViewController,
         super.viewDidAppear(animated)
 
         if let schedulesButton {
-
             scheduleTipPresenter.showIfNeeded(sourceItem: schedulesButton) { [weak self] vc in
                 guard let self else { return }
                 present(vc, animated: animated)
@@ -310,10 +309,11 @@ public class StopViewController: UIViewController,
 
         let filterMenuButton = UIBarButtonItem(title: filterButtonTitle, image: filterButtonImage, menu: filterMenu())
         let moreMenuButton = UIBarButtonItem(title: "MORE", image: UIImage(systemName: "ellipsis.circle"), menu: pulldownMenu())
-        self.schedulesButton = UIBarButtonItem(image: UIImage(systemName: "calendar"), style: .plain, target: self, action: #selector(showScheduleForStop))
-        self.schedulesButton?.accessibilityLabel = Strings.schedules
+        let schedulesBtn = UIBarButtonItem(image: UIImage(systemName: "calendar"), style: .plain, target: self, action: #selector(showScheduleForStop))
+        schedulesBtn.accessibilityLabel = Strings.schedules
+        self.schedulesButton = schedulesBtn
 
-        navigationItem.rightBarButtonItems = [moreMenuButton, filterMenuButton, schedulesButton!]
+        navigationItem.rightBarButtonItems = [moreMenuButton, filterMenuButton, schedulesBtn]
 
         self.moreMenuButton = moreMenuButton
     }
@@ -1285,4 +1285,3 @@ public class StopViewController: UIViewController,
         }
     }
 }
-


aaronbrethorst (changes requested):
Thanks for working on this feature, Ved! The two-level toggle system is a nice UX improvement that makes schedules more discoverable. I have some feedback on a few issues that should be addressed before merging.

---

## Critical Issues

### 1. Missing Localization for New UI Strings

The PR adds hardcoded English strings that should use `OBALoc()` for localization:

```swift
Text("Stop Schedule").tag(false)
Text("Full Route Schedule").tag(true)
Text("To: \(departure.headsign)")
```

**Suggested fix** - Add to `StopScheduleStrings` enum:
```swift
static let stopSchedule = OBALoc("stop_schedule_view.stop_schedule", value: "Stop Schedule", comment: "Tab label for stop-only schedule view")
static let fullRouteSchedule = OBALoc("stop_schedule_view.full_route_schedule", value: "Full Route Schedule", comment: "Tab label for full route schedule view")
static let toDestination = OBALoc("stop_schedule_view.to_destination_fmt", value: "To: %@", comment: "Headsign label showing destination")
```

### 2. NavigationStack Removal - Clarification Needed

The diff shows removal of the `NavigationStack` wrapper from `ScheduleForStopView`, which also removes the navigation title, close button, and toolbar. Can you clarify where navigation is coming from now? If `ScheduleForStopViewController` provides it via UIKit hosting, that's fine - just want to make sure the close button and title still work in the modal presentation.

### 3. Unrelated Change in RegionPickerView.swift

The `Button` → `TaskButton` change in `RegionPickerView.swift` appears unrelated to the schedule feature. Could you either:
- Submit it as a separate PR for cleaner git history, OR
- Add a note in the PR description explaining why it was included (e.g., bug found during testing)

---

## Important Issues

### 4. Potential Duplicate IDs in StopDeparture List

Using `tripID` alone as the `id` could cause SwiftUI List rendering issues if a trip serves the same stop multiple times (loop routes):

```swift
let departure = StopDeparture(
    id: stopTime.tripID,  // Could be non-unique
    ...
)
```

**Suggested fix** - Use a composite ID:
```swift
id: "\(stopTime.tripID)-\(stopTime.departureDate.timeIntervalSince1970)"
```

### 5. Selected Route Can Become Invalid After Date Change

When the date changes and new schedule data is fetched, the previously selected route may not exist in the new data (e.g., a route that doesn't run on weekends). Consider validating `selectedRouteID` after fetching:

```swift
// After setting scheduleData, validate route still exists
if let routeID = selectedRouteID,
   !(scheduleData?.stopRouteSchedules.contains(where: { $0.routeID == routeID }) ?? false) {
    selectedRouteID = scheduleData?.stopRouteSchedules.first?.routeID
}
```

### 6. Accessibility: Segmented Picker Has Empty Label

```swift
Picker("", selection: $isShowingFullRouteSchedule)
```

**Suggested fix**:
```swift
Picker("Schedule Type", selection: $isShowingFullRouteSchedule)
    .accessibilityLabel("Choose between stop schedule and full route schedule")
```

---

## Suggestions (Optional)

### 7. Nest StopDeparture Inside ViewModel

Consider nesting the `StopDeparture` struct inside `ScheduleForStopViewModel` to properly scope it and avoid polluting the module namespace.

### 8. Add Hashable Conformance

Adding `Hashable` is free for structs with Hashable properties and improves SwiftUI diffing:
```swift
struct StopDeparture: Identifiable, Hashable { ... }
```

### 9. Consider Renaming StopDeparture

Since `ArrivalDeparture` exists for real-time data, consider `ScheduledDeparture` to clarify this represents scheduled (not real-time) data.

---

## What's Good

- Moving schedules to a prominent navigation bar button addresses the discoverability concern from issue #929
- The two-level toggle (Live↔Schedule, Stop↔FullRoute) is intuitive
- Proper headsign fallback logic (falls back to `directionHeadsign` when `stopHeadsign` is empty)
- Departures are properly sorted by time
- Good use of `@objc` for the selector target

---

Let me know if you have any questions about the feedback!

aaronbrethorst (changes requested):
Hey Ved, excellent work iterating on this PR — you've addressed all the feedback from the previous two rounds cleanly. The localization, composite IDs, route validation, tip re-anchoring, accessibility label, and presentation style consistency all look great. Just a few more items before we can merge:

## Important

1. **Force unwrap of `schedulesButton!` is fragile.** (`StopViewController.swift:316`) The `schedulesButton` is assigned two lines above, so it's non-nil, but force-unwrapping an Optional instance property is crash-prone if this function is ever refactored. Use a local `let` instead:

   ```swift
   let schedulesBtn = UIBarButtonItem(image: UIImage(systemName: "calendar"), style: .plain, target: self, action: #selector(showScheduleForStop))
   schedulesBtn.accessibilityLabel = Strings.schedules
   self.schedulesButton = schedulesBtn

   navigationItem.rightBarButtonItems = [moreMenuButton, filterMenuButton, schedulesBtn]
   ```

2. **Segmented Picker has an empty title string.** (`ScheduleForStopView.swift:161`) The `.accessibilityLabel()` modifier on a segmented `Picker` may not propagate to all child segments on all iOS versions. Since `.pickerStyle(.segmented)` hides the label visually anyway, pass the accessibility string as the Picker title instead of relying solely on the external modifier:

   ```swift
   Picker(StopScheduleStrings.chooseScheduleType, selection: $isShowingFullRouteSchedule) {
   ```

## Fit and Finish

3. **Stray blank line in `viewDidAppear`.** (`StopViewController.swift:224`) There's an empty line between the `if let schedulesButton {` and the first statement in the block. The rest of the file doesn't have leading blank lines inside `if let` blocks — remove it for consistency.

4. **`isShowingFullRouteSchedule` doesn't reset when route changes.** (`ScheduleForStopView.swift:32`) If a user is viewing "Full Route Schedule" for Route A and switches to Route B, they stay in full-route mode, which triggers a new data load they didn't explicitly request. Consider resetting to `false` in the `.onChange(of: stopViewModel.selectedRouteID)` handler, or add a brief code comment explaining that preserving the toggle state across route changes is intentional.

Issue #930, opened 2025-12-22
Schedules for stop - switch between 12 vs 24 hour clock


wkulesza:
Another suggestion to a great and long-awaited feature of displaying schedule for stop would be to be able to display 24 hour clock rather than 12 hour clock - while the US uses 12, most (if not all Europe) uses only 24 hour clocks.

aaronbrethorst:
@Vedd-Patel sure, but you should use `Locale.hourCycle` - https://developer.apple.com/documentation/foundation/locale/components/hourcycle

PR #939, opened 2026-01-02
Add 24h format support with Locale detection

Vedd-Patel:
FIXES [#930](https://github.com/OneBusAway/onebusaway-ios/issues/930)

## Summary
Automatic 12h/24h time display switching based on device locale using `Locale.current`(hourCycle).

### ScheduleForRouteViewModel.swift
- Added computed property `departureTimesDisplay: [[Date?]]`
  - Returns `sortedDepartureTimes` formatted per user's hour cycle
  - Maps dates to `Date?` for optional handling
- New helper `formatDate(_ date: Date?) -> String?`
  - Uses `DateFormatter` with `Locale.current.hourCycle`
  - Handles nil dates gracefully

### GridTimetableView.swift
- Updated `ForEach` binding: `departureTimesDisplay[dayIndex][hourIndex]`
- New `timeText` formatter in view cell:

```
Text(timeFormatter.string(from: time ?? Date.distantPast))
  .opacity(time != nil ? 1 : 0)
```

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,38 @@
+{
+  "permissions": {
+    "allow": [
+      "WebFetch(domain:github.com)",
+      "Bash(find:*)",
+      "Bash(git log:*)",
+      "Bash(gh issue view:*)",
+      "Bash(gh pr list:*)",
+      "Bash(git branch:*)",
+      "Bash(swift package show-dependencies:*)",
+      "Bash(git clone:*)",
+      "Bash(cat:*)",
+      "mcp__sosumi__searchAppleDocumentation",
+      "mcp__sosumi__fetchAppleDocumentation",
+      "Bash(xcodebuild clean build-for-testing:*)",
+      "Bash(xcodebuild build:*)",
+      "Bash(xcodebuild clean build:*)",
+      "Bash(git stash:*)",
+      "mcp__context7__resolve-library-id",
+      "Bash(scripts/generate_project:*)",
+      "Bash(xcrun simctl:*)",
+      "WebSearch",
+      "Bash(open:*)",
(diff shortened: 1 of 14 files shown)


aaronbrethorst (changes requested):
Thanks for tackling this feature! I have some feedback before this is ready to merge.

## Incomplete HourCycle handling

The check `Locale.current.hourCycle == .oneToTwelve` doesn't account for all 12-hour locales. Locale.HourCycle has four cases, and .zeroToEleven is also a 12-hour format used in some regions, like Japan. Users with that locale will incorrectly see the flat 24-hour list instead of AM/PM grouping.

Suggested fix:

```
let is12Hour = Locale.current.hourCycle == .oneToTwelve ||
                Locale.current.hourCycle == .zeroToEleven
```

Consider extracting this to a shared extension like Locale.uses12HourClock since it's checked in three places.

## Duplicate formatters

timeFormatter and timeFormatterWithAMPM are now identical—both use .timeStyle = .short with .locale = .current. The original distinction (compact display vs. accessibility with AM/PM context) has been lost. If this is intentional, consider removing one or renaming timeFormatterWithAMPM to something like accessibilityTimeFormatter since it no longer guarantees AM/PM in 24-hour locales.

## Test will fail in 24-hour locales

test_formatTimeAccessible_withDate_includesAMPM asserts that the result contains "AM" or "PM", but with the new locale-aware implementation, this will fail when tests run on a machine configured with a 24-hour locale. The test needs to be updated to either set an explicit locale or adjust its expectations based on the current locale.

## Indentation in AccessibilityTimetable.swift

The code inside List has inconsistent indentation (looks like ~20 spaces instead of the expected 12). Please clean this up to match the project style.

Once these are addressed, this should be good to go. Let me know if you have questions about any of the feedback.

PR #973, opened 2026-01-08
fix : prevent map bottom sheets from stacking when tapping pins (#972)

### LINKED ISSUE #972: [Bug] : bottom sheets stack indefinitely when tapping multiple map pins
> App Version = 25.4.2
### Description
I have observed a UI navigation issue where bottom sheets stack on top of each other indefinitely when map pins are tapped repeatedly.

The issue occurs because the `MapViewController` fails to dismiss an existing `MapItemViewController` (bottom sheet) before presenting a new one. This results in a **pile-up** of view controllers.

**Same Pin:** Tapping the same pin multiple times opens duplicate sheets on top of each other.

**Different Pins:** Tapping a new pin while a sheet is open stacks the new sheet over the old one, rather than replacing it.

This persists until the user manually dismisses every single stacked sheet one by one to reveal the map again.

### ScreenRec
<video src="https://github.com/user-attachments/assets/f6730bc3-2a17-43a9-a027-c14c6d519147" width="220" controls></video>

### Steps to Reproduce

1. Open the app and navigate to the map.
2. Long-press to drop a pin or tap an existing POI to open the bottom sheet.
3. Without closing the sheet, tap the same pin again multiple times.
4. Alternatively, tap a different pin nearby without closing the first sheet.
5. Attempt to swipe down to dismiss the top sheet.

Result: The sheet dismisses, but reveals another sheet underneath it instead of showing the map. The user must swipe down multiple times to clear the stack.

diveshpatil9104:
### Description

> Fixes #972

Fixes the UI navigation issue where bottom sheets stack on top of each other indefinitely when map pins are tapped repeatedly. The map now properly dismisses existing bottom sheets before presenting new ones, ensuring only one sheet is visible at a time.
### Root Cause
`MapViewController.displayMapItemController(_:userPin:)` method creates and presents new `FloatingPanelController` instances without first dismissing existing `semiModalMapItemController` panels. This causes a **pile-up** effect where:

1. No Dismissal Check: The method immediately creates a new panel without checking if one already exists
2. Multiple Entry Points: Three different code paths can trigger panel creation (direct pin taps, search results, user-dropped pins)
3. Weak Reference Lost: Each new panel overwrites the `semiModalMapItemController` reference, but the old panels remain in the view hierarchy

This creates an invisible stack where users must manually dismiss multiple sheets one-by-one to return to the map view.

### Problem

**Scenario 1 - Same Pin:**
Long-press to drop a pin → Bottom sheet opens
Tap the same pin again → Second sheet opens on top
Tap again → Third sheet stacks on top
User swipes down → First sheet dismisses, revealing second sheet 

**Scenario 2 - Different Pins:**
Drop Pin A → Bottom sheet opens
Drop Pin B nearby → New sheet stacks on top of first
Tap POI marker → Third sheet stacks
User must swipe down 3 times to clear the stack

Result: memory leaks from retained panels, and confused users who can't return to the map view.

### Solution

1. Core Method (`displayMapItemController`):

- Added a check at the start of the method to see if `semiModalMapItemController` is non-nil.
- If a panel exists, it is dismissed immediately `(animated: false) `to prevent visual artifacts before showing the new one.

2. Pin Selection (`mapView(_:didSelect:)` ):

- Added logic to dismiss existing panels when a UserDroppedPin is selected, ensuring the UI resets before centering on the pin.

3. Search Results (`mapPanelController(_:didSelectMapItem:)`):

- Added dismissal logic before presenting search results to maintain a clean navigation state.

### ScreenRec : 

Before (Bug) | After (Fixed)
-- | --
<video src="https://github.com/user-attachments/assets/f6730bc3-2a17-43a9-a027-c14c6d519147" width="400"/> | <video src="https://github.com/user-attachments/assets/a8e88857-5105-4445-8db1-0194349182fe" width="400"/>
Tapping multiple pins causes bottom sheets to stack indefinitely. Users must swipe down multiple times to dismiss all stacked sheets. | Tapping any pin replaces the existing bottom sheet. Only one sheet is visible at a time, and a single swipe returns to the map.

### Modified Files
`OBAKit/OBANext/Map/MapViewController.swift`

### DIFF
--- a/OBAKit/Mapping/MapViewController.swift
+++ b/OBAKit/Mapping/MapViewController.swift
@@ -679,7 +679,10 @@ class MapViewController: UIViewController,
 
     private var semiModalMapItemController: FloatingPanelController?
 
-    /// Helper to dismiss any existing map item controller to avoid stacking
+    /// Dismisses the currently displayed map item controller panel, if one exists.
+    /// Ensures proper cleanup before displaying a new map item or when the associated pin is removed.
+    /// - Parameter animated: Whether to animate the dismissal. Use `false` for immediate replacement,
+    ///   `true` for user-initiated actions like pin removal.
     private func dismissExistingMapItemController(animated: Bool = false) {
         if let existingController = semiModalMapItemController {
             removeSemiModalPanel(existingController, animated: animated)
@@ -728,7 +731,6 @@ class MapViewController: UIViewController,
     }
 
     func mapPanelController(_ controller: MapFloatingPanelController, didSelectMapItem mapItem: MKMapItem) {
-        dismissExistingMapItemController()
         floatingPanel.move(to: .half, animated: false)
 
         let mapDestination = mapItem.placemark.coordinate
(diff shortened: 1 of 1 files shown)


aaronbrethorst (approved):
great work, Divesh! this is ready to merge.

PR #975, opened 2026-01-10
moved regions from userdefaults to disk

### LINKED ISSUE #629: Move regions storage from UserDefaults to disk
Move regions storage from UserDefaults to disk. Storing regions in a database could be overkill since there are only a handful of infrequently-updated regions.

Currently, regions are stored as encoded objects in UserDefaults. We should use `Codable` to store OBA-regions in `/Library/Application Support` and custom regions in `/Documents`.

Files stored in `/Documents` in the app's sandbox are synced via iCloud and may be exposed via Files.app, which could be helpful for sharing custom regions.

The currently selected region would continue to be stored in UserDefaults, identified by its `RegionIdentifier`.

## Possible file structure
```
Library/
  └ Application Support/       ☜ Won't be deleted by clearing cache, but won't be synced via iTunes/iCloud
    └ Regions/
      └ default-regions.json   ☜ This can be refreshed via OBA (not user-generated data)

Documents/
  └ custom-regions/            ☜ Any .json in the directory that can be decoded into a Region object.
    └ region-a.json
    └ region-b.json
```

`Library/Application Support/Regions/default-regions.json` is not a copy of regions.json, it's the encoded-JSON of how `OBAKitCore.Region`.

---

This change should not require an Internet migration, so we don't need to inform the user.

Kartikayy007:
Fixes #629 

- Store regions as JSON files instead of UserDefaults
- Add RegionsServiceFileManagerProtocol for testable file operations
- Store only regionIdentifier (Int) for current region, not full object
- all test passing

Storage Locations
- Default regions: Application Support/Regions/default-regions.json
- Custom regions: Documents/custom-regions/region-{id}.json

### DIFF
--- a/OBAKitCore/Location/Regions/RegionsService.swift
+++ b/OBAKitCore/Location/Regions/RegionsService.swift
@@ -32,8 +32,26 @@ public class RegionsService: NSObject, LocationServiceDelegate {
     private let apiService: RegionsAPIService?
     private let locationService: LocationService
     private let userDefaults: UserDefaults
+    private let fileManager: RegionsServiceFileManagerProtocol
     private let bundledRegionsFilePath: String
     private let apiPath: String?
+
+    private static var regionsDirectory: URL {
+        let appSupport = FileManager.default.urls(
+            for: .applicationSupportDirectory, in: .userDomainMask
+        ).first!
+        return appSupport.appendingPathComponent("Regions")
+    }
+
+    private static var customRegionsDirectory: URL {
+        let documents = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
+            .first!
+        return documents.appendingPathComponent("custom-regions")
+    }
+
+    public static var defaultRegionsFileURL: URL {
+        regionsDirectory.appendingPathComponent("default-regions.json")
+    }
 
     /// Initializes a `RegionsService` object, which coordinates the current region, downloading new data, and storage.
     /// - Parameters:
@@ -43,10 +61,11 @@ public class RegionsService: NSObject, LocationServiceDelegate {
     ///   - bundledRegionsFilePath: The path to the bundled regions file. It is probably named "regions.json" or something similar.
     ///   - apiPath: The path to the remote regions.json file on the server. e.g. /path/to/regions.json
     ///   - delegate: A delegate object for callbacks.
-    public init(apiService: RegionsAPIService?, locationService: LocationService, userDefaults: UserDefaults, bundledRegionsFilePath: String, apiPath: String?, delegate: RegionsServiceDelegate? = nil) {
+    public init(apiService: RegionsAPIService?, locationService: LocationService, userDefaults: UserDefaults, fileManager: RegionsServiceFileManagerProtocol = RegionsServiceFileManager(), bundledRegionsFilePath: String, apiPath: String?, delegate: RegionsServiceDelegate? = nil) {
         self.apiService = apiService
         self.locationService = locationService
         self.userDefaults = userDefaults
+        self.fileManager = fileManager
         self.bundledRegionsFilePath = bundledRegionsFilePath
         self.apiPath = apiPath
 
@@ -55,7 +74,7 @@ public class RegionsService: NSObject, LocationServiceDelegate {
             RegionsService.alwaysRefreshRegionsOnLaunchUserDefaultsKey: false
         ])
 
-        if let regions = RegionsService.loadStoredRegions(from: userDefaults), regions.count > 0 {
+        if let regions = RegionsService.loadStoredRegions(fileManager: fileManager), regions.count > 0 {
             self.regions = regions
         }
         else {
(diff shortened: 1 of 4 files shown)


aaronbrethorst (changes requested):
Thanks for working on this, Kartikay! This is a meaningful architectural improvement. Moving region storage to the file system is the right approach, and I appreciate the work you've put into creating a testable abstraction with `RegionsServiceFileManagerProtocol`.

I've reviewed the changes and have some feedback organized by priority.

---

## Critical Issues

These need to be addressed before we can merge.

### 1. Missing Data Migration for Existing Users

The PR removes the UserDefaults storage keys (`storedRegionsUserDefaultsKey`, `storedCustomRegionsUserDefaultsKey`) and changes how `currentRegion` is stored (from a full encoded `Region` object to just the `regionIdentifier` integer).

However, there's no migration path for existing users. When they update to this version:

- Their downloaded regions list will be lost (they'll fall back to bundled regions)
- Their custom regions will be lost entirely
- Their current region selection will be reset (the old encoded `Region` can't be read as an `Int`)

Per the original issue #629: *"This change should not require an Internet migration, so we don't need to inform the user."* — but that assumes we silently migrate their data, not lose it.

**Required:** Add migration logic in `init` that:
1. Checks if the old UserDefaults keys contain data
2. Migrates that data to the new file-based storage
3. Migrates `currentRegion` from the encoded object to the `regionIdentifier`
4. Cleans up the old UserDefaults keys after successful migration

### 2. Code Formatting Issue in `loadStoredRegions`

The `loadStoredRegions` method has broken indentation that suggests a copy-paste error:

```swift
private static func loadStoredRegions(fileManager: RegionsServiceFileManagerProtocol) -> [Region]? {
let fileURL = defaultRegionsFileURL    // ← Should be indented

guard fileManager.fileExists(atPath: fileURL.path) else {    // ← Should be indented
    return nil
}
// ... rest of method has mixed indentation
}
```

The entire method body needs to be properly indented inside the function.

---

## Important Issues

### 3. Force Unwraps on System Directory URLs

```swift
private static var regionsDirectory: URL {
    let appSupport = FileManager.default.urls(
        for: .applicationSupportDirectory, in: .userDomainMask
    ).first!    // ← Force unwrap
    return appSupport.appendingPathComponent("Regions")
}

private static var customRegionsDirectory: URL {
    let documents = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
        .first!    // ← Force unwrap
    return documents.appendingPathComponent("custom-regions")
}
```

While these directories are effectively guaranteed to exist on iOS, force unwrapping is generally avoided in our codebase. Consider using `guard let` with a `fatalError` that includes a descriptive message, or returning an optional URL.

### 4. Silent Error Handling in `customRegions` Getter

```swift
public var customRegions: [Region] {
    // ...
    do {
        let files = try fileManager.contentsOfDirectory(at: directory)
        for file in files where file.pathExtension == "json" {
            let region = try fileManager.load(Region.self, from: file)
            regions.append(region)
        }
    } catch {
        regions = []    // ← Silently swallows all errors
    }
    return regions
}
```

If a user's custom regions fail to load due to a file system error, they'll see an empty list with no indication of what went wrong. Please add `Logger.error()` calls to capture these failures for debugging.

### 5. `defaultRegionsFileURL` Should Be Internal

```swift
public static var defaultRegionsFileURL: URL {
    regionsDirectory.appendingPathComponent("default-regions.json")
}
```

This is marked `public` but it's only used internally and in tests. Consider making it `internal` (the default) — tests can still access it via `@testable import`.

---

## Fit and Finish

### 6. Test Coverage: Corrupted File Handling

The original `test_corruptedDefaults` test verified that corrupted data in UserDefaults would fall back to bundled regions. The updated test just verifies "no file on disk" → "loads bundled."

Consider adding a test that writes invalid JSON to the mock file system to verify the app gracefully handles corrupted files on disk.

### 7. Consider Atomic Writes

`Data.write(to:)` is not atomic by default. For important user data, consider:

```swift
try data.write(to: destination, options: .atomic)
```

This prevents partial writes if the app crashes during save. Not critical for regions data, but it's a good defensive practice.

### 8. Doc Comment Update Needed

The `add(customRegion:)` method's doc comment was updated to say "File system errors" but the `delete(customRegionIdentifier:)` method's doc comment still says it throws for custom region lookup failures, which is no longer accurate (it now silently succeeds if the file doesn't exist).

---

## What's Good

- **Protocol-based design**: `RegionsServiceFileManagerProtocol` enables clean, isolated testing without touching the real file system
- **Follows Apple's guidelines**: Using Application Support for app-managed data and Documents for user-generated custom regions is correct per Apple's file system documentation
- **Simplified current region storage**: Storing only the `regionIdentifier` instead of the full encoded object reduces storage overhead and eliminates potential data desync issues
- **Well-designed mock**: `FileManagerMock` properly simulates file operations including directory tracking
- **Tests updated correctly**: The tests properly exercise the new behavior

---

Please address the critical issues (data migration and code formatting), and ideally the important issues as well. Once those are resolved, this will be ready to merge.

Thanks again for your work on this!

PR #977, opened 2026-01-10
Fix : re-render glitch in Edit Bookmarks during bottom sheet interaction (#976)

### LINKED ISSUE #976: Edit Bookmarks list flickers and re-animates on sheet interaction
### Description:

> App Version = 25.4.2

When in Edit Bookmarks or Edit Groups screen, interacting with the bottom sheet (dragging or modal transitions) causes the entire list to reload and re-animate, creating a visual glitch where rows slide in repeatedly.

### Steps to Reproduce:
 1. Go to Bookmarks tab
2. Tap "Edit"
3. Drag the bottom sheet down slightly or interact with it
4. Observe the list reloading with slide-in animation

### Expected Behavior:
List should remain stable without re-animating during sheet interactions

### Actual Behavior:
List reloads and re-animates every time `viewWillAppear` is triggered

### Screen Recording:
<video src="https://github.com/user-attachments/assets/a662dd0c-aa1e-4726-b6d7-01f3137b59ed" width="220" controls></video>

### Root Cause:
`loadForm()` is called in `viewWillAppear`, which triggers on every sheet interaction, causing unnecessary form rebuilds.

diveshpatil9104:
> fixes  #976

### Problem
When users interact with the bottom sheet in Edit Bookmarks or Edit Groups screens, the entire list reloads and re-animates, causing visual glitches and unnecessary state resets.

### Solution
Moved form initialization from `viewWillAppear` to `viewDidLoad` to ensure forms load only once during the view controller lifecycle.

### Changes
- Added `isFormLoaded` flag to track initialization state
- Moved `loadForm()` call to `viewDidLoad()`
- Added `reloadData()` method for future external refresh needs
- Applied fix to both `ManageGroupsViewController` and `ManageBookmarksViewController`

### ScreenRec
Before (Bug) | After (Fixed)
-- | --
<video src="https://github.com/user-attachments/assets/a662dd0c-aa1e-4726-b6d7-01f3137b59ed" width="400"/> | <video src="https://github.com/user-attachments/assets/4e142640-ad06-4db1-9153-0d5c6a98ed21" width="400"/>
Interacting with the bottom sheet (dragging or bouncing) triggers a full reload of the list. This causes a jumpy "slide-in" animation and visual glitch every time the sheet moves. | The list is loaded only once. Interacting with the bottom sheet no longer triggers a reload, ensuring the list remains stable and the animation only happens when entering the screen.

### Files Changed
- `ManageGroupsViewController.swift`
- `ManageBookmarksViewController.swift`

### DIFF
--- a/OBAKit/Bookmarks/ManageBookmarksViewController.swift
+++ b/OBAKit/Bookmarks/ManageBookmarksViewController.swift
@@ -13,7 +13,6 @@ import OBAKitCore
 
 class ManageBookmarksViewController: FormViewController {
     private let application: Application
-    private var isFormLoaded = false
 
     init(application: Application) {
         self.application = application
@@ -35,15 +34,6 @@ class ManageBookmarksViewController: FormViewController {
         tableView.setEditing(true, animated: false)
     }
 
-    override func viewWillAppear(_ animated: Bool) {
-        super.viewWillAppear(animated)
-        if !isFormLoaded {
-            loadForm()
-            tableView.setEditing(true, animated: false)
-            isFormLoaded = true
-        }
-    }
-
     // MARK: - Form Builders
 
     /// Creates, loads, and populates data in the Eureka Form object.
@@ -53,7 +43,6 @@ class ManageBookmarksViewController: FormViewController {
         for s in bookmarksSections {
             form +++ s
         }
-        isFormLoaded = true
     }
 
     // MARK: - TableView Delegate Overrides
@@ -153,11 +142,5 @@ class ManageBookmarksViewController: FormViewController {
                 }
             }
         }
-    }
-    
-    func reloadData() {
-        isFormLoaded = false
-        loadForm()
-        tableView.setEditing(true, animated: false)
     }
 }
--- a/OBAKit/Bookmarks/ManageGroupsViewController.swift
+++ b/OBAKit/Bookmarks/ManageGroupsViewController.swift
@@ -13,7 +13,6 @@ import OBAKitCore
 
 class ManageGroupsViewController: FormViewController {
     private let application: Application
-    private var isFormLoaded = false
 
     init(application: Application) {
         self.application = application
@@ -33,15 +32,6 @@ class ManageGroupsViewController: FormViewController {
         tableView.setEditing(true, animated: false)
     }
 
-    override func viewWillAppear(_ animated: Bool) {
-        super.viewWillAppear(animated)
-        if !isFormLoaded {
-            loadForm()
-            tableView.setEditing(true, animated: false)
-            isFormLoaded = true
-        }
-    }
-
     override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
 
@@ -54,7 +44,6 @@ class ManageGroupsViewController: FormViewController {
     private func loadForm() {
         form.removeAll()
         form +++ groupsSection
-        isFormLoaded = true
     }
 
     // MARK: - Groups Section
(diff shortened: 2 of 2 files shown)


aaronbrethorst (changes requested):
Thanks for tackling this, Divesh! The visual glitch you identified is a real problem, and moving initialization out of `viewWillAppear` is generally the right approach. However, the current implementation has some issues that need to be addressed.

---

## Critical

### 1. Duplicate Initialization Logic

The PR adds `loadForm()` to **both** `viewDidLoad()` and `viewWillAppear()`:

```swift
override func viewDidLoad() {
    super.viewDidLoad()
    loadForm()                              // ← Called here
    tableView.setEditing(true, animated: false)
}

override func viewWillAppear(_ animated: Bool) {
    super.viewWillAppear(animated)
    if !isFormLoaded {
        loadForm()                          // ← And also here
        tableView.setEditing(true, animated: false)
        isFormLoaded = true
    }
}
```

Since `viewDidLoad()` is always called before `viewWillAppear()`, and `loadForm()` sets `isFormLoaded = true`, the code in `viewWillAppear()` will never execute. This is dead code.

The fix should simply move the logic from `viewWillAppear()` to `viewDidLoad()` and remove the flag entirely:

```swift
override func viewDidLoad() {
    super.viewDidLoad()
    loadForm()
    tableView.setEditing(true, animated: false)
}

// Remove viewWillAppear override entirely, or keep it empty if needed for future use
```

This applies to both `ManageBookmarksViewController` and `ManageGroupsViewController`.

### 2. `isFormLoaded` Flag is Redundant

The `isFormLoaded` flag is set in two places:
- In `loadForm()` at the end
- In `viewWillAppear()` after calling `loadForm()`

Since `loadForm()` already sets the flag, setting it again in `viewWillAppear()` is redundant. But more importantly, if you follow the fix above, you won't need the flag at all.

---

## Important

### 3. `reloadData()` Method Concerns

The `reloadData()` method is added for "future external refresh needs":

```swift
func reloadData() {
    isFormLoaded = false
    loadForm()
    tableView.setEditing(true, animated: false)
}
```

Two issues:

1. **Unused code**: This method isn't called anywhere. Per our coding guidelines, we avoid adding code for hypothetical future needs. If this is genuinely needed, please point to where it will be called. Otherwise, remove it.

2. **Access control**: If this method is intended for external use, it should be marked `internal` explicitly or `public` if needed outside the module. Currently it has default `internal` access, which is fine, but a comment explaining its intended purpose would help.

### 4. `loadForm()` Access Level in ManageBookmarksViewController

Looking at the diff, `loadForm()` remains `private` but `reloadData()` would need to call it. This works since they're in the same class, but if you keep `reloadData()`, consider whether `loadForm()` should remain private or be extracted differently.

---

## Fit and Finish

### 5. Trailing Whitespace

There's trailing whitespace before the `reloadData()` method in both files:

```swift
    }

    func reloadData() {
```

Please remove the trailing whitespace on the blank line.

---

## What's Good

- **Correct diagnosis**: Moving initialization from `viewWillAppear()` to `viewDidLoad()` is the right fix. Per Apple's documentation, `viewWillAppear(_:)` is called every time the view is about to be displayed, while `viewDidLoad()` is called once after the view hierarchy is loaded into memory.
- **Consistent fix**: Applying the same pattern to both affected view controllers is good.
- **Great PR description**: The before/after videos clearly demonstrate the problem and fix.

---

## Summary

The core fix is correct, but the implementation is more complex than it needs to be. Please:

1. Remove the duplicate `viewWillAppear()` logic (keep only `viewDidLoad()`)
2. Remove the `isFormLoaded` flag (it's unnecessary with the simplified approach)
3. Either remove `reloadData()` or explain where it will be used
4. Clean up trailing whitespace

diveshpatil9104:
Thanks @aaronbrethorst for the thorough review ☺
you are right - I overcomplicated this. i have simplified the implementation:

**Critical Issues:**
1. Removed duplicate initialization - `loadForm()` now only called in `viewDidLoad()`
2. Removed `isFormLoaded` flag entirely - unnecessary with simplified approach

**Important Issues:**
3. Removed unused `reloadData()` methods from both controllers
4. N/A - `loadForm()` stays private ...only used internally

**Fit & Finish:**
5. Cleaned up trailing whitespace

The fix is now much cleaner: just move initialization to `viewDidLoad()` where it belongs, no flags or conditional logic needed. Both `ManageBookmarksViewController` and `ManageGroupsViewController` now follow the same simplified pattern :)

PR #1015, opened 2026-01-16
Surface DecodingError context for easier debugging

### LINKED ISSUE #709: Surface DecodingError.Context in UI or in Crashlytics
We frequently have server-side data scheme changes that may cause a non-optional field to be missing (for example, #588 and #706). Currently, this causes the user to see an unhelpful "The data couldn't be read because it is missing." alert.

To triage, a developer needs to (1) attach to the app, (2) reproduce the unexpected response, (3) output the `DecodingError` object in the debugger, and finally (4) update the Swift Decodable model. This process is costly and may cause us to not catch other more serious errors immediately.

In particular, we are interested in`DecodingError.Context`, where it contains which key is missing/corrupted/type mismatched/etc. This should be shown in the UI where the error description text is shown, and/or [uploaded to Crashlytics](https://firebase.google.com/docs/crashlytics/customize-crash-reports?platform=ios#log-excepts) where we may proactively catch decoding errors.

diveshpatil9104:
## Problem

> fixes #709

Server-side schema changes such as missing fields or type changes can cause JSON decoding to fail. Currently, these failures surface as a generic error message (e.g. “The data couldn’t be read because it is missing”), which makes
triaging difficult.
Developers must attach a debugger and manually inspect the `DecodingError` to identify which key failed and why.

---

## Solution

This PR improves decoding error visibility by intercepting `DecodingError`s at the API boundary and extracting useful context information, including:

- The type of decoding failure (missing key, type mismatch, corrupted data)
- The specific key involved
- The full JSON coding path

This information is logged in a human-readable format and can be easily forwarded to Crashlytics for proactive monitoring.

---

## Implementation

- Added `DecodingErrorReporter`, a small utility that converts `DecodingError`
  into a readable message.
- Updated `APIService+SendData` to:
  - Catch `DecodingError` explicitly
  - Log detailed decoding context with human-readable messages
  - Re-throw the original error to maintain compatibility

All decoding failures are now handled consistently in one place.

---

## Testing

A runtime decoding error was intentionally triggered to verify behavior.

### Before this change
- Only a generic error message was shown
- No information about the failing key or JSON path was visible

### After this change
- Console logs clearly show the decoding error type, failing key, and full JSON
  coding path

Example output:
```text
DecodingError.typeMismatch
Path: data → references → situations → 0 → severity 
Context: Expected to decode Int but found a string instead.
```

<details>
<summary>Console output screenshot</summary>

Decoding error context

</details>

This confirms decoding failures are now immediately diagnosable without attaching a debugger.

---

### DIFF
--- a/OBAKitCore/Network/APIService/APIService+SendData.swift
+++ b/OBAKitCore/Network/APIService/APIService+SendData.swift
@@ -47,6 +47,7 @@ extension APIService {
         request.httpMethod = method.value
         request.httpBody = requestData
         request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+        request.setValue("en-US", forHTTPHeaderField: "Accept-Language")
 
         let (data, response): (Foundation.Data, URLResponse)
 


aaronbrethorst (changes requested):
Thanks for working on this, Divesh! This is a valuable debugging improvement. Having clear, actionable error messages when decoding fails will significantly reduce triage time. The `DecodingErrorReporter` utility is well-structured and covers all four `DecodingError` cases.

---

## Critical

### 1. Inconsistent Coverage: `APIService+GetData.swift` Not Updated

The PR only adds detailed error reporting to `APIService+SendData.swift`, but `APIService+GetData.swift` also decodes JSON responses at lines 84-88:

```swift
do {
    return try decoder.decode(T.self, from: data)
} catch {
    logger.error("Decoder failed for \(url, privacy: .public): \(error, privacy: .public)")
    throw error
}
```

This is actually the more commonly used code path (for GET requests), so it should also benefit from the `DecodingErrorReporter`. Please update this method to use the same pattern:

```swift
do {
    return try decoder.decode(T.self, from: data)
} catch let error as DecodingError {
    let message = DecodingErrorReporter.message(from: error)
    logger.error("DECODING ERROR for \(url, privacy: .public): \(message, privacy: .public)")
    throw error
} catch {
    logger.error("Decoder failed for \(url, privacy: .public): \(error, privacy: .public)")
    throw error
}
```

---

## Important

### 2. Missing Unit Tests

The `DecodingErrorReporter` utility is a self-contained, pure function that's straightforward to test. Please add unit tests that verify each case produces the expected output:

- `keyNotFound` - verify the key name and path appear in the message
- `typeMismatch` - verify the expected type and path appear
- `valueNotFound` - verify the expected type and path appear
- `dataCorrupted` - verify the path and context appear

This will ensure the utility doesn't regress and documents the expected output format.

### 3. Issue #709 Mentions Crashlytics Integration

The original issue requests that these errors be *"uploaded to Crashlytics where we may proactively catch decoding errors."*

The current implementation only logs to the console. While logging is a good first step, should we also integrate with Crashlytics? If that's planned for a follow-up PR, that's fine—just want to confirm the intent.

---

## Fit and Finish

### 4. Log Message Prefix Consistency

In `APIService+SendData.swift`, the log messages use different prefixes:

```swift
logger.error("DECODING ERROR: \(message, privacy: .public)")
// vs
logger.error("UNEXPECTED DECODING ERROR: \(error.localizedDescription, privacy: .public)")
```

The second catch block handles non-`DecodingError` errors during the decode call. Consider whether "UNEXPECTED DECODING ERROR" is the right phrasing—it's not really a decoding error if it's not a `DecodingError`. Perhaps just "ERROR" or "DECODE CALL FAILED" would be clearer.

Also, in `APIService+GetData.swift`, the existing log uses "Decoder failed for [url]". For consistency across the codebase, consider using a similar format in `SendData`.

### 5. Access Control for `DecodingErrorReporter`

The `DecodingErrorReporter` enum is `internal` by default. If this utility might be useful for other modules or for testing from `OBAKitTests`, consider making it `public`. If it's intentionally internal-only, that's fine.

---

## What's Good

- **Comprehensive error handling**: All four `DecodingError` cases are covered, plus the `@unknown default` for future-proofing
- **Clean utility design**: `DecodingErrorReporter` is a focused, stateless utility that's easy to understand and reuse
- **Good path formatting**: The `→` separator in the coding path makes the JSON structure easy to follow
- **Privacy-aware logging**: Using `.public` privacy for error messages is appropriate since they don't contain user data
- **Well-documented PR**: The before/after comparison and example output make the improvement clear

---

## Summary

Please:

1. Add the same error reporting pattern to `APIService+GetData.swift` (lines 84-88)
2. Add unit tests for `DecodingErrorReporter`
3. Clarify whether Crashlytics integration is planned for this PR or a follow-up
4. Consider the log message consistency suggestions

aaronbrethorst (approved):
Hey Divesh, you've done excellent work on this `DecodingErrorReporter` implementation. The utility is clean, well-structured, and covers all four `DecodingError` cases with future-proofing via `@unknown default`. The test coverage is comprehensive—nine tests that verify each error type, nested paths, and context information. You've addressed all the feedback from both review rounds:

- Added `DecodingErrorReporter` to both `GetData` and `SendData` paths
- Wrote thorough unit tests with clear test model structures
- Standardized log message formatting across both files
- Made the utility `public` for testability
- Successfully rebased onto main to include the `Accept-Language` header fix

There's nothing left to change—this PR is ready to merge. This will make debugging server-side schema issues significantly easier for developers.

PR #1018, opened 2026-01-18
Fix ServiceAlert TimeWindow decoding for milliseconds 

### LINKED ISSUE #687: ServiceAlert/Situation TimeWindows are decoded as seconds instead of milliseconds.
Screenshot 2023-11-01 at 12 05 41 PM

However, one of the test case fixtures has seconds instead of milliseconds, only for the TimeWindow:

https://github.com/OneBusAway/onebusaway-ios/blob/main/OBAKitTests/Modeling/Model%20Unit%20Tests/ReferencesTests.swift#L80-L82

https://github.com/OneBusAway/onebusaway-ios/blob/main/OBAKitTests/fixtures/arrival-and-departure-for-stop-MTS_11589.json#L122-L127

diveshpatil9104:
## Summary
Fixes #687

`ServiceAlert.TimeWindow` previously assumed all Unix timestamps were expressed in seconds. However, some OneBusAway API responses provide timestamps in milliseconds, which caused decoded dates to appear tens of thousands of years in the future.

This PR adds automatic detection of seconds vs. milliseconds and normalizes both formats during decoding, while preserving backward compatibility with existing fixtures.

---

## Changes
- Added a `decodeUnixTimestamp()` helper to `ServiceAlert.TimeWindow`
- Detects seconds vs. milliseconds based on timestamp magnitude
- Normalizes millisecond values before decoding into `Date`
- Added a regression test `test_serviceAlerts_withMilliseconds()`

---

## Testing
-  Existing tests continue to pass (seconds-based timestamps)
- New test verifies millisecond-based timestamps
- Both formats decode to identical `Date` values

---

## Technical Details
A threshold of `10_000_000_000` (≈ November 2286 in seconds) is used to distinguish between timestamp formats:
- Values above the threshold are treated as milliseconds
- Values at or below the threshold are treated as seconds

This ensures correct decoding of real-world API data without breaking existing test fixtures.

---

### DIFF
--- a/OBAKitTests/Modeling/Model Unit Tests/ReferencesTests.swift
+++ b/OBAKitTests/Modeling/Model Unit Tests/ReferencesTests.swift
@@ -186,44 +186,4 @@ class ReferencesTests: OBATestCase {
         expect(trip.shapeID) == "Hillsborough Area Regional Transit_38042"
         expect(trip.headsign) == "Downtown to UATC via 15th St"
     }
-    func test_serviceAlerts_withMilliseconds() throws {
-        // Milliseconds-based JSON
-        let millisecondsJSON = """
-        {
-            "activeWindows": [
-                {
-                    "from": 1539781200000,
-                    "to": 1539826200000
-                }
-            ]
-        }
-        """.data(using: .utf8)!
-
-        // Seconds-based JSON (same moments in time)
-        let secondsJSON = """
-        {
-            "activeWindows": [
-                {
-                    "from": 1539781200,
-                    "to": 1539826200
-                }
-            ]
-        }
-        """.data(using: .utf8)!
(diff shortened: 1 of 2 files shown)


aaronbrethorst (approved):
Hey Divesh, you've done a great job addressing all the feedback on this timestamp fix. The implementation is solid and the test coverage is now comprehensive.

You moved the test to `ServiceAlertTests.swift`, added actual value assertions, and included the boundary test I suggested. The use of `beCloseTo` for the floating-point comparison at the threshold boundary is a smart touch.

There's nothing left to change—this PR is ready to merge. Thanks for your patience working through the review, and for fixing this long-standing bug that was causing dates to appear in the year 32,000+!

PR #1019, opened 2026-01-18
Fix genstrings extraction for OBALoc formatted data migration strings

### LINKED ISSUE #664: `genstrings` not working correctly with `OBALoc()` `format` variant
The variant of `OBALoc` that includes a `format:` parameter isn't being correctly interpreted by `genstrings`, which is invoked in the `extract_strings` script. It's not clear based on a cursory glance whether this is something that can be corrected with `genstrings`, or if we need to just remove this `OBALoc` variant and do things the hard way with it. 

Repro:

1. Run `script/extract_strings`
2. Look at the diff for OBAKit/Strings/en.lproj/Localizable.strings — specifically, look at "data_migration_bulletin.report_summary_number_of_successes" and "data_migration_bulletin.report_summary_number_of_failures"

Expected: 

These two keys have the values `"%d successful"` and `"%d failures"`, respectively.

Actual:

They don't.

diveshpatil9104:
### Summary

Fixes #664

Fixes an issue where formatted localization strings using the `OBALoc(..., format:)` variant were not correctly interpreted by `genstrings` when running `scripts/extract_strings`.

Specifically, the following keys were not being extracted with their expected `%d` format values:
- `data_migration_bulletin.report_summary_number_of_successes`
- `data_migration_bulletin.report_summary_number_of_failures`

---
### What changed

- Removed usage of the `format:` parameter in `OBALoc` for data migration summary strings
- Applied formatting at runtime using `String(format:)`
- Re-ran `scripts/extract_strings` and committed the updated `Localizable.strings` files

### Result

After this change, running `scripts/extract_strings` correctly generates:

```text
"%d successful"
"%d failures"

### DIFF
--- a/OBAKit/Strings/Localization.swift
+++ b/OBAKit/Strings/Localization.swift
@@ -14,7 +14,3 @@ fileprivate class Localization: NSObject {}
 internal func OBALoc(_ key: String, value: String, comment: String) -> String {
     return NSLocalizedString(key, tableName: nil, bundle: Bundle(for: Localization.self), value: value, comment: comment)
 }
-
-internal func OBALoc(_ key: String, format: String, comment: String, _ arguments: CVarArg...) -> String {
-    return String(format: OBALoc(key, value: format, comment: comment), arguments)
-}


diveshpatil9104:
Thanks for the review, Aaron!
I've removed the unused` OBALoc(format:)` function .

PR #1021, opened 2026-01-19
Fix map z-fighting when bookmarking both a Stop and a Trip at the same time

### LINKED ISSUE #1020: Bug: Garbled Map Labels When Stop and Trip Are Both Bookmarked
### Description

I have observed a UI rendering issue where the map displays two overlapping annotations for the same location if a user bookmarks both a Stop and a Trip associated with that stop. This causes "z-fighting," where the label appears bold, distorted, or garbled because two views are drawn on top of each other.

---
### Steps to Reproduce

1. Open the app and select a stop.
2. Add a bookmark for the Stop.
3. Add a second bookmark for a Trip  at the same stop.
4. Return to the Map View.
5. Observation: The stop label appears garbled compared to others.

---

### ScreenREC 
<video src="https://github.com/user-attachments/assets/f331ba5e-c335-4d73-a9c1-5cde393527ba" width="220" controls></video>

---
### Root Cause
The `displayUniqueStopAnnotations` method filters the bookmarks array directly. Since the array contains separate entries for the Stop bookmark and Trip bookmark , the map adds two separate annotation objects for the same coordinate.

---

diveshpatil9104:
### Description

> Fixes #1020

This PR resolves a rendering glitch where the stop annotation label renders with overlapping text. Previously, when a user had multiple bookmarks associated with the same location (e.g., bookmarking a Stop and a Trip at that same Stop), the stop label for that location appeared distorted with overlapping or doubled text.

---

### The Problem
The `displayUniqueStopAnnotations` method previously filtered the raw bookmarks array. Since the array can contain multiple bookmark objects for the same `stopID` (e.g., one `.stop` type and one `.trip` type), the map was adding multiple `MKAnnotation` objects to the exact same coordinate. TThis caused the map to render two views on top of each other, resulting in visual "z-fighting" artifacts where the text looked bold, jagged, or unreadable.

---

### The Fix 

**Deduplication**: The logic was updated to use a `bookmarksHash` dictionary. We now iterate over `bookmarksHash.values ` when adding annotations to the map. This guarantees that only one annotation is ever added per `stopID`, regardless of how many bookmarks exist for that location.

---

### Testing 

- Verified that adding both a Stop Bookmark and a Trip Bookmark for the same location now results in a single, clean annotation and text on the map.
- Verified that z-fighting artifacts are no longer present.

---

### ScreenREC
Before (Bug) | After (Fixed)
-- | --
<video src="https://github.com/user-attachments/assets/f331ba5e-c335-4d73-a9c1-5cde393527ba" width="400"/> | <video src="https://github.com/user-attachments/assets/03a44c51-ccb5-42dd-ab00-8881ba4af3ec" width="400"/>
Overlapping annotations caused "z-fighting," making the text distorted, or garbled. | Text remains crisp and readable without overlapping label.

### DIFF
--- a/OBAKit/Mapping/MapRegionManager.swift
+++ b/OBAKit/Mapping/MapRegionManager.swift
@@ -141,7 +141,7 @@ public class MapRegionManager: NSObject,
     private let mapViewShowsHeadingKey = "mapRegionManager.mapViewShowsHeadingKey"
 
     /// Provides storage for the last visible map rect of the map view.
-    /// 
+    ///
     /// In the event that this value is unavailable, the getter will try to offer up an alternative,
     /// such as the current region's service rect.
     public var lastVisibleMapRect: MKMapRect? {
@@ -384,6 +384,7 @@ public class MapRegionManager: NSObject,
 
     private func displayUniqueStopAnnotations() {
         var bookmarksHash = [StopID: Bookmark]()
+        // When multiple bookmarks exist for the same stop, the last one in the bookmarks array takes precedence
         for bm in bookmarks {
             bookmarksHash[bm.stopID] = bm
         }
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
@diveshpatil9104 

> Should I open a separate issue and PR to track this map refresh bug?

yes please

PR #1051, opened 2026-02-18
Fix: Remove notification observers in StopAnnotationView and WalkTimeView deinit

diveshpatil9104:
### Summary

`StopAnnotationView` and `WalkTimeView` each register a notification observer in `init()` but neither class implements `deinit` to remove them. This PR adds the missing cleanup.

### Problem

**StopAnnotationView** registers a VoiceOver status observer:
- `UIAccessibility.voiceOverStatusDidChangeNotification`

**WalkTimeView** registers a device orientation observer:
- `UIDevice.orientationDidChangeNotification`

While modern NotificationCenter (iOS 9+) uses zeroing-weak references and cleans up automatically on deallocation, explicit removeObserver calls are the established convention in this codebase (see StopArrivalView, SearchListViewController, ProgressHUD) and are widely accepted defensive practice.

### Fix

Added `deinit` to both classes with `NotificationCenter.default.removeObserver(self)`.

### Why this matters

This follows the same pattern established in the recent `MapViewController` fix and the existing `subscribeToTripPlannerNotifications` / `unsubscribeFromTripPlannerNotifications` convention in the codebase. Every `addObserver` should have a corresponding `removeObserver`.

### DIFF
--- a/OBAKit/Mapping/StopAnnotationView.swift
+++ b/OBAKit/Mapping/StopAnnotationView.swift
@@ -85,6 +85,10 @@ class StopAnnotationView: MKAnnotationView {
 
     required init?(coder aDecoder: NSCoder) { fatalError("init(coder:) has not been implemented") }
 
+    deinit {
+        NotificationCenter.default.removeObserver(self)
+    }
+
     // MARK: - Annotation View Overrides
 
     public override func prepareForReuse() {
--- a/OBAKit/Stops/WalkTimeView.swift
+++ b/OBAKit/Stops/WalkTimeView.swift
@@ -83,6 +83,10 @@ class WalkTimeView: UIView {
 
     required init?(coder: NSCoder) { fatalError("init(coder:) has not been implemented") }
 
+    deinit {
+        NotificationCenter.default.removeObserver(self)
+    }
+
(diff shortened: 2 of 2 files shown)


aaronbrethorst (approved):
Thanks for the cleanup work here, Divesh — nice to see the observer hygiene being applied consistently across the codebase.

## Feedback

### Fit and finish

1. **PR description overstates the problem (not a code issue)**
   The description says the target-action `addObserver` API "holds an **unsafe unretained** reference to the observer." That was true before iOS 9, but since iOS 9 (and this project targets iOS 17+), `NotificationCenter` uses zeroing-weak references for target-action observers and [automatically cleans up registrations when the observer is deallocated](https://developer.apple.com/documentation/foundation/notificationcenter/1413994-removeobserver). So the observers aren't actually "leaking" or "accumulating" — they're cleaned up by the runtime.

   That said, the `removeObserver(self)` calls are **harmless**, follow the existing codebase pattern (see `StopArrivalView`, `SearchListViewController`, `ProgressHUD`), and represent widely-accepted defensive practice. I'm fine merging this as-is. Just wanted to flag the nuance for your awareness going forward.

PR #1056, opened 2026-02-19
(Surveys) Fix hero question state inconsistency in External Survey

mosliem:
Fixes hero question state to ensure the view is removed from the hierarchy when opening an external survey by properly resetting hero state in `SurveysViewModel`.

> 1. showHeroQuestion not reset on external survey path -- hero popup stays visible on map


aaronbrethorst (approved):
Hey Mohamed — good eye catching this. `clearExternalSurveyState()` was the one call site that bypassed `clearHeroQuestionState()` and nil'd `heroQuestion` directly, leaving `showHeroQuestion` stuck as `true` so the hero popup lingered on the map. Using the existing helper that resets both properties is the correct fix, and it's consistent with how `skipSurvey()`, `remindMeLater()`, and `setRemainingSurveyQuestions()` already handle this. Looks good to merge!

PR #1057, opened 2026-02-19
Fix surveys view hosting protocol observation

mosliem:
This PR resolves two observation issues related to `SurveyViewHostingProtocol` 

> observeSurveyLoadingState() missing [weak self] -- ProgressHUD can get stuck indefinitely.

> observeSurveyToastMessage reads toast twice.

### DIFF
--- a/OBAKit/Mapping/MapViewController.swift
+++ b/OBAKit/Mapping/MapViewController.swift
@@ -1106,6 +1106,7 @@ class MapViewController: UIViewController,
 
     func stopObserveSurveysState() {
         observationActive = false
+        ProgressHUD.dismiss()
     }
 
     func observeSurveyHeroQuestion() {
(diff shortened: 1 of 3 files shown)


aaronbrethorst (approved):
Hey Mohamed — three solid fixes here, all addressing real issues with the Observation-based pattern.

The missing `[weak self]` in `observeSurveyLoadingState()` was the most consequential — without it, the view controller would be retained by the observation closure, and if the VC was deallocated mid-observation the ProgressHUD would stay stuck indefinitely. The fix brings it in line with every other `withObservationTracking` call in the file.

The toast double-read is a subtle one — reading `surveysVM.toast` multiple times across the `switch` and case branches means each access independently participates in the Observation tracking. Binding it once to `let toast` is both cleaner and semantically correct.

And adding `ProgressHUD.dismiss()` to `stopObserveSurveysState()` in both `MapViewController` and `StopViewController` is good belt-and-suspenders cleanup — ensures no stuck HUD if the observation cycle is interrupted for any reason.

This looks good to merge!

PR #1059, opened 2026-02-19
Fix Survey URL builder in application

mosliem:
Fixes this points: 

> 5. CoreApplication.externalSurveyURLBuilder uses concrete type instead of protocol

> 9. Missing space in MARK comment
CoreApplication.swift


aaronbrethorst (approved):
Hey Mohamed — clean as a whistle. Using `ExternalSurveyURLBuilderProtocol` instead of the concrete `ExternalSurveyURLBuilder` on the property type is the right call — it follows dependency inversion and makes the property mockable for tests. The MARK comment spacing fix is a nice touch too. Looks good to merge!

PR #1073, opened 2026-03-02
Improve Weather Card Details

### LINKED ISSUE #71: Improve weather feature
The weather display right now is incredibly basic. It'd be great to enhance the feature with the following items:

- [ ] Show both temperature and conditions on the map.
- [ ] Display today's forecast in a modal view controller. It should give the user current conditions and an hour-by-hour forecast.

Vedd-Patel:
## Summary

Closes #71

This PR improves the weather feature UI in the map view to closely 
match the Apple Maps weather experience, replacing the previous 
full-screen modal with a compact, floating weather card.

### WeatherCardView.swift (new file)
- Created a new compact floating card view that appears over the map
- Frosted glass background using `.ultraThinMaterial` matching the 
  map's aesthetic
- Header shows weather icon, region name, condition, chance of rain, 
  current temperature, and H/L range
- Horizontal scrollable hourly forecast row with "Now" highlighted
- Bottom row showing wind speed, precipitation, and feels-like 
  temperature
- Close button below the card as a clear affordance
- Tapping anywhere on the card also dismisses it
- All units (temperature, wind speed) respect the user's locale 
  (°F/mph for US, °C/km/h elsewhere)

### MapViewController.swift
- Replaced full-screen modal sheet with an inline floating card 
  anchored to the center of the map view
- Weather button moved to top-left corner with frosted glass styling
  matching the right toolbar
- Button shows current weather icon + temperature at a glance
- Tapping the button toggles the card; tapping outside dismisses it
- Button background adapts correctly to light and dark mode

### Weather Card
| Light Mode | Dark Mode |
|---|---|
| light mode | dark mode |


aaronbrethorst (changes requested):
Hey Ved, thanks for overhauling the weather experience — the floating card design with hourly scrolling and the frosted-glass button are a big step up from the old `UIAlertController`. The accessibility coverage throughout `WeatherCardView` is particularly well done, with combined labels, hints, and proper header traits. Before we can merge this, I will need you to make a couple changes:

## Critical

### 1. `masksToBounds = true` makes the weather button shadow invisible

`OBAKit/Mapping/MapViewController.swift:286-290`

The button sets `layer.masksToBounds = true` and then configures shadow properties. Shadows render outside the layer bounds, so `masksToBounds` clips them entirely — those four lines of shadow configuration are dead code.

```swift
button.layer.masksToBounds = true          // clips everything outside bounds
button.layer.shadowColor = UIColor.black.cgColor  // never visible
button.layer.shadowOpacity = 0.15                 // never visible
button.layer.shadowOffset = CGSize(width: 0, height: 2)
button.layer.shadowRadius = 6
```

Since the blur subview already has its own `layer.cornerRadius = 12` and `layer.masksToBounds = true` to handle clipping, remove `masksToBounds = true` from the button itself. The button's `cornerRadius` will still round the background color, and the shadow will render correctly.

### 2. `loadWeather()` accesses UIKit off the main actor in the error path

`OBAKit/Mapping/MapViewController.swift:427`

The success path correctly uses `await MainActor.run { ... }`, but the `catch` block sets `weatherButton.isHidden = true` without any main actor dispatch. This is a UIKit mutation off the main thread — undefined behavior that produces runtime warnings in Xcode 15+.

Additionally, this `Task` captures `self` strongly, unlike every other `Task` in this file which uses `Task { @MainActor [weak self] in ... }`. Match the existing convention:

```swift
Task { @MainActor [weak self] in
    do {
        self.forecast = try await apiService.getWeather()
    } catch {
        weatherButton.isHidden = true
        Logger.error(error.localizedDescription)
    }
}
```

### 3. `.replacingOccurrences(of: "%@")` instead of `String(format:)` for localized string

`OBAKit/Mapping/MapViewController.swift:399-403`

```swift
.replacingOccurrences(of: "%@", with: formattedTemp)
```

Every other localized format string in this PR correctly uses `String(format:, ...)`. This one uses `.replacingOccurrences`, which will break if a translator includes a literal `%@` in their translation, or if the format string evolves to have multiple arguments. Use `String(format:)` for consistency and correctness.

### 4. Fade-in animation fires on every `forecast` update, not just the first

`OBAKit/Mapping/MapViewController.swift:407-410`

```swift
weatherButton.alpha = 0
UIView.animate(withDuration: 0.3, delay: 0, options: .curveEaseOut) {
    self.weatherButton.alpha = 1
}
```

`loadWeather()` is called from `viewDidAppear`, which fires every time the user switches back to the map tab. Each time a new forecast arrives, the button snaps to `alpha = 0` and fades back in — a visible flicker. Guard the animation on the initial appearance:

```swift
let wasHidden = weatherButton.isHidden
weatherButton.isHidden = false
if wasHidden {
    weatherButton.alpha = 0
    UIView.animate(withDuration: 0.3, delay: 0, options: .curveEaseOut) {
        self.weatherButton.alpha = 1
    }
}
```

This also eliminates the duplicate `weatherButton.isHidden = false` on lines 397 and 405.

## Important

### 5. Move logic out of MapViewController

`OBAKit/Mapping/MapViewController.swift`

MapViewController is already a large file, and this PR adds ~200 lines of weather-specific logic directly to it: icon mapping, temperature/wind formatting, button configuration, card presentation/dismissal, and animation. This makes MapViewController harder to maintain and the weather logic impossible to unit test.

Please extract as much as possible:

- **Icon mapping and condition text** belong on the `WeatherForecast` model (or a helper alongside it in OBAKitCore). Both the button and the card need these, and having them on the model eliminates the duplication noted in item #6 below.
- **Formatting helpers** (`formatTemp`, `formatWindSpeed`, `formatTime`) could live on `WeatherForecast` or a small `WeatherFormatter` helper — they're pure functions with no view dependencies.
- **Card presentation/dismissal** could be extracted into a lightweight coordinator or presenter class, but at minimum the formatting and data-mapping logic should not live in the view controller.

The goal is for MapViewController to do little more than call `showWeatherCard(for: forecast)` and `dismissWeatherCard()`, with the rest living in testable, focused types.

### 6. Duplicate weather-icon-to-SF-Symbol mapping in two places

`OBAKit/Mapping/MapViewController.swift:433-448` and `OBAKit/Weather/WeatherCardView.swift:387-398`

`systemImageName(for:)` and `iconConfigMap` both maintain the identical 10-entry mapping of weather icon strings to SF Symbol names. If a new icon is added or a name changes, both must be updated in lockstep. Extract this to a shared location — either a static method on `WeatherForecast` or by having `MapViewController` reference `WeatherCardView`'s mapping.

### 6. `DateFormatter` created per cell render

`OBAKit/Weather/WeatherCardView.swift:335-341`

`formatTime()` allocates a new `DateFormatter` on every call. With 10 hourly cells, that's 10 allocations per render pass, and SwiftUI re-renders frequently (trait changes, rotation, dark mode). Use a `static let` cached formatter instead.

### 7. No max width constraint on the weather card

`OBAKit/Mapping/MapViewController.swift:326-329`

The card width is `view.widthAnchor - 32`, which on iPad or Stage Manager could be 1000+ points. The card's horizontal layout will stretch with vast whitespace. Add a max width:

```swift
hostingController.view.widthAnchor.constraint(lessThanOrEqualToConstant: 500),
hostingController.view.widthAnchor.constraint(
    equalTo: view.widthAnchor, constant: -32
).priority(.defaultHigh),
```

### 8. Use `removeChildController(_:)` instead of manual child removal

`OBAKit/Mapping/MapViewController.swift:359-361`

The `dismissWeatherCard` completion block manually calls the three-step child removal dance:

```swift
card.willMove(toParent: nil)
card.view.removeFromSuperview()
card.removeFromParent()
```

We already have a `UIViewController.removeChildController(_:)` extension (in `OBAKitCore/Extensions/UIKitExtensions.swift`) that does exactly this. Use it:

```swift
self.removeChildController(card)
```

### 9. Background dismiss tap gesture fires through the card

`OBAKit/Mapping/MapViewController.swift:346-349`

The dismiss gesture is added to the parent `view` with `cancelsTouchesInView = false`, and `gestureRecognizer(_:shouldRecognizeSimultaneouslyWith:)` unconditionally returns `true`. This means tapping the close button or hourly scroll fires both the SwiftUI `.onTapGesture` on the card and the UIKit gesture on the parent view. The `guard` in `dismissWeatherCard` prevents a crash, but the double-fire is wasteful and could cause animation glitches. Either add a delegate check to skip taps inside the card, or remove the SwiftUI `.onTapGesture` and rely on the UIKit gesture alone.

## Fit and Finish

### 10. "// THE FIX" dev comment

`OBAKit/Weather/WeatherCardView.swift:117`

```swift
// THE FIX: String(format:) maps the first %@ to High, and the second %@ to Low sequentially
```

This reads like a debugging note. Either remove it (the code is self-explanatory) or rewrite as a normal comment.

### 11. Redundant configuration properties re-set in `forecast` didSet

`OBAKit/Mapping/MapViewController.swift:393-394`

`imagePlacement` and `imagePadding` are set to the same values as the initial button configuration (lines 244-245). Since they never change, remove them from the `didSet`.

aaronbrethorst (changes requested):
Hey Ved, you're making great progress on the weather card overhaul. Nice work addressing the critical issues from the first round — the main-actor fix, the `wasHidden` fade guard, the `String(format:)` correction, and the max-width constraint all look correct. Creating `WeatherFormatter` in OBAKitCore was the right structural move. Before we can merge this, I need you to make a couple more changes:

## Important

### 1. `WeatherCardView` doesn't use `WeatherFormatter` — duplicates all its methods

`OBAKit/Weather/WeatherCardView.swift:331-350`

You created `WeatherFormatter` with `formatTime`, `formatTemp`, and `formatWindSpeed` — but `WeatherCardView` still has its own private copies of all three. The card's `formatTime` (line 331) also recreates a `DateFormatter` on every call (the original item #7 from last review), while `WeatherFormatter` correctly caches one.

Delete the three private methods from `WeatherCardView` and call `WeatherFormatter` instead:

```swift
// Before
private func formatTemp(_ temp: Double) -> String {
    MeasurementFormatter.unitlessConversion(temperature: temp, unit: .fahrenheit, to: locale)
}

// After — just call the shared helper
WeatherFormatter.formatTemp(temp, locale: locale)
```

Same for `formatTime` and `formatWindSpeed`.

### 2. Dead `systemImageName(for:)` method in MapViewController

`OBAKit/Mapping/MapViewController.swift:421-436`

The `forecast` didSet on line 382 correctly calls `WeatherFormatter.systemImageName(for:)`, but 40 lines below there's still a private `systemImageName(for:)` with the same mapping. This is now dead code — nothing calls it. Delete lines 421–436.

### 3. `WeatherFormatter` imports SwiftUI unnecessarily

`OBAKitCore/Weather/WeatherFormatter.swift:12`

`WeatherFormatter` only uses Foundation types (`DateFormatter`, `MeasurementFormatter`, `Locale`). Remove `import SwiftUI`. OBAKitCore is marked extension-safe, so keeping imports minimal is good practice.

### 4. Thread-unsafe locale mutation on shared `DateFormatter`

`OBAKitCore/Weather/WeatherFormatter.swift:23-26`

```swift
public static func formatTime(_ date: Date, locale: Locale) -> String {
    hourlyTimeFormatter.locale = locale  // mutates shared static
    return hourlyTimeFormatter.string(from: date)
}
```

`hourlyTimeFormatter` is a `static let`, but `formatTime` mutates its `.locale` on every call. If two callers on different threads pass different locales, this is a data race. Since this app likely only calls it from the main thread today, this isn't urgent — but the simplest fix is to just create a local formatter inside the method (matching the current `formatWindSpeed` pattern), or pass locale into the static initializer if it never changes.

### 5. Background dismiss gesture fires through the card

`OBAKit/Mapping/MapViewController.swift` (showWeather / gestureRecognizer delegate)

The dismiss tap gesture is added to the parent `view` with `cancelsTouchesInView = false`, and `gestureRecognizer(_:shouldRecognizeSimultaneouslyWith:)` unconditionally returns `true`. Tapping the close button fires both the SwiftUI button action and the UIKit gesture — `dismissWeatherCard` runs twice. The `guard` prevents a crash, but the double invocation is wasteful and can cause animation glitches.

The cleanest fix: in the gesture recognizer delegate, check whether the tap is inside the card and skip it:

```swift
public func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer,
                              shouldRecognizeSimultaneouslyWith other: UIGestureRecognizer) -> Bool {
    if gestureRecognizer == weatherCardDismissTapGesture,
       let cardView = weatherCardViewController?.view,
       cardView.point(inside: gestureRecognizer.location(in: cardView), with: nil) {
        return false
    }
    return true
}
```

## Fit and Finish

### 6. `conditionText(for:)` belongs in `WeatherFormatter`

`OBAKit/Weather/WeatherCardView.swift:352-373`

Since you've already extracted `systemImageName(for:)` and the formatting helpers into `WeatherFormatter`, `conditionText(for:)` is the one remaining weather data-mapping function still living in the view. Moving it to `WeatherFormatter` would make it reusable (the button could show condition text too) and keep `WeatherCardView` focused on layout.

aaronbrethorst (changes requested):
Hey Ved, you're making great progress on the weather card overhaul — the third iteration is looking much cleaner with `WeatherFormatter` properly integrated, the gesture delegate fix, and `conditionText` in its rightful home. Before we can merge this, I need you to make a couple more changes:

## Critical

### 1. Dismiss tap gesture has no delegate — taps inside the card dismiss it

`OBAKit/Mapping/MapViewController.swift:376-379`

The `weatherCardDismissTapGesture` is created and added to the view, but its `delegate` is never set to `self`. The guard in `gestureRecognizer(_:shouldRecognizeSimultaneouslyWith:)` at line 462 that checks whether the tap is inside the card will never be called. This means tapping anywhere — including inside the weather card — will trigger `dismissWeatherCard()`, making the hourly scroll view and other card interactions unusable.

```swift
let tapGesture = UITapGestureRecognizer(target: self, action: #selector(dismissWeatherCard))
tapGesture.cancelsTouchesInView = false
// Missing: tapGesture.delegate = self
```

Additionally, even with the delegate set, `shouldRecognizeSimultaneouslyWith` is the wrong method. It only controls whether *two* gestures can recognize at the same time — it does not prevent a *single* gesture from recognizing on its own. When the user taps inside the card and no competing gesture fires, the dismiss tap still recognizes because `shouldRecognizeSimultaneouslyWith` is never consulted.

The correct approach is to implement `gestureRecognizerShouldBegin(_:)` instead:

```swift
tapGesture.delegate = self

// Then in the delegate:
func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
    if gestureRecognizer == weatherCardDismissTapGesture,
       let cardView = weatherCardViewController?.view,
       cardView.point(inside: gestureRecognizer.location(in: cardView), with: nil) {
        return false  // Don't begin — tap was inside the card
    }
    return true
}
```

You can keep the existing `shouldRecognizeSimultaneouslyWith` returning `true` as-is (it's needed for the long-press pin gesture), and remove the weather-specific guard from it.

### 2. Same localization key used with two different format strings

`OBAKit/Mapping/MapViewController.swift:286` and `OBAKit/Mapping/MapViewController.swift:417`

Both sites use the key `"map_controller.show_weather_button"`, but with different `value:` strings:

```swift
// Line 286 — initial button setup
value: "Show Weather Forecast"

// Line 417 — forecast didSet update
value: "Show Weather Forecast for %@"
```

`OBALoc` wraps `NSLocalizedString`. When a translation exists for this key, both call sites get the *same* translated string. One expects a plain string, the other expects a format string with `%@`. If the translation contains `%@`, the first call site will display a literal `%@`. If it doesn't, the `String(format:)` call at line 416 will silently drop the temperature argument.

Use a distinct key for the format variant:

```swift
weatherButton.accessibilityLabel = String(
    format: OBALoc(
        "map_controller.show_weather_button_with_temp",
        value: "Show Weather Forecast, currently %@",
        comment: "Accessibility label for the weather button showing current temperature"
    ),
    formattedTemp
)
```

## Important

### 3. No tests for `WeatherFormatter`

`OBAKitCore/Weather/WeatherFormatter.swift`

`WeatherFormatter` contains five public static methods — all pure functions with no side effects — and zero tests. In particular:

- `systemImageName(for:)` is a dictionary lookup with a fallback default. If a mapping entry is deleted or misspelled during a future refactor, the UI silently shows the wrong icon. A test for each of the 10 known keys plus an unknown key is trivial to write and prevents regressions.
- `conditionText(for:)` has grouped switch cases (`"clear-day"` and `"clear-night"` both return `"Clear"`). These are easy to break accidentally.
- `formatTemp` and `formatWindSpeed` involve locale-dependent conversion. Testing with `en_US` vs `en_GB` would verify the unit system logic works correctly.

Please add a `WeatherFormatterTests.swift` in `OBAKitTests/`.

### 4. Duplicate icon-to-SF-Symbol mapping

`OBAKitCore/Weather/WeatherFormatter.swift:36-48` and `OBAKit/Weather/WeatherCardView.swift:339-350`

`WeatherFormatter.systemImageName(for:)` and `WeatherCardView.iconConfigMap` both maintain the identical 10-entry mapping from weather icon strings to SF Symbol names. If a new icon is added or a name changes, both must be updated.

The simplest fix: delete `iconConfigMap` from `WeatherCardView` and have it call `WeatherFormatter.systemImageName(for:)` for the symbol name. Keep only a color-only mapping in `WeatherCardView` since `Color` is a SwiftUI type that can't live in OBAKitCore:

```swift
private static let iconColors: [String: (primary: Color, secondary: Color)] = [
    "clear-day": (.yellow, .clear),
    "clear-night": (.gray, .yellow),
    "rain": (.gray, .blue),
    // ...
]

private func styledWeatherIcon(for weatherIcon: String) -> some View {
    let name = WeatherFormatter.systemImageName(for: weatherIcon)
    let colors = Self.iconColors[weatherIcon] ?? (.gray, .clear)
    // ...
}
```

This keeps the symbol name as a single source of truth in `WeatherFormatter`.

## Fit and Finish

### 5. `blurView.alpha = 1.0` is redundant

`OBAKit/Mapping/MapViewController.swift:302`

Views default to `alpha = 1.0`. This line does nothing — remove it.

### 6. Strong `self` capture in `showWeather` animation closures

`OBAKit/Mapping/MapViewController.swift:364-367`

```swift
UIView.animate(withDuration: 0.2) {
    self.weatherButton.alpha = 0
}
```

The other animation closure in `dismissWeatherCard` also captures `self` strongly (line 385), but that's less concerning because the completion block cleans up state. In `showWeather`, the strong capture is harmless in practice (the view controller owns the button) but using `[weak self]` would be more consistent with the `loadWeather` pattern established earlier.

Finally, please squash the commits in this branch into one. If you haven't done this before, I have a guide that tries to demystify this here: https://www.brethorsting.com/blog/2026/01/git-rebase-for-the-terrified/

Issue #400, opened 2021-06-23
Automate Production of TestFlight builds


aaronbrethorst:
I should be able to use Fastlane or something similar to create TestFlight builds directly from merges to the `master` branch here on GitHub. I imagine it'll look something like this:

1. A PR is opened.
2. The PR is reviewed and eventually merged to `master`.
3. The merge to `master` triggers a GitHub Action that builds and submits the app to Apple for publication on TestFlight.

More info:

* https://byteable.dev/migrate-fastlane-to-use-auth-keys-9113a666cd5d
* https://byteable.dev/build-and-deploy-your-ios-app-with-github-actions-and-fastlane-48c328cc5541

PR #1074, opened 2026-03-02
Add automated TestFlight deployment via GitHub Actions and Fastlane

diveshpatil9104:
#400
## Summary
- Add Fastlane configuration to build and upload the OneBusAway app to TestFlight
- Add a new GitHub Actions workflow that triggers on merges to `main`, automating the build-and-deploy pipeline
- Update `Gemfile` to include the `fastlane` gem

Currently, merges to `main` only trigger unit tests via `obakittests.yml`. There is no automated path to get builds into TestFlight for testers. This means every TestFlight build requires a manual archive and upload from a developer's machine.

With this change, the flow becomes:
```
PR merged to main -> GitHub Actions builds the app -> Uploaded to TestFlight automatically
```

## What Changed

| File | Change |
|---|---|
| `Gemfile` | Added `fastlane` gem (existing `jazzy` gem preserved) |
| `fastlane/Appfile` | New — app identifier and team ID |
| `fastlane/Fastfile` | New — `beta` lane that builds and uploads to TestFlight |
| `.github/workflows/testflight.yml` | New — deploy workflow triggered on push to `main` |

## How It Works
1. A PR is merged to `main`
2. The `TestFlight Deploy` workflow starts (alongside the existing `OBAKitTests` workflow)
3. The workflow installs the code signing certificate and provisioning profiles from GitHub Secrets
4. Fastlane generates the Xcode project via `scripts/generate_project`, builds the app, and uploads to TestFlight
5. The temporary keychain is cleaned up regardless of success or failure

## Design Decisions
- **App Store Connect API Key auth** (not session-based) — avoids 2FA issues on CI, as recommended by [Apple](https://developer.apple.com/documentation/appstoreconnectapi) and [Fastlane docs](https://docs.fastlane.tools/app-store-connect-api/)
- **`GITHUB_RUN_NUMBER` as build number** — auto-increments with each workflow run, no conflicts with local builds
- **`ENV.fetch()` instead of `ENV[]`** — raises a clear error if a secret is missing, so failures are visible rather than silent
- **Concurrency group** — prevents two deploys from running simultaneously if PRs merge in quick succession
- **`if: always()` on keychain cleanup** — ensures the temporary keychain is deleted even if the build fails
- **`timeout-minutes: 45`** — prevents stuck builds from burning CI minutes indefinitely
- **Xcode version verification** — `xcodebuild -version` after `xcode-select` to catch missing Xcode versions early with a clear log
- **Ruby gem caching** — caches `vendor/bundle` keyed on `Gemfile.lock` to avoid reinstalling fastlane on every run
- **Widget provisioning profile support** — installs the OBAWidget profile (`org.onebusaway.iphone.OBAWidget`) if provided via `WIDGET_PROVISIONING_PROFILE_BASE64`, skips gracefully if using a wildcard profile
- **Matches existing CI setup** — same macOS 15 runner, Xcode 26.2, SPM cache strategy, and `generate_project` step as `obakittests.yml`
- **Branch confirmed as `main`** — verified via `git remote show origin` (the issue description mentions `master` but the actual repo default branch is `main`)

## Admin Setup Required 
This PR will not function until the following GitHub Secrets are added at **Settings > Secrets and variables > Actions**:

### Required Secrets

| Secret | Source |
|---|---|
| `APP_STORE_CONNECT_KEY_ID` | App Store Connect > Users and Access > Integrations > API |
| `APP_STORE_CONNECT_ISSUER_ID` | Same page as above |
| `APP_STORE_CONNECT_KEY_CONTENT` | Base64-encoded `.p8` file from above |
| `CERTIFICATE_BASE64` | Base64-encoded `.p12` distribution certificate |
| `CERTIFICATE_PASSWORD` | Password used when exporting the `.p12` |
| `KEYCHAIN_PASSWORD` | Any strong password (for the temporary CI keychain) |
| `PROVISIONING_PROFILE_BASE64` | Base64-encoded App Store distribution `.mobileprovision` for `org.onebusaway.iphone` |

### Optional Secret (depends on profile type)

| Secret | When Needed |
|---|---|
| `WIDGET_PROVISIONING_PROFILE_BASE64` | Required if using explicit (non-wildcard) profiles. This covers the widget bundle ID `org.onebusaway.iphone.OBAWidget`. Not needed if the main profile is a wildcard that covers both bundle IDs. |

### DIFF
--- a/.github/workflows/testflight.yml
+++ b/.github/workflows/testflight.yml
@@ -1,8 +1,10 @@
 name: TestFlight Deploy
 
 on:
-  push:
-    branches: [ main ]
+  workflow_run:
+    workflows: ["OBAKitTests"]
+    types: [completed]
+    branches: [main]
 
 # Prevent multiple deploys from running simultaneously.
 concurrency:
@@ -11,6 +13,7 @@ concurrency:
 
 jobs:
   deploy:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
     runs-on: macos-15
     timeout-minutes: 45
 
@@ -50,16 +53,16 @@ jobs:
         CERTIFICATE_PASSWORD: ${{ secrets.CERTIFICATE_PASSWORD }}
         KEYCHAIN_PASSWORD: ${{ secrets.KEYCHAIN_PASSWORD }}
       run: |
-        KEYCHAIN_PATH=$RUNNER_TEMP/app-signing.keychain-db
-        security create-keychain -p "$KEYCHAIN_PASSWORD" $KEYCHAIN_PATH
-        security set-keychain-settings -lut 21600 $KEYCHAIN_PATH
-        security unlock-keychain -p "$KEYCHAIN_PASSWORD" $KEYCHAIN_PATH
+        KEYCHAIN_PATH="$RUNNER_TEMP/app-signing.keychain-db"
+        security create-keychain -p "$KEYCHAIN_PASSWORD" "$KEYCHAIN_PATH"
+        security set-keychain-settings -lut 21600 "$KEYCHAIN_PATH"
+        security unlock-keychain -p "$KEYCHAIN_PASSWORD" "$KEYCHAIN_PATH"
 
-        CERT_PATH=$RUNNER_TEMP/certificate.p12
-        echo -n "$CERTIFICATE_BASE64" | base64 --decode -o $CERT_PATH
-        security import $CERT_PATH -P "$CERTIFICATE_PASSWORD" -A -t cert -f pkcs12 -k $KEYCHAIN_PATH
-        security set-key-partition-list -S apple-tool:,apple: -k "$KEYCHAIN_PASSWORD" $KEYCHAIN_PATH
-        security list-keychain -d user -s $KEYCHAIN_PATH
+        CERT_PATH="$RUNNER_TEMP/certificate.p12"
+        echo -n "$CERTIFICATE_BASE64" | base64 --decode -o "$CERT_PATH"
+        security import "$CERT_PATH" -P "$CERTIFICATE_PASSWORD" -A -t cert -f pkcs12 -k "$KEYCHAIN_PATH"
+        security set-key-partition-list -S apple-tool:,apple: -k "$KEYCHAIN_PASSWORD" "$KEYCHAIN_PATH"
+        security list-keychain -d user -s "$KEYCHAIN_PATH"
 
     - name: Install provisioning profiles
       env:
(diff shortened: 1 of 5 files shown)


aaronbrethorst (changes requested):
This is a well-thought-out CI/CD setup, Divesh — the PR description is exceptionally thorough, and the design decisions (API key auth, `ENV.fetch`, concurrency group, `if: always()` cleanup) are all the right calls. One thing to fix before merging.

## Critical Issues

None.

## Important Issues

1. **`Gemfile.lock` not updated** (`Gemfile`): The PR adds `gem 'fastlane', '~>2.225'` to the Gemfile, but the committed `Gemfile.lock` still only lists `jazzy` and its dependencies. Fastlane pulls in ~100+ transitive dependencies — without locking them, every CI run resolves versions fresh, making builds non-reproducible. A fastlane update that ships a breaking change could silently break deploys.

   Fix: run `bundle install` locally and commit the updated `Gemfile.lock` with this PR.

## Fit-and-Finish

None — the workflow, Fastfile, and Appfile are all clean and well-structured.

## Strengths

- `ENV.fetch()` over `ENV[]` is a great pattern — a missing secret fails fast with a clear error instead of producing a confusing `nil`-related failure later in the lane.
- The concurrency group with `cancel-in-progress: false` is the right choice for deploys — you don't want to cancel a half-uploaded build, and queueing ensures sequential delivery.
- The widget provisioning profile is handled gracefully — optional install that skips for wildcard profiles.
- `sh("../scripts/generate_project", "OneBusAway")` correctly accounts for Fastlane's working directory being `fastlane/`.
- The Fastlane `api_key` is stored in the lane context, so `build_app` can use it for automatic code signing without explicit passthrough.
- Mirroring the existing CI setup (same runner, Xcode version, SPM cache strategy) keeps the infrastructure consistent.

## Recommended Action

1. Run `bundle install` and commit the updated `Gemfile.lock`

aaronbrethorst (changes requested):
Hey Divesh, you're making great progress on the TestFlight deployment pipeline — nice work getting the `Gemfile.lock` updated. Before we can merge this, I need you to make a couple more changes:

## Important

1. **TestFlight deploy should be gated on tests passing** (`.github/workflows/testflight.yml`): Right now, both workflows (`OBAKitTests` and `TestFlight Deploy`) trigger independently on push to `main`. If a broken commit lands, the TestFlight deploy will proceed even if the tests fail — shipping a potentially broken build to testers.

   Fix: switch to `workflow_run` so the deploy only triggers after the test workflow succeeds:

   ```yaml
   on:
     workflow_run:
       workflows: ["OBAKitTests"]
       types: [completed]
       branches: [main]
   ```

   And add a condition on the job:

   ```yaml
   jobs:
     deploy:
       if: ${{ github.event.workflow_run.conclusion == 'success' }}
   ```

2. **Unquoted shell variable expansions** (`.github/workflows/testflight.yml`): Several variables (`$KEYCHAIN_PATH`, `$CERT_PATH`, `$APP_PP_PATH`, `$WIDGET_PP_PATH`) are used unquoted in shell commands. While `$RUNNER_TEMP` is unlikely to contain spaces today, quoting these is standard shell hygiene and prevents subtle breakage if GitHub ever changes the temp path format. For example:

   ```bash
   # Before
   security create-keychain -p "$KEYCHAIN_PASSWORD" $KEYCHAIN_PATH
   # After
   security create-keychain -p "$KEYCHAIN_PASSWORD" "$KEYCHAIN_PATH"
   ```

   Apply the same quoting to all uses of `$KEYCHAIN_PATH`, `$CERT_PATH`, `$APP_PP_PATH`, and `$WIDGET_PP_PATH`.

PR #1084, opened 2026-03-08
SearchList SwiftUI Migration 

mosliem:
Migrates `SearchListViewController` and its UIKit/OBAList infrastructure to SwiftUI.

## Motivation

The previous implementation required every search row to be wrapped in a type-erased `AnyOBAListViewItem`, registered with a `UICollectionView`, and bridged through `OBAListViewSection` and `SearchPlacemarkViewModel` just to display results. The delegation chain between `SearchListViewController`, `SearchDelegate`, and `SearchInteractor` added indirection with no meaningful benefit.

The new implementation replaces all of this with plain Swift structs and a single `@Observable` class. The view accesses `SearchInteractor` directly — no delegate, no view model, no type erasure.

### Added
- **`SearchListModels.swift`** — New pure Swift models `SearchListSection`, `SearchListRow`, and `SearchListSectionKind` with no UIKit or OBAList imports.
- **`SearchListView.swift`** — New pure SwiftUI view that observes `SearchInteractor` directly via `@Observable`.

### Modified
- **`SearchInteractor.swift`** — Editted to be `@Observable`, publishing `sections: [SearchListSection]` directly instead of returning `[OBAListViewSection]`.

### Deleted
- `SearchListViewController.swift`
- `SearchPlacemarkViewModel.swift`
- `SearchPlacemarkTableCell.swift`
- `SearchPlacemarkContentConfiguration.swift`

## Vidoe

https://github.com/user-attachments/assets/7e6ce215-05df-422b-aa77-50f48f074caa

## Next 
- [ ] `SearchResultViewController` Migration.


aaronbrethorst (changes requested):
Hey Mohamed, nice work tackling the UIKit-to-SwiftUI migration of the search list. The overall direction is right -- replacing the delegate/`OBAListViewSection` machinery with `@Observable` and a plain SwiftUI `List` is a clear win for readability. There are a few issues to sort out before this can merge, though.

## Critical

1. **`MKPointOfInterestCategory.symbolName` does not exist -- this won't compile.**
   `SearchListRow.swift:94` calls `poi.symbolName`, but `MKPointOfInterestCategory` has no such property in any shipping iOS SDK. This is a build blocker. Either create an extension that maps known categories to SF Symbol names, or simplify to a default icon:
   ```swift
   static func systemImageForMapItem(_ mapItem: MKMapItem) -> Icon {
       return .system("mappin")
   }
   ```

2. **`@Observable` properties mutated off the main thread.**
   `SearchInteractor.swift:196-221` -- `MKLocalSearch.start` calls its completion on an arbitrary queue. Setting `placemarkSearchState` and `cachedPlacemarks` there will trigger SwiftUI view updates from a background thread, causing runtime warnings or crashes. Wrap the mutations:
   ```swift
   search.start { [weak self] response, error in
       DispatchQueue.main.async {
           // ... state mutations here
       }
   }
   ```
   (This also fixes the retain cycle in the current code, where the closure captures `self` strongly.)

3. **Bookmarks use wrong row kind.**
   `SearchInteractor.swift:118` -- bookmarks are created with `kind: .recentStop` instead of `kind: .bookmark`. The `.bookmark` case exists in the enum but is never used anywhere.

## Important

4. **Commented-out protocol method should be deleted, not commented.**
   `SearchInteractor.swift:19` -- `// func searchInteractorNewResultsAvailable(...)` is dead code from the old delegate flow. Delete it.

5. **`@Binding var sections` should be `let sections`.**
   `SearchListView.swift:30` -- `SearchListContentView` never writes to `sections`, so `@Binding` is misleading. Change to `let sections: [SearchListSection]` and update the call site on line 22.

6. **`placemarkSearchState.didSet` and `cachedPlacemarks.didSet` both call `searchModeObjects`, causing double rebuilds.**
   `SearchInteractor.swift:46-48, 58-62` -- On a successful search, `.success` is set (triggers rebuild), then `cachedPlacemarks` is set (triggers rebuild again). Consider setting `cachedPlacemarks` first, then `placemarkSearchState`, and removing the `didSet` from `cachedPlacemarks`. Or consolidate into a single rebuild trigger.

7. **Timer and search not cleaned up on deinit.**
   `SearchInteractor` has `debounceTimer` and `localSearch` that are never invalidated/cancelled in a `deinit`. Add:
   ```swift
   deinit {
       debounceTimer?.invalidate()
       localSearch?.cancel()
       placemarkSearchTask?.cancel()
   }
   ```
   Note: `placemarkSearchTask` (line 57) appears to be unused dead code -- if it's not needed, remove it.

8. **`SearchListSection` properties should be `let`, not `var`.**
   `SearchListSection.swift:12-16` -- `id`, `title`, and `content` are never mutated after construction. Mark them `let`.

## Fit and Finish

9. **Remove unused parameters.**
   - `SearchListRowView.swift:129` -- `badgedIcon` accepts `imagePadding` but never uses it.
   - `SearchListRowView.swift:146` -- `iconImage` accepts `padding` (default `0`) but it's never called with a non-zero value. The `size - padding` math is always `size - 0`.

10. **`isPlacemark` computed property is inconsistent with the rest of the view.**
    `SearchListRowView.swift:105-108` -- The rest of the file uses `switch row.kind`; this one check uses a separate boolean. Consider moving `isPlacemark` onto `SearchListRow.Kind` where it's co-located with the type, or just inline the check.

11. **Missing explicit `import UIKit`.**
    `SearchListRow.swift` uses `UIImage` but only imports `Foundation` and `MapKit`. It works because MapKit transitively imports UIKit on iOS, but an explicit import is clearer.

12. **Error rows have no retry mechanism.**
    When placemark search fails, the user sees a static error row with no way to retry. Consider adding a tap-to-retry action on error rows, or at minimum a subtitle like "Tap to retry."

## Type Design Note

The `SearchListRow` model stores display data (`title`, `subtitle`, `icon`) independently of `Kind`, which means they can get out of sync (and already have -- see item 3). Long-term, consider making `Kind` the single source of truth for display data, or adding factory methods that enforce consistency. This isn't blocking for this PR, but it's worth thinking about as you continue the migration.

---

**Recommendation:** Fix the critical items (1-3) and the important items (4-8), then this should be good to go. The fit-and-finish items can be addressed in this PR or a follow-up.

aaronbrethorst (changes requested):
Hey Mohamed, you're making great progress on the SwiftUI search migration. Most of the feedback from the first review has been addressed well -- the main thread dispatch, bookmark kind fix, deinit cleanup, double rebuild, and `@Binding` -> `let` are all solid. A few things still need attention before we can merge:

## Critical

1. **`MKPointOfInterestCategory.symbolName` still does not exist -- build blocker.**
   `SearchListRow.swift:101` calls `poi.symbolName`, but `MKPointOfInterestCategory` has no such property in any shipping iOS SDK (I checked the Apple docs). In your earlier comment you mentioned "there is already an extension in this file `MKPointOfInterestCategory.swift`" but that file does not exist anywhere in the repository. This needs to be resolved -- either create an extension that maps known categories to SF Symbol names, or simplify to a default icon:
   ```swift
   static func systemImageForMapItem(_ mapItem: MKMapItem) -> Icon {
       return .system("mappin")
   }
   ```

2. **Retain cycle in `mapPlacemarkItems` action closure.**
   `SearchInteractor.swift:350-352` -- the outer `compactMap` closure uses `[weak self]`, but after `guard let self`, the inner action closure captures that strong `self`:
   ```swift
   ) {
       self.delegate?.showMapItem(mapItem)  // strong capture
   }
   ```
   Since `SearchInteractor` holds `sections` -> `SearchListRow.action` -> back to `SearchInteractor`, this is a retain cycle. The other action closures in the file (recentStops, bookmarks, quickSearch, clearRecents) all correctly use `[weak self]` directly. Fix this one the same way:
   ```swift
   ) { [weak self] in
       self?.delegate?.showMapItem(mapItem)
   }
   ```

3. **MKLocalSearch cancellation errors are shown to the user as real errors.**
   `SearchInteractor.swift:197-202` -- when a new search starts, the previous `MKLocalSearch` is cancelled on line 178. The cancelled search's completion handler fires with an error, which gets assigned to `placemarkSearchState = .error(error)`. This causes a brief flash of an error message before the new results arrive. Filter out cancellation errors:
   ```swift
   if let error {
       let nsError = error as NSError
       if nsError.domain == NSURLErrorDomain && nsError.code == NSURLErrorCancelled {
           return  // intentional cancellation, ignore
       }
       self.placemarkSearchState = .error(error)
       return
   }
   ```

## Important

4. **`@Bindable` is unnecessary on `SearchListView.searchInteractor`.**
   `SearchListView.swift:14` -- `@Bindable` is for creating `$`-prefixed bindings. This view never creates any bindings; it only reads `searchInteractor.sections`. A plain `var searchInteractor: SearchInteractor` is sufficient for `@Observable` observation to work.

5. **UUID-based row identity causes SwiftUI to recreate all rows on every search update.**
   `SearchListRow.swift:40,59` -- every time `searchModeObjects` rebuilds sections, new `SearchListRow` instances get fresh `UUID()` values. SwiftUI uses `id` for diffing; new UUIDs mean every row is treated as brand-new, defeating diff optimizations and potentially causing visual glitches (lost scroll position, unnecessary animations). Derive `id` from content instead -- e.g., a combination of `kind` and `title`.

6. **`sections` should be `private(set)`.**
   `SearchInteractor.swift:58` -- `var sections: [SearchListSection] = []` is publicly writable. External code should not be able to overwrite the interactor's output. Mark it `private(set)`.

7. **`localSearch` reference assigned too late for correct cancellation.**
   `SearchInteractor.swift:227` -- `self.localSearch = search` is set inside the completion handler, not after calling `search.start`. Between starting the search and the callback firing, `localSearch` still references the previous search. Move the assignment to right after `search.start`:
   ```swift
   let search = MKLocalSearch(request: searchRequest)
   self.localSearch = search
   search.start { [weak self] response, error in
       // ...
   }
   ```

## Fit and Finish

8. **Variable shadowing in `clearRecentsRow` is confusing.**
   `SearchListRowView.swift:78` -- `if let icon = row.icon, case let .system(icon) = icon` rebinds `icon` three times in one expression. Consider using distinct names like `rowIcon` and `systemName`.

9. **`default` case in `SearchListRowView.body` silently catches new `Kind` cases.**
   `SearchListRowView.swift:22-23` -- if a new `Kind` variant is added, it will fall into `actionRow` without a compiler warning. Exhaustive matching would be safer.

10. **Empty title fallback produces blank rows.**
    `SearchInteractor.swift:346` -- `mapItem.name ?? mapItem.placemark.title ?? ""` falls back to an empty string. Consider filtering out map items with no name/title inside the `compactMap`, since it already supports returning `nil`.

11. **`guard recentStops.count > 0` should use `!isEmpty`.**
    `SearchInteractor.swift:109,125` -- Swift convention favors `.isEmpty` over `.count > 0`.

aaronbrethorst (approved):
Hey Mohamed, thanks for working through the feedback from the last round -- the vast majority of items are addressed cleanly. The `@Observable` threading fix, the bookmark kind correction, the `didSet` consolidation, exhaustive switch matching, `.isEmpty` usage, and deinit cleanup all look great. Solid follow-through.

No critical issues remaining. A few smaller items to consider.

## Critical Issues (0 found)

None.

## Important Issues (1 found)

None.

## Suggestions (2 found)

1. **Unused `imagePadding` / `padding` parameters in icon helpers.**
   `SearchListRowView.swift:124` -- `badgedIcon` accepts `imagePadding` and passes it to `iconImage` as `padding`, but the math `size - padding` just produces `size - 0` in most cases (only placemarks pass 14, which makes the icon 18pt inside a 32pt badge -- verify this is intentional). If it is intentional, consider renaming `imagePadding` to something like `iconInset` to make the purpose clearer.

2. **Error rows still have no retry mechanism.**
   When placemark search fails, the user sees a static error row with no way to retry. Consider adding a tap action that re-triggers `searchModeObjects`, or at minimum a subtitle like "Tap to retry." This can be a follow-up.

## Strengths

- Clean `@Observable` integration -- the removal of the delegate-based `searchInteractorNewResultsAvailable` callback in favor of direct observation is a real improvement.
- The `isUpdatingSections` / `defer` guard prevents re-entrant `didSet` rebuilds cleanly.
- `DispatchQueue.main.async` in the `MKLocalSearch` completion properly handles thread safety.
- `mapPlacemarkItems` now filters out items with empty titles via `compactMap`, which was a good catch.
- The deinit cleanup of timer and search is correct.
- Exhaustive switch matching in `SearchListRowView.body` means new `Kind` cases will produce a compiler error rather than silently falling through.
- Good use of OTPKit's `MKPointOfInterestCategory.symbolName` extension for contextual POI icons.

## Recommended Action

Consider the suggestions for a follow-up or this PR.

Issue #389, opened 2021-06-13
OBAListView still has identity problems, causing crashes


ualch9:
Despite #383, still has crashing or runtime warnings by UIKit.
 
Creating issue for tracking, draft pr soon.

ualch9:
Plan is to use synthesized Hashable instead of manually creating it. Requires reworking how `OBAListViewItem` does action callbacks, as callbacks cannot be Hashable.

ualch9:
After #391, all `applyData` calls result in the list view reloading all data. See attached video.

https://user-images.githubusercontent.com/22162410/123139623-44447d00-d40b-11eb-9b13-ef16cb967260.mov

ualch9:
Additionally, DiffableDataSource is flooding stdout with item index errors of header items (but not crashing):
```
2021-06-24 09:58:09.850758-0700 App[13091:4345657] [DiffableDataSource] Failed to find index of item OBAKit.AnyOBAListViewItem(_anyEquatable: OBAKit.(unknown context at $1080a4320).AnyEquatable(value: OBAKit.OBAListViewHeader(id: "agency_alerts_Washington State Ferries", title: Optional("Washington State Ferries"), isCollapsible: true), isEqual: (Function)), _id: (Function), _configuration: (Function), _separatorConfiguration: (Function), _onSelectAction: (Function), _onDeleteAction: (Function), _leadingContextualActions: (Function), _trailingContextualActions: (Function), _type: AnyHashable(OBAKit.OBAListViewHeader(id: "agency_alerts_Washington State Ferries", title: Optional("Washington State Ferries"), isCollapsible: true)))
```

ualch9:
Opened #420 and #421.

PR #1089, opened 2026-03-09
OBAListView - Implement smart diffable updates

### LINKED ISSUE #420: OBAListView always reloading data (instead of performing diff)
After #391, all `applyData` calls result in the list view reloading all data. See attached video.

https://user-images.githubusercontent.com/22162410/123139623-44447d00-d40b-11eb-9b13-ef16cb967260.mov

_Originally posted by @ualch9 in https://github.com/OneBusAway/OBAKit/issues/389#issuecomment-867014676_

mosliem:
Fixes #420 
Replaces the existing `applyData` implementation with a smart diffable update pipeline that avoids unnecessary full rebuilds by computing the minimal required update strategy before applying changes.

### `OBAListView.swift`

- Added `UpdateStrategy` enum — four modes: `.noChange`, `.contentUpdate`, `.sectionReload`, `.fullRebuild`
- Added `applyDataSmartly(to:animated:scrollBehavior:)` — orchestrates updates based on computed strategy
- Added `getUpdateStrategy(oldSections:newSections:)` — compares old and new snapshots to determine the cheapest valid update path
- Added `applyFullRebuild`, `applySectionReload`, `applyContentUpdate` — isolated update handlers per strategy
- Added `ScrollBehavior` and `ScrollState` — preserves scroll position across updates, with opt-in scroll-to-bottom behavior

### Current State
- Avoids unnecessary full rebuilds.
- Add scroll preserve and apply it in `StopViewController`, leading to auto scroll when loading more data.

### Video 

https://github.com/user-attachments/assets/94ec4a59-7f89-4050-904d-f823c956a2d2

### Feature Tasks
- [ ] Fully migrate to this component to SwiftUI
- [ ] Migrate all cells used by it to SwiftUI

## Release Notes

* **New Features**
  * List view now supports customizable scroll behavior when applying data—choose between preserving current scroll position or automatically scrolling to the bottom.

* **Bug Fixes**
  * Improved list update handling to more intelligently manage section and item changes, reducing unnecessary full rebuilds.

* **Tests**
  * Added comprehensive test coverage for list view update strategy logic.


aaronbrethorst (changes requested):
Hey Mohamed, thanks for taking the time to work on improving OBAListView's update performance. The idea of computing a minimal update strategy instead of always doing a full rebuild is exactly the right direction, and the scroll-state preservation is a nice touch for the "load more" UX. Before we can merge this, I will need you to make a couple changes:

## Critical

1. **Race condition in `applyFullRebuild` due to `DispatchQueue.main.async`** (`OBAListView.swift:484-490`): The section-level snapshot is applied synchronously, but the section snapshots (which carry the actual items) are deferred to the next run-loop tick via `DispatchQueue.main.async`. This creates a window where the collection view has section headers registered but zero items inside them. If `applyData()` is called again during that window (e.g., from a timer or refresh), `getUpdateStrategy` will diff against the already-updated `lastDataSourceSnapshot` and attempt to reconfigure or reload items that don't exist yet in the diffable data source. The original code applied everything synchronously, and `applySectionReload` and `applyContentUpdate` in this PR also apply section snapshots synchronously — the `async` dispatch here is inconsistent and should be removed.

2. **`getUpdateStrategy` misclassifies partial item-set changes as `.contentUpdate`** (`OBAListView.swift:447-452`): The `isDisjoint(with:)` check only returns `.sectionReload` when the old and new item ID sets share *zero* members. If even one item ID is shared (e.g., "load more" adds 9 new arrivals but 1 existing one is still present), the strategy falls through to `.contentUpdate`. `applyContentUpdate` then calls `reconfigureItems` only on the matching items and silently ignores the new ones, leaving stale data in the list. The check should be `oldIDs != newIDs` (or similar) rather than `isDisjoint`, so that any change in the item ID set triggers a section reload.

## Important

3. **`applyDataSmartly` should not be `public`** (`OBAListView.swift:383`): This method bypasses all the guards in `applyData` — snapshot validation, empty-section filtering, and empty-data-source handling. Any external caller using `applyDataSmartly` directly can inject duplicate IDs or empty sections without any defensive checks. It should be `internal` (or `private`) since `applyData` is the intended public entry point.

4. **`ScrollState.restore` captures `listView` strongly in async closures** (`OBAListView.swift:649-659`): Both the `.preserve` and `.scrollToBottom` branches dispatch async closures that capture the `listView` parameter by strong reference. If the view is deallocated before the closure fires (especially the `.scrollToBottom` case with its 200ms delay), the closure extends the collection view's lifetime unnecessarily and calls `setContentOffset` on a view that's no longer in any window hierarchy. Use `[weak listView]` capture lists in both closures.

5. **Hardcoded 0.2s delay for `scrollToBottom`** (`OBAListView.swift:656`): The `asyncAfter(deadline: .now() + 0.2)` assumes layout completes within 200ms. On slower devices or under memory pressure, `contentSize` may still reflect the pre-update value, causing the scroll to land in the wrong position or do nothing. Consider using `layoutIfNeeded()` before calculating the scroll offset, or using the `performBatchUpdates` completion handler instead of a time-based heuristic.

6. **Post-filter empty array not handled** (`OBAListView.swift:297-310`): The `sections.isEmpty` guard at line 297 runs against the *original* sections array from the data source. After `handleEmptySections` filters out headerless empty sections at line 303, the resulting array could be empty — but `handleEmptyDataSource()` was never called. This means `lastDataSourceSnapshot` and the diffable data source can get out of sync while the empty-data background view is incorrectly shown. Add an empty check after the filter step.

## Fit and Finish

7. **`UpdateStrategy` is wider than necessary** (`OBAListView.swift:371`): The enum is `internal` but only consumed by `private` methods within the same file. It could be `private` to limit unintended coupling as the module grows.

8. **`.scrollToBottom` fires even on `.noChange` strategy** (`OBAListView.swift:655-659`): The `.preserve` branch correctly guards against `.noChange` and `.fullRebuild`, but `.scrollToBottom` dispatches unconditionally. When nothing changed, scrolling the user to the bottom is incorrect.

9. **No tests for the new strategy logic**: The `getUpdateStrategy` method is the decision-making core of this feature, and wrong classifications silently produce stale UI. There are currently zero tests for `OBAListView` anywhere in the test suite. I'm not going to block the PR on this, but tests for `getUpdateStrategy` (at minimum) would significantly increase confidence in this change.

aaronbrethorst (changes requested):
Mohamed, this round is a real step up. You worked through every item from the last review carefully — the `DispatchQueue.main.async` race in `applyFullRebuild` is gone, the `isDisjoint` check became a proper `oldIDs != newIDs` set comparison, `applyDataSmartly` is now `private`, both scroll closures capture `[weak listView]`, the post-filter empty check is in, `.scrollToBottom` guards `.noChange`, and you added a focused 233-line test for `getUpdateStrategy` that even includes the load-more scenario. All 15 tests pass and the project builds clean. Thank you for that.

But fixing the classifier surfaced a deeper problem that the previous review didn't reach: the `.contentUpdate` *application* path is incompatible with how OBAListView identifies items, and it fails silently. I need that resolved before we merge.

## Critical Issues (1 found)

**1. `applyContentUpdate` silently drops the items that actually changed** [`OBAListView.swift:524-549`]

`AnyOBAListViewItem`'s identity is *full-value*: `hash(into:)` combines `_type` (the whole underlying model) and `==` compares every value via `AnyEquatable` (`OBAListViewItem.swift:183-189`). So to the diffable data source, an item with the same `id` but a changed value is a **different identity**. That's why the original `applyData` always re-applied a fresh per-section snapshot — it let the data source compute delete-old + insert-new, which renders correctly.

`applyContentUpdate` instead does this:

```swift
let existingIdentifiers = Set(snapshot.itemIdentifiers)        // OLD-valued items
let itemsToReconfigure = newSections.flatMap { $0.contents }
    .filter { existingIdentifiers.contains($0) }              // full-value containment
```

`getUpdateStrategy` routes to `.contentUpdate` precisely when the section structure and the set of item `id`s are unchanged but values differ — the value-changed case. But `existingIdentifiers.contains($0)` only returns `true` for items whose entire value is unchanged, so:

- An item that **changed** (e.g. an arrival ticking 3 min → 2 min) is filtered **out** and never reconfigured.
- An item that **didn't change** survives the filter and gets a no-op reconfigure.

Even if a changed item slipped through, `reconfigureItems` re-runs the cell provider against the identifier already in the snapshot (the old value), so it still couldn't render the new content. The result is stale cells with no log, no error, no fallback.

It's partially masked today: when *every* item in a section changes, `itemsToReconfigure` is empty and the guard at line 533 falls back to `applySectionReload` (which works). So this only corrupts the **partial-change** case — some items change, some don't — which is the hardest to spot in QA. That masking is itself a hazard: the fallback turns a deterministic bug into an intermittent one.

The core issue is that `reconfigureItems` is only sound when item identity is decoupled from item content, which isn't true here. My recommendation for this PR is to **fold `.contentUpdate` into `.sectionReload`** (apply fresh section snapshots and let the diffable data source compute the minimal diff — it already does this efficiently), or drop the `reconfigureItems` shortcut entirely. Truly decoupling identity from content (id-only `Hashable` on `AnyOBAListViewItem`) is a much larger change with ripples across the module, and not something I'd want to take on inside this PR. Worth noting separately: arrival rows mint a fresh `UUID()` per build (`StopArrivalItem.swift:25`), so `.contentUpdate` never even engages for the stop arrivals list — the headline use case — while remaining a live hazard for screens like Bookmarks.

## Important Issues (3 found)

**2. The `type(of:)` guard in `getUpdateStrategy` is dead code** [`OBAListView.swift:447-452`]

`OBAListViewSection.contents` is `[AnyOBAListViewItem]` (`OBAListViewSection.swift:37`), so `type(of: $0)` is the constant `AnyOBAListViewItem.self` for every element. The `zip(...).allSatisfy { $0 == $1 }` check can never be `false` and never forces a `.sectionReload`. It reads like a safety net for "the underlying cell type changed but the id stayed the same" — exactly the kind of identity mismatch that caused the historical #389 crashes — but it enforces nothing. Either compare the wrapped model type (the erased `_type` that `hash` already uses) or delete the check; a guard that silently does nothing is worse than no guard.

**3. `validateSnapshot` failure leaves the UI silently stale** [`OBAListView.swift:312-315`, `594-610`]

On duplicate section or item IDs, `applyData` logs and `return`s, discarding the whole new data set; the previously rendered content stays until some later refresh happens to be clean. `Logger.error` only writes to the OS log — no Sentry, no offending IDs surfaced (even though `validateSnapshot` knows exactly which ID collided). Consider returning the colliding IDs and either de-duping defensively so the user still sees current data, or routing through the Sentry path so the upstream data defect actually gets fixed.

**4. `applyData(animated:)` is now honored where it previously wasn't — please verify visually** [`OBAListView.swift:294`]

The old signature was `applyData(animated: Bool = true)`, but the old body always passed `animatingDifferences: false`, so animation never happened anywhere. Your version actually honors `animated`. Several call sites pass `animated: true` — `BookmarksViewController.swift:319`, `RecentStopsViewController.swift:67/112/148`, `StopViewController.swift:1185` — and those screens will now animate where they were previously static. That's plausibly an improvement, but it's an app-wide behavior change. Please confirm it looks right on Bookmarks in particular, which is historically the screen most sensitive to OBAListView identity churn.

## Suggestions (4 found)

**5. Add apply-level tests, not just classification tests.** The current suite is a clean test of `getUpdateStrategy` (the decision), but nothing exercises `applyContentUpdate`/`applySectionReload`/`applyFullRebuild` (the execution), which is where #1 lives. You can drive a real `OBAListView` through `applyData()` with a stub `obaDataSource` and assert on `diffableDataSource.snapshot()` (the same `snapshot()` `scrollTo(section:)` already reads). A test that applies a value change and asserts the cell/snapshot reflects the new value would have caught #1 immediately. While there, add a `collapseState` transition case (it's a real `.sectionReload` trigger and `OBAListViewSection.==` deliberately excludes `collapseState`) and a header-title-only change case.

**6. Tighten `ScrollState`.** Adding `layoutIfNeeded()` before computing the offset is a good improvement over last round. Two further nits: (a) have `ScrollState` capture a `weak var listView` in `init(from:)` so `restore` can't be handed a *different* view than the one it captured from — right now that's a caller convention, not a guarantee; (b) the `0.2`s `asyncAfter` is still a magic delay tied to layout settling — extract it to a named constant with a comment, or drive the bottom-scroll off the `apply(completion:)` callback instead of a timer. Also note the `.preserve` strategy guard is a deny-list, so any future `UpdateStrategy` case is preserve-eligible by omission.

**7. Make the internal fallbacks observable.** `applySectionReload`'s `pairs.isEmpty` → full rebuild (line 508) and `applyContentUpdate`'s empty → section reload (line 533) both fire silently. The first in particular signals that the live snapshot diverged from `lastDataSourceSnapshot` (the documented "source of truth") — exactly the desync this class exists to prevent. A `Logger.warn` before each fallback would turn an invisible recovery into a breadcrumb.

**8. Document the cross-file invariant and clean up naming.** The correctness of `.contentUpdate` (once fixed) and the section-ID-set comparison both depend on `AnyOBAListViewItem`'s full-value identity, which lives in a different file. One comment at `OBAListViewItem.swift:183` noting that `OBAListView.getUpdateStrategy` depends on full-value `hash`/`==` would warn the next person who's tempted to "optimize" the hash to id-only. Smaller items: `test_noChange_whenSectionsAndItemsAreIdentical` (`OBAListViewUpdateStrategyTests.swift:80`) actually asserts `.contentUpdate` — rename it; `.noChange` is only ever reachable for empty input, which `applyData` short-circuits before `applyDataSmartly`, so it's nearly vestigial (a `// bothEmpty` note or rename would help); and there's a stray blank line in `StopViewController.loadData` around line 579.

## Strengths

- Every item from the previous review was addressed, and addressed correctly — the async race, the set-equality fix, the access-control tightening, the weak captures, and the `.noChange` scroll guard are all exactly right.
- Replacing the `OBAListViewDelegate` + `didApplyData` callback + the `shouldScrollToBottom…` mutable flag with a single `ScrollBehavior` parameter is the cleanest part of this PR. The scroll intent now travels *with* the data application instead of through out-of-band view-controller state — that's a genuine reduction in coupling and kills a whole class of stale-flag races. The delegate removal is clean, with no orphaned references anywhere in the tree.
- `getUpdateStrategy` is a pure, deterministic function with good branch coverage, and grounding a test in the real load-more scenario was a nice touch.

## Recommended Action

1. Resolve Critical #1 — fold `.contentUpdate` into `.sectionReload` (or otherwise stop using `reconfigureItems` against full-value identity).
2. Address the Important items: remove/repair the dead `type(of:)` guard (#2), reconsider the silent `validateSnapshot` return (#3), and visually verify the `animated` behavior change on Bookmarks (#4).
3. Consider the suggestions — apply-level tests (#5) are the highest-value follow-on.
4. Re-request review after the fixes.

**Verdict: Request Changes.**

aaronbrethorst:
### Code review

Found 2 issues:

1. The new test file is written with XCTest + Nimble, but `main` has moved the suite to Swift Testing and Nimble is gone from the project entirely (no package pin, no other test file imports it), so `import Nimble` won't resolve. (CLAUDE.md says "`OBAKitTests/`: Unit tests for all frameworks. Written with **Swift Testing** (`@Suite` / `@Test` / `#expect`), not XCTest" and "Testing: Swift Testing (first-party; Nimble was removed)")

https://github.com/OneBusAway/onebusaway-ios/blob/c9a6b1977eedd984bc5b837601692734d5c4e94b/OBAKitTests/Controls/OBAListViewUpdateStrategyTests.swift#L10-L14

2. `isLoadingMore` is assigned but never read anywhere, and no call site passes the new `scrollBehavior:` argument — `dataDidReload()` still calls plain `listView.applyData(animated: false)`, which defaults to `.preserve`. Since this PR also deletes `didApplyData` (which did the `scrollTo(section: dataAttributionSection, at: .bottom)`), the "scroll to the bottom after tapping LOAD MORE" behavior is dropped and `ScrollBehavior.scrollToBottom` is dead code. (bug due to `OBAKit/Stops/StopViewController.swift` — the flag's only two occurrences are the declaration at L1011 and the write at L1028)

https://github.com/OneBusAway/onebusaway-ios/blob/c9a6b1977eedd984bc5b837601692734d5c4e94b/OBAKit/Stops/StopViewController.swift#L1027-L1029

aaronbrethorst (changes requested):
The June round addressed most of what I raised, and addressed it well — folding `.contentUpdate` into `.sectionReload` removed the case where value-only changes could be dropped, the dead `type(of:)` guard is gone, the full-rebuild fallback now logs before it fires, and the cross-file invariant is documented. Thank you for working through all of it.

Three things now stand between this and merging.

**The test file can't compile.** `OBAListViewUpdateStrategyTests.swift` imports `XCTest` and `Nimble`. `main` moved the suite to Swift Testing in July: Nimble has no package pin any more, and the only remaining `import XCTest` is `PolylinePerformanceTests`, which is a documented exception because `measure` has no Swift Testing equivalent. This needs rewriting as `@Suite` / `@Test` / `#expect`. That's on me for the timing — the migration landed after your last push.

**Scroll-to-bottom is regressed.** `isLoadingMore` is declared and assigned in `StopViewController` and never read. Nothing anywhere passes `scrollBehavior:` to `applyData` — `dataDidReload()` calls it with the `.preserve` default — so `ScrollBehavior.scrollToBottom` is unreachable. Combined with deleting `didApplyData`/`OBAListViewDelegate`, that means the LOAD MORE → scroll-to-bottom behavior is gone. The PR description says the scroll behavior is applied in `StopViewController`, so I think this is an oversight rather than a decision.

**The branch conflicts with `main`**, which has moved a fair amount under it — Bookmarks is SwiftUI now and no longer uses OBAListView at all, so point 4 from my June review is moot.

One design question I'd like your read on, because it may undercut the whole premise: `applySectionReload` calls `snapshot.reloadSections(...)` on every changed section before applying the per-section snapshots. Since `applySectionSnapshots` already computes a minimal diff, that first pass reloads every cell in every changed section — which looks a lot like the "everything reloads" behavior #420 set out to eliminate, on the exact path that matters (arrivals refreshing every 30s). It may also be why `ScrollState.preserve` is needed at all. Am I reading that right?

Smaller, for whenever you're back in here: `getUpdateStrategy` never compares `section.title`, so a header-title-only change classifies as `.noChange` and renders stale.

I know this has been open a long while. If you'd rather I take the rebase and the test rewrite off your hands, say so and I will.

Issue #455, opened 2021-08-06
Help users track their proximity to their final destination


aaronbrethorst:
This is a tracking task that needs to be fleshed out and split into multiple subtasks.

I've been asked by users on several occasions to let them better track when they are approaching their final destination. I can imagine this might be decomposed into a few different stories:

- As a user on an unfamiliar route, I want to feel confident that I'll get off at the right stop. I want to be able to tell OBA which stop I'm going to get off at, and then be able to see where I am in relation to that stop. I want to be able to look at it on a map to see how far away I am, and also see a list of how many stops are in between me and it.
- As a user who sometimes falls asleep when I'm on the bus, I want to know that I'll be woken up when I'm approaching my stop. I depend on OBA making my phone (or watch) buzz with a notification when I'm approaching my stop. 

Look into the location manager region monitoring capabilities to see how these might be done: https://developer.apple.com/reference/corelocation/cllocationmanager

This library would be useful and save me some trouble here: https://github.com/malcommac/SwiftLocation

barbeau:
FWIW, we have "destination reminders" on OBA Android that is effectively this feature:
https://github.com/OneBusAway/onebusaway-android/blob/master/DESTINATION_REMINDERS.md

It's actually a generalized port of some of our previous research at USF designed to help individuals with intellectual disabilities use public transportation by providing precise reminders of where to exit the transit vehicle. There is a description of the project with a YouTube video here:
https://www.usf.edu/news/2019/federally-funded-technology-eases-fears-for-riders-too-nervous-to-take-the-bus.aspx

Battery impact is definitely a concern when using GPS, but we ended up relying on GPS because of precision and accuracy requirements. We also designed an algorithm to detect the exact vehicle departure from the stop prior to your destination stop, which is patented by USF - but we convinced USF to allow us to use the algorithm and open-source the software within OBA. If you're interested in checking out the algorithm and the main navigation logic it's in this class:
https://github.com/OneBusAway/onebusaway-android/blob/master/onebusaway-android/src/main/java/org/onebusaway/android/nav/NavigationServiceProvider.java

We also collected a lot of GPS data and annotated each trip on the exact coordinates that should trigger our "Get ready" and "Exit the vehicle now" audio/vibration reminders and implemented unit tests based on this data (so we know if any changes to the nav logic breaks behavior on real data). That data might be useful for automated testing on iOS regardless of the exact implementation of the feature, and is all the `nav_trip*.csv` files here:
[...]

PR #1092, opened 2026-03-11
Add geofence region monitoring to LocationService

diveshpatil9104:
## Summary
- Extend `LocationManager` protocol with `startMonitoring(for:)`, `stopMonitoring(for:)`, and `monitoredRegions`
- Add `didEnterMonitoredRegion` and `monitoringDidFailFor` delegate methods to `LocationServiceDelegate`
- Implement `startMonitoringProximity`, `stopMonitoringProximity`, and `stopMonitoringAllProximityAlerts` on `LocationService`
- Handle `CLLocationManagerDelegate` region callbacks with non-circular region guard and error logging
- Update both mock location managers and `LocDelegate` for testability
- 14 unit tests covering region creation, removal, delegate callbacks, edge cases, and safety

## Context
**PR 2 of 4** for issue #455 (destination proximity alerts). Builds on the `ProximityAlert` model from PR 1.

| PR | Scope | Status |
|----|-------|--------|
| PR 1 | `ProximityAlert` model + `UserDataStore` persistence | Merged |
| **PR 2** | `LocationService` geofencing | **This PR** |
| PR 3 | `ProximityAlertManager` + local notifications | Next |
| PR 4 | `StopViewController` UI integration | Upcoming |

## Design Decisions
- **`String` identifier in delegate** — delegates receive `identifier: String`, not `CLRegion`. Keeps the protocol clean and avoids exposing CoreLocation types to consumers
- **`notifyOnEntry = true`, `notifyOnExit = false`** — we only care about arriving at the destination, not leaving
- **`stopMonitoringAllProximityAlerts` uses UUID check** — only removes regions with UUID identifiers, preserving any other monitored regions the system might have
- **`guard region is CLCircularRegion`** — ignores beacon regions or other non-circular region types
- **`Logger.error` in `monitoringDidFailFor`** — logs before notifying delegates, per Aaron's error visibility standards
- **Both mocks updated** — `LocationManagerMock` and `MockAuthorizedLocationManager` both implement region monitoring so existing tests remain unaffected

## Files Changed
| File | Change |
|------|--------|
| `OBAKitCore/Location/Location/LocationManagerProtocol.swift` | +3 protocol methods |
| `OBAKitCore/Location/Location/LocationService.swift` | +2 delegate methods, +2 notify helpers, +3 public methods, +2 CLLocationManager delegates |
| `OBAKitTests/Helpers/Mocks/LocationServiceMocks.swift` | Region monitoring in mock + 2 delegate capture properties |
| `OBAKitTests/Helpers/Mocks/MockAuthorizedLocationManager.swift` | Region monitoring in mock |
| `OBAKitTests/Location/LocationServiceRegionMonitoringTests.swift` | **New** — tests |

Closes part of #455

### DIFF
--- a/OBAKitCore/Location/Location/LocationManagerProtocol.swift
+++ b/OBAKitCore/Location/Location/LocationManagerProtocol.swift
@@ -43,6 +43,11 @@ public protocol LocationManager {
     var isHeadingAvailable: Bool { get }
     func startUpdatingHeading()
     func stopUpdatingHeading()
+
+    // MARK: - Region Monitoring
+    func startMonitoring(for region: CLRegion)
+    func stopMonitoring(for region: CLRegion)
+    var monitoredRegions: Set<CLRegion> { get }
 }
 
 extension CLLocationManager: LocationManager {
--- a/OBAKitCore/Location/Location/LocationService.swift
+++ b/OBAKitCore/Location/Location/LocationService.swift
@@ -16,6 +16,8 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
     @objc optional func locationService(_ service: LocationService, locationChanged location: CLLocation)
     @objc optional func locationService(_ service: LocationService, headingChanged heading: CLHeading?)
     @objc optional func locationService(_ service: LocationService, errorReceived error: Error)
+    @objc optional func locationService(_ service: LocationService, didEnterMonitoredRegion identifier: String)
+    @objc optional func locationService(_ service: LocationService, monitoringDidFailFor identifier: String?, error: Error)
 }
 
 @objc(OBALocationService) public class LocationService: NSObject, CLLocationManagerDelegate {
@@ -100,6 +102,18 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
     private func notifyDelegatesErrorReceived(_ error: Error) {
         for delegate in delegates.allObjects {
             delegate.locationService?(self, errorReceived: error)
+        }
+    }
+
+    private func notifyDelegatesDidEnterMonitoredRegion(_ identifier: String) {
+        for delegate in delegates.allObjects {
+            delegate.locationService?(self, didEnterMonitoredRegion: identifier)
+        }
+    }
+
+    private func notifyDelegatesMonitoringDidFail(_ identifier: String?, error: Error) {
+        for delegate in delegates.allObjects {
+            delegate.locationService?(self, monitoringDidFailFor: identifier, error: error)
         }
     }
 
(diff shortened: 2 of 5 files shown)


aaronbrethorst (changes requested):
Hey Divesh, thanks for the clean breakdown of the proximity alerts feature into well-scoped PRs — this layered approach makes review much easier. The code is well-structured, follows existing `LocationService` patterns closely, and the test suite covers the important happy paths and edge cases nicely (especially the non-UUID region preservation test). Before we can merge this, I will need you to make a couple changes:

## Important

1. **Missing authorization guard in `startMonitoringProximity`** (`LocationService.swift:272`): Every other state-mutating method in `LocationService` guards on capability before calling the underlying manager — `startUpdatingLocation` checks `isLocationUseAuthorized`, `startUpdatingHeading` checks `isHeadingAvailable`. `startMonitoringProximity` calls `locationManager.startMonitoring(for:)` unconditionally. At minimum, add a guard consistent with the existing pattern. Worth noting: region monitoring requires `.authorizedAlways` for background delivery (which is the whole point of proximity alerts). The current `isLocationUseAuthorized` check accepts `.authorizedWhenInUse`, so you may want a dedicated check or at least a comment documenting that the caller (PR 3's `ProximityAlertManager`) is responsible for ensuring Always authorization.

2. **UUID heuristic in `stopMonitoringAllProximityAlerts` is collision-prone** (`LocationService.swift:295`): Using `UUID(uuidString: region.identifier) != nil` as a proxy for "this is our proximity region" will silently remove any region whose identifier happens to be a valid UUID — including regions registered by third-party SDKs or other app components. The same issue affects `didEnterRegion` (`LocationService.swift:300-302`), which fires `notifyDelegatesDidEnterMonitoredRegion` for *any* `CLCircularRegion`, not just ones created by proximity alerts. A simple fix for both: use a namespaced prefix like `"oba.proximity.\(alert.id.uuidString)"` in `startMonitoringProximity`, then filter on that prefix in `stopMonitoringAllProximityAlerts` and `didEnterRegion`. This makes ownership unambiguous and both filtering paths consistent.

## Fit and Finish

3. **`stopMonitoringProximity` silent no-op** (`LocationService.swift:284-289`): When the region isn't found, the method returns silently. The `monitoringDidFailFor` callback uses `Logger.error` for its failure path — consider adding a `Logger.warn` here for consistency. This would help diagnose cases where an upstream silent failure (e.g., the authorization issue above) prevented the region from ever being registered.

4. **A few test gaps worth closing**: (a) calling `startMonitoringProximity` twice with the same alert to document the deduplication contract, (b) calling `stopMonitoringAllProximityAlerts` with no monitored regions (trivial empty-input boundary test), and (c) verifying multiple delegates both receive the same callback (since the multicast `NSHashTable` pattern is the whole point of the delegate infrastructure).

diveshpatil9104:
Thanks Aaron! All four issues fixes:

- **Authorization guard:** Added guard `isLocationUseAuthorized` to `startMonitoringProximity`, matching the existing pattern. Added a doc comment noting that the caller (PR 3's `ProximityAlertManager`) will be responsible for ensuring Always authorization for background delivery.
- **Namespaced prefix:** Replaced UUID heuristic with `"oba.proximity."` prefix. Both `stopMonitoringAllProximityAlerts` and `didEnterRegion` now filter on this prefix, making ownership unambiguous and both paths consistent.
- **`Logger.warn`:** Added to `stopMonitoringProximity` when no matching region is found.
- **New tests (5 added, 19 total):** duplicate alert deduplication, empty `stopAll` boundary, multiple delegate multicast, unauthorized monitoring guard, and non-prefixed circular region filtering in `didEnterRegion`.

aaronbrethorst (approved):
Hi Divesh — thanks for turning the round-one feedback around so cleanly. I re-checked all four items from my last pass and they're each resolved exactly as I'd hoped: the `isLocationUseAuthorized` guard now matches the rest of `LocationService`, the UUID heuristic is gone in favor of the `oba.proximity.` namespace (applied consistently in *both* the stop-all filter and `didEnterRegion`), the `Logger.warn` no-op is in place, and the test suite grew to cover dedup, the empty-input boundary, the multicast delegate, the unauthorized guard, and prefix filtering. I built the project and ran `LocationServiceRegionMonitoringTests` — **17/17 green**.

The layered PR-of-4 structure continues to pay off here. This change is small, correct, and reads like the code already around it.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None that block this PR. Worth being explicit about why, since a deep pass surfaced a few things that *sound* alarming but are correctly scoped out:

- **Silent no-op on unauthorized `startMonitoringProximity`.** This is the `guard … else { return }` pattern I explicitly asked you to match last round, and it's consistent with `startUpdatingLocation`. The legitimate concern — that region monitoring (unlike location updates) has no self-healing re-arm, so a dropped alert stays dead — is real, but it's the **caller's** job to handle, and the caller (`ProximityAlertManager`) lands in PR 3. Your doc comment already names that responsibility. Tracking it for PR 3 rather than blocking here.
- **`.authorizedAlways` vs `.authorizedWhenInUse`.** Same story: the guard accepts when-in-use, the doc comment notes background delivery needs Always, and PR 3 owns enforcing that. Documented, deferred.

## Suggestions (4 found — all deferred to follow-up issue)

1. **`monitoringDidFailFor` doesn't filter by prefix** [`LocationService.swift:314`]. `didEnterRegion` filters on `oba.proximity.`, but the failure callback forwards *every* region's failure to proximity delegates. Symmetry would be nice, and classifying `CLError.Code` would let PR 3 distinguish "too many regions" from "denied." → PR 3.
2. **20-region CoreLocation cap** is unhandled; the 21st region fails asynchronously with a generic error. → PR 3.
3. **`radiusMeters` isn't validated** against `maximumRegionMonitoringDistance`; `CLCircularRegion` silently clamps. Belongs on the `ProximityAlert` model. → follow-up.
4. **Non-circular but prefixed region in `didEnterRegion`** could log instead of silently returning — it's a "can't happen" bug indicator. Minor. → follow-up.

## Strengths

- Prefix-namespacing for region ownership is the right call, and the **`test_stopMonitoringAllProximityAlerts_preservesNonPrefixedRegions`** test proves teardown never clobbers another subsystem's regions — exactly the invariant that matters for a shared `LocationService`.
- The mocks back `monitoredRegions` with a real `Set<CLRegion>`, so the duplicate-replace test genuinely exercises `CLRegion`'s identifier-based equality rather than faking it.
- `monitoringDidFailFor` is tested with both a region and a `nil` region — the nil case is commonly missed.
- The `CLBeaconRegion`-with-matching-prefix test is a sharp touch: it proves the `region is CLCircularRegion` check does real work, not just the prefix check.

## Recommended Action

Nothing required for this PR. The follow-up suggestions are captured in #1150 for the PR 3 / model work.

Issue #444, opened 2021-08-06
De-emphasize past stops polyline on TripViewController's map


ualch9:
Similar to how Link's on-board display shows next stops (image below)
Similar: onebusaway/onebusaway-iphone#319

![Screen Shot 2021-08-06 at 9 47 10 AM](https://user-images.githubusercontent.com/22162410/128544572-01d6c048-65da-4164-8e1a-f030528accbb.png)
Image from @aaronbrethorst lol

Borisserz:
This landed on `main` in #1305 (`TripRouteOverlays`: spent half gray/thinner, ahead in route color; no invented split when progress is unknown). The PR used `Refs` rather than `Closes`, so GitHub left this open.

Okay to close?

Borisserz:
Closing: delivered in #1305 (`TripRouteOverlays` spent/ahead split). Left open only because that PR used `Refs`.

Borisserz:
Closing: delivered in #1305 (`TripRouteOverlays` spent/ahead split). Left open only because that PR used `Refs`.

Borisserz:
@aaronbrethorst could you close this? Delivered in #1305; I don't have close permission.

PR #1093, opened 2026-03-11
Enhance trip view with stop progress tracking and past/future distinction

### LINKED ISSUE #571: Improve vehicle route transfer experience
The current screen looks unfinished and buggy.

Letter

diveshpatil9104:
Closes #571 , #444

## Summary

- Add **TripProgressView** to the floating panel header showing "Stop X of Y" with a progress bar and ETA to the user's destination stop
- **Dim past stops** in the stop list using `secondaryLabel` text color so users instantly see where the vehicle is relative to their stop
- **Color-code the spine line** in `TripSegmentView` — gray for past stops, brand color for upcoming stops

## Screenshots

| Before | After |
|--------|-------|
|  |  |

### `OBAKit/Trip/TripStopListItem.swift`
- Added `TripStopTemporalState` enum (`.past`, `.current`, `.future`)
- Extended `TripStopViewModel` with `temporalState`, `stopIndex`, `totalStops`; init accepts `closestStopIndex` to compute temporal state
- Updated `hash(into:)` and `==` to include all new fields
- Added `applyTemporalStateStyling()` — past stops use `secondaryLabel` text color (not alpha dimming, which washes out in light mode), current stop uses bold `.headline` font, user destination uses bold
- Added "Passed stop" VoiceOver accessibility label for past stops
- Reset font, textColor, and `temporalState` in `prepareForReuse()`

### `OBAKit/Trip/TripSegmentView.swift`
- Added `temporalState` property and `pastLineColor` computed property (`secondaryLabel` at 40% opacity)
- `drawRegularTripSegment()` draws gray spine line and squircle stroke for past stops, brand color for current/future
- Top line: gray for `.past`/`.current`, brand for `.future`
- Bottom line and squircle stroke: gray for `.past`, brand for `.current`/`.future`

### `OBAKit/Trip/TripFloatingPanelController.swift`
- Added `TripProgressView` (stop count label, ETA label, `UIProgressView`) — `isAccessibilityElement = true` with combined label
- Added `tripProgressWrapper` with readable content guide constraints
- Extracted `closestStopIndex(in:)` helper to avoid duplicate computation between list section and progress view
- Updated `tripStopListSection()` to compute and pass `closestStopIndex`, `stopIndex`, `totalStops` to each `TripStopViewModel`
- Progress view hidden in `.tip` panel state, shown in `.half`/`.full`
- ETA uses `arrivalDeparture.arrivalDepartureMinutes` — the same real-time predicted value displayed in the "Xm" badge at the top of the card
- Three ETA states: "~X min to your stop", "Arriving now", "Passed your stop"
- Progress view hidden entirely when no real-time data (`closestStopID` unavailable)

## Edge cases handled

- No real-time data → all stops show as future, progress bar hidden
- Bus already passed user's stop → shows "Passed your stop"
- Bus at user's stop → shows "Arriving now"
- Bus late (predicted arrival is negative or zero minutes) → shows "Arriving now"
- No `arrivalDeparture` (opened from vehicle view) → shows "Stop X of Y" only, no ETA
- Empty stop times / missing `closestStopID` → progress bar hidden

## Summary by CodeRabbit

* **New Features**
  * Added a trip progress header showing the current stop count and ETA.
  * Added a visual progress bar to track journey progress.
  * Enhanced stop displays with distinct past, current, and upcoming states.
  * Improved accessibility labels for progress details and passed stops.

* **Tests**
  * Added coverage for progress calculations, ETA messaging, and stop-state classification.

### DIFF
--- a/OBAKit/Trip/TripFloatingPanelController.swift
+++ b/OBAKit/Trip/TripFloatingPanelController.swift
@@ -383,12 +383,13 @@ class TripFloatingPanelController: UIViewController,
         )
 
         // Compute ETA to user's destination stop, if available.
+        // Uses arrivalDepartureMinutes from ArrivalDeparture which already
+        // accounts for predicted/real-time arrival — same value as the "Xm" badge.
         var etaText: String?
         if let arrivalDeparture = tripConvertible?.arrivalDeparture {
             let userStopIndex = tripDetails.stopTimes.firstIndex { $0.stopID == arrivalDeparture.stopID }
             if let userStopIndex = userStopIndex {
                 if userStopIndex < currentIndex {
-                    // Vehicle has already passed the user's stop.
                     etaText = OBALoc(
                         "trip_progress.passed_your_stop",
                         value: "Passed your stop",
@@ -401,17 +402,15 @@ class TripFloatingPanelController: UIViewController,
                         comment: "Shown when the vehicle is arriving at the user's stop"
                     )
                 } else {
-                    let userStopTime = tripDetails.stopTimes[userStopIndex]
-                    let secondsUntilArrival = userStopTime.arrivalDate.timeIntervalSince(Date())
-                    let minutesUntilArrival = Int(ceil(secondsUntilArrival / 60.0))
-                    if minutesUntilArrival > 0 {
+                    let minutes = arrivalDeparture.arrivalDepartureMinutes
+                    if minutes > 0 {
                         etaText = String(
                             format: OBALoc(
                                 "trip_progress.eta_to_stop_fmt",
                                 value: "~%d min to your stop",
                                 comment: "Estimated time of arrival to the user's stop. e.g. ~8 min to your stop"
                             ),
-                            minutesUntilArrival
+                            minutes
                         )
                     } else {
                         etaText = OBALoc(
--- a/OBAKit/Trip/TripFloatingPanelController.swift
+++ b/OBAKit/Trip/TripFloatingPanelController.swift
@@ -15,7 +15,7 @@ import FloatingPanel
 
 /// Testable value type encapsulating the "Stop X of Y" header data.
 /// Returns `nil` when progress data is unavailable and the header should be hidden.
-struct TripProgressViewModel {
+nonisolated struct TripProgressViewModel {
     let stopCountText: String
     let etaText: String?
     let progress: Float
@@ -54,12 +54,19 @@ struct TripProgressViewModel {
                     ),
                     minutes
                 )
-            } else {
+            } else if userStopIndex - closestStopIndex == 1 {
+                // The prediction is non-positive but the vehicle is at the adjacent
+                // stop, so "Arriving now" is justified by position alone.
                 etaText = OBALoc(
                     "trip_progress.arriving_now",
                     value: "Arriving now",
                     comment: "Shown when the vehicle is arriving at the user's stop"
                 )
+            } else {
+                // The prediction is stale (e.g. the vehicle is running late) and it is
+                // still several stops away — omit the ETA rather than show a
+                // misleading "Arriving now".
+                etaText = nil
             }
         } else {
             etaText = nil
--- a/OBAKit/Trip/TripFloatingPanelController.swift
+++ b/OBAKit/Trip/TripFloatingPanelController.swift
@@ -24,6 +24,7 @@ class TripFloatingPanelController: UIViewController,
         didSet {
             if isLoadedAndOnScreen {
                 listView.applyData(animated: false)
+                updateProgressView()
             }
         }
     }
@@ -90,6 +91,7 @@ class TripFloatingPanelController: UIViewController,
         super.viewWillAppear(animated)
 
         listView.applyData(animated: false)
+        updateProgressView()
     }
 
     // MARK: - Public Methods
@@ -122,15 +124,18 @@ class TripFloatingPanelController: UIViewController,
         switch panelState {
         case .tip:
             self.separatorView.isHidden = true
+            self.tripProgressWrapper.isHidden = true
             self.stopArrivalView.normalInfoStack.forEach { $0.isHidden = isAccessibility }
             self.stopArrivalView.accessibilityInfoStack.forEach { $0.isHidden = true }
         case .half:
             self.separatorView.isHidden = false
+            self.tripProgressWrapper.isHidden = false
             self.stopArrivalView.normalInfoStack.forEach { $0.isHidden = isAccessibility }
             self.stopArrivalView.accessibilityInfoStack.forEach { $0.isHidden = true }
             self.stopArrivalView.accessibilityMinimalInfoStack.forEach { $0.isHidden = !isAccessibility }
         case .full:
             self.separatorView.isHidden = false
+            self.tripProgressWrapper.isHidden = false
             self.stopArrivalView.normalInfoStack.forEach { $0.isHidden = isAccessibility }
             self.stopArrivalView.accessibilityInfoStack.forEach { $0.isHidden = !isAccessibility }
         case .hidden: break
@@ -186,7 +191,32 @@ class TripFloatingPanelController: UIViewController,
         return view
     }()
 
-    private lazy var outerStack = UIStackView.verticalStack(arrangedSubviews: [topPaddingView, stopArrivalWrapper, separatorView, listView])
+    private lazy var tripProgressView: TripProgressView = {
+        let view = TripProgressView.autolayoutNew()
+        view.isHidden = true
+        return view
+    }()
+
+    private lazy var tripProgressWrapper: UIView = {
+        let wrapper = tripProgressView.embedInWrapperView(setConstraints: false)
+        NSLayoutConstraint.activate([
+            tripProgressView.topAnchor.constraint(equalTo: wrapper.topAnchor, constant: ThemeMetrics.compactPadding),
+            tripProgressView.leadingAnchor.constraint(equalTo: wrapper.readableContentGuide.leadingAnchor),
+            tripProgressView.trailingAnchor.constraint(equalTo: wrapper.readableContentGuide.trailingAnchor),
+            tripProgressView.bottomAnchor.constraint(equalTo: wrapper.bottomAnchor, constant: -ThemeMetrics.compactPadding)
+        ])
+        return wrapper
+    }()
+
+    private lazy var outerStack = UIStackView.verticalStack(arrangedSubviews: [topPaddingView, stopArrivalWrapper, tripProgressWrapper, separatorView, listView])
+
+    // MARK: - Helpers
+
+    /// Returns the index of the vehicle's closest stop within the trip's stop list, or nil if unavailable.
+    private func closestStopIndex(in tripDetails: TripDetails) -> Int? {
+        guard let closestStopID = tripDetails.status?.closestStopID else { return nil }
+        return tripDetails.stopTimes.firstIndex { $0.stopID == closestStopID }
+    }
 
     // MARK: - ListAdapterDataSource (Data Loading)
     func canCollapseSection(_ listView: OBAListView, section: OBAListViewSection) -> Bool {
@@ -293,9 +323,21 @@ class TripFloatingPanelController: UIViewController,
             contents.append(AdjacentTripItem(order: .previous, trip: previousTrip, onSelectAction: selectAdjacentTripAction).typeErased)
         }
 
+        let closestStopIdx = closestStopIndex(in: tripDetails)
+
         // Stop times
         let selectTripStopAction: OBAListViewAction<TripStopViewModel> = { [unowned self] item in self.onSelectTripStop(item) }
-        let stopTimes: [AnyOBAListViewItem] = tripDetails.stopTimes.map { TripStopViewModel(stopTime: $0, arrivalDeparture: arrivalDeparture, onSelectAction: selectTripStopAction).typeErased }
+        let totalStops = tripDetails.stopTimes.count
+        let stopTimes: [AnyOBAListViewItem] = tripDetails.stopTimes.enumerated().map { index, stopTime in
+            TripStopViewModel(
+                stopTime: stopTime,
+                arrivalDeparture: arrivalDeparture,
+                stopIndex: index,
+                totalStops: totalStops,
+                closestStopIndex: closestStopIdx,
+                onSelectAction: selectTripStopAction
+            ).typeErased
+        }
         contents.append(contentsOf: stopTimes)
 
         // Next trip, if any.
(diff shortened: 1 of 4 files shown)


aaronbrethorst (changes requested):
Hi Divesh — this is a genuinely nice piece of product thinking. "Stop X of Y" with an ETA to *my* stop, dimmed past stops, and a color-coded spine line together make the trip card much more legible at a glance, and the screenshots sell it well. I especially like that you reached for `secondaryLabel` color instead of alpha dimming (which really does wash out in light mode), wired the new state into `Hashable`/`Equatable` *and* `prepareForReuse`, and thought through the edge cases in the description. The `closestStopIndex(in:)` helper to avoid recomputing the index in two places is exactly the right instinct.

There are a few things I'd like buttoned up before this merges — one is a visible layout gap in a common case, and the other two are correctness/reuse hazards. None are hard to fix.

## Critical Issues (0 found)

None — nothing here crashes or corrupts data.

## Important Issues (3 found)

- **Empty padded gap when there's no real-time data, in `.half`/`.full` panel states.** [OBAKit/Trip/TripFloatingPanelController.swift:197-208, :120-141]
  Two independent `isHidden` flags drive the same visual element: `tripProgressWrapper.isHidden` is set purely by panel state in `configureView` (`false` in `.half`/`.full`), while the inner `tripProgressView.isHidden` is set by data availability in `updateProgressView` (`true` when `closestStopIndex` is nil). The inner view is pinned to the wrapper's top/bottom with `compactPadding` (lines 200-203). A view with `isHidden = true` still participates fully in Auto Layout and still reports its intrinsic size — `isHidden` only collapses layout for *arranged subviews of a `UIStackView`*, and here the hidden view is a plain subview of the wrapper, not an arranged subview. So when there's no real-time `closestStopID` (the common scheduled-only case), the wrapper stays visible and retains the inner view's height plus `2 × compactPadding`, leaving a blank padded band above the separator. Fix by giving visibility a single source of truth — e.g. have `updateProgressView` set `tripProgressWrapper.isHidden` (combined with panel state), or make `tripProgressView` an arranged subview of a stack so its `isHidden` collapses the layout.

- **Headline font can leak onto reused cells.** [OBAKit/Trip/TripStopListItem.swift:286-304]
  `applyTemporalStateStyling` bumps `titleLabel.font` to `.headline` for `.current` and for user-destination `.future` stops, but the `.past` and non-destination `.future` branches never set the font back to `.body`. The only reset lives in `prepareForReuse` (line 173), and in-place reconfiguration of an already-visible cell re-invokes `apply(_:)` *without* recycling — so a cell that transitions current → past as the vehicle advances can keep its `.headline` font while only the color changes. Make the font assignment exhaustive inside `applyTemporalStateStyling` (explicitly set `.body` in the branches that currently leave it untouched) rather than relying on `prepareForReuse`.

- **Two sources of truth for "current".** [OBAKit/Trip/TripStopListItem.swift:100-101 vs TripFloatingPanelController.swift:214]
  `isCurrentVehicleLocation` is derived from `arrivalDeparture?.tripStatus?.closestStopID`, while `temporalState == .current` is derived from `tripDetails.status?.closestStopID` (via `closestStopIndex`). Those are two different status objects. If they ever disagree — a stale `arrivalDeparture`, a trip with a duplicated stop ID — a row can be styled `.current` while `isCurrentVehicleLocation == false`, or vice versa, and the cell styling and the segment-view glyph would split. Derive both from the same status object (or assert they agree at construction) so the type can't represent the inconsistent state.

## Suggestions (6 found)

- **No tests for the new logic, and the most interesting logic isn't testable as written.** The temporal-state classification in `TripStopViewModel` (TripStopListItem.swift:104-115) is a pure function and trivially unit-testable today — the boundary cases (`closestStopIndex == nil` → all `.future`; `closestStopIndex == 0`; `closestStopIndex == totalStops - 1`; the `<` / `==` / `>` transitions) are exactly where an off-by-one would silently mis-color the whole list. More importantly, the richest logic — the "Stop X of Y" string, the progress fraction, and the three-way ETA branch — is trapped inside the private `updateProgressView()` on a `UIViewController` (TripFloatingPanelController.swift:348-426). Consider extracting it into a small value-type view model with a failable initializer (`init?` returning `nil` to mean "hide"), taking primitives rather than reaching through a live, time-relative `ArrivalDeparture`. That one refactor makes the off-by-one in "Stop X of Y" (1-based display: index 0 → "Stop 1 of N", last index → "Stop N of N"), the `minutes <= 0` → "Arriving now" fallback, and the guard ladder all cheap, deterministic tests. Note: the existing `trip_details_1_18196913.json` fixture has `status: null`, so you'll want a status-bearing variant (or a `dictionaryToModel`-built `TripStatus`) to cover the `.current` path end-to-end.

- **`temporalState` lacks `didSet { setNeedsDisplay() }`.** [OBAKit/Trip/TripSegmentView.swift:49]
  Unlike `image` (lines 40-44), setting `temporalState` doesn't invalidate the display. It's safe today only by call-order luck — `apply(...)` sets `temporalState` (TripStopListItem.swift:255) right before `setDestinationStatus(...)` (line 256), and the latter calls `setNeedsDisplay()`. Reorder those two lines and the temporal coloring silently stops updating. Add `didSet { setNeedsDisplay() }` so the redraw is enforced by the type, not by caller discipline.

- **`stopIndex` / `totalStops` are stored but never read** beyond feeding `temporalState` and bloating `==`/`hash`. They look like speculative state for a future "stop N of M" row label. If there's no near-term consumer, consider dropping them from the stored surface (keep them as `init` parameters that produce `temporalState`); if you keep position, a small validating `struct StopPosition { let index, total: Int }` would close the "loose `Int` pair with no enforced `0 <= index < total`" gap.

- **`progressBar.setProgress(progress, animated: true)`** [TripFloatingPanelController.swift, `TripProgressView.configure`] animates from 0 the first time the view is unhidden, which reads as a sweep on first appearance. Consider `animated: false` for the initial configure.

- **Positional format specifiers.** The `trip_progress.stop_count_fmt` ("Stop %d of %d") uses bare `%d`; `%1$d`/`%2$d` is more localization-safe for languages that reorder the numbers. Matches existing patterns in the file, so optional.

- **Past + user-destination loses its emphasis.** In `applyTemporalStateStyling`, `.past` wins the switch, so a destination stop the vehicle has already passed drops the bold treatment. Likely fine (it's in the past), just flagging the intentional trade-off.

## Strengths

- Strong product instinct — the combined progress header + dimming + spine coloring genuinely improves legibility, and the before/after screenshots make the case.
- `TripStopTemporalState` is a clean, minimal, `Hashable` enum that makes illegal states unrepresentable — the strongest type in the PR.
- Temporal state is derived once in the initializer and frozen as `let`; no post-hoc mutation to drift out of sync.
- Cell-reuse awareness: the new field is reset in `prepareForReuse` and added to `Hashable`/`Equatable` so the list re-renders as the vehicle moves.
- Accessibility was considered throughout — the "Passed stop" VoiceOver label and the combined `accessibilityLabel` on `TripProgressView`.
- The `closestStopIndex(in:)` helper removes a duplicate computation between the list section and the progress view.

## Recommended Action

1. Fix the empty-gap layout issue — give the progress wrapper's visibility a single source of truth.
2. Make the font assignment in `applyTemporalStateStyling` exhaustive so `.headline` can't leak across reuse.
3. Reconcile the two sources of truth for "current" (same status object for both `isCurrentVehicleLocation` and `temporalState`).
4. Consider extracting the progress/ETA computation into a testable value type and adding unit tests for the temporal-state and ETA boundaries.

aaronbrethorst:
### Code review

Found 1 issue:

1. The new test file is written with XCTest + Nimble. CLAUDE.md says OBAKitTests is "Written with **Swift Testing** (`@Suite` / `@Test` / `#expect`), not XCTest" (the sole exception is `PolylinePerformanceTests`, which needs `measure`), and that Nimble was removed — it is no longer in `Apps/Shared/app_shared.yml`'s `packages:` nor in any `Package.resolved`, so `import Nimble` will not resolve on `main`. `OBAKitTests/project.yml` uses `sources: ["."]`, so this file is picked up automatically and will break the test target. It needs to be rewritten as a `@Suite(.serialized)` with `@Test` / `#expect`; the `beCloseTo` calls have a first-party replacement in `OBAKitTests/Helpers/Matchers/Approximately.swift`. Note that this branch's last CI run predates the Swift 6 / Swift Testing migration, so the green check on it is not meaningful.

https://github.com/OneBusAway/onebusaway-ios/blob/993078bc06b58e73eec1c38c140e61efde4c97de/OBAKitTests/ViewModels/TripStopListItemTests.swift#L9-L18

aaronbrethorst (changes requested):
All three Important items from June are genuinely fixed, and I verified each at head.

The padded-gap problem is gone — `tripProgressView.isHidden` is no longer set anywhere, `tripProgressWrapper` is the single source of truth, and the new `isPanelExpandedEnoughForProgress` flag is set in every `configureView(for:)` branch including `.hidden`, which used to fall through. `applyTemporalStateStyling` is now exhaustive across all three states for font *and* both text colors, so nothing leaks onto a reused cell. And `isCurrentVehicleLocation` is derived from the same `closestStopIndex` that produces `temporalState`, which kills the two-sources-of-truth problem — that one's actually a net win, since a trip opened from a vehicle never rendered the glyph before and now does.

The `TripProgressViewModel` extraction is exactly what I had in mind, and `TripStopTemporalState.classify(stopIndex:closestStopIndex:)` as a pure static is the right seam.

**Two things stop it building on current `main`.** The branch's only CI run is from June 9, which predates the Swift 6 migration entirely — treat that green check as meaningless.

**1. The new tests import a package that no longer exists.**

```swift
import XCTest
import Nimble
class TripStopTemporalStateTests: XCTestCase {
```

Nimble was removed from the project in late July; it's gone from `app_shared.yml` and `Package.resolved`. `OBAKitTests/project.yml` uses `sources: ["."]`, so the file is picked up automatically and the import fails. The suite needs rewriting in Swift Testing — `@Suite(.serialized)`, `@Test`, `#expect`. That matters more than usual here, because the tests are the thing that makes the extraction worthwhile, and right now the coverage doesn't actually exist.

**2. `TripStopTemporalState` needs `nonisolated`.** It's declared as a bare `enum … : Hashable` in OBAKit, which builds with `SWIFT_DEFAULT_ACTOR_ISOLATION: MainActor` — so the type is main-actor isolated. But it's consumed by `nonisolated struct TripStopViewModel` (I confirmed that's `nonisolated` on `main` at `OBAKit/Trip/TripStopListItem.swift:24`): `hasher.combine(temporalState)` inside a nonisolated `hash(into:)`, and `classify(...)` called from a nonisolated `init`. With the concurrency diagnostics escalated to errors, both are hard failures. One word fixes it:

```swift
nonisolated enum TripStopTemporalState: Hashable {
```

`TripProgressViewModel` has the same shape, so if any rewritten test touches it directly, either mark it `nonisolated` or make the test `@MainActor`.

Three notes for the rebase, none blocking:

- **`isCurrentVehicleLocation` now reads `tripDetails.status` rather than `arrivalDeparture.tripStatus`.** Those are different objects — if trip-details returns `status: null` while the arrivals payload carried a `tripStatus`, the glyph and its VoiceOver flag regress to absent. It's the unification I asked for and a net improvement, but worth a sanity check against a `status: null` fixture.
- **"Arriving now" when `arrivalDepartureMinutes <= 0`** fires even when the rider's stop is ten stops ahead — so a bus running twenty minutes late reads "Arriving now" at exactly the moment riders are staring at it. Your description lists this as intentional; I'd still reconsider gating it on proximity.
- **`.past` is conveyed by color alone** — `secondaryLabel` text plus a 40%-alpha spine, no weight or glyph change. VoiceOver is covered via "Passed stop" in `accessibilityValue` and `.current` gets a real `.headline`, so this only bites low-vision users not running VoiceOver. Narrow, but a small weight or opacity change on the label would close it.

Rebase, convert the suite, add the `nonisolated`, and get a fresh CI run — the design is in good shape and I'll turn the re-review around quickly.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Approved, and thank you for sticking with this one through several rounds.

Every blocker from both of my previous reviews is genuinely closed, not papered over. I checked each:

- The empty padded gap is gone for the right reason. `tripProgressWrapper` is now an arranged subview of `outerStack` and the single source of truth, so `isHidden` actually collapses the layout instead of leaving a hole.
- `applyTemporalStateStyling` is exhaustive across all three states for `font`, `textColor`, and `alpha`, so the headline font no longer leaks onto reused cells.
- "Current" has one source of truth: `isCurrentVehicleLocation` derives from the same `closestStopIndex` that feeds `classify(...)`.
- The test file is real Swift Testing now — `@Suite(.serialized)`, `@Test`, `#expect`/`#require`, no XCTest and no Nimble — and both `TripStopTemporalState` and `TripProgressViewModel` are `nonisolated`. CI is green at head on a branch that is current with `main`, so those two are verified rather than claimed.

You also went past what I asked on the "Arriving now" case: gating the `minutes <= 0` fallback on `userStopIndex - closestStopIndex == 1` is the right call, and covering it with three tests is better than the note deserved. The `alpha = 0.7` on `.past` addresses the color-alone concern properly.

I went looking for off-by-ones in `classify` and did not find any — `<`/`==`/`>` map cleanly to past/current/future, `closestStopIndex(in:)` and the `enumerated()` in `tripStopListSection` index into the same array, and both `status == nil` and "closest stop not in stopTimes" degrade to all-future with the header hidden. The spine coloring has no seam between adjacent rows either.

Follow-ups worth an issue rather than another round here:

- `tripConvertible.didSet` does not call `updateProgressView()`, so if `$tripConvertible` publishes after `$tripDetails` on a refresh, the ETA text is computed from the prior `ArrivalDeparture` for one cycle. Self-corrects on the next poll.
- `TripProgressView` has no accessibility-size handling: `stopCountLabel` and `etaLabel` sit side by side at `.caption1` with `etaLabel` at `.required` compression resistance, so "Stop 5 of 18" truncates at large Dynamic Type sizes. The rest of the panel does handle this.
- The status-less fixture test hardcodes `closestStopIndex: nil` rather than deriving it from the fixture, so it re-tests `classify(nil)` instead of the derivation path. Not vacuous, just narrower than its name suggests.

Approved on the merits — but one heads-up before this can land. #1241 just merged and it also touched `TripFloatingPanelController.swift` (it swapped the blink `DispatchQueue.main.throttle` call for a `Throttler` instance), so this branch has picked up merge conflicts against `main`. Please merge or rebase `main` in and resolve; the overlap should be small and localized to the property declarations near the top of the class and the blink call site.

Once it is mergeable it is good to go — no re-review needed unless resolving the conflict changes behavior.

aaronbrethorst:
Re-checked this before merging, since you pushed after my approval and I don't want an approval covering code I didn't read.

The post-approval push is the #1241 conflict resolution I asked for, and it's correct. The only delta from the SHA I approved is one hunk in `TripFloatingPanelController.swift`, swapping `DispatchQueue.main.throttle` for the `scrollBlinkThrottler = Throttler()` form — and it wasn't optional: `OBAKit/Extensions/DispatchExtensions.swift` was deleted on `main` by #1241, so the old call doesn't compile anymore. You took main's version verbatim, and the PR's net diff now contains no throttle lines at all, which is the tell that the rebase absorbed it cleanly rather than re-introducing a variant. `TripSegmentView.swift`, `TripStopListItem.swift`, and the test file are unchanged from what I approved.

Merging. Follow-ups from my earlier review still stand as separate work, plus two I'd add:

- The four new `OBALoc` keys (`trip_progress.stop_count_fmt`, `trip_progress.passed_your_stop`, `trip_progress.arriving_now`, `trip_progress.eta_to_stop_fmt`) and `trip_stop.passed_stop.accessibility_label` aren't in the catalogs yet — English is fine via the `value:` defaults, but worth a `scripts/extract_strings` pass.
- `TripProgressView` doesn't handle accessibility text sizes: `stopCountLabel` and `etaLabel` sit side by side at `.caption1` with `etaLabel` at `.required` compression resistance, so "Stop 5 of 18" truncates at large Dynamic Type while the rest of the panel adapts.

Minor: the PR description says `TripStopViewModel` gained `stopIndex`/`totalStops` stored properties, but they're init parameters — the code is right, the description drifted.

PR #1096, opened 2026-03-12
Adds a white-label configurable arrival/departure filter

### LINKED ISSUE #597: Select what arrival/departures to show: either scheduled, estimated or both
For white-label projects, the service owner may dislike to visualization of (possibly unreliable) scheduled arrivals that may highlight a bad service performance. I was asked to omit the visualization of scheduled arrivals unless there are particular conditions (e.g. there are no estimated arrivals).
So I would need options to enable/disable the visualization of scheduled arrivals or (even better) a callback function that could allow the selection of what to add to the list vs. to ignore it.

diveshpatil9104:
## Summary

Adds a white-label configurable arrival/departure filter that lets users choose which departures to display: **All**, **Real-Time Only**, or **Scheduled Only**. Closes #597.

**Problem:** Transit agencies using the white-label framework wanted to control whether scheduled (non-real-time) arrivals are shown. Some agencies have unreliable schedule data, and displaying "arrives 5 min late" highlights poor service performance. They needed the ability to hide scheduled arrivals when real-time data is unavailable.

**Solution:** A global `ArrivalDepartureFilter` setting accessible from both the **Settings page** and a **Departure Type menu** on each stop view. Brands can configure the default via `CoreAppConfig.defaultArrivalDepartureFilter`.

## Changes

- **`StopPreferences.swift`** — Added `ArrivalDepartureFilter` enum (`.all`, `.estimatedOnly`, `.scheduledOnly`)
- **`CoreAppConfig.swift`** — Added `defaultArrivalDepartureFilter` property for white-label brand default + shared UserDefaults key
- **`AppConfig.swift`** — Pass-through for new config parameter
- **`ArrivalDeparture.swift`** — Added `filter(by:)` extension on `[ArrivalDeparture]` using the `predicted` property
- **`StopViewController.swift`** — Reads global filter from UserDefaults, applies to both time-sorted and route-sorted views, adds Departure Type submenu
- **`SettingsViewController.swift`** — Added Arrival Display section with `AlertRow` for global filter preference
- **`ArrivalDepartureFilterTests.swift`** — Unit tests for filter logic, enum, empty arrays, complementary counts, and UserDefaults round-trips

## Screenshots / Demo

  https://github.com/user-attachments/assets/261529cf-f77f-411f-90e4-71b46ed31a47

## Test Plan

-  Unit tests pass (`ArrivalDepartureFilterTests`, `ArrivalDepartureFilterEnumTests`, `ArrivalDepartureFilterUserDefaultsTests`)
-  Open a stop with mixed real-time and scheduled arrivals
-  Tap **⋯ menu → Departure Type → Real-Time Only** — verify only real-time arrivals remain
-  Tap **Departure Type → Scheduled Only** — verify only scheduled arrivals remain
-  Tap **Departure Type → All Departures** — verify all arrivals return
-  Navigate to a different stop — verify same filter applies (global setting)
-  Open **Settings → Arrival Display → Show Departures** — verify AlertRow shows current selection
-  Change filter in Settings, go back to a stop — verify filter is applied
-  Switch sort mode (time vs route) — verify filter works in both modes
-  Kill and relaunch app — verify filter preference persists

## Summary by CodeRabbit

* **New Features**
  * Added a Departure Type filter with All, Real-Time Only, and Scheduled Only options in Settings and the stop screen menu.
  * Filter selections persist across sessions and use the app’s configured default when no preference is saved.
  * Stop results update according to the selected filter, with an explanatory empty state when no departures match.
  * Added localized labels and updated departure type and sorting icons.

* **Tests**
  * Added coverage for filtering, serialization, saved preferences, and default configuration propagation.


aaronbrethorst (changes requested):
Hey Divesh, great job putting together this filtering feature — the `ArrivalDepartureFilter` enum is clean, the filter logic on `[ArrivalDeparture]` is well-placed, and the test coverage with the complementary-counts test is a nice touch. The menu pattern is consistent with the existing sort menu too. Before we can merge this, I will need you to make a couple changes:

## Important

1. **`defaultArrivalDepartureFilter` is never used as the fallback** (`StopViewController.swift:507`, `SettingsViewController.swift:138`): You've wired `defaultArrivalDepartureFilter` through the entire `CoreAppConfig` → `AppConfig` chain, but both `activeArrivalDepartureFilter` and `currentArrivalFilterDisplayTitle()` fall back to `.all` when no UserDefaults value is saved — instead of using `application.config.defaultArrivalDepartureFilter`. This means white-label apps that set a non-`.all` default (the whole point of the config property) won't see their configured default take effect. Both fallbacks should read from the config:

   In `StopViewController.swift:507`: `return application.config.defaultArrivalDepartureFilter` instead of `return .all`

   In `SettingsViewController.swift:138`: `?? application.config.defaultArrivalDepartureFilter` instead of `?? .all`

## Fit and Finish

2. **Duplicate localization strings** (`StopViewController.swift:513-515`, `SettingsViewController.swift:143-149`): The display titles "All Departures", "Real-Time Only", and "Scheduled Only" are defined with different localization keys in both view controllers. Consider extracting them into a shared helper (or shared localization keys) so translators only need to translate each string once and the labels stay consistent.

3. **Missing test fixture in the repo?**: The test references `arrivals_and_departures_for_stop_1_10020_no_realtime.json`. I can see it exists in the fixtures directory, but `test_filterByEstimatedOnly_noRealtimeData_returnsEmpty` is named "returns empty" yet the assertion is `expect(result.count) == predictedCount` — which could be non-zero if the fixture has some predicted entries. The test name should match the actual assertion, or the fixture should truly have zero predicted entries for the test to be meaningful.

diveshpatil9104:
Thanks Aaron! All three changes are:

- **White-label default fallback:** Now used in both places.
- **Display titles:** Extracted into a shared `displayTitle` property on the enum.
- **Test assertion:** Updated to match its name.

aaronbrethorst (changes requested):
Thanks for turning this around so quickly, Divesh — all three items from the last pass are cleanly resolved. The display titles are now consolidated into a single `ArrivalDepartureFilter.displayTitle` source of truth, both the stop menu and Settings honor `application.defaultArrivalDepartureFilter` as the fallback, and the `..._noRealtimeData_returnsEmpty` test now asserts what its name promises (I confirmed the `1_10020_no_realtime` fixture really does have zero predicted entries, so that test is meaningful). I also verified the filter is applied *before* `filteringTerminalDuplicates()` in both sort paths, which is the correct order — dedup prefers predicted entries, so filtering first avoids dropping scheduled rows whose predicted twin was removed.

A couple of things still need attention before this can merge.

## Critical Issues (0 found)

None.

## Important Issues (2 found)

1. **The 7 new localization keys aren't in the string catalog** [`OBAKit/Strings/en.lproj/Localizable.strings`]. The keys introduced here — `arrival_departure_filter.all_departures` / `.real_time_only` / `.scheduled_only` (`ModelExtensions.swift:109-113`), `settings_controller.arrival_display_section.title` / `settings_controller.arrival_filter.title` / `.selector_title` (`SettingsViewController.swift:144-150`), and `stop_controller.arrival_filter.menu_title` (`StopViewController.swift:522`) — aren't present in `Localizable.strings`. Every other `OBALoc` key in the app is checked in there (e.g. the `settings_controller.map_section.*` keys), and that file is what feeds Transifex. English renders fine because `OBALoc` carries the inline `value:` default, but until these are extracted, no other locale can translate them. Please run `scripts/extract_strings` and commit the result.

2. **Settings persists the filter by round-tripping through the localized display string** [`SettingsViewController.swift:119-130`]. The save path reads the selected `AlertRow<String>` title and maps it back to an enum via `arrivalFilterFromDisplayTitle(_:)`, which returns `.all` on no match. It works today because the row's options are generated from the same `displayTitle` values, so a match is guaranteed — but it's fragile: a localization edit, a whitespace/normalization difference, or any future divergence between the stored title and `displayTitle` would silently revert a user's saved preference to "All Departures" with no signal. It's also inconsistent with the read path (`currentArrivalFilterDisplayTitle()`), which correctly falls back to the configured default rather than a hardcoded `.all`. The enum already has stable raw values that everywhere else persists correctly — please persist the `rawValue` here too rather than matching on the display string (and if the title-matching helper stays for any reason, its no-match branch should fall back to `application.defaultArrivalDepartureFilter`, not `.all`).

## Suggestions (2 found)

3. **Add a regression test for the config-default fallback** [`StopViewController.swift:502-508`, `SettingsViewController.swift:136-140`, `CoreApplication.swift:44-47`]. This is the exact class of bug we fixed last round, and it's the white-label feature's whole reason for existing, yet nothing asserts that an unset/invalid UserDefaults value resolves to `application.defaultArrivalDepartureFilter` (not `.all`) when an agency configures a non-`.all` default. `AppConfigTests.swift:30` already shows the construction pattern — a test that `AppConfig(... defaultArrivalDepartureFilter: .scheduledOnly)` propagates through `super.init` to `CoreApplication.defaultArrivalDepartureFilter`, plus a UserDefaults test that nil/garbage resolves to the configured default, would lock this down cheaply.

4. **Add the positive complement to the no-realtime test** [`ArrivalDepartureFilterTests.swift`]. You assert `.estimatedOnly` returns empty on the no-realtime fixture; a symmetric assertion that `.scheduledOnly` (and `.all`) return all 8 rows on that same fixture would pin the "no realtime → everything is scheduled" guarantee for free.

## Strengths

- All three prior review items addressed accurately, with `displayTitle` now a single source of truth.
- `filter(by:)` is well-placed on `Sequence where Element == ArrivalDeparture`, and the `predicted`-based logic is correct and readable.
- Filter-before-dedup ordering is right in both the time-sorted and route-sorted paths.
- OBAKitCore app-extension safety is preserved — the enum and `CoreAppConfig` additions stay Foundation-only, while the `displayTitle`/`OBALoc`/UIKit concerns live in OBAKit.
- Solid leaf-level test coverage: empty arrays, per-element predicate checks, the complementary-counts partition invariant, and the raw-value backward-compat contract are all pinned.

## Recommended Action

1. Run `scripts/extract_strings` and commit so the new keys reach the catalog (Important #1).
2. Persist the filter by `rawValue` in Settings instead of round-tripping the display string (Important #2).
3. Consider the two test additions above.
4. Re-request review after the fixes.

aaronbrethorst:
### Code review

Found 3 issues:

1. The filter is only wired into the legacy `StopViewController`, but `FeatureFlags.isNewStopPageEnabled` defaults to `true` (`userDefaults.object(forKey: useNewStopPageKey) as? Bool ?? true`), so `Router.makeStopController` sends users to `StopPageViewController` / `StopPageView` by default. `StopPageView.filteredDepartures` applies only `filter(preferences:)` + `filteringTerminalDuplicates()` and never calls `filter(by:)`, and the new page has no Departure Type menu. On a default install of either app the new Settings > Arrival Display row is user-visible but changes nothing on the stop screen; the filter only takes effect if the user turns the new stop page off in Settings, or on the transfer-context path. The new stop page landed on main in `aff9954a` (2026-07-10), after this branch was written.

https://github.com/OneBusAway/onebusaway-ios/blob/12ef19aa47582f095a48cdf922c80a4c0a65fd81/OBAKit/Stops/StopViewController.swift#L717-L721

2. `defaultArrivalDepartureFilter` can't actually be configured by a white-label brand. Both shipping apps build their config from the `@objc convenience init(appBundle:userDefaults:analytics:)` in `AppDelegate.m` (`Apps/OneBusAway/AppDelegate.m:50`, `Apps/KiedyBus/AppDelegate.m:37`), and that initializer neither takes the parameter nor derives it from the bundle — so it always falls through to `.all`. Unlike the other per-brand settings it sits next to (`fixedRegionName`, `fixedRegionOBABaseURL`, which are read from `OBAKitConfig` in `FoundationExtensions.swift`), there is no Info.plist key for it, so an agency can only set a non-`.all` default by editing framework source — which is the thing the white-label design exists to avoid.

https://github.com/OneBusAway/onebusaway-ios/blob/12ef19aa47582f095a48cdf922c80a4c0a65fd81/OBAKit/Orchestration/AppConfig.swift#L25-L43

3. The two new "resolves to default" tests are tautologies — they inline `ArrivalDepartureFilter(rawValue: saved ?? "") ?? .scheduledOnly` and assert it equals `.scheduledOnly`, which is true by construction regardless of the app's behavior. No `CoreAppConfig` is constructed with a configured default and neither `CoreApplication.defaultArrivalDepartureFilter` nor the fallbacks in `StopViewController`/`SettingsViewController` is exercised, so these would still pass if the config-default fallback regressed to a hardcoded `.all` — the exact bug from the first review round.

https://github.com/OneBusAway/onebusaway-ios/blob/12ef19aa47582f095a48cdf922c80a4c0a65fd81/OBAKitTests/Modeling/REST%20Model%20Service%20Tests/ArrivalDepartureFilterTests.swift#L204-L213

aaronbrethorst (changes requested):
The filtering core here is good, and you have closed most of what I raised in the earlier rounds. The config default is now a real fallback in both `StopViewController` and `SettingsViewController`, the duplicate strings are consolidated into `ArrivalDepartureFilter.displayTitle`, and replacing `arrivalFilterFromDisplayTitle(_:)` with an `AlertRow<ArrivalDepartureFilter>` that persists `rawValue` is exactly right — persisting through a localized display string was the worst thing in the previous version and it is properly gone.

Three things are blocking.

**1. CI is red on localization parity.** The build did not fail to compile; `LocalizationTests` "Every locale has same keys as english" failed with 12 issues, one per non-English locale. All eight new keys were added only to `en.lproj`:

```
arrival_departure_filter.{all_departures,real_time_only,scheduled_only}
settings_controller.arrival_display_section.title
settings_controller.arrival_filter.{title,selector_title}
stop_controller.arrival_filter.{menu_title,no_results}
```

Run `scripts/extract_strings` and propagate to all 13 locales. There is also a stray double blank line after `settings_controller.arrival_filter.title`, which suggests the file was hand-edited rather than regenerated.

**2. The filter does nothing on the default stop page.** This is the one that matters, and it is not your fault. `FeatureFlags.isNewStopPageEnabled` resolves to `?? true`, nothing registers it `false`, and `Router.makeStopController` is the only construction site for `StopViewController` — it branches to `StopPageViewController` for all normal navigation. The new page's `StopPageView.filteredDepartures` applies `filter(preferences:)` and `filteringTerminalDuplicates()` and never calls your `filter(by:)`, and it has no Departure Type menu.

So on a default install today, the Settings row you added is visible and inert. The flag flipped to default-on in `aff9954a` on 2026-07-10, well after this branch was designed, so this is main-side drift that landed on top of you. It still has to be resolved before this can ship: either wire `filter(by:)` into `StopPageView` and add the menu to the SwiftUI departure context menu, or explicitly scope this PR to the legacy page and say so in the body. I would rather you pick than have me pick for you — tell me which and I will help.

**3. The two "resolves to default" tests are tautologies.** `#expect(ArrivalDepartureFilter(rawValue: saved ?? "") ?? .scheduledOnly == .scheduledOnly)` parses as `(… ?? .scheduledOnly) == .scheduledOnly`, because nil-coalescing binds tighter than `==`. That is true by construction and would still pass if the production fallback regressed to a hardcoded `.all`. This is the same class of bug I flagged in round one, so it is worth fixing properly: build a `CoreAppConfig` with a configured default and assert through `CoreApplication.defaultArrivalDepartureFilter`.

One more that I would like understood even if you defer it: `defaultArrivalDepartureFilter` cannot actually be set by either shipping app. Both `AppDelegate.m` files construct config through the `@objc convenience init`, which does not take the parameter or read it from the bundle, and `CoreAppConfig`'s bundle convenience init hardcodes `.all`. The sibling white-label settings come from `OBAKitConfig` in `FoundationExtensions.swift`; there is no equivalent key here. As it stands the white-label knob that motivates the feature is unreachable for OneBusAway and KiedyBus alike.

Smaller notes:

- Time-sorted mode gets no "no results" empty state when the filter hides everything — only the route-sorted branch adds `arrival_filter_no_results`. Time-sorted always appends an empty section, so the user just sees blank space.
- The route-sorted empty-state message also fires when the user hid all routes via the route filter, telling them to change Departure Type when that is not the cause.
- The file-scope `// swiftlint:disable force_cast` is still there. CodeRabbit raised it too.

Happy to re-review as soon as the locales and the stop-page question are sorted.

aaronbrethorst:
### Code review

Found 2 issues:

1. A Departure Type change made in Settings never reaches a stop page that is already on screen. `StopViewModel` snapshots the app-wide value into a `@Published` property once, in `init`, and only ever mutates it from `updateArrivalDepartureFilter(_:)` — there is no `UserDefaults` observation and no re-read in `viewWillAppear`. Settings is presented modally from the More tab, so a stop page pushed in another tab survives the round trip: the rider changes "Show Departures", returns to the stop, and the list is unchanged while ⋯ → Departure Type still shows the old checkmark. Nothing signals it, and only re-navigating to the stop fixes it. This also departs from how the same screen consumes the sibling app-wide setting `stopUIReducedColors`, which `DepartureRowView` / `GroupedListView` read through `@AppStorage` so it does apply live (and `MapViewModel` subscribes to `UserDefaults.didChangeNotification` for the same reason). It also undercuts the comment on `CoreApplication.effectiveArrivalDepartureFilter` calling itself "the single resolution point for every consumer".

https://github.com/OneBusAway/onebusaway-ios/blob/8a24b736b0a404c8ce0e5b16e8f596167ec77b96/OBAKit/ViewModels/StopViewModel.swift#L209-L211

2. The Arrival Display section title was renamed to cover both arrivals and departures in all 12 non-English locales, but not in English. `8a24b73` updated every translation (`fr` → "Affichage des arrivées et départs", `es` → "Visualización de llegadas y salidas", `zh-Hant` → "到站與離站資訊顯示", etc.), while `en.lproj` and the inline `OBALoc(... value: "Arrival Display" ...)` in `SettingsViewController.swift` still say "Arrival Display". English users get a section titled for arrivals only, over a row and selector that both filter departures — the exact inconsistency the commit set out to remove — and the source string now diverges from every translation of it. `LocalizationTests` only checks key parity, so CI stays green.

https://github.com/OneBusAway/onebusaway-ios/blob/8a24b736b0a404c8ce0e5b16e8f596167ec77b96/OBAKit/Strings/en.lproj/Localizable.strings#L984-L986

aaronbrethorst (changes requested):
This is much closer, and I want to say plainly that every blocker from all three earlier rounds is genuinely resolved — I checked each rather than assuming, and nothing was quietly dropped across five months:

- **March:** the config default is a real fallback via `CoreApplication.effectiveArrivalDepartureFilter`; titles are consolidated into `ArrivalDepartureFilter.displayTitle`; the estimated-only-with-no-realtime test asserts `.isEmpty`.
- **June:** all 10 keys are in all 13 catalogs with real translations rather than English copies; the display-string round-trip is gone in favor of `AlertRow<ArrivalDepartureFilter>` persisting `rawValue`; the config-default regression test and the positive complements are there.
- **August:** the big one — the filter now actually works on the default stop page. `StopPageView` applies `filter(by:)` before `filteringTerminalDuplicates()`, `StopPageToolbar` has a Departure Type submenu, and `StopPageViewController.filterMenu()` nests one for the pushed presentation. The knob is reachable too: `Bundle.defaultArrivalDepartureFilter` reads `OBAKitConfig.DefaultArrivalDepartureFilter` and both `appBundle` inits pass it. The tautological fallback tests are replaced with ones that go through `app.effectiveArrivalDepartureFilter` and would fail on a regression. The `force_cast` disable is gone.

Graceful degradation checks out: neither app sets `DefaultArrivalDepartureFilter`, absent or unrecognized values resolve to `.all`, and both config types default the parameter. Unconfigured agencies get exactly today's behavior.

Two things left, both small. I'm asking for one more round rather than merging because one is a real bug and the other ships wrong copy in the app's primary language.

**1. Changing the setting doesn't reach a stop page that's already open.** `StopViewModel.swift:210` seeds `arrivalDepartureFilter` once in `init` from `environment.effectiveArrivalDepartureFilter`, and the only other writer is the in-page menu at `:512-513`. Nothing observes `UserDefaults`, and `viewWillAppear` doesn't re-read. Settings is presented modally from the More tab, so a stop page sitting in another tab survives that round trip showing both stale results *and* a stale checkmark in its own menu.

I considered whether that's just how settings behave here, and it isn't — this codebase propagates live in two different ways: `@AppStorage(UserDefaultsStore.stopUIReducedColorsKey)` in `TripCardView.swift:33`, and a `UserDefaults.didChangeNotification` subscription in `MapViewModel.swift:114`. Either shape works; the notification one is probably the closer fit since you need to re-run the filter, not just re-render.

**2. The English section title didn't get the rename.** Your last commit renamed `settings_controller.arrival_display_section.title` to cover both arrivals and departures across all 12 non-English locales — fr is "Affichage des arrivées et départs", es is "Visualización de llegadas y salidas" — but `en.lproj:985` and the inline `OBALoc` value in `SettingsViewController.swift` still say "Arrival Display". So English is now the only locale with the old, narrower wording. `LocalizationTests` only checks key parity, so CI can't see this.

Smaller notes, no action needed:

- The sheet toolbar's filter glyph doesn't fill for a non-`.all` Departure Type, while the pushed path deliberately does (`StopPageViewController.swift:588-589`). In the sheet they're sibling submenus so filling the *route* glyph would be wrong — I just want it to be a decision rather than an oversight.
- `departureFilterMenu()` marks selection with `action.image = checkmark` rather than `action.state = .on`. It matches the local convention in `sortMenu()`/`filterMenu()`, but `.state = .on` is what conveys selection to VoiceOver — `BookmarksViewController.swift:154` does it that way.
- The time-sorted no-results branch drops the trailing section, so "Load More" and the walk-time row disappear in that state. Likely intentional with the new empty state, but the original comment says that section is always appended.
- `OBAKitConfig.DefaultArrivalDepartureFilter` isn't documented in the README — worth adding, since the white-label knob is the point of the feature.

Push those two and ping me; I'll merge it the same day. Thanks for staying with this one.

aaronbrethorst (approved):
Landing this. Every blocker from all three earlier rounds is resolved, and I'm taking the two remaining items as follow-ups rather than another round — you've been at this since March and the feature is correct.

Filed for later:

- **#1273** — the stop page keeps a stale filter after a Settings change. `StopViewModel` seeds `arrivalDepartureFilter` once in `init` and nothing observes `UserDefaults`, so a stop page open in another tab survives the modal Settings round trip with stale results and a stale checkmark. Worth fixing because this screen already propagates settings live two different ways (`@AppStorage` in `TripCardView.swift:33`, a `didChangeNotification` subscription in `MapViewModel.swift:114`) — the notification shape is probably the better fit since the filter needs to re-run, not just re-render.
- **#1274** — the English `settings_controller.arrival_display_section.title` didn't get the arrivals-and-departures rename that all 12 other locales did, plus the VoiceOver `.state = .on` nit, the undocumented `DefaultArrivalDepartureFilter` key, and two smaller items.

**One thing you should know: I resolved the merge conflict myself and pushed it to your branch**, rather than sending it back to you. The conflict came from #1258 (implausible epoch date filtering) landing on `main` while this was open — my fault for merging that first.

It was a semantic conflict, not a textual one, so I want the resolution on the record. `main` applies `.filteringTerminalDuplicates()` early; your branch deliberately moved dedup to run *after* the Departure Type filter, and your comment explains exactly why — dedup prefers the predicted half of an arrival/departure pair, so filtering afterward could drop a scheduled row whose predicted twin was already consumed. That reasoning is right, so I kept it:

- `StopPageView.routeVisibleDepartures` now applies `.filteringImplausibleDates()` only; terminal dedup stays downstream at the `departures` binding.
- `StopViewController.stopArrivalsSection`: the time-sorted path filters implausible dates before the route filter. The route-sorted path hoists a `plausible` binding so the grouping pass and the "would dropping the type filter bring groups back" baseline pass start from the same population — otherwise those two could compare unequal sets.

I verified it rather than assuming: `build-for-testing` succeeded and the full `OBAKitTests` suite passed (1945 tests, 212 suites) against the merge locally, and CI is green on the pushed result.

If you disagree with that ordering call, say so and I'll change it — you know this pipeline better than I do.

Thanks for the persistence on this one. Five months and three review rounds is a lot, and the final shape is good: the config default resolves properly through `effectiveArrivalDepartureFilter`, unconfigured agencies get exactly today's behavior, all 10 strings are in all 13 catalogs with real translations, and the fallback tests go through the production API instead of asserting on constants.

PR #1014, opened 2026-01-16
DRAFT: New UI Exploration

aaronbrethorst:
This pull request is a work in progress reimagination of the OBA iOS app UI. I think it represents a meaningful improvement to the product user experience but needs substantially more work before it's ready for release.

PR #1099, opened 2026-03-14
fix: Add persistent close button to StopViewController in PanelUI

mx-pai:
## Problem

In the new sheet-based UI (`useNewUI`), tapping a stop opens `StopViewController` with no way to navigate back. The close button added by `StopViewControllerWrapper` was placed on the right side of the navigation bar, but `StopViewController.updateNavigationBarItems()` overwrites `rightBarButtonItems` on data load, removing the close button entirely.

## Fix

Moved the close button to `leftBarButtonItem`, which is never overwritten by `StopViewController`'s internal logic. This ensures users can always return to the Home panel.

## Testing

- Built and ran on physical device with `useNewUI` enabled
- Verified close button persists after stop data loads
- Verified tapping close returns to Home panel correctly

### DIFF
--- a/OBAKit/PanelUI/Map/VehiclesMapView.swift
+++ b/OBAKit/PanelUI/Map/VehiclesMapView.swift
@@ -65,7 +65,7 @@ struct VehiclesMapView: View {
                     }
                 )
             } else if let stop = selectedStop {
-                StopViewControllerWrapper(
+                PanelStopViewControllerWrapper(
                     application: viewModel.application,
                     stop: stop,
                     onArrivalDepartureTapped: { arrivalDeparture in
--- a/OBAKit/PanelUI/Stops/PanelStopViewControllerWrapper.swift
+++ b/OBAKit/PanelUI/Stops/PanelStopViewControllerWrapper.swift
@@ -16,24 +16,27 @@ struct PanelStopViewControllerWrapper: UIViewControllerRepresentable {
     let application: Application
     let stop: Stop
     var onArrivalDepartureTapped: ((ArrivalDeparture) -> Void)?
-    /// Called when the close button is tapped.
-    var onClose: (() -> Void)?
-
-    @Environment(\.dismiss) private var dismiss
+    var onClose: () -> Void
 
     func makeCoordinator() -> Coordinator {
-        Coordinator(dismiss: dismiss, onClose: onClose)
+        Coordinator(onClose: onClose)
     }
 
     func makeUIViewController(context: Context) -> UINavigationController {
         let stopVC = StopViewController(application: application, stop: stop)
         stopVC.onArrivalDepartureTapped = onArrivalDepartureTapped
-        stopVC.navigationItem.largeTitleDisplayMode = .always
-        stopVC.navigationItem.leftBarButtonItem = UIBarButtonItem(
+
+        let closeButton = UIBarButtonItem(
             barButtonSystemItem: .close,
             target: context.coordinator,
             action: #selector(Coordinator.close)
         )
+        closeButton.accessibilityLabel = OBALoc(
+            "panel_stop_view_controller_wrapper.close_button.accessibility_label",
+            value: "Close",
+            comment: "Accessibility label for the close button on the stop panel."
+        )
+        stopVC.navigationItem.leftBarButtonItem = closeButton
 
         let navController = UINavigationController(rootViewController: stopVC)
         navController.navigationBar.prefersLargeTitles = true
@@ -42,23 +45,22 @@ struct PanelStopViewControllerWrapper: UIViewControllerRepresentable {
     }
 
     func updateUIViewController(_ uiViewController: UINavigationController, context: Context) {
+        // Refresh callbacks so the coordinator always has the latest closures
+        context.coordinator.onClose = onClose
+        if let stopVC = uiViewController.viewControllers.first as? StopViewController {
+            stopVC.onArrivalDepartureTapped = onArrivalDepartureTapped
+        }
     }
 
     final class Coordinator {
-        let dismiss: DismissAction
-        let onClose: (() -> Void)?
+        var onClose: () -> Void
 
-        init(dismiss: DismissAction, onClose: (() -> Void)?) {
-            self.dismiss = dismiss
+        init(onClose: @escaping () -> Void) {
             self.onClose = onClose
         }
 
         @objc func close() {
-            if let onClose {
-                onClose()
-            } else {
-                dismiss()
-            }
+            onClose()
         }
     }
 }
--- a/OBAKit/PanelUI/Stops/PanelStopViewControllerWrapper.swift
+++ b/OBAKit/PanelUI/Stops/PanelStopViewControllerWrapper.swift
@@ -1,0 +1,64 @@
+//
+//  PanelStopViewControllerWrapper.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import UIKit
+import SwiftUI
+import OBAKitCore
+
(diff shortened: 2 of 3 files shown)


diveshpatil9104 (commented):
Hey @mx-pai! I’m a fellow contributor to the project, just sharing a quick review and some suggestions. using `leftBarButtonItem` so `updateNavigationBarItems()` can't overwrite it is the right call.                                                                                 
                                                                                                                                                                                                          
###   A few things to address:  

 **0. CLA not signed (blocker):** The Contributor License Agreement check is failing. Please sign the CLA at the link provided in the PR checks                                                                                                                                                                               
                                                                                                                                                                                                          
  **1. Accessibility label (blocker):** The close button needs a localized accessibility label via `OBALoc()` — this is a strict project rule for VoiceOver support.                                      
                             
  ```swift                                                                                                                                                                                                
  closeButton.accessibilityLabel = OBALoc(
      "panel_stop_view_controller_wrapper.close_button.accessibility_label",
      value: "Close",                                                                                                                                                                                     
      comment: "Accessibility label for the close button on the stop panel."
  )                                                                                                                                                                                                       
  ```                        
                                                                                                                                                                                                          
  **2. Stale `DismissAction`:** `dismiss` is captured at `makeCoordinator()` time and never refreshed. Simplest fix: always require `onClose` and drop the `@Environment(\.dismiss)` path.                
   
  **3. Empty updateUIViewController:** If callbacks change on re-render, the view controller won't pick them up. Worth re-applying onClose and onArrivalDepartureTapped through the coordinator here.                    
 

  **4. Tests:** Would be great to add tests for the close button wiring.

  **5. Screenshots:** Since this is a UI change, could you add before/after screenshots or a short video to the PR description? It helps reviewers see the fix in action.

aaronbrethorst (changes requested):
Hi mx-pai — thanks for chasing this one down. Diagnosing that `updateNavigationBarItems()` was silently clobbering the right-side close button (`StopViewController.swift:406` reassigns `navigationItem.rightBarButtonItems` wholesale on data load) is a genuinely sharp catch, and moving the button to `leftBarButtonItem` is exactly the right fix — that path is never touched by the VC's internal logic. You also handled diveshpatil9104's earlier feedback well: the accessibility label, the non-optional `onClose`, and refreshing callbacks in `updateUIViewController` are all real improvements over the original wrapper.

The fix itself is correct. My hesitation is structural, and it's in scope for this PR because this PR is what created the situation.

## Critical Issues (0 found)

None. The fix works and the retain-cycle story is clean — `selectedStop` is `@State` (a SwiftUI-managed value box, `VehiclesMapView.swift:24`), SwiftUI owns the `Coordinator`, and `updateUIViewController` refreshes `onClose` each cycle. No leaks.

## Important Issues (2 found)

- **Near-duplicate wrapper leaves the original as orphaned dead code.** `PanelStopViewControllerWrapper` is ~90% identical to the existing `public struct StopViewControllerWrapper` (`OBAKit/Stops/StopViewController.swift:16-78`). `VehiclesMapView.swift:68` was the *only* caller of that old wrapper on this branch — now that it points at the new type, the old one has zero callers. Because it's `public`, the compiler won't flag it, so it'll silently rot and the two copies will drift. Please resolve this within the PR, preferably by **consolidating into one wrapper** with a parameter for close-button placement (e.g. `enum CloseButtonPlacement { case leading, trailing }`) and folding your improvements — the accessibility label and the `updateUIViewController` callback refresh — into the unified type. If consolidation feels out of scope, the minimum acceptable alternative is to **delete the now-dead `StopViewControllerWrapper`** in this same PR (and drop the now-unneeded `import SwiftUI` at `StopViewController.swift:13` if nothing else needs it). Leaving both is the worst option. [`OBAKit/PanelUI/Stops/PanelStopViewControllerWrapper.swift`, `OBAKit/Stops/StopViewController.swift:16-78`]

- **New localization key isn't registered — and a registered one already exists.** The key `panel_stop_view_controller_wrapper.close_button.accessibility_label` (`PanelStopViewControllerWrapper.swift:34`) isn't in `OBAKit/Strings/en.lproj/Localizable.strings`, so it won't reach Transifex and VoiceOver users in non-English locales will get an untranslated "Close". The codebase already has `Strings.close` (`OBAKitCore/Strings/Strings.swift:26`, `OBALoc("common.close", value: "Close", …)`) — just reuse that for the accessibility label rather than minting a new key. [`PanelStopViewControllerWrapper.swift:34`]

## Suggestions (2 found)

- **Dropped `largeTitleDisplayMode = .always`.** The old wrapper set this (`StopViewController.swift:42`); the new one only sets `prefersLargeTitles = true` on the nav bar. Without `.always`, the large title can collapse to inline depending on panel context. If the large title was intentional, restore it; if the inline look is what you want in the panel, fine — just make it a deliberate choice rather than a silent drift from the type you copied. [`PanelStopViewControllerWrapper.swift:24-37`]

- **Trim the restating comment** at `PanelStopViewControllerWrapper.swift:48` (`// Refresh callbacks so the coordinator always has the latest closures`) — the code says this plainly. (Likely moot if you consolidate.)

## Strengths

- Correct root-cause diagnosis of a subtle UIKit-overwrites-SwiftUI interaction.
- Making `onClose` non-optional and dropping the stale `@Environment(\.dismiss)` capture is a real correctness improvement.
- Refreshing callbacks in `updateUIViewController` is the more-correct representable pattern (the old wrapper's no-op `updateUIViewController` could go stale).
- License header and file placement (UI framework, not extension-safe Core) are correct.

## Recommended Action

1. Consolidate the two wrappers into one parameterized type (preferred), or delete the now-dead `StopViewControllerWrapper`.
2. Reuse `Strings.close` for the accessibility label instead of the new unregistered key.
3. Decide intentionally about the large-title behavior.
4. Re-run the review after changes.

**Verdict: Request Changes.** The fix is right, but this PR orphans a public wrapper into dead code while adding a near-duplicate, and that's worth cleaning up before it lands rather than after.

aaronbrethorst:
### Code review

Found 2 issues:

1. Switching this call site to `PanelStopViewControllerWrapper` leaves `public struct StopViewControllerWrapper` (`OBAKit/Stops/StopViewController.swift:16-78`) with zero callers anywhere in the repo — it was the only caller on `ui-rethink`, and the type does not exist on `main`. Because it is `public`, neither the compiler nor SwiftLint will flag it, so two ~90% identical wrappers will silently drift. Consolidate into one wrapper with a close-button-placement parameter, or delete the now-dead one in this PR (bug due to `OBAKit/PanelUI/Map/VehiclesMapView.swift:68` orphaning `OBAKit/Stops/StopViewController.swift:16-78`)

https://github.com/OneBusAway/onebusaway-ios/blob/e2f4ad5c1e99a068a526a1c21e3a3f8164d297d7/OBAKit/PanelUI/Map/VehiclesMapView.swift#L66-L70

2. The new key `panel_stop_view_controller_wrapper.close_button.accessibility_label` is not present in any of the ten `Localizable.strings` files on this branch, so `OBALoc` falls back to the English `value:` in every locale. That is a net regression for VoiceOver in es/it/pl/zh-Hans: a system `.close` `UIBarButtonItem` already carries a UIKit-localized label, and this override replaces it with hardcoded English. `Strings.close` (`OBAKitCore/Strings/Strings.swift:26`, `common.close`) already exists and is translated in all four non-English locales — use it instead of minting a new key (bug due to `OBAKit/PanelUI/Stops/PanelStopViewControllerWrapper.swift:34-38`)

https://github.com/OneBusAway/onebusaway-ios/blob/e2f4ad5c1e99a068a526a1c21e3a3f8164d297d7/OBAKit/PanelUI/Stops/PanelStopViewControllerWrapper.swift#L33-L39

Note: this PR targets `ui-rethink`, whose last upstream commit is 2026-02-22, and the draft PR #1014 that would land it has been idle since then and is now `CONFLICTING` against `main`. Worth confirming the base branch is still the right target before spending more time here.

PR #1102, opened 2026-03-14
Show arrival times in the region's timezone.

### LINKED ISSUE #332: Displayed times does not use region's time zone
I'm currently in a different time zone and the displayed bus times for Puget Sound is using my local time zone. It should always be in PST, regardless of where I am.

The screenshot below shows my local time (taipei is 16 hours ahead of seattle), seattle's time, and how OBA displays the time.
Screen Shot 2020-12-13 at 5 09 17 PM

Proposed behavior:
- If I am in PST, it won't show a "PST" badge or other indicator
- When I am outside of PST, it should show the time along with a PST indicator (e.g. "1:34 AM (GMT-8)" or "1:34 AM local time"). Maybe include a toggle for using local/region time zone.

I believe `Agency` and/or `Region` has a timeZone property.

mosliem:
## Summary

Closes #332

When a user's device timezone differs from the transit agency's timezone,
arrival and departure times now display with the agency's timezone
abbreviation (e.g. "3:45 PM ET"), so users travelling across regions
always see times in the correct local context.

## Changes

- **`Agency+regionTimeZone`** — new computed property that safely resolves
  the agency's raw IANA timezone string into a `TimeZone`, returning `nil`
  if empty or invalid
- **`TimeZone+timeZoneAbbreviation`** — new extension that returns a short
  label (e.g. "PT", "ET") only when the agency timezone differs from the
  device's current timezone, suppressing redundant labels
- **`Formatters.updateTimeZone(_:)`** — new method to update the shared
  formatter's timezone; backed by `makeTimeFormatter()` which appends the
  abbreviation to the format string when needed
- **`StopViewController`**, **`TripViewController`**, **`BookmarkDataLoader`**
  — each calls `updateTimeZone` once agency data is available
  
## Screenshots 
Simulator Screenshot - iPhone 16 - 2026-03-15 at 01 35 21
Simulator Screenshot - iPhone 16 - 2026-03-15 at 01 35 16
Simulator Screenshot - iPhone 16 - 2026-03-15 at 01 35 08

## Test

- [x] Open a stop in local timezone — times show without abbreviation
- [x] Open a stop in a different timezone — times show with abbreviation (e.g. "3:45 PM PT")
- [x] Open the Trip view — times reflect the region's timezone
- [x] Open Bookmarks — times reflect the region's timezone

### DIFF
--- a/OBAKit/Stops/StopViewController.swift
+++ b/OBAKit/Stops/StopViewController.swift
@@ -132,6 +132,14 @@ public class StopViewController: UIViewController,
                 dataDidReload() // Refresh UI to reflect the change
             }
         }
+
+        updateTimeZone()
+    }
+
+    private func updateTimeZone() {
+        if let timeZone = stop?.routes.first?.agency.regionTimeZone {
+            application.formatters.updateTimeZone(timeZone: timeZone)
+        }
     }
 
     /// Arrival/Departure data for this stop.
--- a/OBAKit/Trip/TripViewController.swift
+++ b/OBAKit/Trip/TripViewController.swift
@@ -88,6 +88,10 @@ class TripViewController: UIViewController,
         let appearance = UINavigationBarAppearance()
         appearance.configureWithDefaultBackground()
         navigationItem.scrollEdgeAppearance = appearance
+
+        if let regionTimeZone = tripConvertible.arrivalDeparture?.route.agency.regionTimeZone {
+            application.formatters.updateTimeZone(timeZone: regionTimeZone)
+        }
     }
 
     override func viewWillAppear(_ animated: Bool) {
--- a/OBAKitCore/Bookmarks/BookmarkDataLoader.swift
+++ b/OBAKitCore/Bookmarks/BookmarkDataLoader.swift
@@ -61,7 +61,9 @@ public class BookmarkDataLoader: NSObject {
             bookmark.isTripBookmark
         else { return }
 
-        Task(priority: .userInitiated) {
+        Task(priority: .userInitiated) { [weak self] in
+            guard let self else { return }
+
             do {
                 let stopArrivals = try await apiService.getArrivalsAndDeparturesForStop(id: bookmark.stopID, minutesBefore: 0, minutesAfter: 60).entry
 
@@ -72,6 +74,13 @@ public class BookmarkDataLoader: NSObject {
                     }
 
                     self.delegate?.dataLoaderDidUpdate(self)
+
+                    // All bookmarks belong to the user's currently selected region, so every
+                    // concurrent load here resolves to the same agency timezone. Writing it
+                    // multiple times is harmless — each write is an identical value.
+                    if let regionTimeZone = stopArrivals.stop.routes.first?.agency.regionTimeZone {
+                        self.application.formatters.updateTimeZone(timeZone: regionTimeZone)
+                    }
                 }
             } catch {
                 await self.application.displayError(error)
--- a/OBAKitCore/Extensions/TimeZone.swift
+++ b/OBAKitCore/Extensions/TimeZone.swift
@@ -1,0 +1,21 @@
+//
+//  TimeZone.swift
+//  OBAKit
+//
+//  Created by Mohamed Sliem on 11/03/2026.
+//
+
+import Foundation
+
+public extension TimeZone {
+    /// A short, English abbreviation for this time zone when it differs from the user's current time zone.
+    ///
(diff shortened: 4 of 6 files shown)


aaronbrethorst:
@mosliem I'm open to adding this, but I think it needs to be an opt-in option instead of being the default behavior. I'm imagining seeing a UI alert on app launch:

[Region Name] is in a different time zone than you. Do you want to view times in [Region Time Zone] or [My Time Zone]?

[Region Time Zone] [My Time Zone]

## concrete example

Puget Sound is in a different time zone than you. Do you want to view times in US/Pacific or Egypt Standard Time?

[US/Pacific] [Egypt Standard Time]

aaronbrethorst:
### Code review

Found 2 issues:

1. `.shortGeneric` only yields a true abbreviation for North American zones; everywhere else it falls back to a long generic name, so the arrival row renders `3:45 PM Poland Time` / `3:45 PM United Kingdom Time` rather than the `"PT"`/`"ET"` the doc comment promises. Verified against Foundation for `en_US`: `America/Los_Angeles` → `PT`, `America/New_York` → `ET`, but `Europe/Warsaw` → `Poland Time`, `Europe/London` → `United Kingdom Time`, `Africa/Cairo` → `Egypt Time`, `Europe/Rome` → `Italy Time`, `Asia/Tokyo` → `Japan Time`. That string is appended to every departure time in a dense list, and this repo ships a Polish app (`Apps/KiedyBus`). The hardcoded `Locale(identifier: "en_US")` also means non-English users always get the English name, in an app with 13 in-repo locales.

https://github.com/OneBusAway/onebusaway-ios/blob/7226e15300593feaf11dedf6a19851d95b6b0021/OBAKitCore/Extensions/TimeZone.swift#L16-L19

2. The bookmark loader writes the region time zone *after* it notifies the delegate, so the list is re-rendered with the previous formatter and only picks up the new one on a later update. `BookmarksViewController.dataLoaderDidUpdate` calls `listView.applyData(animated: false)`, and the cells format their times through `application.formatters.timeFormatter` during that pass. With a single trip bookmark the times stay in the device's time zone until the next 30-second refresh; with several, which fetch completes first decides what the first paint looks like. Note `StopViewController` gets this ordering right (`updateTimeZone()` runs before `dataDidReload()`), so the two call sites disagree. Moving the `updateTimeZone` block above the `dataLoaderDidUpdate` call fixes it.

https://github.com/OneBusAway/onebusaway-ios/blob/7226e15300593feaf11dedf6a19851d95b6b0021/OBAKitCore/Bookmarks/BookmarkDataLoader.swift#L74-L84

aaronbrethorst (changes requested):
Thanks for your patience on this one — it's sat far longer than it should have.

The underlying request in #332 is legitimate, and the mechanics here are reasonable: adding `Agency.regionTimeZone`, rebuilding the shared `timeFormatter` for the region zone, and appending an abbreviation only when it differs from the device's is a sensible first cut.

Where I've landed is that the shape of the feature needs to change before the code does, which is what I was getting at back in March. Flipping every rider in a visiting region to region-local time by default is a surprising change for people who are looking at a schedule in the timezone they're standing in. I'd want this to be opt-in — a prompt or a setting — rather than the default behavior, and that's a different design rather than a rebase of this one.

That also has an implementation consequence. Mutating a process-wide shared `Formatters` from whichever screen loaded last is a fragile way to carry a user preference; a persisted setting consulted at format time would be cleaner and would sidestep two real problems this version has:

- In `BookmarkDataLoader` the time zone is set *after* `dataLoaderDidUpdate`, so the first paint formats with the stale formatter — `StopViewController` orders it the other way round, and the two disagree.
- The abbreviation is computed once from `Date.now` and baked into `dateFormat` as a quoted literal, so nothing re-derives it across a DST transition or a device timezone change in a long-lived session.

Two other things worth knowing for a future version: `.shortGeneric` isn't short outside North America — `Europe/Warsaw` renders as "Poland Time", not "CET", which would be appended to every row in a dense departure list, and we ship KiedyBus for Poland. And the feature is currently partial: `ScheduleForRouteViewModel` keeps its own formatter pinned to `TimeZone.current`, and the widget builds a `Formatters` that never gets updated, so those surfaces would stay in device time while Stop and Trip flip.

The branch also conflicts with `main` now — `BookmarkDataLoader` has been substantially rewritten with batch IDs and staleness gates since March.

Given all that, I think the most respectful thing I can say is: don't rebase this. If you're still interested in the feature, I'd happily review a fresh PR built around an opt-in setting, and I'm glad to sketch what that would look like first so you're not guessing. If you'd rather hand it off, no hard feelings either — say the word and I'll close it out.

PR #1103, opened 2026-03-15
Migrate region storage from UserDefaults to disk

### LINKED ISSUE #629: Move regions storage from UserDefaults to disk

mosliem:
## Summary

Closes #629

Migrates region data out of `UserDefaults` into proper disk-based
storage, with the following structure:

- **Downloaded/server regions** → `Application Support/Regions/default-regions.json`
- **Custom regions** → `Documents/custom-regions/<id>.json` (one file per region — synced via iCloud and accessible through Files.app)
- **Currently selected region** → `UserDefaults` (identifier only, not the full object)

### `OBAKitCore`
- **New `RegionsFileStorage.swift`** — `RegionsFileStorageProtocol` + concrete
  `RegionsFileStorage` implementation. All directory URL construction uses
  `fileManager.url(for:in:appropriateFor:create:)` (throws, no force-unwraps).
  Corrupted custom region files are individually logged and skipped so other
  regions still load.
- **Updated `RegionsService.swift`**:
  - Accepts a `fileStorage: RegionsFileStorageProtocol` parameter (defaults to
    `RegionsFileStorage()` — no breaking change to existing call sites)
  - `currentRegion` now stores only the `regionIdentifier` (Int) in UserDefaults;
    the full `Region` is looked up at read time via `find(id:)`, preventing stale
    object data on disk
  - `currentRegion` setter compares identifiers directly to avoid a disk read on
    every location update
  - One-time transparent migration from legacy UserDefaults format to disk on
    first launch; legacy keys are removed after migration so it runs only once

### `OBAKitTests`
- **New `RegionsFileStorageTests.swift`** 
- **Updated `RegionsServiceTests.swift`** 

## Video

https://github.com/user-attachments/assets/10c904f5-0754-4822-b10e-8f99fd1d9756


aaronbrethorst (changes requested):
Hey Mohamed, I like the direction of this PR — moving region data out of UserDefaults into proper file storage is the right architectural call, and the protocol/mock design makes it cleanly testable. The `TemporaryDirectoryFileManager` subclass for test isolation is a nice approach. Before we can merge this, I will need you to make a couple changes:

## Critical

1. **Migration deletes legacy UserDefaults keys even when the save fails** (`RegionsService.swift:297`, `314`, `322`): In `migrateFromUserDefaultsIfNeeded`, `userDefaults.removeObject(forKey:)` runs unconditionally — regardless of whether `saveDefaultRegions` threw or whether `try? PropertyListDecoder().decode(...)` returned nil. If the disk write fails (out of storage, sandbox error) or the decode fails (model schema changed between versions), the legacy data is permanently deleted from UserDefaults and nothing was written to disk. On the next launch, the app silently falls back to bundled regions and the user's previously-stored region list is gone forever.

   The fix: only call `removeObject` inside the `do` block after `saveDefaultRegions` succeeds. For the decode failure case, either leave the legacy key intact (so migration retries on next launch) or at minimum log the raw data before deleting. The same pattern applies to all three migration blocks (default regions, custom regions, current region).

2. **`currentRegion` getter triggers synchronous disk I/O on every access** (`RegionsService.swift:155-159`, `200`, `233-234`): The getter calls `find(id:)`, which falls back to `customRegions` (line 200). `customRegions` is a computed property that calls `fileStorage.loadCustomRegions()` — which does `contentsOfDirectory` + `Data(contentsOf:)` for every file — on every single property access. `currentRegion` is read from `MapViewController`, `StopViewController`, `BookmarksViewController`, `SearchInteractor`, and many other hot paths (including `updateCurrentRegionFromLocation`, which fires on every location update). For any user with a custom region selected, this means synchronous disk I/O on every location callback. Even for users with a standard region, `find(id:)` still falls through to the disk read when the identifier isn't found in the in-memory `regions` array.

   The fix: cache custom regions in memory (invalidate on `add`/`delete`), or cache the resolved `currentRegion` object and invalidate when `regions` changes or when the UserDefaults identifier changes.

## Important

3. **`loadCustomRegions` error handling is asymmetric with `loadDefaultRegions`** (`RegionsFileStorage.swift:18` vs `25`): `loadDefaultRegions()` throws errors to the caller. `loadCustomRegions()` silently swallows all errors (URL resolution failure, directory listing failure) and returns `[]`. The caller cannot distinguish "no custom regions exist" from "the Documents directory is inaccessible." Since `saveCustomRegion` and `deleteCustomRegion` both throw, `loadCustomRegions` should also throw for total-failure conditions. Per-file decode errors can still be caught and skipped internally (the `compactMap` pattern is fine for partial failures).

4. **No test for the migration failure path** (`RegionsServiceTests.swift`): The three migration tests only cover the success path. There's no test verifying behavior when `saveDefaultRegions` throws during migration — which is exactly the scenario from Critical issue #1. A test that configures `MockRegionsFileStorage` to throw on save, populates a legacy UserDefaults key, and asserts the legacy key is NOT deleted would catch this data-loss bug and prevent regressions.

## Fit and Finish

5. **`currentRegion` setter silently ignores `nil`** (`RegionsService.swift:162`): `guard let newValue else { return }` means you can never deselect a region. The getter can return `nil` (when no identifier is stored, or when `find(id:)` returns nil because the region was deleted), but the setter refuses `nil`. This asymmetry was pre-existing, but it's worth a brief doc comment on the property explaining the contract.

6. **Throwing computed properties** (`RegionsFileStorage.swift:49-71`): `var defaultRegionsFileURL: URL { get throws }` is valid Swift but unusual — most Swift codebases express fallible URL construction as functions. Consider converting to `func defaultRegionsFileURL() throws -> URL` for consistency with common Swift conventions.

aaronbrethorst (approved):
Hi Mohamed — thanks for the careful turnaround on this. The previous round flagged six items spanning a real data-loss bug, a hot-path performance regression, and a handful of fit-and-finish notes. I went through each one against the current `c2dfe8c8`, and **all six are addressed**, most of them backed by new tests that lock in the behavior. The migration code now reads like something I'd trust with a user's region list, which is exactly where it needed to land.

I rebuilt the project and ran the three regions test suites: **41 tests, 0 failures**. Build succeeds clean.

## Verification of prior feedback

1. **Migration no longer deletes legacy keys when the save fails** ✅ — `migrateDefaultRegions` removes the legacy key only inside the `do` after a successful `saveDefaultRegions` (`RegionsService.swift:365-371`); `migrateCustomRegions` tracks a `failedCount` and `guard failedCount == 0 else { return }` preserves the key if any region save throws (`:387-402`); `migrateCurrentRegion` deletes via `defer`, which is correct here because the data is fully consumed and there's nothing to retry (`:409-420`). The new tests `test_migration_defaultRegions_saveFailure_preservesLegacyKey` and `test_migration_customRegions_saveFailure_preservesLegacyKey` assert the legacy data survives byte-for-byte. This was the critical one, and it's solid.

2. **`currentRegion` no longer does synchronous disk I/O on every access** ✅ — the getter now reads a plain `Int` identifier from UserDefaults (`:195-211`), and `customRegions` is cached behind `customRegionsCacheLock`, invalidated on `add`/`delete` (`:42-46`, `:251`, `:272`). The hot path (`updateCurrentRegionFromLocation` → `currentRegion` setter) compares identifiers without touching disk.

3. **`loadCustomRegions` error handling is now symmetric** ✅ — total-failure conditions (directory access, listing) throw; per-file decode errors are logged and skipped via `compactMap` (`RegionsFileStorage.swift:96-119`). The protocol doc comment spells out the contract, and `test_loadCustomRegions_skipsCorruptedFiles` proves the partial-failure path.

4. **Migration failure now has test coverage** ✅ — the two `_saveFailure_preservesLegacyKey` tests plus the corrupt-data and empty-array tests cover exactly the data-loss scenario from the first review.

5. **`currentRegion` setter `nil` contract is documented** ✅ — the property now carries a clear doc comment explaining the getter can return `nil` while the setter ignores it (`:186-194`).

6. **Throwing computed properties converted to functions** ✅ — `defaultRegionsFileURL() throws -> URL` etc. (`RegionsFileStorage.swift:50-72`).

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None.

## Suggestions (3 found) — all optional, none blocking

- **Encode/decode date-strategy asymmetry** [`RegionsFileStorage.swift:90,124` vs `:84,113`] — saves use a plain `JSONEncoder()` (default `.deferredToDate`) but loads use `JSONDecoder.RESTDecoder()` (`.millisecondsSince1970`). Harmless today because `Region` has no `Date` fields, so nothing round-trips through a date strategy — but if `Region` ever gains one, saved files will silently fail to round-trip. Worth a matching encoder or a one-line comment documenting the constraint. Tracked separately in #1152.

- **`storeRegions()` persistence failure is invisible** [`RegionsService.swift:308-316`] — a failed disk write after a successful network refresh is logged but never surfaced, in contrast to network failures which call `notifyDelegatesDisplayError`. This matches the old fire-and-forget UserDefaults behavior, so it's not a regression, but consider an error ID in the log line for Sentry tracking so a recurring persistence failure is at least diagnosable.

- **`customRegions` returns `[]` on a load error** [`RegionsService.swift:284-287`] — a transient I/O failure makes a user's custom regions briefly disappear from the picker with no signal. Returning `[]` for resilience is a defensible choice (and correctly does *not* cache the empty result, so the next access retries) — a short comment noting that the no-cache-on-error asymmetry is intentional would protect it from a well-meaning future "fix."

## Strengths

- The `TemporaryDirectoryFileManager` subclass is a clean way to redirect Application Support / Documents lookups in tests without touching the real file system — nice.
- One-file-per-custom-region under `Documents/custom-regions/` is the right call: iCloud-syncable and Files.app-visible, and it lets one corrupt file fail in isolation instead of taking down the whole list.
- The `MockRegionsFileStorage` with injectable per-method errors makes the failure-path tests genuinely exercise the error handling rather than mocking it away.
- Protocol injection with a defaulted concrete implementation kept every existing call site unchanged — zero churn at the boundaries.
- `OBAKitCore` extension-safety is preserved (Foundation + FileManager + Logger only; no UIKit).

## Recommended Action

Ship it. The suggestions above are all non-blocking polish — none represents a current bug, and the date-strategy note is purely defensive against a field `Region` doesn't have yet. I'd happily merge this as-is and track the encode/decode symmetry as a separate follow-up.

---

**Verdict: Merge**

_The one leftover suggestion (encode/decode date-strategy symmetry) is tracked as #1152 so it isn't lost._

PR #1113, opened 2026-03-27
add: (SearchList) add tap-to-retry on error rows with enhanced visual…

mosliem:
Static error rows in placemark search now support tap-to-retry, with an enhanced visual design to communicate the failure and recovery action clearly.

## Changes
- `SearchInteractor`: added `retryPlacemarkSearch()` wired as the error row's action.
- `SearchListRowView`: `.error` renders as a tappable `Button` with orange icon, brand-coloured subtitle.
- Refactored `labelStack` into a parameterised function and extended `plainIcon` to eliminate duplicated layout code

## Video

https://github.com/user-attachments/assets/f4f8a775-b7b5-401d-aee5-e97b80744002

### DIFF
--- a/OBAKit/Search/SearchInteractor.swift
+++ b/OBAKit/Search/SearchInteractor.swift
@@ -207,7 +207,7 @@ class SearchInteractor: NSObject {
                 DispatchQueue.main.async {
                     guard let self else { return }
                     if let error {
-                        if (error as NSError).code == NSURLErrorCancelled {
+                        if (error as? URLError)?.code == .cancelled {
                             return
                         }
                         self.placemarkSearchState = .error(error)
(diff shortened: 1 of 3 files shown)


aaronbrethorst (approved):
Hey Mohamed, nice work on this — the tap-to-retry UX is a meaningful improvement over a static error row, and the retry logic is clean (resetting `lastSearchText` to bypass the dedupe guard is the right approach). I also appreciate the component refactoring: parameterizing `labelStack()` and extending `plainIcon()` with size/tint eliminates layout duplication while keeping each row type visually distinct. The `URLError` cast improvement is a good bonus too.

There's nothing left to change — this PR is ready to merge. Well-structured feature with clean refactoring.

PR #1120, opened 2026-03-30
Add destination stop picker for trip sharing

diveshpatil9104:
Adds a destination stop selection step before sharing a trip, so the shared link includes where the passenger is getting off.

Resolves the UI component of [#449](https://github.com/OneBusAway/onebusaway-ios/issues/449).

### What changed

**New file — `DestinationStopPickerController.swift`**

A modal view controller that fetches the trip's stop list via `apiService.getTrip()` and displays only the stops *after* the boarding stop. Uses `OBAListView` with `OBAListRowView.ValueViewModel` (no custom cell), a `.loading` / `.data` / `.error` state machine, and `ErrorClassifier` for user-facing errors. Task is cancelled in `deinit`.

**Modified — `StopViewController.swift`**

- Added a **"Share Trip"** `UIAction` to the long-press context menu (`stopArrivalContextMenu`).
- `shareTripStatus()` now presents the destination picker instead of immediately opening the share sheet.
- New `DestinationStopPickerDelegate` conformance: on stop selection, encodes the URL with `destination_stop_id` and opens `UIActivityViewController`; on cancel, dismisses the modal.
- Added `presentShareError()` — shows a localized alert if URL encoding fails.

### Before / After

| Before | After |
|--------|-------|
| Long-press menu: Bookmark, Schedule | Long-press menu: Bookmark, Schedule, **Share Trip** |
| "Share Trip" opened share sheet immediately (no destination) | "Share Trip" → destination picker → share sheet |
| Shared URL had no destination info | Shared URL includes `destination_stop_id` query param |

<details>
<summary>Screen recording</summary>

https://github.com/user-attachments/assets/73e2088e-3ec9-4e17-84be-522052e6a79a

</details>

### How to test

- Long-press an arrival → "Share Trip" appears in context menu
- Tap "Share Trip" → destination picker modal appears
-  Only stops **after** boarding stop are shown
-  Each row shows stop name + arrival time
-  Tap a stop → modal dismisses → share sheet opens with URL containing `destination_stop_id`
-  Cancel button dismisses modal cleanly
-  Turn off internet → error state shows (not a crash)
-  If URL encoding fails → "Unable to Share" alert appears

---

> **PR Roadmap for [#449](https://github.com/OneBusAway/onebusaway-ios/issues/449)**
>
> | # | PR | Status |
> |---|-----|--------|
> | 1 | Deep link data model (`destination_stop_id`) | `Merged` |
> | 2 | **Destination stop picker UI** | `Current` |
> | 3 | Enhanced trip view + share button | `Next` |
> | 4 | Deep link navigation with destination | `Upcoming` |
> | 5 | Comprehensive tests | `Upcoming` |

## Summary by CodeRabbit

* **New Features**
  * Share a trip from a stop arrival or departure.
  * Choose a destination stop, skip destination selection, or cancel before sharing.
  * View loading, empty, and error states while destination options are retrieved.
  * Retry loading destination options after an error.
* **Bug Fixes**
  * Shows an “Unable to Share” alert when required data is missing or sharing fails.
  * Prevents incomplete or invalid sharing attempts.
* **Tests**
  * Added coverage for picker loading, empty states, errors, selection, and retry behavior.

### DIFF
--- a/OBAKit/Stops/StopPage/Departures/DepartureRowView.swift
+++ b/OBAKit/Stops/StopPage/Departures/DepartureRowView.swift
@@ -203,6 +203,7 @@ struct DepartureRowActions {
     let onSchedule: () -> Void
     let onBookmark: () -> Void
     let onShowTrip: () -> Void
+    let onShareTrip: () -> Void
     /// Lazily builds the long-press preview (a `TripViewController` embedded via a
     /// representable). Built by the hosting VC so `Application`/UIKit stay out of
     /// the view layer; `AnyView` is acceptable here because it lives inside the
@@ -229,6 +230,9 @@ extension View {
                 }
                 Button(action: actions.onBookmark) {
                     Label(Strings.addBookmark, systemImage: "bookmark")
+                }
+                Button(action: actions.onShareTrip) {
+                    Label(OBALoc("stop_page.row.share_trip", value: "Share Trip", comment: "Context menu action that shares the trip after choosing a destination stop"), systemImage: "square.and.arrow.up")
                 }
             }, preview: {
                 actions.makePreview()
--- a/OBAKit/Stops/StopPage/StopPageView.swift
+++ b/OBAKit/Stops/StopPage/StopPageView.swift
@@ -34,6 +34,9 @@ struct StopPageNavigationHandler {
     /// Opens the bookmark editor: `nil` for a stop-level bookmark, an
     /// `ArrivalDeparture` for a trip-level bookmark (row swipe/menu).
     let showBookmarkEditor: (ArrivalDeparture?) -> Void
+    /// Starts the trip-sharing flow for a departure (row context menu):
+    /// destination stop picker, then the system share sheet.
+    let shareTrip: (ArrivalDeparture) -> Void
     /// Presents the alarm lead-time picker bulletin for a departure (the trip
     /// panel's Set-an-alarm button), reusing `AlarmBuilder` from the legacy
     /// stop screen.
@@ -525,6 +528,7 @@ struct StopPageView: View {
             onSchedule: { navigation.showScheduleForRoute(departure) },
             onBookmark: { navigation.showBookmarkEditor(departure) },
             onShowTrip: { navigation.showTrip(departure) },
+            onShareTrip: { navigation.shareTrip(departure) },
             makePreview: { navigation.makeTripPreview(departure) }
         )
     }
--- a/OBAKit/Stops/StopPage/StopPageViewController.swift
+++ b/OBAKit/Stops/StopPage/StopPageViewController.swift
@@ -195,6 +195,7 @@ class StopPageViewController: UIHostingController<StopPageRootView>,
         showWalkingDirections: {},
         showAlertDetail: { _ in },
         showBookmarkEditor: { _ in },
+        shareTrip: { _ in },
         showAlarmPicker: { _ in },
         startLiveActivity: { _ in },
         showExternalSurveyError: {},
@@ -207,6 +208,10 @@ class StopPageViewController: UIHostingController<StopPageRootView>,
         showReportProblem: {},
         closeSheet: {}
     )
+
+    /// Owns the picker → share-sheet trip-sharing flow. Holds `self` weakly,
+    /// so retaining it for the controller's lifetime creates no cycle.
+    private lazy var tripSharingCoordinator = TripSharingCoordinator(application: application, presenter: self)
 
     private func makeNavigationHandler() -> StopPageNavigationHandler {
         StopPageNavigationHandler(
@@ -223,6 +228,7 @@ class StopPageViewController: UIHostingController<StopPageRootView>,
                 self.application.viewRouter.navigateTo(alert: alert, from: self)
             },
             showBookmarkEditor: { [weak self] departure in self?.showBookmarkEditor(for: departure) },
+            shareTrip: { [weak self] departure in self?.tripSharingCoordinator.start(arrivalDeparture: departure) },
             showAlarmPicker: { [weak self] departure in self?.showAlarmPicker(for: departure) },
             startLiveActivity: { [weak self] departure in self?.startLiveActivity(for: departure) },
             showExternalSurveyError: { [weak self] in self?.showExternalSurveyError() },
--- a/OBAKit/Stops/StopViewController.swift
+++ b/OBAKit/Stops/StopViewController.swift
@@ -1065,6 +1065,11 @@ public class StopViewController: UIViewController,
                 actions.append(schedule)
             }
 
+            let shareTrip = UIAction(title: OBALoc("stop_controller.share_trip", value: "Share Trip", comment: "Context menu button that allows the user to share their trip status."), image: Icons.share) { [weak self] _ in
+                self?.shareTripStatus(viewModel: viewModel)
+            }
+            actions.append(shareTrip)
+
             // Create and return a UIMenu with all of the actions as children
             return UIMenu(title: viewModel.name, children: actions)
         }
@@ -1347,25 +1352,31 @@ public class StopViewController: UIViewController,
 
     // MARK: - Share Trip Status
     func shareTripStatus(viewModel: ArrivalDepartureItem) {
-        guard let arrivalDeparture = arrivalDeparture(forViewModel: viewModel) else { return }
+        guard let arrivalDeparture = arrivalDeparture(forViewModel: viewModel) else {
+            Logger.error("No arrivalDeparture found for share trip status view model.")
+            return
+        }
         shareTripStatus(arrivalDeparture: arrivalDeparture)
     }
 
     func shareTripStatus(arrivalDeparture: ArrivalDeparture) {
-        guard
-            let region = application.currentRegion,
-            let appLinksRouter = application.appLinksRouter
-        else {
-            return
-        }
+        let picker = DestinationStopPickerController(
+            application: application,
+            arrivalDeparture: arrivalDeparture
+        )
+        picker.delegate = self
+        let nav = application.viewRouter.buildNavigation(controller: picker)
+        self.present(nav, animated: true)
+    }
 
-        let url = appLinksRouter.encode(arrivalDeparture: arrivalDeparture, region: region) as Any
-
-        let activityController = UIActivityViewController(activityItems: [self, url], applicationActivities: nil)
-
-        // Use self.presnt because when using application.viewRouter.present(:_),
-        // it disables UIActivityViewController's "tap anywhere to dismiss".
-        self.present(activityController, animated: true)
+    private func presentShareError() {
+        let alert = UIAlertController(
+            title: OBALoc("stop_controller.share_error.title", value: "Unable to Share", comment: "Alert title when trip sharing fails."),
+            message: OBALoc("stop_controller.share_error.message", value: "Something went wrong while preparing the trip link. Please try again.", comment: "Alert message when trip sharing fails."),
+            preferredStyle: .alert
+        )
+        alert.addAction(UIAlertAction(title: Strings.dismiss, style: .default))
+        present(alert, animated: true)
     }
 
     // MARK: - Schedules
@@ -1562,6 +1573,50 @@ private extension StopViewController {
             self.showAllTransferDepartures = true
             self.listView.applyData()
         }
+    }
+}
+
+// MARK: - DestinationStopPickerDelegate
+
+extension StopViewController: DestinationStopPickerDelegate {
+    func destinationStopPicker(
+        _ controller: DestinationStopPickerController,
+        didSelectStop stopTime: TripStopTime,
+        for arrivalDeparture: ArrivalDeparture
+    ) {
+        dismiss(animated: true) { [weak self] in
+            guard let self else { return }
+            guard
+                let region = self.application.currentRegion,
+                let appLinksRouter = self.application.appLinksRouter
+            else {
+                Logger.error("Missing region or appLinksRouter for trip sharing.")
+                self.presentShareError()
+                return
+            }
+
+            guard let url = appLinksRouter.encode(
+                arrivalDeparture: arrivalDeparture,
+                region: region,
+                destinationStopID: stopTime.stopID
+            ) else {
+                Logger.error("Failed to encode trip sharing URL.")
+                self.presentShareError()
+                return
+            }
+
+            let activityController = UIActivityViewController(
+                activityItems: [self, url],
+                applicationActivities: nil
+            )
+            // Use self.present because when using application.viewRouter.present(:_),
+            // it disables UIActivityViewController's "tap anywhere to dismiss".
+            self.present(activityController, animated: true)
+        }
+    }
+
+    func destinationStopPickerDidCancel(_ controller: DestinationStopPickerController) {
+        dismiss(animated: true)
     }
 }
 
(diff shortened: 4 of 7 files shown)


aaronbrethorst (changes requested):
Hi Divesh — this is a clean, well-scoped slice of the #449 roadmap. The new `DestinationStopPickerController` faithfully follows the established `state`-enum + `OBAListView` + async-`Task` pattern from `CurrentTripViewController`, your concurrency and memory hygiene are correct (weak captures throughout, `deinit` cancellation, `MainActor.run` before every UI mutation), and every external symbol you call checks out against the real APIs. The error alert, the cancellation guard, and the localized strings all show real care. Nice work.

I verified the whole surface area by hand — `encode(...destinationStopID:)`, `getTrip`, `TripStopTime`/`TripDetails`, `ValueViewModel`, `EmptyData`, `buildNavigation`, `Logger` — and there are no crashes or compile concerns. The items below are about correctness in the edge cases and a few cheap design tightenings.

## Critical Issues (0 found)

None.

## Important Issues (2 found)

- **Boarding-stop-not-found fallback can produce an incorrect shared destination.** `OBAKit/Trip/DestinationStopPickerController.swift:119-126`: when the boarding stop isn't located in the trip's stop times, the fallback shows *all* stops — including stops at or before where the rider boarded. The entire feature is "where are you getting off?", so this lets the user generate and share a link whose destination is behind them, and the list looks identical to the correct filtered one. A `Logger.warn` to the device log isn't user feedback (and there's no Sentry behind `Logger`, so it's invisible in production). Please either treat a not-found boarding stop as an `.error` state ("couldn't determine your boarding point on this trip") or make a permissive fallback an explicit, communicated decision. In practice a trip always contains its own boarding stop, so failing closed here costs nothing.

- **Misleading error for a missing API service.** `OBAKit/Trip/DestinationStopPickerController.swift:99-103`: a nil `apiService` surfaces `APIError.noResponseBody`, which renders to the user as *"The server unexpectedly didn't return any data…"* — but the server was never contacted; this is a local configuration state. `CurrentTripViewController` handles the same case with a purpose-built `NSError` ("Unable to connect to the transit service."). Worth matching for a message that points at the real cause.

## Suggestions (6 found)

- **Model empty as its own state.** `DestinationStopPickerController.swift:38-42`, `155-158`, `198-210`: `.data([])` is smuggled in as "empty," forcing two independent `isEmpty` checks (`items(for:)` and `emptyData(for:)`) to stay in sync. Add a `case empty` and set `forwardStops.isEmpty ? .empty : .data(forwardStops)`. Then `items(for:)` is a clean `guard case .data` and the switch is exhaustive over real states. The sibling `CurrentTripViewController` already uses explicit empty-ish cases (`.noResults`, `.noLocation`), so this is the more consistent choice here too.
- **Drop the redundant `arrivalDeparture` from the delegate callback.** `DestinationStopPickerController.swift:16-20`: the delegate handed the picker its `arrivalDeparture` when constructing it, so echoing it back in `didSelectStop:for:` only introduces a trivially-true invariant the delegate must reason about. The controller reference is sufficient identity; `stopTime` is the genuinely new information.
- **Make the main-actor invariant structural.** Annotating the controller `@MainActor` lets you delete the explicit `MainActor.run` hops and converts "correct because I remembered to wrap each write" into "correct by construction."
- **No retry on the error screen.** `DestinationStopPickerController.swift:196-197`: this is a modal whose only escape from a transient network error is Cancel, then re-triggering the whole share flow. `EmptyData`'s `buttonConfig:` supports a reload affordance — consider one.
- **Silent dead-end in `shareTripStatus(viewModel:)`.** `OBAKit/Stops/StopViewController.swift:1355-1359`: the new `Logger.error` is an improvement, but the user still taps "Share Trip" and nothing happens. Since `presentShareError()` now exists, this path could use it. (Low severity — the guard should be unreachable — but "unreachable + silent" is how mystery bug reports are born.)
- **Enrich log context.** The `Logger.error`/`warn` lines omit the tripID/stopID. With no Sentry behind `Logger`, the device log is the only forensic trail, so include the identifiers.

## Strengths

- Correct, idiomatic Swift concurrency: off-main `getTrip`, `MainActor.run` before UI writes, `CancellationError` early-return, `deinit` cancel.
- No retain cycles — `weak` delegate, `[weak self]` in the row action, the dismiss completion, and the `UIAction`.
- Clean state machine driving all rendering through a single `didSet`; strong encapsulation (everything private, enum private).
- Proper `OBALoc` localization with sensible defaults and comments on all seven new strings.
- Correct presentation ordering: dismiss the picker, then present the share sheet / error alert in the completion.

## Recommended Action

1. Fix the boarding-stop-not-found fallback so it can't share a behind-you destination.
2. Swap the no-service error for a configuration-appropriate message.
3. Fold in the `.empty` case, the delegate-signature trim, and `@MainActor` while you're in here — they're cheap and the delegate is brand new, so trimming it costs nothing.
4. Re-request review.

**Verdict: Request Changes.**

aaronbrethorst:
### Code review

Found 1 issue:

1. The new test suite is written against the pre-migration XCTest world and no longer matches `main`. It subclasses `OBATestCase` with `override func setUp()` / `override func tearDown()` and calls `super.setUp()`, but `OBATestCase` on `main` is a plain `@MainActor` class whose `init() async throws` and `deinit` explicitly replace `setUp`/`tearDown` — there is nothing for those to override. Even setting that aside, the `func test_…` methods would never be discovered, so the suite would silently contribute zero tests. (CLAUDE.md says tests are "Written with **Swift Testing** (`@Suite` / `@Test` / `#expect`), not XCTest" and that "`OBATestCase` is a plain `@MainActor` base class, not an `XCTestCase`. Suites that need its fixtures inherit it and override `init() async throws` (where `setUp` used to go); teardown goes in `deinit`.") Note that the green `build` check on this PR ran on 2026-06-09, before the Swift 6 / Swift Testing migration landed, so it does not cover this.

https://github.com/OneBusAway/onebusaway-ios/blob/1c87578ceadbd671a618c61c1fc937fe87b1ef74/OBAKitTests/Trip/DestinationStopPickerControllerTests.swift#L15-L29

aaronbrethorst (changes requested):
The feature code came back in good shape — both Important items from June are genuinely fixed, and I checked them at head rather than taking the description's word for it.

Failing closed on the boarding stop is the right call: `guard let boardingIndex = …firstIndex(where:)` with a real error instead of the permissive fallback means we can no longer share a destination that's behind the rider. That was the one that actually worried me. The `apiService == nil` path now raises "Unable to connect to the transit service." instead of a bogus `noResponseBody`, `.empty` is modeled as its own state so `items(for:)` is a clean `guard case .data`, the delegate lost its redundant `arrivalDeparture`, the class is `@MainActor` with the `MainActor.run` hops deleted, there's a Try Again button on the error screen, and `shareTripStatus(viewModel:)` no longer dead-ends silently. That's seven of eight, and you added a test pinning the boarding-stop-not-found path.

**The blocker is that this can't build on current `main`.** The branch is seven weeks old and its only CI run is from June 9, which predates the entire Swift 6 migration — so that green check is telling you nothing. Concretely:

`OBAKitTests/Trip/DestinationStopPickerControllerTests.swift` is XCTest-shaped:

```swift
import XCTest
class DestinationStopPickerControllerTests: OBATestCase {
    override func setUp() {
        super.setUp()
```

`OBATestCase` on `main` is now a plain `@MainActor class` with no superclass and no `setUp`/`tearDown` — I checked, there is nothing to override or call `super` on. The suite needs to become Swift Testing: `@Suite(.serialized)`, `@Test`, `#expect`, with setup moved into `override init() async throws` and teardown into `deinit`.

Good news: I went looking for Swift 6 concurrency problems in the *production* code and didn't find any. The `deinit { fetchTask?.cancel() }` on a `@MainActor` class is fine (`RentalLayerCoordinator` does the same), `RESTAPIService` is an actor so the capture is Sendable, and binding the non-Sendable `RESTAPIResponse<TripDetails>` across the await is allowed under region isolation — `MapRegionManager` already does that shape. So this really is just the test file plus a rebase.

Three things worth folding in while you're there, none blocking:

- **`.empty` is a dead end.** With "no remaining stops on this trip" the only exit is Cancel — there's no way to share at all. Not a regression (the old `shareTripStatus` had no callers), but a destination-less share would be reasonable there.
- **The three new `Logger.error` lines in `StopViewController`** (`:1211`, `:1470`, `:1480`) still carry no tripID/stopID. The picker's own logging does now — this was the one suggestion only partly picked up.
- **`if error is CancellationError`** won't match `URLError.cancelled`, which is how URLSession actually reports cancellation. There's an `error.isCancellation` helper for exactly this (see `TripViewModel.swift:115-128` and the comment there). Effectively unreachable as written, since `guard let self` makes the capture strong for the request's duration — which also means `deinit`'s cancel is mostly decorative.

Rebase onto `main`, convert the suite, and get a real CI run — then I'll re-review promptly. The feature itself is close.

aaronbrethorst:
### Code review

Found 1 issue:

1. The new "Share Trip" action is added only to the legacy `StopViewController`'s arrival context menu, which is not the screen users see by default. On `main`, `FeatureFlags.isNewStopPageEnabled` resolves to `true` when the toggle has never been touched (`OBAKitCore/Orchestration/FeatureFlags.swift:26-28`), and `ViewRouter.makeStopController(stop:)` / `makeStopController(stopID:)` route to `StopPageViewController` in that case — `StopViewController` is constructed nowhere else in the app. The SwiftUI departure row has its own `.contextMenu` (`OBAKit/Stops/StopPage/Departures/DepartureRowView.swift:216-232`) offering only Schedules and Add Bookmark. So in a default build the destination picker is unreachable: it shows up only after switching off Settings > Experimental > new stop page, or via the transfer-context path. That flag defaulted to enabled on 2026-07-10 (aff9954a), after this branch was cut, so this is drift rather than an oversight — but as it stands the PR ships the feature dark. (bug due to `OBAKit/Stops/StopViewController.swift` adding the action to the legacy menu only)

https://github.com/OneBusAway/onebusaway-ios/blob/67c0e7631af7472655ea7d4e02b2037d5d9dc118/OBAKit/Stops/StopViewController.swift#L879-L884

aaronbrethorst (changes requested):
This is much improved, and the blocker from my last review is properly closed. The test suite is real Swift Testing now — `@Suite(.serialized)` on `OBATestCase`, `override init() async throws`, `@Test`/`#expect`/`#require`, using the repo's own `poll(until:)` — and I confirmed against CI run 30665374005 at head that all seven tests actually executed and passed inside a green 1698-test run. That is verified, not claimed.

The other items are handled well too:

- `if Task.isCancelled || error.isCancellation` now matches the pattern in `TripViewModel` and `MapRegionManager`, and more importantly the task captures its identifiers up front and unwraps `[weak self]` only after the await — so `deinit`'s cancel is load-bearing instead of decorative. The comment saying exactly that is a nice touch.
- "Share Without Destination" turns the `.empty` state from a dead end into a real escape, with tests pinning it.
- The logger lines now carry `tripID`/`stopID`/`destinationStopID`.
- Presentation goes through `ViewRouter`, and the blanket `force_try` disable is gone.

What is holding this is reachability, and like #1096 it is main-side drift rather than anything you did. "Share Trip" is added only to `StopViewController`'s context menu, and `FeatureFlags.isNewStopPageEnabled` now resolves to `?? true`. `Router.swift` is the only place `StopViewController` is built, and it branches to `StopPageViewController` whenever that flag is on. The SwiftUI `DepartureRowView` context menu has Schedules and Add Bookmark and no share entry. So as merged, only users who turn off the experimental toggle — or arrive through a transfer context — can ever reach the picker you built. The flag flipped default-on in `aff9954a` on 2026-07-10, after this branch was cut.

Two ways forward, and I am fine with either: add the action to the new page's departure context menu, or state plainly in the PR body that the picker is intentionally legacy-only until the redesigned page catches up. What I do not want is to merge 290 lines of good UI that nobody can open.

While you are in here, two things worth fixing:

- `UIActivityViewController` is presented without a `popoverPresentationController` source, which throws on iPad. We target iPad, and the codebase knows this — `MapItemViewModel` has an explicit "iPad support" block and `LogViewerViewController` does the same. Those two lines are copied verbatim from the pre-existing dead code so you did not introduce the bug, but giving `shareTripStatus` a menu entry is what makes it reachable for the first time.
- The boarding stop is located with `firstIndex(where: { $0.stopID == boardingStopID })` rather than `arrivalDeparture.stopSequence`. On a loop trip that visits the same stop twice, the first occurrence wins and the picker offers stops the rider has already passed — which is the exact failure the June review asked you to close. `stopSequence` is documented as the index into the trip's stop sequence, `isLastStopOnTrip` already relies on it, and `AppLinksRouter.encode` already emits it as `stop_sequence`.

Minor: `fetchTask?.cancel()` sits after the `apiService` guard, so a nil-service early return leaves a prior in-flight task alive to overwrite the `.error` state. CodeRabbit raised this five minutes after your last push, so you have not had a chance at it yet.

Sort out the reachability question and I will get this landed.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
All four items from my last pass are fixed, and I checked each at head rather than from the description.

**Reachability — the blocker — is closed.** `DepartureRowActions` gained `onShareTrip`, `DepartureRowView`'s context menu has the button, `StopPageView` wires it to `navigation.shareTrip`, and `StopPageViewController` owns a `TripSharingCoordinator` that starts the flow. I confirmed the default path is actually live rather than merely present: `StopPreferences.init()` sets `sortType = .time`, `StopPageView:344` branches to `ChronologicalListView`, and that's the view applying `.departureRowActions(actions)`. So a default-configuration user on the new stop page can reach the picker.

**The loop-trip fix is the one I'm happiest with.** Anchoring on `arrivalDeparture.stopSequence` is right, and the `allStopTimes[boardingSequence].stopID == boardingStopID` cross-check with a bounds guard and a `Logger.error` makes it fail closed rather than silently listing the wrong half of the trip. That's better than what I asked for — I'd have accepted the sequence lookup alone.

iPad popover anchoring matches the `MapItemViewModel.swift:229-234` pattern, and `fetchTask?.cancel()` is now the first thing `loadStopTimes()` does.

I also checked the tests aren't the tautological kind that's been going around today. They're not: `Loop trip anchors on stopSequence` builds `["1_x", boarding, "1_y", boarding, "1_z"]`, where the old `firstIndex(where:)` would produce 3 rows and select `1_y` instead of 1 row selecting `1_z` — it fails on revert. `Stop list shorter than the boarding sequence` uses a list where an ID search would have succeeded and yielded `.empty` rather than the asserted `.error`, so it fails on revert too.

On privacy, since this ships a shareable link: the URL carries region ID, stop ID, trip ID, service date, stop sequence, and the destination the user picked. No coordinates, no device or user identity, nothing the user didn't choose. That's the right envelope.

Four follow-ups, none blocking:

- **Duplicate keys for the same words.** `stop_page.row.share_trip` and `stop_controller.share_trip` are both "Share Trip" for the same action — two keys, two translations, free to drift across 13 locales. A shared `Strings` symbol would fix it.
- **Route-grouped mode still can't share.** `GroupedListView` applies no context menu at all, so with `sortType == .route` there's no Share Trip — nor Show Trip, Alarm, Schedule, or Bookmark. Pre-existing gap in the new page rather than something you introduced, but it does mean one of two sort modes stays share-less. Worth its own issue.
- **`.error` has no share escape.** `.empty` offers "Share Without Destination"; a network failure leaves only Try Again and Cancel. Not a regression, just an asymmetry — the destination is optional, so a failed fetch needn't block sharing.
- **`TripSharingCoordinator` is untested.** The nine tests all target the picker; the `destination_stop_id` encoding, the "Unable to Share" alert, and the dismiss-then-present ordering aren't covered. That coordinator is the part both stop pages now depend on.

Merging. Nice work getting this one over the line — it's been a long haul.

PR #1126, opened 2026-04-06
Add stop bookmark support to widget

### LINKED ISSUE #1125: Widget to support Bookmarks for a station
When I bookmarked a stop, a train station for example, nothing shows up in the onebusaway widget 

App | Widget 
----|----
Image | Image

The proposal is to show something like the board that you would see on the platform 
Image

tomtclai:
Closes #1125

### DIFF
--- a/OBAWidget/Views/WidgetRowView.swift
+++ b/OBAWidget/Views/WidgetRowView.swift
@@ -13,6 +13,7 @@ import OBAKitCore
 private enum Constants {
     static let minutes: UInt = 60
     static let maxDeparturesToShow = 3
+    static let maxStopDeparturesToShow = 2
     static let rowWidth: CGFloat = 180
     static let fontSize: CGFloat = 13
 }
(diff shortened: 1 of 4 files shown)


OBAWidget/Views/WidgetRowView.swift
```diff
@@ -27,6 +28,10 @@ struct WidgetRowView: View {
         bookmark?.name ?? " "
     }
 
+    private var isStopBookmark: Bool {
+        bookmark?.isTripBookmark == false
```
Vedd-Patel, line 32: if bookmark happens to be nil, this evaluates to false (since nil == false is false)
this means a nil bookmark will default to rendering the tripBookmarkView
assuming tripBookmarkView handles a nil bookmark gracefully this is fine, but it's worth double checking that this behavior is intentional

Vedd-Patel (commented):
hey @tomtclai 
good work!
here are some reviews from my side(i'm not an official reviewer here so you can ignore these if you want)

you should create a dedicated fear/fix/test... branch for each PR as pushing from main will lead you to some difficulties in some cases!


Localization Strings Check:
In stopBookmarkView, the empty state uses:
```
String(format: LocalizationKeys.noDeparturesInNextNMinutes, String(Constants.minutes))
```
just verify that the corresponding string key for noDeparturesInNextNMinutes has been properly added to the Localizable.strings files, as that change isn't visible in the diffs provided

tomtclai:
Thanks for the pointer for the feature branching... I will open a new pr from there.

PR #1127, opened 2026-04-06
Add stop bookmark support to widget

### LINKED ISSUE #1125: Widget to support Bookmarks for a station

tomtclai:
Closes #1125

## Summary by CodeRabbit

- **New Features**
  - Added stop-based bookmarks to widget displays.
  - Stop bookmarks show up to two upcoming departures with route, destination, time, and schedule status.
  - Widgets now include all bookmarks available in the current region.
  - Trip bookmarks continue displaying trip-specific arrival and departure information.
  - Added fallback messaging when departure information is unavailable.

- **Bug Fixes**
  - Correctly distinguishes stop bookmarks from trip bookmarks when loading arrival data.
  - Filters stop departures to show only upcoming services in chronological order.


aaronbrethorst (changes requested):
Hi Tom — thanks for this. Stop bookmarks silently rendering nothing in the widget (issue #1125) has been a real papercut, and the approach here is the right one: branch on `isTripBookmark`, keep the existing trip layout untouched, and give stops their own "platform board" style row. The dual-dictionary design and the `stopArrDepDic` reset in `loadData()` are clean, and routing both bookmark types through the same fetch is sensible. A few things to tighten before this goes in.

## Critical Issues (0 found)

None — no correctness bugs or crashes. Concurrency is safe (dictionary writes happen inside `MainActor.run` within the `TaskGroup`).

## Important Issues (1 found)

- **Stop bookmarks with no upcoming departures render a blank row.** `lookupStopArrivals(for:)` returns `[]` (never nil), so for a stop with no service in the next 60 minutes, `departures` is empty, the `if let departures` succeeds, and the `ForEach` produces nothing — the user sees just the stop name and empty space. [OBAWidget/Views/WidgetRowView.swift:80-102] The trip layout handles this case gracefully via `nextDepartureLabel` ("No departures in next 60 minutes" / "Tap for more information"). Please give the stop layout an equivalent empty state so a quiet stop reads as intentional rather than broken. This is the one item I'd like fixed in this PR, since it's the exact feature being added.

## Suggestions (4 found)

- **Drop the `isStopBookmark` double-negative.** `bookmark?.isTripBookmark == false` reads awkwardly and means a `nil` bookmark routes to the stop layout by accident (harmless today — `nil` only happens in the preview — but unintentional). Branch directly on the positive form, matching `loadArrivalDeparture` and `fetchArrivalData` which both lead with `bookmark.isTripBookmark`:
  ```swift
  var body: some View {
      if bookmark?.isTripBookmark == true {
          tripBookmarkView
      } else {
          stopBookmarkView
      }
  }
  ```
  [OBAWidget/Views/WidgetRowView.swift:31-48]

- **Consider filtering past departures on the stop path.** The list is sorted ascending by `arrivalDepartureDate` but not filtered, so a very-recently-departed trip (negative minutes) could surface as one of the two shown rows. `minutesBefore: 0` makes this unlikely but not impossible. A `.filter { $0.temporalState != .past }` before sorting would match what riders expect from a platform board. [OBAWidget/Provider/WidgetDataProvider.swift:90-91]

- **Redundant fetches when a stop has both a stop and trip bookmark.** Each bookmark triggers its own `getArrivalsAndDeparturesForStop` call, so a stop bookmarked both ways is fetched twice and each fetch fills only one dictionary. It works correctly; it's just wasteful. Not worth blocking on, but worth a note for a future pass. [OBAWidget/Provider/WidgetDataProvider.swift:76-93]

- **Tidy two small things in the view:** the `// MARK: - Trip Bookmark Layout (existing)` marker — the "(existing)" is a PR-authoring artifact that won't age well [WidgetRowView.swift:51] — and the identical title `Text` block duplicated across both layouts, which could become a shared `private var titleText` [WidgetRowView.swift:56-59, 82-85].

## On the tests

The two new tests (`testStopBookmarkIsNotTripBookmark`, `testTripBookmarkIsTripBookmark`) are correct and pin the `isTripBookmark` discriminator the whole feature keys off of — a sensible minimal safety net. The genuinely new logic (the `getBookmarks()` filter change, the `fetchArrivalData` branch routing, the sort predicate) is untested, but that's a structural limitation: `WidgetDataProvider` is a `shared` singleton that builds its own `CoreApplication`/`UserDefaults`/`CLLocationManager`, and the test target doesn't depend on `OBAWidget`. I'm not asking you to refactor that here — the view-body rendering is fairly untestable anyway. A worthwhile future cleanup (separate PR) would extract the sort/filter logic into a free function or type in OBAKitCore so it's reachable from the test target.

## Strengths

- Correctly resets `stopArrDepDic` alongside `arrDepDic` in `loadData()`, avoiding stale cross-refresh data.
- Existing trip-bookmark layout left fully intact; the change is cleanly additive.
- Reuses `routeAndHeadsign`, `shortFormattedTime(until:)`, and `backgroundColorForScheduleStatus` — consistent with the rest of the codebase.
- Added a unit test rather than shipping the discriminator change untested.

## Recommended Action

1. Add an empty state to `stopBookmarkView` for stops with no upcoming departures.
2. Simplify the `isStopBookmark` double-negative.
3. Consider the past-departure filter and the minor view tidy-ups.
4. Re-request review after the empty-state fix.

**Verdict: Request Changes** — only the blank-row empty state is blocking; everything else is polish.

aaronbrethorst:
### Code review

Found 1 issue:

1. The new `Trip bookmark is a trip bookmark` test crashes the test process. `Fixtures.dictionaryToModel` decodes with a bare `JSONDecoder`, so `loadReferences(_:regionIdentifier:)` is never called and the resulting `ArrivalDeparture` has a nil `stop` (`public var stop: Stop!`). `Bookmark.init(name:regionIdentifier:arrivalDeparture:)` then does `self.init(..., stop: arrivalDeparture.stop)`, which implicitly force-unwraps that nil and traps — taking the whole OBAKitTests run down with it, not just this case. The established pattern for a references-backed `ArrivalDeparture` is `Fixtures.loadRESTAPIPayload(type: StopArrivals.self, fileName: "arrivals-and-departures-for-stop-1_10914.json")` followed by `.arrivalsAndDepartures.first`, as in `BookmarksViewModelTests`. (bug due to `Bookmark.swift:65-66` force-unwrapping `arrivalDeparture.stop`)

https://github.com/OneBusAway/onebusaway-ios/blob/ac0d2f68806164e85ebafe54ee42aa7a6f8fc4d8/OBAKitTests/Modeling/UserData/Bookmarks/BookmarkTests.swift#L92-L96

aaronbrethorst (changes requested):
The widget work itself looks good, and the blocking item from my last review is properly fixed — a stop bookmark with no upcoming departures now renders the fallback label instead of a blank row. The `isStopBookmark` double negative is gone, the past-departure filter is in, and pulling the duplicated title into `titleText` tidied it up nicely.

One thing blocks it, and it's small but fatal: the new `Trip bookmark is a trip bookmark` test crashes the test host.

`Fixtures.dictionaryToModel` decodes with a bare `JSONDecoder`, so `loadReferences` never runs and `ArrivalDeparture.stop` — declared `Stop!` — stays nil. `Bookmark.init(name:regionIdentifier:arrivalDeparture:)` passes `arrivalDeparture.stop` into a non-optional `Stop` parameter, so constructing the bookmark force-unwraps nil and takes down the whole OBAKitTests process, not just that test.

The fix is to build the fixture the way `BookmarksViewModelTests` does — `Fixtures.loadRESTAPIPayload(type: StopArrivals.self, ...)` — which exists precisely because these models need their references loaded.

Worth flagging for your own sake: no build or test check has ever run on this branch. The only checks present are CodeRabbit and the CLA, so nothing would have caught this before merge. If you rebase onto current `main` the build workflow should pick it up and give you a real signal.

- `stopBookmarkView` colors the countdown with `formatters.backgroundColorForScheduleStatus(...)` as a foreground color. Every other departure-time label in the app uses `colorForScheduleStatus`; the `*Background` variants are documented as badge fills. Only the `.unknown` case actually differs, but it's worth matching.
- The new stop rows have no accessibility label on the countdown, so VoiceOver reads a literal "5m". `DepartureTimeBadgeView` and the trip `departureLabel` both provide spoken equivalents.

Swap the fixture and I'll re-review.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Everything from both prior rounds is fixed. The fixture swap is the right one:

```swift
let stopArrivals = try Fixtures.loadRESTAPIPayload(
    type: StopArrivals.self,
    fileName: "arrivals-and-departures-for-stop-1_10914.json"
)
```

That routes through `RESTAPIResponse.init(from:)`, which calls `loadReferences`, so `ArrivalDeparture.stop` is populated and the `Stop!` force-unwrap in `Bookmark.init` no longer takes down the test host. I also checked the fixture actually has `routeId`, `routeShortName`, and `tripHeadsign`, so `isTripBookmark` is genuinely true rather than passing vacuously.

Both smaller notes are done too — the countdown now uses `colorForScheduleStatus` instead of the `*Background` variant, and the stop rows carry an accessibility label built from `formatters.explanation(from:)`, which is the same spoken form `DepartureTimeBadgeView` uses. The Swift Testing conversion follows the conventions in CLAUDE.md.

**Now the part that's on me, not you.** I said last time that no build or test check had ever run on this branch. I now know why, and it isn't anything you did: every `OBAKitTests` run on this branch — including the one at `be1052fd` — ends in `action_required`. That's GitHub holding fork-PR workflows until a maintainer approves them. Merging `main` in didn't help and never would.

So I'm approving the code on the merits, but this shouldn't land until that run actually executes. I'll approve the pending workflow run so we get a real signal. Given that last round's blocker was precisely a crash that only execution catches, I'd rather see it green than merge on a careful read.

Two things I noticed that aren't yours to fix here:

- The stop path filters `temporalState != .past` once at fetch time, but `BookmarkTimelineProvider` reuses that snapshot across six hours of half-hour entries, so late entries can show already-departed trips. The trip badge path has the identical staleness on `main`, so this is a pre-existing shape rather than something you introduced.
- `fallbackLabel`'s doc comment says the nil case means "no data yet", but `lookupStopArrivals(for:)` returns `[]` rather than nil on failure, so a *failed* fetch reads as "no departures in the next 60 minutes". `lookupArrivalDeparture` behaves the same today — worth untangling sometime, not here.

Thanks for sticking with this one across several rounds.

PR #1128, opened 2026-04-07
Implement customizable walking speed with HealthKit

### LINKED ISSUE #217: Let user customize their walking speed
The user should be able to customize their speed of movement, which is used to determine how quickly they can reach their transit stop. 

The user should be able to customize this value in two ways:

1. By typing a new number into a text box in Settings.

2. By retrieving this data from HealthKit in iOS 14

HealthKit in iOS 14 adds support for determining and accessing the user's walking speed. There should be an option for retrieving this data from the HealthKit store instead of simply using the hardcoded value of 3.1 miles/hour I've included.

https://developer.apple.com/videos/play/wwdc2020/10182/

mosliem:
Fixes #217
Previously, walking speed was hardcoded to 1.4 m/s. This PR allows users to customize their walking speed for more accurate arrival and trip planning estimates.
### Changes
* **Settings UI**: Added a new "Walking Speed" section where users can manually pick a speed preset (Slow, Average, Fast).
* **HealthKit Sync**: Added a toggle to sync actual walking speed directly from the Health app (fetches the latest valid speed from the past 30 days).
* **Trip Planning**: Updated `WalkingDirections` to calculate travel times using the newly customized velocity.
* **Auto-Sync**: Automatically syncs the latest HealthKit walking data in the background on app launch for opted-in users.
* **Config/Permissions**: Linked the `HealthKit` framework and added necessary usage description strings & entitlements for OneBusAway.
* **Tests**: Added coverage for `UserDataStore` persistence and `WalkingDirections` calculation edges cases.

## Summary by CodeRabbit

* **New Features**
  * Walking Speed settings: three presets (slow/average/fast), manual entry, and a Health app toggle that disables manual controls when enabled
  * App requests Health permission, can auto-sync walking speed (including on launch), and uses the value for walk-time estimates
  * HealthKit entitlement enabled for apps, HealthKit framework included, and iOS usage description/localized labels added; shows error toast on sync failure

* **Tests**
  * Added tests for settings decisions, sync behavior, validation/clamping, and travel-time calculations

### DIFF
--- a/Apps/KiedyBus/project.yml
+++ b/Apps/KiedyBus/project.yml
@@ -19,6 +19,7 @@ targets:
       properties:
         com.apple.security.application-groups:
           - group.pl.kiedybus.iphone
+        com.apple.developer.healthkit: true
     info:
       path: Apps/KiedyBus/Info.plist
       properties:
--- a/Apps/OneBusAway/project.yml
+++ b/Apps/OneBusAway/project.yml
@@ -25,6 +25,7 @@ targets:
           - applinks:sidecar.onebusaway.org
         com.apple.developer.in-app-payments:
           - merchant.org.onebusaway.iphone
+        com.apple.developer.healthkit: true
     info:
       path: Apps/OneBusAway/Info.plist
       properties:
--- a/Apps/Shared/app_shared.yml
+++ b/Apps/Shared/app_shared.yml
@@ -42,6 +42,7 @@ targets:
           - twitter
           - comgooglemaps
         NSLocationWhenInUseUsageDescription: See where you are in relation to transit, and help you navigate more easily.
+        NSHealthShareUsageDescription: OneBusAway uses your walking speed from the Health app to give you more accurate arrival time estimates.
         NSLocationTemporaryUsageDescriptionDictionary:
           MapStatusView: See where you are in relation to transit, and help you navigate more easily.
         UILaunchStoryboardName: LaunchScreen
--- a/OBAKit/Orchestration/Application.swift
+++ b/OBAKit/Orchestration/Application.swift
@@ -90,6 +90,9 @@ public class Application: CoreApplication, PushServiceDelegate {
 
     lazy var toastManager = ToastManager()
 
+    @MainActor
+    lazy var walkingSpeedManager = WalkingSpeedManager(userDataStore: userDataStore)
+
     @objc lazy var userActivityBuilder = UserActivityBuilder(application: self)
 
     /// Handles all deep-linking into the app.
@@ -374,6 +377,10 @@ public class Application: CoreApplication, PushServiceDelegate {
         reportAnalyticsUserProperties()
 
         configureTipKit()
+
+        if userDataStore.walkingSpeedSource == .healthKit {
+            Task { await walkingSpeedManager.requestHealthKitAuthorizationAndSync() }
+        }
     }
 
     @objc public func applicationDidBecomeActive(_ application: UIApplication) {
--- a/OBAKit/Orchestration/Application.swift
+++ b/OBAKit/Orchestration/Application.swift
@@ -379,7 +379,7 @@ public class Application: CoreApplication, PushServiceDelegate {
         configureTipKit()
 
         if userDataStore.walkingSpeedSource == .healthKit {
-            Task { await walkingSpeedManager.requestHealthKitAuthorizationAndSync() }
+            Task { await walkingSpeedManager.refreshFromHealthKitIfPossible() }
         }
     }
 
--- a/OBAKit/Settings/SettingsViewController.swift
+++ b/OBAKit/Settings/SettingsViewController.swift
@@ -254,8 +254,7 @@ class SettingsViewController: FormViewController {
    // MARK: - Privacy
 
     private let privacySectionReportingEnabled = "privacySectionReportingEnabled"
-    private let alwaysShowSurveysOnStops = "alwaysShowSurveysOnStops"
-    
+
     // MARK: - Walking Speed Keys
     private let walkingSpeedMetersPerSecondKey = "walkingSpeedMetersPerSecond"
     private let walkingSpeedUseHealthKitKey = "walkingSpeedUseHealthKit"
--- a/OBAKit/Settings/SettingsViewController.swift
+++ b/OBAKit/Settings/SettingsViewController.swift
@@ -9,6 +9,7 @@
 
 import Eureka
 import Foundation
+import HealthKit
 import OBAKitCore
 import UIKit
 
@@ -38,6 +39,7 @@ class SettingsViewController: FormViewController {
             +++ mapSection
             +++ alertsSection
             +++ accessibilitySection
+            +++ walkingSpeedSection
             +++ surveySection
             +++ debugSection
 
@@ -58,7 +60,9 @@ class SettingsViewController: FormViewController {
             RegionsService.alwaysRefreshRegionsOnLaunchUserDefaultsKey: application.userDefaults.bool(forKey: RegionsService.alwaysRefreshRegionsOnLaunchUserDefaultsKey),
             MapRegionManager.mapViewShowsStopAnnotationLabelsDefaultsKey: application.userDefaults.bool(forKey: MapRegionManager.mapViewShowsStopAnnotationLabelsDefaultsKey),
             debugModeEnabled: application.userDataStore.debugMode,
-            alwaysShowSurveysOnStops: application.userDataStore.alwaysShowSurveysOnStops
+            alwaysShowSurveysOnStops: application.userDataStore.alwaysShowSurveysOnStops,
+            walkingSpeedMetersPerSecondKey: snapToPreset(application.userDataStore.walkingSpeedMetersPerSecond),
+            walkingSpeedUseHealthKitKey: application.userDataStore.walkingSpeedSource == .healthKit
         ])
     }
 
@@ -114,6 +118,23 @@ class SettingsViewController: FormViewController {
         } else {
             application.userDefaults.set(false, forKey: RegionsService.alwaysRefreshRegionsOnLaunchUserDefaultsKey)
         }
+
+        // Walking Speed — save source first to avoid race condition
+        if let useHK = values[walkingSpeedUseHealthKitKey] as? Bool {
+            application.userDataStore.walkingSpeedSource = useHK ? .healthKit : .manual
+        }
+
+        if let speed = values[walkingSpeedMetersPerSecondKey] as? Double,
+           application.userDataStore.walkingSpeedSource == .manual {
+            application.userDataStore.walkingSpeedMetersPerSecond = speed
+        }
+
+        // Snap to nearest preset when toggling HealthKit OFF
+        if let useHK = values[walkingSpeedUseHealthKitKey] as? Bool, !useHK {
+            application.userDataStore.walkingSpeedMetersPerSecond = snapToPreset(
+                application.userDataStore.walkingSpeedMetersPerSecond
+            )
+        }
     }
 
     // MARK: - Map Section
@@ -164,6 +185,59 @@ class SettingsViewController: FormViewController {
         return section
     }()
 
+    // MARK: - Walking Speed
+
+    private let walkingSpeedPresets: [Double] = [0.9, 1.4, 1.8]
+
+    private func snapToPreset(_ speed: Double) -> Double {
+        walkingSpeedPresets.min(by: { abs($0 - speed) < abs($1 - speed) }) ?? 1.4
+    }
+
+    private lazy var walkingSpeedSection: Section = {
+        let section = Section(OBALoc("settings_controller.walking_speed_section.title", value: "Walking Speed", comment: "Settings > Walking Speed section title"))
+
+        section <<< SegmentedRow<Double> {
+            $0.tag = walkingSpeedMetersPerSecondKey
+            $0.title = OBALoc("settings_controller.walking_speed.title",
+                              value: "Walking speed",
+                              comment: "Settings > Walking Speed section > Speed picker")
+            $0.options = walkingSpeedPresets
+            $0.displayValueFor = { speed in
+                switch speed {
+                case 0.9: return OBALoc("settings_controller.walking_speed.slow", value: "Slow (~2 mph)", comment: "")
+                case 1.8: return OBALoc("settings_controller.walking_speed.fast", value: "Fast (~4 mph)", comment: "")
+                default:  return OBALoc("settings_controller.walking_speed.avg", value: "Average (~3 mph)", comment: "")
+                }
+            }
+            $0.disabled = Condition.function([walkingSpeedUseHealthKitKey], { [weak self] form in
+                guard let self = self else { return false }
+                return (form.rowBy(tag: self.walkingSpeedUseHealthKitKey) as? SwitchRow)?.value ?? false
+            })
+        }
+
+        if HKHealthStore.isHealthDataAvailable() {
+            section <<< SwitchRow {
+                $0.tag = walkingSpeedUseHealthKitKey
+                $0.title = OBALoc("settings_controller.walking_speed.use_healthkit",
+                                  value: "Use Health app data",
+                                  comment: "Settings > Walking Speed section > HealthKit toggle")
+                $0.onChange { [weak self] row in
+                    guard let self, row.value == true else { return }
+                    Task {
+                        let granted = await self.application.walkingSpeedManager.requestHealthKitAuthorizationAndSync()
+                        if !granted {
+                            row.value = false
+                            row.reload()
+                            self.application.userDataStore.walkingSpeedSource = .manual
+                        }
+                    }
+                }
+            }
+        }
+
+        return section
+    }()
+
     // MARK: - Agency Alerts
 
     private lazy var alertsSection: Section = {
@@ -180,6 +254,11 @@ class SettingsViewController: FormViewController {
    // MARK: - Privacy
 
     private let privacySectionReportingEnabled = "privacySectionReportingEnabled"
+    private let alwaysShowSurveysOnStops = "alwaysShowSurveysOnStops"
+    
+    // MARK: - Walking Speed Keys
+    private let walkingSpeedMetersPerSecondKey = "walkingSpeedMetersPerSecond"
+    private let walkingSpeedUseHealthKitKey = "walkingSpeedUseHealthKit"
 
     private lazy var privacySection: Section = {
         let section = Section(OBALoc("settings_controller.privacy_section.title", value: "Privacy", comment: "Settings > Privacy section title"))
--- a/OBAKit/Settings/SettingsViewController.swift
+++ b/OBAKit/Settings/SettingsViewController.swift
@@ -187,10 +187,8 @@ class SettingsViewController: FormViewController {
 
     // MARK: - Walking Speed
 
-    private let walkingSpeedPresets: [Double] = [0.9, 1.4, 1.8]
-
     private func snapToPreset(_ speed: Double) -> Double {
-        walkingSpeedPresets.min(by: { abs($0 - speed) < abs($1 - speed) }) ?? 1.4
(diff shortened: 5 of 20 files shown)


aaronbrethorst (changes requested):
Hi Mohamed! This is a genuinely nice feature, and the shape of the implementation is right: a clean `WalkingSpeedManager` seam, a `WalkingSpeedSource` enum to model the manual/HealthKit distinction, the `velocity` default param that keeps `WalkingDirections` backward-compatible, and real persistence tests for `UserDataStore`. The HealthKit plumbing — `isHealthDataAvailable()` gating, the `withCheckedContinuation` bridge with no `self` captured in the background callback, the `0.5...5.0` sanity range — shows you thought carefully about the platform.

There's one design issue at the heart of the HealthKit flow that I want to fix before this merges, plus a small cluster of correctness/cleanup items. None of them require rethinking the feature — they harden the unhappy paths.

## Critical Issues (1 found)

- **HealthKit "authorization succeeded" is treated as "read access granted" — but Apple guarantees you can't know that, so a denying user gets a silently dead toggle.** [OBAKit/TripPlanning/WalkingSpeedManager.swift:31-43] Per Apple's documented privacy model (verified against the official *Authorizing access to health data* guide): *"your app doesn't know whether someone granted or denied permission to read data from HealthKit. If they denied permission, attempts to read data return only samples that your app successfully saved."* `requestAuthorization(toShare: [], read: [speedType])` therefore does **not** throw when the user taps "Don't Allow" — it succeeds. So `requestHealthKitAuthorizationAndSync()` returns `true`, the caller [SettingsViewController.swift:227-232] reads that as `granted`, leaves the toggle ON, and disables the manual segmented control [SettingsViewController.swift:212-215]. The subsequent query returns zero samples (no error), the speed never updates, and the user is left believing Health sync is driving their ETAs when it isn't — with no way to fix it short of toggling Health back off. Because denial and "no data yet" are the *most likely* outcomes, not edge cases, this is the default experience for many users.
  - **Fix:** make success contingent on actually retrieving a usable sample rather than on the prompt completing. Have `requestHealthKitAuthorizationAndSync()` return `true` only when `syncLatestWalkingSpeed` produced a valid speed; otherwise log, revert `walkingSpeedSource` to `.manual`, and let the toggle fall back OFF. One product call to make explicitly: a user who genuinely granted access but currently has no walking-speed samples (e.g. no Apple Watch) would also be bounced to `.manual` — I think that's the right tradeoff, but decide it on purpose.

## Important Issues (4 found)

- **Dead code: the `walkingVelocity` constant is now orphaned.** [OBAKit/TripPlanning/WalkingDirections.swift:17] After the refactor, `private static let walkingVelocity = 1.4` is referenced nowhere (confirmed by grep across OBAKit/OBAKitCore/OBAKitTests) — the implementation divides by the `velocity` parameter whose default is the literal `1.4`. Delete it, or better, make it the source of that default (see next item).

- **`velocity: 0` (or negative) silently yields `+Infinity` / negative travel time — and the public API doesn't guard it.** [OBAKit/TripPlanning/WalkingDirections.swift:29] `distance / velocity` does not trap on zero in Swift. It's unreachable *today* because the only caller [StopViewController.swift:1139] sources the value from `walkingSpeedMetersPerSecond` (registered default 1.4, UI-constrained to presets) and HealthKit samples are range-filtered. But `travelTime` is `public` with a permissive `Double`, so the safety lives entirely in callers. Add `guard velocity > 0 else { return nil }` — and ideally clamp at the store boundary: enforce the `0.5...5.0` range in the `walkingSpeedMetersPerSecond` setter [UserDataStore.swift:1028-1035] so no consumer can ever read a divide-hostile value. Promote that range to one shared constant instead of letting it live only inside `WalkingSpeedManager`.

- **The `HKSampleQuery` completion handler throws away its `Error?`.** [OBAKit/TripPlanning/WalkingSpeedManager.swift:63] `{ _, samples, _ in ...}` collapses a genuine query failure (corrupt store, execution error) into the same "no samples" path as an empty result — a silent failure, and inconsistent with the rest of the file, which logs the auth error on line 36. Bind the error and `Logger.error` it before resuming with `nil`.

- **Denied/failed HealthKit gives the user no explanation — the switch just snaps back OFF.** [OBAKit/Settings/SettingsViewController.swift:224-234] When the request fails, the toggle silently reverts with no message. The user can't tell whether it's a bug, a denial, or something they fix in iOS Settings → Privacy → Health. The app already has a `ToastManager`; surface a brief message pointing them there. (This becomes more important once the Critical fix routes denials through this same path.)

## Suggestions (4 found)

- **Collapse the four copies of `1.4` into one constant.** The default speed is hardcoded independently at UserDataStore.swift:380, WalkingDirections.swift:23 (and the dead :17), and SettingsViewController.swift:193 (`?? 1.4`) plus the `default:` "Average" case. If anyone ever changes the default, these diverge silently.

- **Consider a `WalkingSpeedPreset` enum to consolidate the presets.** [OBAKit/Settings/SettingsViewController.swift:188-210] The literals `[0.9, 1.4, 1.8]` appear as a bare array, again as exact-equality `switch` cases in `displayValueFor` (`case 0.9:`, `case 1.8:`), and again in `snapToPreset`. The float-equality switch is safe today only because Eureka hands back the unmodified option literals — a derived value like `1.4000001` would silently fall through to "Average." An enum (`case slow = 0.9, average = 1.4, fast = 1.8` with a `localizedTitle` and a `nearest(to:)`) owns the value, label, and snapping in one place and removes the fragility. It belongs in OBAKit (it carries localized strings); the raw `Double` stays in OBAKitCore.

- **Pin the enum's raw values and document the persistence contract.** [OBAKitCore/Models/UserData/UserDataStore.swift:17-20] `WalkingSpeedSource` is persisted by `rawValue`, so reordering or inserting a case silently corrupts every existing user's stored setting. `@objc Int` is the right representation — just write `case manual = 0` / `case healthKit = 1` and add a one-line "do not reorder" comment.

- **New `OBALoc` strings aren't extracted, and this branch predates main's localization migration.** None of the `settings_controller.walking_speed.*` keys exist in any `.strings` file (`OBALoc`'s inline `value:` fallback prevents a crash, but they're untranslatable until `scripts/extract_strings` runs). Separately, heads-up: `main` has since migrated localization from `.lproj/Localizable.strings` to `Localizable.xcstrings`, so you'll hit a conflict on rebase — worth syncing with `main` before final extraction. Also minor: the "non-capture explicit" / "runs on a background thread" comments at WalkingSpeedManager.swift:54,64 restate what the code already shows and will go stale; I'd drop them.

## Strengths

- Correct concurrency: `@MainActor` manager, `withCheckedContinuation` bridging the background `HKSampleQuery`, `limit: 1` guaranteeing a single resume, and no `self` capture in the callback.
- `isHealthDataAvailable()` gating correctly hides the toggle on unsupported devices, and the auth-failure `catch` (line 35-39) is the right pattern — specific to the throwing call, logged, and reverts to `.manual`.
- The persistence tests are meaningful, not box-checking: `test_walkingSpeed_roundTrip` builds a fresh `UserDefaultsStore` to verify true cross-instance persistence, and the default-value test pins the `1.4` registration that's the most likely regression on a future merge.
- The `0.5...5.0` HealthKit range filter sensibly rejects garbage samples, and the out-of-range no-op is honestly documented.

## Recommended Action

1. Fix the Critical: make HealthKit sync success mean "a sample was actually read," reverting to `.manual` (and informing the user) otherwise.
2. Delete the dead `walkingVelocity`; guard `velocity > 0` (and/or clamp at the store); log the swallowed query error.
3. Add user-facing feedback when the toggle can't engage.
4. Consider the suggestions (single `1.4` constant, preset enum, pinned raw values), run `scripts/extract_strings`, and sync with `main` for the `.xcstrings` migration.

---

**Verdict: Request Changes.** The feature is well-architected and close, but the HealthKit flow currently presents an active-looking toggle that does nothing for any user who denies read access or has no Health data — and that's the common case, not the edge. Closing that hole (plus the small correctness cleanups) gets this to merge.

aaronbrethorst (changes requested):
Hi Mohamed — this is looking great. Every item from the last round landed cleanly: the HealthKit denial path now means "we actually read a sample" rather than "the prompt closed," the swallowed query error logs, `travelTime` guards `velocity > 0`, the `1.4` literals collapsed into `WalkingSpeed.defaultMetersPerSecond`, the presets became a real `WalkingSpeedPreset` enum that owns value + label + snapping, and `WalkingSpeedSource` got its "do not reorder" persistence contract. The active-vs-passive asymmetry — active toggle may downgrade to `.manual`, launch refresh never does — is now both implemented correctly and documented where a future maintainer will see it. I also generated both `OneBusAway` and `KiedyBus` and confirmed `NSHealthShareUsageDescription` merges into each app's `Info.plist`, so there's no first-use crash. Nicely done.

One substantive gap keeps this from merging, plus a trivial comment fix. Both are about protecting the work you already did, not redoing it.

## Critical Issues (0 found)

None. There are no current correctness bugs — the logic is right as written.

## Important Issues (1 found)

- **The HealthKit denial/sync state machine — the exact code last round's Critical fix introduced — has zero test coverage and no seam to add it.** [OBAKit/TripPlanning/WalkingSpeedManager.swift:17] `healthStore` is a hardcoded `private let healthStore = HKHealthStore()`, so the load-bearing contract you wrote this class to enforce is completely unguarded: that `requestHealthKitAuthorizationAndSync()` returns `false` and forces `walkingSpeedSource = .manual` when no usable sample comes back (lines 49-53), that an out-of-range sample is discarded and never writes `.healthKit` (lines 99-102), and that `refreshFromHealthKitIfPossible()` *never* downgrades the source (the asymmetry documented at lines 56-60). A future refactor that flips this back to "auth succeeded ⇒ return true" would silently recreate the precise bug — a dead toggle for a denying user — that we agreed was the Critical issue last time, and nothing would catch it. Extract a small protocol around the two operations you actually use (`requestAuthorization` and "fetch latest in-range sample"), inject it, and cover four cases: denial/no-sample → `.manual` + `false`; in-range sample → value written + `.healthKit` + `true`; out-of-range sample → no write; passive refresh with no sample → source unchanged. That's the difference between the prior fix being permanent and being one careless edit away from regressing.

## Suggestions (6 found — none blocking)

- **Misleading comment.** [OBAKit/TripPlanning/WalkingSpeedManager.swift:71] The doc says a miss "keeps current manual preset," but on the passive `refreshFromHealthKitIfPossible` path the user is already on `.healthKit`, so a miss keeps the *previously-synced HealthKit value*, not a manual preset — and it contradicts the (correct) rationale you wrote 11 lines above. Reword to describe the method's own contract: "leaves the stored speed and source untouched and returns `false`; the caller decides how to react." (This one's quick — worth folding in even though it's not blocking.)

- **`saveWalkingSpeedValues` branch logic is also untestable as written.** [OBAKit/Settings/SettingsViewController.swift:125-143] The source-first → manual-only-write → snap-on-off ordering is correct, but it's a private `FormViewController` method, so the interaction can't be exercised. If you extract the pure decision — `(currentSource, useHK: Bool?, segmentSpeed: Double?) -> (newSource, newSpeed)` — into a free function or onto `WalkingSpeedPreset`, you get the same testability win as the manager refactor for the second-most-subtle piece of logic in the PR.

- **Clamp the read path too, or reify the value.** [OBAKitCore/Models/UserData/UserDataStore.swift:1031-1039] The `0.5...5.0` invariant is enforced only in the setter; the getter returns a bare `userDefaults.double(...)`, which yields `0.0` if the key is ever cleared. `travelTime` then returns `nil` (the row just hides), so it's not a crash — but the invariant guards writes and not reads. Either clamp on read as well, or longer-term wrap the value in a `struct WalkingSpeed { init(clamping:) }` so the guarantee travels with the value instead of being re-derived at each boundary.

- **Collapse the duplicated HealthKit guard.** [OBAKit/TripPlanning/WalkingSpeedManager.swift:34-39, 62-66] `isHealthDataAvailable()` + `quantityType(forIdentifier: .walkingSpeed)` appears verbatim in both entry points. A `private var walkingSpeedType: HKQuantityType?` helper removes the duplication while leaving the two different failure actions explicit at the call sites.

- **`Calendar.current.date(byAdding: .day, value: -30, to: Date())!`** [OBAKit/TripPlanning/WalkingSpeedManager.swift:74] is the one force-unwrap in the new code. It won't realistically fail, but `?? Date.distantPast` removes the theoretical trap.

- **The new `settings_controller.walking_speed.*` keys still aren't extracted.** They live only as `OBALoc` inline `value:` fallbacks, so nothing crashes, but they're untranslatable until `scripts/extract_strings` runs. Worth doing before this merges so the translators pick them up.

## Strengths

- The denial-as-failure redefinition is exactly right, and the doc comment at lines 24-31 explains Apple's privacy model and the deliberate "no Apple Watch ⇒ `.manual`" tradeoff so the next reader can't mistake it for a bug.
- Clean concurrency: `@MainActor` manager, `withCheckedContinuation` bridging the background query, `limit: 1` guaranteeing a single resume, no `self` capture in the callback, both error paths logged.
- `WalkingSpeedSource` is an exemplary persisted type — `@objc Int`, pinned raw values, `?? .manual` fallback on read, and the "do not reorder" comment placed exactly where it's needed.
- The pure-logic tests are genuinely behavioral, not box-checking: `WalkingDirectionsTests` recomputes expected times from distance/velocity and asserts the slower-is-longer relationship; `UserDataStoreTests` re-instantiates the store to prove real persistence and asserts clamping against `validRange.lowerBound/upperBound` rather than literals.
- `StopViewController.swift:973` actually threads the customized speed into the travel-time calc — the feature is wired end to end, not just stored.

## Recommended Action

1. Add a test seam to `WalkingSpeedManager` (protocol + injection) and cover the denial / in-range / out-of-range / passive-refresh cases. This is the one blocking item.
2. Fix the misleading comment at `WalkingSpeedManager.swift:71`.
3. Optionally take the cheap suggestions (extract the save decision, clamp on read, dedup the guard, drop the force-unwrap) and run `scripts/extract_strings`.
4. Re-request review.

---

**Verdict: Request Changes.** The feature is correct and well-built — this isn't about a bug in the code. It's that the safety-critical denial path you fixed last round is currently protected by zero tests and has no seam to add them, so it's one edit away from silently regressing. A small protocol abstraction locks it in permanently; pair that with the one-line comment fix and this is ready.

aaronbrethorst (approved):
Mohamed — this is ready. Last round left exactly one blocking item: the HealthKit denial/sync state machine, the load-bearing logic introduced to fix the dead-toggle bug, had zero test coverage and no seam to add it. You did precisely the right thing — extracted a narrow `WalkingSpeedHealthKitProviding` protocol, injected it into `WalkingSpeedManager`, and wrote eight tests that pin down every branch that matters: denial/no-sample → `.manual` + `false`, in-range → value written + `.healthKit` + `true`, out-of-range → no write, auth-throws → `.manual`, HK-unavailable → `.manual`, and all three passive-refresh paths including the asymmetry that a passive launch refresh must *never* downgrade the source. That contract is now permanent instead of one careless edit away from regressing. I built both apps, ran the suite, and all 31 walking-speed tests pass.

You also folded in every optional suggestion from last round: the comment at `WalkingSpeedManager.swift:65-67` now describes the method's own contract instead of contradicting itself; `WalkingSpeedSettingsDecision.compute` extracts the settings branching into a pure, `Equatable`, unit-tested function (five tests); the read path is clamped alongside the write; the duplicated availability guard collapsed into `walkingSpeedType` + `isAvailable`; the date force-unwrap became `?? .distantPast`; and the localization keys are extracted into `Localizable.strings`. I also confirmed the HealthKit entitlement flows correctly into *both* apps — `scripts/generate_project OneBusAway` regenerates `OneBusAway.entitlements` with `com.apple.developer.healthkit` from `project.yml` (the checked-in copy is a gitignored build artifact), and `NSHealthShareUsageDescription` merges into `Info.plist`, so no first-use crash.

## Critical Issues (0 found)

None. I specifically chased the one scenario that could have blocked this — a dismissal race where the user toggles HealthKit ON and leaves Settings before the async auth resolves, potentially persisting `.healthKit` with no usable sample (a dead toggle). I traced it through the actual code: both `viewWillDisappear` → `saveWalkingSpeedValues` and the `onChange` `Task` run on the MainActor, so they can only interleave at `await` suspension points. On any failure ordering, the manager's `.manual` write (`WalkingSpeedManager.swift:50`) and the `onChange` row-reset run as one synchronous continuation with no `await` between them, so `saveWalkingSpeedValues` can never slip between them — the persisted source always lands on `.manual`. The dead-toggle outcome is not reachable. The only residual is cosmetic (see Suggestions).

## Important Issues (0 found)

None.

## Suggestions (6 found — none blocking)

- **Getter-side clamp is effectively untested.** [OBAKitCore/Models/UserData/UserDataStore.swift:1031-1034] The value is clamped on both read and write, but the clamp tests at `UserDataStoreTests.swift:300-308` assign through the setter — which already clamps — so the getter's own `min/max` never sees an out-of-range value. Delete it tomorrow and every test still passes. A three-line test that writes a bad value straight to `UserDefaults` (bypassing the setter) and asserts the getter returns the clamped bound would close the one defense-in-depth path the current tests give false confidence about.
- **Dropped error toast on mid-flight dismissal.** [OBAKit/Settings/SettingsViewController.swift:218-235] If the user toggles HealthKit ON and dismisses Settings before the auth request resolves, the failure toast is shown into a `nil` window and silently dropped. Persisted state is still correct (`.manual`), so this is purely missed feedback in an edge case — worth a follow-up, not a fix here.
- **`saveWalkingSpeedValues` form-adapter casts are untested.** [OBAKit/Settings/SettingsViewController.swift:125-135] `compute` is well-tested in isolation, but the `as? Bool` / `as? Double` bridge from Eureka's form dictionary is the fragile glue: if the `SegmentedRow`'s value type ever drifts, a manual speed change is silently dropped with no crash. A focused test feeding `form.values()` through the save path would guard it.
- **Collapse the duplicated clamp.** [OBAKitCore/Models/UserData/UserDataStore.swift:1031-1040] The identical `min(max(...))` lives in both the getter and setter four lines apart. A `WalkingSpeed.clamped(_:)` helper gives one definition, two call sites — and keep the manager's *reject* (not clamp) semantics for out-of-range HealthKit samples deliberately distinct.
- **Guard constant drift.** [OBAKitCore/Models/UserData/WalkingSpeed.swift:21-25] A one-line test asserting `validRange.contains(defaultMetersPerSecond)` keeps the two constants from drifting apart in a future edit.
- **Log the impossible branch.** [OBAKit/TripPlanning/WalkingSpeedHealthKitProviding.swift:46] `walkingSpeedType == nil` is a should-never-happen case that currently routes to `.manual` with no diagnostic trail. A `Logger.warn` would distinguish it from ordinary "no recent samples."

## Strengths

- The denial/sync state machine — the riskiest logic in the PR — is now the best-covered: all six branches plus the active-vs-passive asymmetry are asserted, with a comment at `WalkingSpeedManagerTests.swift:123-137` explaining *why* a passive refresh must not downgrade. That's a test that catches a real regression, not box-checking.
- `WalkingSpeedHealthKitProviding` is a textbook seam — narrow (three members), the two genuinely surprising HealthKit behaviors (silent denial, no built-in validation) documented exactly where a reader needs them, and the `FakeProvider` is honest rather than over-mocked.
- `WalkingSpeedSettingsDecision` is the design highlight: pulling the form-save branching into a pure, `Equatable` decision is exactly how you make "settings logic trapped in a view controller" testable.
- `WalkingSpeedSource` remains an exemplary persisted type — `@objc Int`, pinned raw values, "do not reorder" comment, `?? .manual` fallback on read.
- The feature is wired end to end: `StopViewController.swift:973` threads the customized speed into the travel-time calc, and `Application.swift:380-383` does the best-effort launch refresh only for opted-in users.

## Recommended Action

1. Merge.
2. Track the six suggestions in a follow-up issue (none block this PR).

---

**Verdict: Merge.** Last round's one blocking item is fully resolved — the safety-critical denial path now has a proper test seam and comprehensive coverage, so the fix is permanent rather than fragile. Every optional suggestion landed too. The remaining items are genuine but minor (a missing defense-in-depth test, a dropped-toast edge case, and small polish), and belong in a tracking issue rather than another review round. Build succeeds, all 31 tests pass. Nicely done.

Issue #515, opened 2021-09-07
Bus Stop Overlap (UX)


roverman:
Issue: 
There are certain stops that’s overlapping each other at the same location.

Recommendation: 
Combine the bus stops together if they are at the exact same location.

![Desktop - 13](https://user-images.githubusercontent.com/68605140/132273692-3002efa4-40cc-4e02-9026-f037f182c51a.png)

Aniket-pd:
Maybe we can handle it by showing one grouped pin when stops share the same location. Tapping it could open a small picker sheet/list of those stops, then route to whichever stop the rider selects. Feels cleaner on the map without losing access to individual stops

aaronbrethorst:
@Aniket-pd I think that would be a great approach. would you be up for tackling it?

Aniket-pd:
Yeah sure i can go for it

Aniket-pd:
Hey @aaronbrethorst   how should we determine whether two stops are at the same location?
Should we use exact latitude/longitude, or round coordinates (e.g., to ~5 meters)?

Open to your thoughts

aaronbrethorst:
@Aniket-pd I'd imagine that you're best off just using `clusteringIdentifier` and `MKClusterAnnotation` and delegating all of the responsibilities for making that decision to MKMapView.

PR #1130, opened 2026-04-08
Improve map stop clustering reliability and cluster selection UX 

Aniket-pd:
Problems Addressed #515 
This update makes stop pins on the map feel much more reliable and easier to use.
Instead of custom grouping, it now uses native MapKit clustering, so pins stay consistent even when you zoom or move around. Tapping a cluster shows a simple list of stops, letting you quickly pick one and navigate to it.

Pins no longer flicker during zoom, and they reappear instantly when you zoom back in, so the map feels smooth and responsive. Clusters also stay stable instead of breaking into messy overlapping pins.

Overall, the map behaves more predictably, and selecting the right stop is faster and clearer

RESULTS: 
1. Clusters remain stable after map interactions, 
2. Dense areas no longer degrade into overlapping unclustered pins.

| Before | After |
|--------|--------|
|  |  |

## Summary by CodeRabbit

* **New Features**
  * Added a cluster stop picker interface that appears when tapping grouped stops on the map, allowing easier selection from co-located stops.
  * Improved stop annotation visibility behavior with hysteresis thresholds and delayed hiding for smoother map interactions.

* **Tests**
  * Added comprehensive test coverage for cluster stop picker deduplication, sorting, and UI formatting logic.
  * Added tests for stop annotation view clustering behavior and visibility hysteresis logic.

### DIFF
--- a/OBAKit/Mapping/MapRegionManager.swift
+++ b/OBAKit/Mapping/MapRegionManager.swift
@@ -70,6 +70,7 @@ public class MapRegionManager: NSObject,
     private let application: Application
 
     private var regionChangeRequestTimer: Timer?
+    private var stopHideTimer: Timer?
 
     private var userLocationAnnotationView: PulsingAnnotationView? {
         didSet {
@@ -223,6 +224,7 @@ public class MapRegionManager: NSObject,
         application.locationService.removeDelegate(self)
         application.regionsService.removeDelegate(self)
         regionChangeRequestTimer?.invalidate()
+        stopHideTimer?.invalidate()
 
         // Cancel all ongoing geocoding operations
         for geocoder in activeGeocoders.values {
@@ -415,7 +417,7 @@ public class MapRegionManager: NSObject,
         }
     }
 
-    private func displayUniqueStopAnnotations() {
+    private func displayUniqueStopAnnotations(notifyDelegates: Bool = true) {
         var bookmarksHash = [StopID: Bookmark]()
         // When multiple bookmarks exist for the same stop, the last one in the bookmarks array takes precedence
         for bm in bookmarks {
@@ -484,7 +486,9 @@ public class MapRegionManager: NSObject,
         }
         mapView.addAnnotations(stopsToAdd)
         refreshAnnotationViews(for: Array(affectedStopIDs))
-        notifyDelegatesStopsChanged()
+        if notifyDelegates {
+            notifyDelegatesStopsChanged()
+        }
     }
 
     private func refreshAnnotationViews(for affectedStopIDs: [StopID]) {
@@ -503,6 +507,7 @@ public class MapRegionManager: NSObject,
                 continue
             }
             view.prepareForReuse()
+            view.clusteringIdentifier = StopAnnotationView.stopClusterIdentifier
             view.annotation = annotation
             view.delegate = self
             mapViewDelegate?.mapRegionManager(self, customize: view)
@@ -511,13 +516,25 @@ public class MapRegionManager: NSObject,
     // MARK: - Zoom In Warning
 
     private static let requiredHeightToShowStops = 40000.0
+    private static let stopHideThreshold = 42000.0
+    private static let stopShowThreshold = 38000.0
+    private static let stopHideDelay: TimeInterval = 0.35
+    private var shouldShowStopAnnotations = true
 
     public var zoomInStatus: Bool {
         mapView.visibleMapRect.height > MapRegionManager.requiredHeightToShowStops
     }
 
-    private func updateZoomWarningOverlay(mapHeight: Double) {
-        notifyDelegatesZoomInStatus(status: zoomInStatus)
+    static func shouldShowStopAnnotations(previouslyShowing: Bool, mapHeight: Double) -> Bool {
+        if previouslyShowing {
+            return mapHeight <= stopHideThreshold
+        } else {
+            return mapHeight <= stopShowThreshold
+        }
+    }
+
+    private func updateZoomWarningOverlay(showZoomWarning: Bool) {
+        notifyDelegatesZoomInStatus(status: showZoomWarning)
     }
 
     // MARK: - Search
@@ -673,12 +690,29 @@ public class MapRegionManager: NSObject,
             return
         }
 
-        updateZoomWarningOverlay(mapHeight: mapView.visibleMapRect.height)
+        let mapHeight = mapView.visibleMapRect.height
+        shouldShowStopAnnotations = Self.shouldShowStopAnnotations(previouslyShowing: shouldShowStopAnnotations, mapHeight: mapHeight)
+        updateZoomWarningOverlay(showZoomWarning: !shouldShowStopAnnotations)
 
-        guard mapView.visibleMapRect.height <= MapRegionManager.requiredHeightToShowStops else {
-            mapView.removeAnnotations(type: Stop.self)
+        guard shouldShowStopAnnotations else {
+            if stopHideTimer == nil {
+                stopHideTimer = Timer.scheduledTimer(
+                    timeInterval: MapRegionManager.stopHideDelay,
+                    target: self,
+                    selector: #selector(hideStopAnnotations),
+                    userInfo: nil,
+                    repeats: false
+                )
+            }
+            regionChangeRequestTimer?.invalidate()
             return
         }
+
+        stopHideTimer?.invalidate()
+        stopHideTimer = nil
+
+        //show existing stops instantly while new data is loading in the background
+        displayUniqueStopAnnotations(notifyDelegates: false)
 
         let visibleStops = mapView.annotations(in: mapView.visibleMapRect).filter(type: Stop.self)
         for s in visibleStops {
@@ -690,6 +724,17 @@ public class MapRegionManager: NSObject,
         regionChangeRequestTimer?.invalidate()
 
         regionChangeRequestTimer = Timer.scheduledTimer(timeInterval: 0.25, target: self, selector: #selector(requestDataForMapRegion(_:)), userInfo: nil, repeats: false)
+    }
+
+    @objc private func hideStopAnnotations() {
+        stopHideTimer?.invalidate()
+        stopHideTimer = nil
+
+        guard !shouldShowStopAnnotations else {
+            return
+        }
+
+        mapView.removeAnnotations(type: Stop.self)
     }
 
     private var isHidingRegions: Bool? {
@@ -793,6 +838,7 @@ public class MapRegionManager: NSObject,
         }
 
         if let stopAnnotation = annotationView as? StopAnnotationView {
+            stopAnnotation.clusteringIdentifier = StopAnnotationView.stopClusterIdentifier
             stopAnnotation.delegate = self
             mapViewDelegate?.mapRegionManager(self, customize: stopAnnotation)
         }
--- a/OBAKit/Mapping/MapViewController.swift
+++ b/OBAKit/Mapping/MapViewController.swift
@@ -27,6 +27,7 @@ class MapViewController: UIViewController,
     MapRegionMapViewDelegate,
     ModalDelegate,
     MapPanelDelegate,
+    ClusterStopPickerViewControllerDelegate,
     UIContextMenuInteractionDelegate,
     UILargeContentViewerInteractionDelegate,
     UIGestureRecognizerDelegate {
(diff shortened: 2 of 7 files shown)


Aniket-pd:
@aaronbrethorst , I’d appreciate your feedback on this.

Aniket-pd:
Note that this is currently just a prototype. the UI/UX will be enhanced after receiving feedback

OBAKit/Mapping/MapRegionManager.swift
```diff
@@ -673,13 +690,30 @@ public class MapRegionManager: NSObject,
...
+            if stopHideTimer == nil {
+                stopHideTimer = Timer.scheduledTimer(
+                    timeInterval: MapRegionManager.stopHideDelay,
+                    target: self,
+                    selector: #selector(hideStopAnnotations),
+                    userInfo: nil,
+                    repeats: false
+                )
+            }
+            regionChangeRequestTimer?.invalidate()
```
Vedd-Patel, line 707: `stopHideTimer` only creates the timer when its nil 
i wanna say it means only the first zoom-out within 0.35 s arms the timerr
if the user zooms out, zooms back in (cancels should fire), then zooms out again quickly, the second zoom out never re arms the timer because the first was invalidated but the nil check path was not reevaluated in the same run-loop tick
in order to this stops remain visible past the hide-threshold indefinitely until a third event triggers it

OBAKit/Mapping/MapRegionManager.swift
```diff
@@ -673,13 +690,30 @@ public class MapRegionManager: NSObject,
...
             return
         }
 
+        stopHideTimer?.invalidate()
+        stopHideTimer = nil
```
Vedd-Patel, line 712: `Timer.scheduledTimer` must be called on the main run loop, if `mapView(_:regionDidChangeAnimated:)` is ever dispatched off-main, the timer fires on a background thread and mutates the annotation layer??

OBAKit/Mapping/MapViewController.swift
```diff
@@ -897,6 +1043,10 @@ class MapViewController: UIViewController,
             // and just go directly to pushing the stop onto the navigation stack.
             application.analytics?.reportEvent(pageURL: "app://localhost/map", label: AnalyticsLabels.mapStopAnnotationTapped, value: nil)
             show(stop: stop)
+        } else if let cluster = view.annotation as? MKClusterAnnotation {
+            let members = Self.clusterStopMembers(from: cluster.memberAnnotations)
+            displayClusterStopPicker(members: members)
+            mapView.deselectAnnotation(cluster, animated: false)
```
Vedd-Patel, line 1049: `clusterStopMembers(from: cluster.memberAnnotations) `silently drops non-stop annotations (bookmarks that aren't associated with a stop, user-dropped pins, etc.) but never surfaces that to the caller and if all members are non-stop the picker is shown with 0 item!!

OBAKit/Mapping/MapRegionManager.swift
Vedd-Patel, line 707: it calls `regionChangeRequestTimer?.invalidate()` inside the block that arms the stopHideTimer
this cancels any pending map region data fetch whenever the user zooms out far enough to trigger pin hiding

Vedd-Patel (changes requested):
@Aniket-pd 
will check in more detail soon!
would love to hear your POV on these reviews!!

OBAKit/Mapping/MapRegionManager.swift
Aniket-pd: This invalidation is intentional .
When we are in the “hide stop annotations” state, pending stop-region fetches are canceled on purpose to avoid loading data for a zoom level where stop pins are not shown.

Aniket-pd:
@aaronbrethorst could you take a look at this PR when you get a chance?

Vedd-Patel:
@coderabbitai full review

aaronbrethorst (changes requested):
Hi Aniket — thanks for this. The core idea here is genuinely nice: leaning on native `MKMapView` clustering instead of hand-rolled grouping, adding hysteresis so pins stop flickering at the zoom boundary, and a clean SwiftUI sheet for picking among co-located stops. The production code is well-organized — extracting `shouldShowStopAnnotations(previouslyShowing:mapHeight:)` as a pure static function was exactly the right testability move, and `StopAnnotationViewTests` targets the precise regression this PR fixes.

But I can't merge it: **the test target doesn't compile.** I generated the project and ran `xcodebuild build-for-testing` against this branch — it fails with exit code 65. That means CI is red and *none* of the new tests actually run. It also tells me the test files were written against an earlier iteration of the production API and never re-built. Let's get them green.

## Critical Issues (1 found)

**The test target fails to compile — three distinct errors.** Verified locally via `xcodebuild build-for-testing -scheme App` (exit 65). Because these all live in the test target, the whole suite fails to build and CI cannot run a single test.

1. **`MockStopAnnotationDelegate` can't conform to an `@objc` protocol as a plain Swift class** — `OBAKitTests/Application/Mapping/StopAnnotationViewTests.swift:73`. `StopAnnotationDelegate` refines `NSObjectProtocol`, so the mock must inherit `NSObject`:
   ```
   error: cannot declare conformance to 'NSObjectProtocol' in Swift;
   'MockStopAnnotationDelegate' should inherit 'NSObject' instead
   ```
   Fix: `private final class MockStopAnnotationDelegate: NSObject, StopAnnotationDelegate`.

2. **`ClusterStopMember` initializer is missing `directionText`** — `OBAKitTests/Application/Mapping/MapViewControllerGroupedStopPickerTests.swift:39-41, 50-51`. The production struct (`MapViewController.swift:639-645`) has five stored fields — `stopID, title, directionText, subtitle, isBookmarked` — with no custom init, so the synthesized memberwise initializer requires all five. Every test call site omits `directionText` (and passes `subtitle` positionally where `directionText` is expected). Add `directionText:` to each.

3. **`MapViewController.clusterStopActionTitle(for:)` doesn't exist** — `MapViewControllerGroupedStopPickerTests.swift:53-54`. This is a stale reference to a renamed/removed API; the production equivalent is `clusterStopPickerSecondaryText(for:)` (`MapViewController.swift:728`). It also has *different semantics*: it returns only the secondary line formatted as `"%1$@ · %2$@"` or `"Direction: %@"`, never `"Pine & 3rd (NW)"`. So renaming the call isn't enough — the test needs rewriting against the four real branches of `clusterStopPickerSecondaryText` (direction+subtitle, direction-only, subtitle-only, neither).

## Important Issues (2 found)

- **`test_zoomWarning_matchesVisibilityState` is tautological** — `MapRegionManagerTests.swift:122-134`. Its name implies it verifies the zoom-warning overlay tracks visibility, but it just calls `shouldShowStopAnnotations` again and double-negates the result (`expect(!showPins).to(beFalse())`). It never touches `updateZoomWarningOverlay` or any delegate. Either wire up a delegate spy and assert the emitted status, or drop it.

- **The highest-risk new behavior is untested: the deferred hide/show flow.** `reloadStopAnnotations` (`MapRegionManager.swift:690-737`) is where the real user-visible risk lives — zooming out schedules `stopHideTimer` (0.35s) which removes annotations only if still hidden; zooming back in must invalidate that timer and re-display via `displayUniqueStopAnnotations(notifyDelegates: false)`. The tests only exercise the pure `shouldShowStopAnnotations` function. A regression here (timer not invalidated on re-show → pins vanish after zooming back in) would sail through. Worth a test that drives the hide→show transition.

## Suggestions (4 found)

- **Unused localization string** — `OBAKit/Strings/en.lproj/Localizable.strings:338`. `"map_controller.grouped_stop_picker.cancel" = "Cancel";` isn't referenced anywhere (grep confirms); the SwiftUI sheet uses an `xmark` close button with `Strings.close`. Leftover from the earlier action-sheet design — remove it.
- **Three identical `makeStop(id:name:)` helpers** — `MapRegionManagerTests.swift:170`, `MapViewControllerGroupedStopPickerTests.swift:57`, `StopAnnotationViewTests.swift:53`. Consolidate into a single `Fixtures.stop(id:name:)` (CodeRabbit flagged this too).
- **No coverage for `clusterStopPickerSecondaryText`'s four branches** once the broken `clusterStopActionTitle` test is removed — this is the most logic-dense new function (including `nilifyBlankValue` blank-vs-nil handling). Add it when you rewrite test #3.
- **`swiftlint` directive typo** — `MapRegionManagerTests.swift:16` reads `// swiftlintXdisable force_try` (a no-op). Harmless since each force-try has its own inline disable, but misleading.

## Strengths

- Production code compiles clean and reads well; the hysteresis band + deferred-hide approach is a sound fix for the flicker problem.
- `StopAnnotationViewTests` directly guards the regression this PR fixes — the clustering identifier surviving `init`, `prepareForReuse`, and `prepareForDisplay` for both stops and bookmarks.
- `test_mapView_viewForStop_resetsClusteringIdentifierOnReuse` tests the reuse path through the real `MapRegionManager`, complementing the view-level tests.
- `stopHideTimer` is correctly invalidated in the cleanup path (`MapRegionManager.swift:227`).
- The dedup/bookmark-priority logic in `clusterStopMembers(from:)` is correct, and the two dedup tests that *do* match the real API are behaviorally sound (once the file compiles).

## Recommended Action

1. Fix the three compile errors so the test target builds (critical — start here).
2. Rewrite the `clusterStopActionTitle` test against `clusterStopPickerSecondaryText`'s four branches; add a hide/show-timer transition test.
3. Replace the tautological `test_zoomWarning_matchesVisibilityState` with a real delegate-spy assertion or remove it.
4. Sweep the suggestions (unused string, dedup the three `makeStop` helpers, swiftlint typo).
5. Run `xcodebuild build-for-testing` + the unit tests locally to confirm green, then re-request review.

**Verdict: Request Changes.**

aaronbrethorst:
### Code review

Found 1 issue:

1. The new/extended tests are written against a test stack that no longer exists on `main`: both new files are `XCTestCase` subclasses importing `XCTest`, and the additions to `MapRegionManagerTests.swift` use Nimble's `expect(...).to(beTrue())` plus `XCTFail`. On `main`, `MapRegionManagerTests` is a Swift Testing `@Suite(.serialized)` over a non-`XCTestCase` `OBATestCase`, and Nimble is gone (the only remaining `import XCTest` is `PolylinePerformanceTests`, the documented exception). As written these files can't compile or run, so none of the new coverage actually executes. (CLAUDE.md says "`OBAKitTests/`: Unit tests for all frameworks. Written with **Swift Testing** (`@Suite` / `@Test` / `#expect`), not XCTest. The single exception is `PolylinePerformanceTests`" and "**Testing**: Swift Testing (first-party; Nimble was removed)")

https://github.com/OneBusAway/onebusaway-ios/blob/b97f9cd46fca3033172962f0c93b3dcd2a5e148b/OBAKitTests/Application/Mapping/MapViewControllerGroupedStopPickerTests.swift#L9-L16

aaronbrethorst (changes requested):
Coming back to this after a while, so let me start with what I still think is right about it: using MapKit's own clustering with one identifier re-applied in `init`, `prepareForReuse`, and `prepareForDisplay`, plus a hysteresis band instead of a single threshold, is a genuinely good answer to the flicker and decluster problems in #515. That core mechanic is maybe sixty lines and I'd like it in the app.

The trouble is that everything around it has gone stale, and I want to be straight with you about the size of the gap rather than send you off to chase a moving target.

- The three compile errors I described in June are all still present at the head: `MockStopAnnotationDelegate` needs to inherit `NSObject` for the `@objc` `StopAnnotationDelegate`, the `ClusterStopMember(...)` call sites omit `directionText`, and `clusterStopActionTitle(for:)` doesn't exist — the real API is `clusterStopPickerSecondaryText(for:)`, with different semantics.
- On top of that, and this one is new since you last pushed: the test files here are `XCTestCase` subclasses using Nimble matchers, and `main` migrated the whole suite to Swift Testing in July. Nimble isn't a dependency any more. So the 261 lines of tests would need rewriting as `@Suite` / `@Test` / `#expect` before any of them could run.
- `MapRegionManager.swift`, `MapViewController.swift`, and `en.lproj/Localizable.strings` all conflict with `main`.
- `main` has also since made `requiredHeightToShowStops` a deliberately shared constant so the UIKit map and the SwiftUI panel agree on when stops appear. The 38000/42000 band here applies only to the UIKit path, so after a rebase the two surfaces would disagree inside the band. `RentalAnnotationView` has also established a clustering convention this would need to fit into.

Honestly assessed, that's closer to a rewrite than a rebase.

My suggestion — and it is a suggestion, your call entirely — is to close this and open a small PR against current `main` with just the clustering identifier and the hysteresis band, no picker UI and no test-target changes beyond a Swift Testing suite for the threshold logic. That lands the part with the clearest value, gets you a green build to work from, and the cluster picker can follow separately once the mechanic is in.

If you'd rather push this one forward as-is, that's fine too and I'll review it — I just didn't want you to sink an evening into a rebase without knowing what's on the other side. Either way, thanks for the original diagnosis; it was correct and it's still worth shipping.

Issue #10, opened 2019-06-12
Add bookmark management features


aaronbrethorst:
# Features:

## Groups

- [x] Rename group
- [x] Rearrange groups
- [x] Delete group
- [x] Add new group
- [x] Add new group and immediately move it

## Bookmarks

- [x] Rename
- [x] Delete
- [x] Add new group and immediately add a bookmark to it

PR #1135, opened 2026-05-16
Shared ViewModel Layer (Phases 1–3)

mosliem:
## Summary

Extracts business logic from five heavyweight UIKit view controllers into shared `@MainActor ObservableObject` ViewModels. The UIKit VCs become thin, Combine-bound UI shells with identical behavior. The same ViewModels will be consumed directly by incoming SwiftUI sheets, eliminating duplication when the new UI layer is built.

---

## Motivation

OneBusAway iOS has ~38 view controllers that mix business logic and UI code. The app is getting a new SwiftUI sheet-based UI. To avoid duplicating logic, all business logic must live in platform-agnostic ViewModels that both UIKit and SwiftUI can share.

**Architecture target:**
```
Services (unchanged)
       ↓
ViewModels (@MainActor + ObservableObject, NO UIKit/SwiftUI imports)
       ↓                    ↓
Old UIKit VCs          New SwiftUI Sheets
(thin Combine shells)  (future — consumes same VMs)
```

---

### New Files — ViewModels

| File | Extracted From | Key State |
|------|---------------|-----------|
| `OBAKit/ViewModels/StopViewModel.swift` | `StopViewController` | stop, stopArrivals, isLoading, operationError, isBrokenBookmark, statusText, stopPreferences, isListFiltered, minutesAfter, isLoadMoreExhausted, surveysRefreshToken, navigationIntent |
| `OBAKit/ViewModels/MapViewModel.swift` | `MapViewController` | weather, isLoadingWeather, showZoomWarning, mapType, locationAuthStatus |
| `OBAKit/ViewModels/MapPanelViewModel.swift` | `MapFloatingPanelController` | nearbyStops, highSeverityAlerts, requestedPanelDetent, isSearchMode, searchResponse |
| `OBAKit/ViewModels/TripViewModel.swift` | `TripViewController`| tripDetails, serviceAlerts, isLoading, mapPolyline, shouldSkipProgrammaticRefresh |
| `OBAKit/ViewModels/BookmarksViewModel.swift` | `BookmarksViewController` | updateToken, sortByGroup |
| `OBAKit/ViewModels/Models/NavigationIntent.swift` | — | `showStop`, `showArrivalDeparture`, `showAlert`, `showSchedule`, `showNearbyStops` |

### Modified Files — VC Wiring

| File | Change |
|------|--------|
| `OBAKit/Stops/StopViewController.swift` | Holds `StopViewModel`; `bindViewModel()` with Combine sinks |
| `OBAKit/Trip/TripViewController.swift` | Holds `TripViewModel`; injects VoiceOver closure via `shouldSkipProgrammaticRefresh` |
| `OBAKit/Mapping/MapViewController.swift` | Holds `MapViewModel`; `bindViewModel()` for mapType + locationAuthStatus; `toggleMapType()` delegates to VM |
| `OBAKit/Mapping/MapFloatingPanelController.swift` | Holds `MapPanelViewModel`; `bindViewModel()` for stops, alerts, panel detent |
| `OBAKit/Bookmarks/BookmarksViewController.swift` | Holds `BookmarksViewModel`; `$updateToken` triggers `applyData()` + widget reload |
| `OBAKit/Recent/RecentStopViewModels.swift` | Renamed `StopViewModel` list-row type to `StopRowItem` to avoid collision with the new class |

---

### Phase 1 — `StopViewModel`
Highest-value extraction. `StopViewController` was 1568 lines with timers, preference persistence, auto-refresh, filter logic, section building, and analytics all mixed into the VC.

**Extracted to `StopViewModel`:**
- `refresh()` — API fetch + error handling + broken-bookmark detection
- `startAutoRefresh()` / `stopAutoRefresh()` — 15 s timer, 30 s stale threshold
- `loadMore()` — extends time window in 30–60 min increments up to 12 h
- `updatePreferences(_:)` — hidden routes, sort order persistence
- `toggleFilter()` — switches filtered/unfiltered views
- `reportStopViewed()` / `recordRecentStop()` — analytics + user data store
- `fetchSurveys()` — increments `surveysRefreshToken` so VC can call `applyData()`
- Navigation via `@Published var navigationIntent: NavigationIntent?`

**Stayed in `StopViewController`:**
- `OBAListView` setup, cell registration, layout
- Navigation bar buttons, swipe actions, context menus
- Modal presentation (AlarmBuilder, BookmarkEditor, StopPreferencesView)
- Pull-to-refresh, idle timer, `firstLoad` collapsed-section restore

### Phase 2 — `MapViewModel` + `MapPanelViewModel`
`MapViewController` wraps `MapRegionManager` (which already handles much of the stop-loading). The ViewModel consolidates the scattered state that the VC owned directly.

**`MapViewModel` (NSObject subclass — required for `@objc` delegate protocols):**
- Adopts  `LocationServiceDelegate`
- `loadWeather()` — async fetch from ObacoService
- `toggleMapType()` — persists choice to UserDefaults
- Publishes `mapType`, `locationAuthStatus`, `weather`, `showZoomWarning`

**`MapPanelViewModel`:**
- Receives nearby stops from `MapRegionManager`
- `loadAlerts()` — fetches high-severity agency alerts
- Platform-agnostic `PanelDetent` enum (UIKit maps to `FloatingPanelState`)
- `isSearchMode` drives the panel expansion on search activation

### Phase 3 — `TripViewModel` + `BookmarksViewModel`
Completes Tier-1 extraction.

**`TripViewModel`:**
- `loadTrip()` — fetches trip details + shape + vehicle status
- `startAutoRefresh()` / `stopAutoRefresh()` — 30 s timer
- Injectable `shouldSkipProgrammaticRefresh: (() -> Bool)?` closure — VC injects a VoiceOver check without the VM needing UIKit

**`BookmarksViewModel`:**
- `updateToken` integer — incremented when bookmarks change; VC observes to call `applyData()` and reload widgets
- `sortByGroup` flag — drives section ordering, persisted via `UserDefaults`

---

## Architecture Rules Established

These rules apply to all future ViewModels in this project:

1. **No UIKit or SwiftUI imports** in any ViewModel file — Foundation, Combine, MapKit, OBAKitCore only.
2. All state mutations are `@MainActor`; `@objc` delegate callbacks use `nonisolated func ... { Task { @MainActor in ... } }`.
3. `NSObject` subclass **only** when the ViewModel must adopt an `@objc` delegate protocol.
4. Navigation via `@Published var navigationIntent: NavigationIntent?` — UI observes, handles, then nils it.
5. Panel state via `PanelDetent` enum — UI maps to platform-specific detent types.
6. Injectable closures instead of UIKit imports for platform-specific checks.

---

## What's Not Included (Next Steps)

Per the migration plan, Tier-2 and Tier-3 extractions remain:

- `RecentStopsViewModel`
- `SearchViewModel`
- `CurrentTripViewModel`
- `SettingsViewModel`, `MoreViewModel`
- SwiftUI sheet layer consuming the shared ViewModels (P0 screens: Map, Stop, Bookmarks, Search, Map Panel)

---

## Testing

- All existing `OBAKitTests` pass after each phase.
- Manual QA verified the old UIKit UI is visually and behaviorally identical after each refactor.
- No new public API surface — changes are internal to `OBAKit`.

## Summary by CodeRabbit

* **Refactor**
  * Migrated many screens to Combine-driven view models for clearer lifecycle, state, and UI bindings.
  * Centralized map, bookmarks, stop, and trip logic into dedicated view models for consistency and testability.
  * Improved loading state handling and refresh behavior to avoid stuck or duplicate loads.

* **Bug Fixes**
  * Prevented cancellation-related errors from surfacing to users and ensured refresh indicators end reliably.

* **Tests**
  * Added comprehensive view-model tests, fixtures, and mocks covering sorting, loading, refresh, and edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OneBusAway/onebusaway-ios/pull/1135?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


mosliem:
@coderabbitai review

Vedd-Patel:
@mosliem prefer "**@coderabbbitai full review**" instead of review
(intentionally written 3 'b' instead of 2 in coderabbit 😅 )

mosliem:
@coderabbitai full review

aaronbrethorst (changes requested):
Mohamed — this is a serious piece of work. A clean extraction of business logic out of five of the heaviest controllers in this codebase, with sensible architectural rules ("no UIKit imports in VMs", "NSObject only when adopting `@objc` delegates", `PanelDetent` as a platform-agnostic shim), and the framing of why we're doing it now (the SwiftUI sheet layer) is exactly right. The bones are good.

That said, this PR isn't ready to merge. There are two user-visible regressions, a substantial chunk of dead infrastructure, and zero test coverage for a refactor whose whole premise is "the VMs are now the single source of truth for behavior, which both UIKit and SwiftUI will share." The first time the SwiftUI consumers land on top of these VMs, untested behavior becomes load-bearing in two UI layers at once. Let's lock the behavior in before that happens.

Verdict: **Request Changes.**

---

## Critical Issues (2 found)

### 1. Analytics and "recent stops" now fire on every refresh — and again on each auto-extend step
`OBAKit/ViewModels/StopViewModel.swift:157-164`

In the old code, `reportStopViewed` and `addRecentStop` only ran when the stop value actually changed (`stop != oldValue` in `didSet`). Effectively once per visit.

In the new VM they fire inside `refresh()` on every successful fetch. The auto-refresh timer ticks every 15 s and `shouldRefresh` gates on a 30 s threshold, so a user sitting on a stop for five minutes generates about ten `stop_viewed` events and re-adds the same stop to recents on each one. It gets worse when `arrivalsAndDepartures.isEmpty` — `autoExtendWindowIfNeeded()` re-enters `refresh()` walking the window 60 → 120 → 240 → 480 → 720 minutes, firing analytics and recents at each step. A late-night stop with no arrivals can rack up 5+ side-effect calls per single visit.

Gate both calls on a "first time we've seen this stop ID in this VM instance" check, or move them out of `refresh()` entirely into a `$stop` sink that uses `removeDuplicates` on `stop.id`.

### 2. Saved "all routes hidden" preference renders an empty list on first load
`OBAKit/ViewModels/StopViewModel.swift:194,257-261`

The old `stopUpdated(stop)` had this check: if `isListFiltered` and every route in the stop is in `stopPreferences.hiddenRoutes`, turn the filter off so the user doesn't see an empty screen. That ran on every stop assignment.

The new `disableFilterIfAllRoutesHidden()` is private and only called from `updateStopPreferences(_:)`. The init at line 108 loads previously-saved preferences but never re-checks the invariant. So a user who hid every route at a stop, left, and returns will see an empty list with no obvious recovery path other than navigating into the filter menu.

Call `disableFilterIfAllRoutesHidden()` once inside `refresh()` after `self.stop = loadedStop` is assigned (or in the VC's `$stop` sink before the first `applyData`).

---

## Important Issues (8 found)

### 3. `NavigationIntent` is fully dead infrastructure as merged
`OBAKit/ViewModels/Models/NavigationIntent.swift`, `OBAKit/ViewModels/StopViewModel.swift:69`, `OBAKit/Stops/StopViewController.swift:1353-1376`

The five-case enum, the `@Published var navigationIntent`, the 24-line dispatch sink in `bindNavigationIntents` — nothing in the codebase ever sets `viewModel.navigationIntent` to a non-nil value. All the navigation paths (`showStop`, `showArrivalDeparture`, `showAlert`, `showSchedule`, `showNearbyStops`) are still handled by direct calls on `application.viewRouter` from the VC. This is scaffolding for the SwiftUI sheets, but as merged it adds API surface for zero current behavior.

Either delete it and re-introduce when the SwiftUI consumer lands in the same PR, or actually route at least one navigation case through it so the wiring is exercised end-to-end.

While we're here: the publishing shape is wrong even when populated. `@Published var navigationIntent: NavigationIntent?` requires the consumer to reset it to `nil` after handling. That's an event-as-state pattern that re-fires on every new subscription. A `PassthroughSubject<NavigationIntent, Never>` exposed as `AnyPublisher` would (a) make nil unrepresentable, (b) drop the manual reset, (c) not replay to new subscribers. Same pattern applies to `surveysRefreshToken` and `BookmarksViewModel.updateToken` — adopt one rule for the VM layer: state is `@Published`, events are `AnyPublisher` backed by `PassthroughSubject`.

### 4. `MapViewModel.loadWeather()` silently swallows all errors with no logging
`OBAKit/ViewModels/MapViewModel.swift:74-83`

The pre-refactor code (still present, now dead, at `MapViewController.swift:394-408`) logged via `Logger.error(error.localizedDescription)`. The new path catches everything and just sets `weather = nil`. An obacoService outage, a schema change in `WeatherForecast`, an SSL hiccup — all indistinguishable from "no weather available for this region." Add `Logger.error("Failed to load weather: \(error.localizedDescription)")` in the catch.

### 5. `TripViewModel.loadData()` partial success leaves the UI in a Frankenstein state
`OBAKit/ViewModels/TripViewModel.swift:90-119`

The three child tasks each mutate published state directly (`tripConvertible`, `tripDetails`, `routePolylineCoordinates`). If `loadTripDetails()` succeeds and `loadTripConvertible()` then throws, the user sees fresh trip-status against a stale arrival/departure, and `displayError(error)` doesn't disclose the half-update.

Stage results into local `let` constants inside the task group and assign to `@Published` properties only after `waitForAll()` succeeds. If a subtask throws, no published state mutates and the error path is clean.

### 6. Dead protocol method and implementation: `mapPanelControllerDisplaySearch`
`OBAKit/Mapping/MapFloatingPanelController.swift:22`, `OBAKit/Mapping/MapViewController.swift:866-868`

Declared, implemented, no callers. Panel state changes now flow through `viewModel.$requestedPanelDetent` → `mapPanelController(_:moveTo:)`. Remove both the protocol requirement and the impl.

### 7. Dead code in MapViewController: `loadWeather()` and the `forecast` plumbing
`OBAKit/Mapping/MapViewController.swift:381-408`

`loadWeather()` is no longer called from anywhere — weather is loaded inside `MapViewModel.loadWeather()` and observed via the `$weather` sink. The local `forecast` property is now only mutated from that sink, but `loadWeather()` still exists as dead code. Delete it.

### 8. Auto-extend window's "fragile sequencing" comment is a test specification, not documentation
`OBAKit/ViewModels/StopViewModel.swift:231-243`

The inline comment is honest:

> If `refresh()` is ever refactored to yield before that assignment, auto-extension will silently stop working.

That's exactly the kind of silent failure mode that survives a code review and dies six months later when someone touches the surrounding control flow. Cheap fix: pull the extension request out of the spawned `Task` and drive it from a single linear caller — set a private flag inside `refresh()`, then trigger `loadMore` after `isLoading = false` in the same scope. Express the invariant in code, not prose. At a minimum, write a unit test that asserts the 60 → 120 → 240 → 480 → 720 walk and the `isLoadMoreExhausted` flag flip at the cap.

### 9. No tests added for a 1246-line refactor whose entire thesis is testability
`OBAKitTests/` — `git diff main..HEAD -- OBAKitTests/` is empty.

The whole point of `@MainActor ObservableObject` with no UIKit imports is that you can unit-test it. The existing test infrastructure (`OBATestCase`, `MockDataLoader`, `AnalyticsMock`, the `Application` construction pattern in `OBAKitTests/Schedules/ScheduleForStopViewModelTests.swift`) already supports what these VMs need. Highest-value tests to add before merge:

- `StopViewModel.loadMore()` caps at 720 minutes and flips `isLoadMoreExhausted`.
- `StopViewModel.autoExtendWindowIfNeeded` walks 60→120→240→480→720 (regression test for #8).
- `StopViewModel`'s analytics/recents fire at most once per stop ID (regression test for #1).
- `StopViewModel.disableFilterIfAllRoutesHidden` runs on initial stop load (regression test for #2).
- `StopViewModel.shouldRefresh` threshold logic (nil → true, <30 s → false, >30 s → true).
- `TripViewModel.loadData()` does not surface a user-facing error when the only failure was `CancellationError`.
- `BookmarksViewModel` reads and writes the `OBABookmarksController_SortBookmarksByGroup` UserDefaults key correctly.

Follow the shape of `ScheduleForStopViewModelTests` — no new mocking infrastructure required.

### 10. PR description claims `MapViewModel` adopts `MapRegionDelegate` — it doesn't
`OBAKit/ViewModels/MapViewModel.swift:24`, `OBAKit/Mapping/MapViewController.swift:27`

`MapViewModel` adopts only `LocationServiceDelegate`. `MapRegionDelegate` is still owned by `MapViewController` (and `MapFloatingPanelController`). Either move the delegate to the VM as the description claims, or correct the description before merge.

---

## Suggestions (5 found)

### S1. `@Published` `$stop` re-emits on every refresh
`OBAKit/Stops/StopViewController.swift:1304-1311`

`refresh()` assigns `self.stop = loadedStop` unconditionally, and `@Published` re-emits regardless of value equality. The `$stop` sink fires every 30 s, re-running `listView.applyData(animated: false)`, `configureTabBarButtons()`, and re-setting `title` to the same string. Combined with the `$stopArrivals` and `$stopPreferences` sinks (which also call `applyData`), each successful refresh triggers multiple redundant list rebuilds. Either `.removeDuplicates()` on `$stop` (Stop is Equatable) or only assign when changed.

### S2. Retain-cycle pattern in `StopViewModel.startAutoRefresh` is the same shape you fixed in `TripViewModel` in `d3d2eaf2`
`OBAKit/ViewModels/StopViewModel.swift:284-289`

```swift
Timer.scheduledTimer ... { [weak self] _ in
    Task { @MainActor in
        if self.shouldRefresh { await self.refresh() }
    }
}
```

The inner `Task` captures `self` strongly. Apply the same fix you made in `TripViewModel` — move `[weak self]` onto the inner Task and re-guard.

### S3. `BookmarksViewModel` has no `deinit` to cancel the data loader's repeating timer
`OBAKit/ViewModels/BookmarksViewModel.swift`

`BookmarkDataLoader` schedules a repeating 30 s timer on the run loop. If `deactivate()` is never called (a future SwiftUI consumer forgets, or the VM is reparented), the timer fires forever. The closure is `[weak self]` so it's a no-op leak, but the RunLoop entry persists. Add `deinit { dataLoader.cancelUpdates() }`.

Related: `dataLoaderDidUpdate` is declared `nonisolated` with an inner `Task { @MainActor [weak self] }`. `BookmarkDataLoader` already calls the delegate inside `await MainActor.run`, so the round-trip is redundant. Either make the method `@MainActor` directly, or document why you're defending against a future loader change.

### S4. `BookmarksViewController.refreshControlPulled` ends the spinner on a fixed 2 s timer, not on real completion
`OBAKit/Bookmarks/BookmarksViewController.swift:154-162`

This isn't new in this PR, but it's adjacent and worth fixing while you're here: the spinner is stopped after 2 s regardless of whether the refresh succeeded, failed, or is still in flight. The VM already has `$updateToken`; tie the spinner to that, or expose an `isLoading: Bool` derived from the data loader state and observe it.

### S5. `PanelDetent` doc-comment promises platform portability that the three-case enum can't deliver
`OBAKit/ViewModels/MapPanelViewModel.swift:14-20`

The comment says "UIKit maps this to FloatingPanel states; SwiftUI maps to `presentationDetents`." SwiftUI's `PresentationDetent` supports `.fraction(0.3)` and `.height(180)` — anything beyond tip/half/full forces a new case and a new mapping on both sides. Either add `.fraction(Double)` / `.height(CGFloat)` cases now (since the SwiftUI sheets are imminent), or rename to `MapPanelPosition` and drop the SwiftUI framing.

---

## Strengths

A lot of the design choices here are exactly right and worth calling out:

- The "no UIKit/SwiftUI imports in any ViewModel file" rule is the single most important constraint and you held the line on it across all five VMs.
- The `@MainActor` plus `nonisolated func ... { Task { @MainActor in ... } }` pattern for `@objc` delegate callbacks is the right idiom for Swift 6 — and `MapViewModel`/`BookmarksViewModel` only inherit `NSObject` where that delegate adoption requires it.
- The injectable `shouldSkipProgrammaticRefresh` closure in `TripViewModel` is a clean way to let the VM stay accessibility-aware without importing `UIKit` — same shape works for any host-environment query the SwiftUI consumers will need.
- Splitting the binding code into `bindListData() / bindLoadingState() / bindNavigationIntents()` etc. in private extensions is exactly how I'd want this organized.
- The d3d2eaf2 commit fixing the spurious sort update and retain cycles caught real bugs, and the commit message documenting *why* is gold for whoever touches this in six months.
- The acknowledged-fragility comment on `autoExtendWindowIfNeeded` is intellectually honest. You're showing your work. Now turn it into a test (#8) and the trade is fully closed.

---

## Recommended Action

1. **Fix the two regressions** (#1, #2) — both are user-visible.
2. **Decide on `NavigationIntent`** (#3): delete it, or wire one navigation path through it so the round-trip is exercised. As-is it's a misleading API surface.
3. **Remove the dead code** (#6, #7) and **correct the PR description** (#10) — one cleanup commit.
4. **Restructure `TripViewModel.loadData`** (#5) and **harden `autoExtendWindowIfNeeded`** (#8) — both are correctness fixes that are cheap now and expensive later.
5. **Add the `MapViewModel.loadWeather` Logger.error** (#4) — one line.
6. **Land the minimum test set** (#9). Pick the 5-7 highest-leverage assertions above and pattern-match `ScheduleForStopViewModelTests`. Without these, the SwiftUI sheets will be the first integration test these VMs ever get.
7. Re-request review after fixes.

The architectural call is sound. Polish to the standard that this PR's own quality of thinking deserves, and this is a strong merge.

aaronbrethorst (approved):
Mohamed — this is a textbook response to a substantive code review. Every critical and important issue from the prior round is addressed with intent, the new test suite is exactly the shape I asked for, and several of the fixes (the `pendingExtensionMinutes` linear recursion, the `try await (a, b, c)` atomic apply in `TripViewModel`, the `MainActor.assumeIsolated` cleanup in `BookmarksViewModel`) are actually nicer than what I sketched. The architectural rules you established in the original PR description now have running code and tests backing them up.

Verdict: **Merge.**

Filed [#1141](https://github.com/OneBusAway/onebusaway-ios/issues/1141) for three small fit-and-finish items that don't need to gate this merge.

---

## Prior review items — all addressed

### Critical

- **#1 Analytics / recents fire once per VM lifetime** — `hasPerformedInitialStopSetup` gate in `performInitialStopSetupIfNeeded(for:)` (`StopViewModel.swift:202`). Regression test `test_analyticsFiresExactlyOnceAcrossRefreshes` and `test_recentStop_recordedExactlyOnce` pin the behavior.
- **#2 "All routes hidden" filter on returning visit** — `disableFilterIfAllRoutesHidden()` now also runs inside `performInitialStopSetupIfNeeded(for:)` (`StopViewModel.swift:210`). Test `test_disableFilter_runsOnInitialLoadWhenAllRoutesHidden` exercises the exact scenario.

### Important

- **#3 `NavigationIntent`** — fully deleted, and `surveysDidRefresh` / `BookmarksViewModel.didUpdate` migrated to `AnyPublisher` backed by `PassthroughSubject`. Exactly the rule I wanted (state is `@Published`, events are `PassthroughSubject`-backed `AnyPublisher`).
- **#4 `MapViewModel.loadWeather()` logging** — `Logger.error("Failed to load weather: ...")` in the catch (`MapViewModel.swift:88`).
- **#5 `TripViewModel.loadData` atomicity** — `try await (convertible, details, polyline)` rethrows the first failure and cancels siblings; all three `@Published` properties only mutate after the tuple resolves (`TripViewModel.swift:96-108`). Comment on line 100-101 documents *why*. Cancellation gets its own branch so it doesn't surface as a user-visible error, and there's a regression test for that.
- **#6 / #7 Dead code** — `mapPanelControllerDisplaySearch` removed from the protocol and impl; dead `MapViewController.loadWeather()` is gone. The `forecast` `didSet` survives because it's the button-rendering sink — that's correct.
- **#8 Auto-extend window fragile sequencing** — replaced with `pendingExtensionMinutes: UInt?` set inside the `do` block, awaited *after* `isLoading = false` in the same scope (`StopViewModel.swift:150, 178-180`). The invariant is now expressed in code, not prose. `test_autoExtend_walksToCapAndFlipsExhausted` is the regression net.
- **#9 No tests** — five new test files, 21 tests, all passing. Coverage hits every regression I asked for: cap walk, analytics dedup, recents dedup, filter invariant on load, `$stop` re-emit guard, `shouldRefresh` threshold, cancellation suppression, sortByGroup defaults + persistence + reset, `lastBatchHadError` set/reset across batches, weather success + error, `mapType` toggle, location-auth delegate, panel detent transitions.
- **#10 PR description correction** — now correctly says `MapViewModel` adopts only `LocationServiceDelegate`.

### Suggestions

- **S1 `$stop` re-emit guard** — `if self.stop != stop { self.stop = stop }` in `applySuccessfulFetch` (`StopViewModel.swift:192-194`). The block comment explains the trade. `test_stop_doesNotReEmitWhenUnchangedAcrossRefreshes` locks it in.
- **S2 Retain-cycle pattern in auto-refresh** — `[weak self]` is now on the inner `Task @MainActor` for both `startAutoRefresh` and `startStatusTimer` (`StopViewModel.swift:326-331, 350-354`). Same shape as the `TripViewModel` fix in `d3d2eaf2`.
- **S3 `BookmarksViewModel` deinit** — `deinit { dataLoader?.cancelUpdates() }` (`BookmarksViewModel.swift:64-66`). Bonus: `dataLoaderDidUpdate` and `dataLoader(_:isLoadingChanged:)` switched from the round-trip `Task { @MainActor }` to `MainActor.assumeIsolated`, which is the right tool — it both removes the hop and *traps* if the loader contract ever changes. Good defensive programming.
- **S4 Refresh spinner ends on fixed timer** — fully reworked. `bindLoadingState()` in `BookmarksViewController` (`BookmarksViewController.swift:418-435`) ties `refreshControl.endRefreshing()` and the haptic to `viewModel.$isLoading` transitioning to `false`, gated by `isUserRefreshing` so the 30 s background tick doesn't end a non-existent pull spinner. `lastRefreshHadError` from `BookmarkDataLoader` feeds the success/failure haptic so a partial-failure batch doesn't masquerade as a success buzz. `test_refresh_setsAndResetsLastRefreshHadError` covers the set-and-reset across two batches.

---

## Additional things I noticed this round

### Strengths

- **`BookmarkDataLoader` batch-ID staleness gate** (`BookmarkDataLoader.swift:79-86, 112, 125-131`) — the `currentBatchID &+= 1` advance retires in-flight per-bookmark Tasks from cancelled or superseded batches. Both the success path and the error path check the batch ID. The error-path comment explicitly notes the gate is also error-suppression for `cancelUpdates()`, which is the right framing — you're not just preventing a data race, you're preventing the user from seeing a stale "Bad Server Response" alert from a refresh they already left behind. This is the kind of subtle infrastructure work that pays off the third time someone wonders why an unrelated error is flashing.
- **`MapBaseType` enum** (`MapViewModel.swift:18-21`) — adding this alongside `PanelDetent` keeps the VM properly MapKit-free. The doc-comment ("UIKit maps `.standard` → `.mutedStandard` and `.hybrid` → `.hybrid`; SwiftUI can map directly to `MapStyle`") is the contract a future SwiftUI consumer needs.
- **Route `Application.apiService` through `config.dataLoader`** (`CoreApplication.swift:174-176`) — this is the unblock that lets the VM tests assert real behavior against `MockDataLoader`. Quietly important; `obacoService` and `regionsService` were already doing this, so the inconsistency was a latent test gap.
- **`AnalyticsMock.stopViewedCount` / `lastReportedStopID`** — minimal, focused mock surface added only where the regression tests need observation. Didn't bloat the mock infrastructure.
- **`assumeIsolated` over `Task { @MainActor in ... }`** — this is the right idiom when you *know* a callback's caller has already hopped to the main actor, and your comment in `dataLoaderDidUpdate` documents the assumption.

### Filed as #1141 (not blocking merge)

- SwiftLint warnings introduced by the controller shrink: `bindListData()` is 52 lines over the 50-line `function_body_length` cap, and the `swiftlint:disable file_length type_body_length` at the top of `StopViewController.swift` now has a superfluous `type_body_length`.
- `PanelDetent` is still three closed cases — fine for the current FloatingPanel mapping, but the SwiftUI portability promise in the doc-comment is something to revisit (or narrow) when the sheet layer lands.
- `refreshSurveys()` is invoked on every `refresh()` call (`StopViewModel.swift:167`). `surveyService.fetchSurveys()` has an internal "already loading" guard so it isn't broken, but it's the same shape as the analytics double-fire that was fixed pre-merge. Worth gating to first-fetch (or to a coarser cadence) in a follow-up.

---

## Recommended Action

1. Merge.
2. Pick up [#1141](https://github.com/OneBusAway/onebusaway-ios/issues/1141) as a quick cleanup pass — or fold the lint-warning bits into the next PR that touches `StopViewController.swift`.

Verified: build succeeded, all 21 ViewModel tests pass (`StopViewModelTests`, `TripViewModelTests`, `BookmarksViewModelTests`, `MapViewModelTests`, `MapPanelViewModelTests`).

This was a real revision, not a checkbox pass. Nicely done.

PR #1138, opened 2026-05-24
Bookmarks ViewModel Layer

mosliem:
## Summary

  - Extract business logic from bookmark management view controllers (`EditBookmarkViewController`, `ManageBookmarksViewController`, `ManageGroupsViewController`) into three new
  `@MainActor` ViewModels
  - VCs become thin Eureka form shells responsible only for UI setup, alert presentation, and delegate callbacks
  - Eliminate force casts, fix a retain cycle in the duplicate-bookmark alert closure, and skip a redundant network call when bookmarking from `StopViewController`

  ## Motivation

  These VCs mixed data persistence, analytics reporting, duplicate detection, and name resolution directly into UIKit callbacks. This made the logic untestable in isolation and tightly
  coupled to Eureka form internals. The extraction follows the same pattern established by `StopViewModel`, `TripViewModel`, `BookmarksViewModel`, and `MapViewModel`.

  ## Changes

  ### New files
  | File | Role |
  |------|------|
  | `EditBookmarkViewModel` | Form initial values, `BookmarkSource` tagged enum, `SaveOutcome` enum, duplicate detection, persistence + analytics |
  | `ManageBookmarksViewModel` | Bookmark CRUD, delete analytics, name persistence, transit-name restore, reorder |
  | `ManageGroupsViewModel` | Group list access, row-to-model conversion, replace-all mutation |
  | `EditBookmarkViewModelTests` | 20 tests |
  | `ManageBookmarksViewModelTests` | 13 tests |
  | `ManageGroupsViewModelTests` | 9 tests |

  ### Key improvements
  - **`BookmarkSource` tagged enum** — eliminates the dual-optional `(stop: Stop?, arrivalDeparture: ArrivalDeparture?)` pattern; impossible state is unrepresentable
  - **`SaveOutcome` enum** — makes save-flow branching (ready vs. duplicate) explicit and exhaustive via `switch`
  - **Safe casts** — `as! Bool` → `(as? Bool) ?? true`; `as! [NameRow]` → `compactMap`
  - **Retain cycle fix** — duplicate alert closure now captures `[weak self]` with a guard
  - **`preloadedArrivals` optimization** — `AddBookmarkViewController` accepts already-fetched arrivals from `StopViewModel`, skipping a redundant network call
  - **Dead code removal** — `fileprivate extension Array where Element == NameRow`, `dataObjectName` computed property, `addBookmarkToStore` helper, and `originalTransitName` all deleted

  ## Behavior change worth flagging

  - **Analytics emission timing for new bookmarks** — the `addBookmark` event now fires *after* the user confirms past the duplicate-bookmark alert, not before it. Previously, an add was reported the moment the user tapped Save, so attempts cancelled at the duplicate prompt were still counted. The new behavior is more accurate (analytics now reflect what actually happened), but anyone tracking add-bookmark conversion rate will see a quiet step-change when this lands.

  ## Test plan

  - [x] 42 new unit tests pass (`EditBookmarkViewModelTests`, `ManageBookmarksViewModelTests`, `ManageGroupsViewModelTests`)
  - [x] Create a new stop bookmark → verify name, group, and favorite toggle persist correctly
  - [x] Create a new trip bookmark → verify analytics event fires
  - [x] Edit an existing bookmark's name and group → verify changes persist
  - [x] Clear the name field and save → verify transit-derived name is restored
  - [x] Trigger duplicate detection → confirm alert appears and "Create Duplicate" works
  - [x] Open Add Bookmark from Stop screen → verify no extra network call (arrivals pre-loaded)
  - [x] Manage Bookmarks screen: reorder, delete, rename → all persist correctly
  - [x] Manage Groups screen: add, rename, reorder, delete groups → all persist correctly
  
  This PR must be merged after #1135 

## Summary by CodeRabbit

* **New Features**
  * Shows an error alert when saving a bookmark fails because region info is unavailable
  * Add-book flow can preload arrival times to avoid an extra network fetch

* **Bug Fixes**
  * Improved duplicate-detection handling when saving bookmarks
  * Fixed concurrent state issues in alerts storage to prevent corruption/crashes
  * Corrected a localization key for duplicate-alert button text

* **Refactor**
  * Reworked bookmark and group management to use dedicated view models for clearer flows

[Review Change Stack](https://app.coderabbit.ai/change-stack/OneBusAway/onebusaway-ios/pull/1138?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


aaronbrethorst (changes requested):
Hi Mohamed,

This is the cleanest installment in the VM extraction series so far. `BookmarkSource` finally puts the "stop or arrival, never both" rule into the type system instead of relying on dual-optional convention, the `@MainActor` annotations are placed at the right granularity, and the AgencyAlertsStore data race fix is exactly the kind of thing that needed a real lock rather than `@unchecked Sendable` hand-waving. The 42 new tests cover the happy paths and the most important nil-region edge case.

There are two issues I'd like fixed before merge — neither is large, but each is in the new code so it's better to land it correctly the first time than to file follow-up issues. After that, a handful of suggestions worth considering, plus some fit-and-finish for the documentation.

## Critical Issues (2 found)

- **`selectedGroupID` is unused in `EditBookmarkViewModel.prepareToSave`** [`OBAKit/ViewModels/EditBookmarkViewModel.swift:110`]. The parameter is declared but never referenced in the body — group lookup happens later in `persist(_:to:isNewBookmark:)`. As written, it misleads readers (and future callers, including yourself in three months) into thinking duplicate detection is group-aware when it isn't. Either remove it from the signature, or actually use it (passing the group into `checkForDuplicates` would arguably be a feature improvement, since the same stop in two different groups is a legitimate scenario today).

- **`deleteAgencyAlerts` writes UserDefaults outside `stateLock`** [`OBAKitCore/Models/AgencyAlertsStore.swift:213`]. `userDefaults.set([], forKey: …)` runs *before* the lock is acquired, while `markAlertRead` at line 170 writes the same key *inside* the lock. The interleaving:
  1. `deleteAgencyAlerts` queue clears UserDefaults to `[]`
  2. Main thread `markAlertRead` takes the lock, inserts an id, writes `[id]` back to UserDefaults
  3. `deleteAgencyAlerts` queue takes the lock, clears in-memory `readAlertIDs`

  Net result: on-disk has `[id]`, in-memory has `[]`. Next launch, the lazy initializer re-hydrates `[id]` and the read state the user just cleared comes back. One-line fix: move the `userDefaults.set` call inside the existing `stateLock.withLock` block alongside the in-memory clears.

## Important Issues (3 found)

- **Analytics emission timing changed silently for new bookmarks.** The old `EditBookmarkViewController.save()` reported the `addBookmark` event during bookmark construction, before the duplicate-confirmation dialog. The new flow in `EditBookmarkViewModel.persist` only fires analytics *after* the user confirms past the duplicate alert. Bookmark adds that are cancelled at the duplicate prompt will no longer appear in analytics. The new behavior is arguably the correct one — analytics should reflect what actually happened — but call it out in the PR description so it doesn't surprise anyone tracking the metric.

- **`prepareToSave` returns `nil` for two distinct failure modes** [`OBAKit/ViewModels/EditBookmarkViewModel.swift:111` and `:129`]. The first is "no current region"; the second is "edit mode but `existingBookmark` is nil", which is a programmer-error state (the init populated it). The VC at line 179 maps both to the same "Unable to Save — current region is not available" alert. If the second branch ever fires, the user sees a misleading message and the invariant violation is invisible. Either `assertionFailure` on the edit-mode nil branch, or add a third `case` to `SaveOutcome` (see Suggestions).

- **`saveNameChange` silently drops the user's edit when `findBookmark` returns nil** [`OBAKit/ViewModels/ManageBookmarksViewModel.swift:67-70`]. The empty-name branch is intentional and documented. The missing-bookmark branch is a desync between Eureka rows and the data store, and dropping the edit silently leaves the user with no signal. Match the pattern at `ManageBookmarksViewController.swift:114` and `Logger.warn` when the bookmark lookup fails.

## Suggestions (8 found)

- **Hoist `isNewBookmark` out of `SaveOutcome`** [`EditBookmarkViewModel.swift:30`]. Both cases carry it identically, forcing the VC to pattern-match the value out twice. Either a `struct SaveOutcome { let status: Status; let bookmark: Bookmark; let isNewBookmark: Bool }`, or `(SaveOutcome, isNew: Bool)` tuple return.

- **Add `case regionUnavailable` to `SaveOutcome`** and drop the `Optional` return on `prepareToSave`. The VC's `switch` becomes exhaustive without a leading `guard let`, and the failure mode is named instead of being smuggled past the switch as `nil`.

- **Defer the edit-mode mutation of `bookmark.name`/`isFavorite`** from `prepareToSave` into `persist` [`EditBookmarkViewModel.swift:130-131`]. As-is, calling `prepareToSave` and then *not* calling `persist` leaves the in-memory `Bookmark` (a reference type the store also vends) dirtied. The doc comment notes this; the type system doesn't enforce it.

- **Replace the `""` sentinel for "no group"** with `UUID?` [`EditBookmarkViewModel.swift:57`, `:93`]. Stringly-typed sentinels are exactly the thing this refactor is supposed to be moving away from.

- **The inner `let routeShortName, let tripHeadsign` rebinds in `originalTransitName`** [`ManageBookmarksViewModel.swift:82-89`] are unreachable: `Bookmark.isTripBookmark` is defined as `routeShortName != nil && routeID != nil && tripHeadsign != nil` (`Bookmark.swift:178-180`), so the guard implies the optional binds always succeed. Either trust `isTripBookmark` and force-unwrap, or drop `isTripBookmark` and let the binds carry the predicate.

- **`groups(from:)` uses `.whitespaces`** [`ManageGroupsViewModel.swift:43`] which does not trim newlines. A row containing `"\n"` would today be persisted as a group named `"\n"`. Use `.whitespacesAndNewlines` and add a test for `"\n"`, `"\t"`, `"\r\n"`.

- **Two test gaps worth filling:**
  - The `preloadedArrivals` optimization on `AddBookmarkViewController` is entirely untested. A test that verifies the network call is skipped when arrivals are provided (and *not* skipped when nil) would lock down the savings.
  - The `AgencyAlertsStore` data race fix has no concurrent-access regression test. A fanout test under TSan that hits `markAlertRead` + `recentHighSeverityAlerts` + `update()` concurrently would catch a future "simplification" that removes the lock.

- **Pre-existing typo worth swatting while you're here:** `OBALoc("edit_book_controller.duplicate_alert.affirmative_button", …)` at `EditBookmarkViewController.swift:200` — sibling keys use `edit_bookmark_controller…`. This exists on main too (verified), so not introduced by this PR, but the file is being rewritten anyway. Fixing it should be coordinated with the translation team since existing localizations are keyed on the typo.

## Documentation Suggestions

A few of the new doc comments will rot. None are blocking, but worth considering as you finalize:

- The "Consumed by `…ViewController` (UIKit, via direct calls)" / "Eureka form UX stays in the VC" / "Contains no UIKit or SwiftUI imports" framing in each VM's class doc reads like a migration-plan note rather than durable documentation. Replace with what the VM *owns and guarantees*.
- `EditBookmarkViewModel.currentGroupID()` doc prescribes "Call this in `viewWillAppear`" — pins the VM to one caller's lifecycle. Describe the semantics ("returns the data store's live view, not the cached `initialGroupID` from init") instead.
- `EditBookmarkViewModel.prepareToSave` doc never mentions the `selectedGroupID` parameter — resolved if you take Critical #1.
- The `AgencyAlertsStore` `stateLock` comment is otherwise excellent — it explains the *why*, points at the symptom (TSan flag, segfault), and states the constraint ("Never hold this lock across an `await`"). One nit: the framing of "main thread vs. background `Task` executor" is slightly off since `update()` is `async` and inherits the caller's actor — "main vs. non-main contexts" would be more accurate.

## Strengths

- `BookmarkSource` tagged enum is a textbook win — eliminates a real class of bug.
- The `[weak self]` capture in the duplicate-alert handler at `EditBookmarkViewController.swift:200` is an improvement over the strong-self closure on main.
- `as!` force casts replaced with `as?` + `??` / `compactMap`.
- The `AgencyAlertsStore` lock fix takes a real snapshot of `agencies` *before* the network `await`, instead of letting child tasks read mutating state — exactly the right pattern.
- 42 tests using a real `Application`/`UserDataStore` stack rather than dependency-injected mocks of the VM's collaborators. This catches integration regressions that mock-heavy tests miss.
- `createApplicationWithoutRegion` is a thoughtful helper for the nil-region branch — exactly the kind of failure path most tests skip.
- Analytics non-emission is verified, not just emission.

## Recommended Action

1. Fix the two critical issues (one-line each).
2. Add a sentence to the PR description noting the analytics-timing change for cancelled-at-duplicate bookmark adds.
3. Add a `Logger.warn` to the missing-bookmark branch in `saveNameChange`.
4. Decide whether you want to take the type-design suggestions (hoisting `isNewBookmark`, adding `regionUnavailable` to `SaveOutcome`, replacing the `""` sentinel) in this PR or a follow-up. I lean toward follow-up so this PR stays focused, but I won't push back if you'd rather land them together.
5. Squash and merge.

**Verdict: Request changes.**

Thanks,
Aaron

aaronbrethorst (changes requested):
Mohamed, this is a genuinely strong piece of work. The `BookmarkSource` enum is a textbook "make illegal states unrepresentable" move — it deletes the old dual-optional `(stop: Stop?, arrivalDeparture: ArrivalDeparture?)` pattern along with the `fatalError()` and force-unwrap that rode along with it. The two-phase `prepareToSave`/`persist` split keeps validation side-effect-free, the retain-cycle fix in the duplicate alert is correct, and the `regionUnavailable` alert replaces a genuinely silent `guard … else { return }`. I verified the build and ran the new suites locally — **all 42+ tests pass** (`TEST SUCCEEDED`), and `OBAKitCore` stays extension-safe. The concurrency rework is also correct: I traced every `withLock` call site and confirmed there's no reentrancy, no lock held across an `await`, and the task group operates on an immutable snapshot.

That said, there's a short list of things I want tightened before this lands, mostly because this VM layer is going to be the pattern the next several controllers copy.

## Critical Issues (0 found)

None. No correctness bugs, no crashes, no data-loss paths.

## Important Issues (4 found)

- **Dead `isFavorite` write that contradicts its own doc comment** [OBAKit/ViewModels/EditBookmarkViewModel.swift:122]. In add mode `prepareToSave` sets `bookmark.isFavorite = isFavorite`, but `persist` sets it again at line 138, and nothing reads it in between (`checkForDuplicates` doesn't consider favorite state). The doc comment at lines 107–109 explicitly promises form values are applied *inside `persist`* — so the line is both dead and a contradiction of the contract. Drop line 122 and let `persist` be the single writer.

- **`SaveOutcome` reintroduces the impossible-state smell the PR fights elsewhere** [OBAKit/ViewModels/EditBookmarkViewModel.swift:30-33]. `.ready(Bookmark, isNew: Bool, isDuplicate: Bool)` advertises four combinations, but `(isNew: false, isDuplicate: true)` is unreachable — duplicate detection only runs in add mode. That's the same two-optionals problem `BookmarkSource` just eliminated, re-expressed as a transposable boolean pair one level up. Promote the confirmation branch to its own case, e.g. `.regionUnavailable` / `.readyToSave(Bookmark, isNew: Bool)` / `.duplicateRequiresConfirmation(Bookmark)`. It mirrors the three-way branching the controller already does and makes the bad state unrepresentable.

- **`isAddMode` + `existingBookmark` encode one fact in two fields, with a misleading failure fallback** [OBAKit/ViewModels/EditBookmarkViewModel.swift:46,72,126-129]. The "edit ⇒ has bookmark" invariant is only runtime-asserted, and when it's violated `prepareToSave` returns `.regionUnavailable` — so an internal logic error would surface to the user as "The current region is not available. Please try again," which is a flatly wrong diagnosis that retrying can't fix. A private `enum Mode { case add; case edit(Bookmark) }` makes the edit branch unconditionally carry the bookmark, deletes the `assertionFailure` and the misleading return, and collapses two fields into one. (`isAddMode` survives as a computed property for the VC's title logic.)

- **The `AgencyAlertsStore` concurrency fix is out of scope for a bookmarks PR** [OBAKitCore/Models/AgencyAlertsStore.swift]. The NSLock rework + its 148-line `AgencyAlertsStoreConcurrencyTests.swift` is a self-contained `OBAKitCore` crash fix with nothing to do with the bookmarks extraction. The PR description flags it, but flagging doesn't make it in-scope — bundling it makes the change harder to review and impossible to revert independently. I'd split it into its own PR. (If CI was red without it and that's why it's stacked here, say so and I'll treat it as a known exception.)

## Suggestions (7 found)

- **`stateLock` doc comment states unverifiable forensics as fact** [OBAKitCore/Models/AgencyAlertsStore.swift:65-72]. "Thread Sanitizer flags `agencies`" and "the corruption segfaults under load on CI" are specific claims a future reader can't confirm and that rot the moment CI changes. Keep the valuable, verifiable part (which contexts touch the state, and "never hold this lock across an `await`") and drop the forensic anecdote.

- **Two more doc comments drifted from the code.** `prepareToSave`'s "form values are applied … inside `persist`" [EditBookmarkViewModel.swift:104-109] is contradicted by the add-mode `isFavorite` write (resolved if you take the dead-code fix above). And `groups(from:)`'s "renamed in-place" [ManageGroupsViewModel.swift:35-38] is inaccurate — it constructs new `BookmarkGroup` instances reusing the UUID; the in-place merge happens later in `replaceBookmarkGroups`. The class-level doc gets this right ("preserving identity"); align the method doc.

- **Extract the duplicated "re-save bookmark to its current group" block** [OBAKit/ViewModels/ManageBookmarksViewModel.swift:72-74,93-95]. `saveNameChange` and `restoreTransitName` both do the identical `findGroup` + `add` dance verbatim. A `private func resaveInPlace(_:)` names the intent and removes the duplication.

- **Regional-alert fetch failures are silently dropped** [OBAKitCore/Models/AgencyAlertsStore.swift:122-124]. `getAlerts(agencies:)` is non-throwing and returns `[]` on failure, so if OBA alerts fail but Obaco succeeds, `update()` "succeeds" with a silent gap and no log. Pre-existing, but you're touching these exact lines — a `Logger.warn` in the failure branch would make partial degradation diagnosable.

- **Inconsistent logging in `bookmarkForBookmarkIndexPath`** [OBAKit/Bookmarks/ManageBookmarksViewController.swift:117-119]. The non-`NameRow` guard logs `Logger.warn`; the sibling bad-UUID guard returns `nil` silently. Same class of form/model desync — give it a matching `Logger.warn`.

- **Empty vs. nil `preloadedArrivals` is a conscious distinction worth a comment** [OBAKit/Bookmarks/AddBookmarkViewController.swift:110-112, StopViewController.swift:1215]. An empty `[]` short-circuits the network fetch and shows "no upcoming departures." Since the call site only passes `nil` when `stopArrivals` is itself nil, this is fine today — but a one-liner noting that empty deliberately suppresses the fallback would prevent a future foot-gun.

- **Test gaps worth a follow-up** (none blocking): the `recentHighSeverityAlerts` 8-hour/severity boundary is never exercised (the concurrency test only seeds "now" alerts); analytics assertions check event *count* but never the `value` payload, so a transposed `routeID`/`stopID` would pass; `currentGroupID()`'s stale-vs-live divergence — the entire reason the method exists — is only tested on the happy path where they agree; and the concurrency test never drives `update()` or `deleteAgencyAlerts()`, so the `agencies` race named in the lock's own docstring has zero coverage (its value is real but contingent on a TSan CI run).

## Strengths

- `BookmarkSource` is the best-designed type here — exhaustive `switch`es mean a future third bookmark kind forces the compiler to flag every site.
- Tests run against a real `UserDataStore`, asserting on observable persisted state rather than mock call-recording — they'll survive refactoring.
- `test_prepareToSave_editMode_doesNotMutateBookmarkUntilPersist` pins the prepare/persist contract precisely.
- The `MockDataLoader` request-recording + `resetRecordedRequestURLs()` is the right way to assert a network short-circuit ("no call happened"), and its lock correctly handles the alerts task-group fan-out.
- Net reduction in crash surface: a `fatalError()`, an `as! Bool` force-cast, and an `arrivalDeparture!` force-unwrap all deleted.
- The localization key fix (`edit_book_controller…` → `edit_bookmark_controller…`) corrects a real namespace bug.

## Recommended Action

1. Remove the dead `isFavorite` write (EditBookmarkViewModel.swift:122).
2. Decide on the `SaveOutcome` and `Mode` type-design changes (or at minimum fix the misleading `.regionUnavailable` fallback).
3. Fix the three drifted comments (stateLock forensics, `prepareToSave`, `groups(from:)`).
4. Either split the `AgencyAlertsStore` change into its own PR or confirm it's an intentional CI-unblocking stack.
5. Suggestions and test-gap items can be deferred to follow-ups.

**Verdict: Request Changes.** The core refactor is excellent and I want it in — but this layer sets the template for the controllers that follow, so the type-design and dead-code/comment cleanups are worth doing now rather than copy-pasting forward.

mosliem:
Re: the `AgencyAlertsStore` concurrency fix — taking the "known exception." CI was red without it, so the fix had to land here to keep the bookmarks work green.

aaronbrethorst (approved):
Mohamed, this is a genuinely exemplary refactor. You took three view controllers that had data persistence, duplicate detection, analytics, and name resolution all tangled into Eureka callbacks, and pulled that logic out into three focused `@MainActor` ViewModels — and then backed it with 45 passing tests. The `BookmarkSource` tagged enum replacing the dual-optional `(stop:arrivalDeparture:)` pattern is exactly the kind of "make illegal states unrepresentable" move I love to see, and your PR description's callout of the analytics-timing behavior change shows real care. I verified the build, ran every new test suite, and dug into each finding below; none of them block the merge.

**Verification performed:**
- `build-for-testing` on iPhone 17 Pro (iOS 26.3.1): **BUILD SUCCEEDED**
- All five new suites pass: `EditBookmarkViewModelTests` (20), `ManageBookmarksViewModelTests` (13), `ManageGroupsViewModelTests` (9), `AddBookmarkViewControllerTests` (2), `AgencyAlertsStoreConcurrencyTests` (1) — **45 tests, 0 failures**
- SwiftLint clean on all changed files
- Confirmed `viewModel.stopArrivals` and `Set.allObjects` resolve (CodeRabbit's two earlier findings are fully addressed)

---

## Critical Issues (0 found)

None. The `AgencyAlertsStore` locking is correct — no lock is held across an `await`, every read/write of `agencies`/`alerts`/`readAlertIDs` is consistently serialized, and there's no re-entrancy (NSLock is non-recursive and no locked closure re-acquires it). As a bonus, routing first access to the `lazy var readAlertIDs` through the lock makes its previously-unsafe lazy init thread-safe.

## Important Issues (0 found)

Nothing that must change before merge. The items below are polish.

## Suggestions (6 found)

- **Dead `isFavorite` parameter on `prepareToSave(name:isFavorite:)`** [`OBAKit/ViewModels/EditBookmarkViewModel.swift:118`]. The parameter is never read in the method body — the flag is only applied in `persist(...)`, as the method's own doc comment notes. Your own test `test_prepareToSave_editMode_doesNotMutateBookmarkUntilPersist` confirms it. Drop it from the signature.

- **`isNewBookmark` has two sources of truth** [`EditBookmarkViewModel.swift:118,142` + `EditBookmarkViewController.swift:185,209`]. `prepareToSave` decides newness and carries it in `SaveOutcome.readyToSave(_, isNew:)`, then the VC threads that boolean back into `persist(..., isNewBookmark:)` and *separately* hard-codes `true` on the duplicate path. Nothing structurally forces these to agree. Consider deriving newness inside `persist` from the outcome/mode, or splitting `.readyToSave` into `.readyToSaveNew`/`.readyToSaveExisting` — the same discipline your `BookmarkSource` enum already shows.

- **The headline behavior change isn't tested where it lives** [`EditBookmarkViewController.swift:189-212`]. The "analytics fires *after* duplicate confirmation" guarantee depends entirely on `persist` being wired only into the "Create Duplicate" action's closure. The VM tests cover that `persist` emits and that `prepareToSave` returns the duplicate outcome without persisting — but nothing asserts that *cancelling* the duplicate alert suppresses the `addBookmark` event. A VC-level test (or extracting the alert routing into a testable unit) would lock down the exact behavior the PR headlines.

- **Concurrency test only bites under Thread Sanitizer** [`AgencyAlertsStoreConcurrencyTests.swift:100`]. The fanout only exercises `markAlertRead`/`isAlertUnread`/reads, and its end-state assertions hold even with a lost update (idempotent set inserts), so without TSan it can pass while broken. It also never concurrently exercises `deleteAgencyAlerts()`'s `removeAll()` — the mutation most likely to deterministically corrupt state during a concurrent read. Worth confirming CI runs OBAKitTests with `-enableThreadSanitizer YES`, and adding a delete-vs-read fanout.

- **Comment imprecision on the empty-array short-circuit** [`AddBookmarkViewController.swift:110`]. It says an empty `preloadedArrivals` "surfaces the 'no upcoming departures' view," but there's no dedicated empty-state view here — an empty array renders the inline "route bookmarks unavailable" explanation in the trip section while still showing the whole-stop bookmark row. Minor reword.

- **Logging parity** [`ManageBookmarksViewController.swift:188`]. `saveNameChange` now nicely `Logger.warn`s when a form bookmark ID is missing from the store, but the parallel not-found case in `restoreEmptyBookmarkNames` silently `continue`s. Make them consistent.

## Notes (not action items)

- **Scope:** the `AgencyAlertsStore` NSLock change (+ its 148-line concurrency test) and the `MockDataLoader` request-recording additions are unrelated to a "Bookmarks ViewModel" PR. The locking is a legitimate fix and you flagged it in the description, so I'm fine landing it together — just noting that a future reader bisecting an `OBAKitCore` concurrency regression won't expect to find it in a UI-refactor PR.
- **Pre-existing, not a regression:** `persist`'s group lookup (`groupID.flatMap { findGroup }`) silently falls back to "(No Group)" if the selected group was deleted on another screen mid-edit. `main` behaved identically via `selectedBookmarkGroup`, so this is not introduced here — but it'd be a reasonable thing to log someday.
- **Behavior change (intentional, good):** `groups(from:)` now skips whitespace-only group names (old code only skipped truly empty), and the new `region_error` strings exist only in `en` for now (other locales fall back to `OBALoc` defaults until Transifex catches up). Both fine.
- **Verified safe:** the `routeShortName!`/`tripHeadsign!` force-unwraps in `ManageBookmarksViewModel.originalTransitName` are guarded by `isTripBookmark`, which `Bookmark.swift:178-180` defines as exactly those three properties being non-nil. The comment's safety claim is accurate.

## Strengths

- `BookmarkSource` and the private `Mode` enum are textbook sum-type design — illegal states genuinely unrepresentable, exhaustive switches, no `default` fallthrough hiding a third case.
- Analytics coverage is tested from all three angles: fires for a new trip bookmark, suppressed when editing an existing trip bookmark, suppressed for stop bookmarks.
- `test_currentGroupID_reflectsLiveMove_divergingFromInitialGroupID` is a great behavioral test of a subtle staleness invariant, and the doc comment explaining live-vs-cached group IDs is excellent.
- The `preloadedArrivals` test that leaves the endpoint *unstubbed* (so any network call trips `MockDataLoader`'s `fatalError`) plus asserts `recordedRequestURLs == 0` is exactly the right way to prove a "must not hit the network" contract.
- Comments throughout earn their place — they explain *why* (concurrency invariants, ID staleness, the deliberate empty-array short-circuit), not *what*.

## Recommended Action

1. Optional in-scope: delete the dead `isFavorite` parameter (one-line cleanup).
2. The remaining suggestions can land as fast-follows — tracked in #1145.
3. The stated merge-order dependency (#1135) is now satisfied — #1135 has merged.

---

**Verdict: Merge.**

PR #1139, opened 2026-05-25
(Recent - Nearby) Stops - Search ViewModels 

mosliem:
### Summary 

Extracts three ViewModels from their ViewControllers as part of the Tier 2 ViewModel layer plan (PR T2-4). Each VC becomes a thin Combine-binding shell; all business logic moves into a testable `@MainActor ObservableObject`.

  | New File |
  |----------|
  | `RecentStopsViewModel` | 
  | `NearbyStopsViewModel` 
  | `SearchViewModel` | 

### Notable fixes included

- **`RecentStopsViewModel.loadData()`** — guards against nil `currentRegion` with an early return to `recentStops = []`, preventing unintended `nil == nil` filter matches
- **`RecentStopsViewModel.delete(alarm:)`** — a failed background remote delete is logged via `Logger.error` rather than surfaced as a modal: the alarm is already gone locally and the server failure isn't user-actionable
- **`UserDataStore.delete(alarm:)`** — matches on the stable `url` instead of full equality. `Alarm.isEqual` compares `tripDate`/`alarmDate`, which lose sub-microsecond precision on the `timeIntervalSince1970` round-trip through `UserDefaults`, so a stored alarm never equalled its in-memory counterpart and delete was a no-op (in production as well as tests)
- **`SearchViewModel.selectVehicle()`** — resets both `vehicleError` and `vehicleSearchResponse` at the start of each call; adds an `isLoading` guard against concurrent taps
- **`SearchError`** — adds `Equatable` conformance solely for the `vehicleError as? SearchError == .noTripsAvailable` test assertion (`SearchViewModelTests.swift:235`); no production consumer.
- **`NearbyStopsViewController`** — removes a 3-second debug sleep that had been left in.

 ### Test coverage

 - `RecentStopsViewModelTests` — initial state, `loadData` region filtering (including nil-region edge case), deleteAll, delete stop, delete alarm (local + remote success + remote failure
  paths)
  - `NearbyStopsViewModelTests` — initial state, nil API service guard, success/failure load, concurrent double-load guard via `CountingDataLoader`
  - `SearchViewModelTests` — subtitle formatting, `results`, `response(substituting:)` (including `boundingRegion`/`error` preservation), `selectVehicle` (nil service, success, network
  error, keyNotFound, concurrent guard, error-then-success state transition)
  
  
  ### What's deferred

  `NearbyStopsListViewController` rewire (replace `MapRegionManager` delegate with `MapPanelViewModel.$nearbyStops` sink) is blocked on `feature/viewModel-layer` merging first, since
  `MapPanelViewModel` lives there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed alarm equality/hashing to prevent mismatches after saving/loading.

* **Improvements**
  * Recent Stops, Search, and Nearby Stops screens are now more responsive and consistent (better loading indicators, refreshes, and delete flows).
  * Vehicle selection and search subtitles/results behave more reliably, with fewer duplicate network calls and clearer error handling.

* **Localization**
  * Added user-facing messages prompting selection of a region in Settings.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OneBusAway/onebusaway-ios/pull/1139?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/OBAKit/Recent/RecentStopsViewController.swift
+++ b/OBAKit/Recent/RecentStopsViewController.swift
@@ -6,6 +6,7 @@
 //
 
 import UIKit
+import Combine
 import OBAKitCore
 
 /// Provides an interface to browse recently-viewed information, mostly `Stop`s.
@@ -16,12 +17,15 @@ public class RecentStopsViewController: UIViewController,
 
     let application: Application
 
-    let listView = OBAListView()
+    private let viewModel: RecentStopsViewModel
+    private let listView = OBAListView()
+    private var cancellables = Set<AnyCancellable>()
 
     // MARK: - Init
 
     public init(application: Application) {
         self.application = application
+        self.viewModel = RecentStopsViewModel(application: application)
 
         super.init(nibName: nil, bundle: nil)
 
@@ -47,13 +51,28 @@ public class RecentStopsViewController: UIViewController,
         view.addSubview(listView)
         listView.contextMenuDelegate = self
         listView.pinToSuperview(.edges)
+
+        bindViewModel()
+    }
+
+    private func bindViewModel() {
+        Publishers.CombineLatest(viewModel.$alarms, viewModel.$recentStops)
+            .sink { [weak self] _ in self?.listView.applyData(animated: true) }
+            .store(in: &cancellables)
+
+        viewModel.$deletionError
+            .compactMap { $0 }
+            .sink { [weak self] error in
+                guard let self else { return }
+                Task { await self.application.displayError(error) }
+            }
+            .store(in: &cancellables)
     }
 
     public override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
 
-        application.userDataStore.deleteExpiredAlarms()
-        listView.applyData()
+        viewModel.loadData()
     }
 
     // MARK: - Actions
@@ -62,9 +81,7 @@ public class RecentStopsViewController: UIViewController,
         let title = OBALoc("recent_stops.confirmation_alert.title", value: "Are you sure you want to delete all of your recent stops?", comment: "Title for a confirmation alert displayed before the user deletes all of their recent stops.")
 
         let alertController = UIAlertController.deletionAlert(title: title) { [weak self] _ in
-            guard let self = self else { return }
-            self.application.userDataStore.deleteAllRecentStops()
-            self.listView.applyData(animated: true)
+            self?.viewModel.deleteAllRecentStops()
         }
 
         present(alertController, animated: true, completion: nil)
@@ -104,21 +121,15 @@ public class RecentStopsViewController: UIViewController,
         }
     }
 
-    func onDeleteAlarm(_ viewModel: AlarmViewModel) {
-        Task {
-            try? await self.application.obacoService?.deleteAlarm(url: viewModel.alarm.url)
-        }
-        self.application.userDataStore.delete(alarm: viewModel.alarm)
-        self.listView.applyData(animated: true)
+    func onDeleteAlarm(_ alarmViewModel: AlarmViewModel) {
+        viewModel.delete(alarm: alarmViewModel.alarm)
     }
 
     // MARK: - Sections
 
-    private var alarms: OBAListViewSection? {
-        let alarms = application.userDataStore.alarms
-        guard alarms.count > 0 else {
-            return nil
-        }
+    private var alarmSection: OBAListViewSection? {
+        let alarms = viewModel.alarms
+        guard alarms.count > 0 else { return nil }
 
         let rows = alarms.compactMap { alarm in
             return AlarmViewModel(withAlarm: alarm, onSelect: onSelectAlarm, onDelete: onDeleteAlarm)
@@ -128,15 +139,9 @@ public class RecentStopsViewController: UIViewController,
         return OBAListViewSection(id: "alarms", title: title, contents: rows)
     }
 
-    private var stops: OBAListViewSection? {
-        guard let currentRegion = application.currentRegion else {
-            return nil
-        }
-
-        let stops = application.userDataStore.recentStops.filter { $0.regionIdentifier == currentRegion.regionIdentifier }
-        guard stops.count > 0 else {
-            return nil
-        }
+    private var stopsSection: OBAListViewSection? {
+        let stops = viewModel.recentStops
+        guard stops.count > 0 else { return nil }
 
         let rows = stops.map { stop -> StopRowItem in
             let onSelect: OBAListViewAction<StopRowItem> = { [unowned self] viewModel in
@@ -144,21 +149,20 @@ public class RecentStopsViewController: UIViewController,
             }
 
             let onDelete: OBAListViewAction<StopRowItem> = { [unowned self] _ in
-                self.application.userDataStore.delete(recentStop: stop)
-                self.listView.applyData(animated: true)
+                self.viewModel.delete(recentStop: stop)
             }
 
             return StopRowItem(withStop: stop, onSelect: onSelect, onDelete: onDelete)
         }
 
-        let title = application.userDataStore.alarms.count > 0 ? Strings.recentStops : nil
+        let title = viewModel.alarms.count > 0 ? Strings.recentStops : nil
         return OBAListViewSection(id: "recent_stops", title: title, contents: rows)
     }
 
     // MARK: - OBAListView
 
     public func items(for listView: OBAListView) -> [OBAListViewSection] {
-        return [alarms, stops].compactMap { $0 }
+        return [alarmSection, stopsSection].compactMap { $0 }
     }
 
     public func emptyData(for listView: OBAListView) -> OBAListView.EmptyData? {
--- a/OBAKit/Recent/RecentStopsViewController.swift
+++ b/OBAKit/Recent/RecentStopsViewController.swift
@@ -56,16 +56,12 @@ public class RecentStopsViewController: UIViewController,
     }
 
     private func bindViewModel() {
+        // Deliver on the main queue so the sink runs *after* @Published commits its new
+        // value (it publishes in willSet). Otherwise items(for:) would read stale
+        // alarms / recentStops and the list would render one update behind.
         Publishers.CombineLatest(viewModel.$alarms, viewModel.$recentStops)
+            .receive(on: DispatchQueue.main)
             .sink { [weak self] _ in self?.listView.applyData(animated: true) }
-            .store(in: &cancellables)
-
-        viewModel.$deletionError
-            .compactMap { $0 }
-            .sink { [weak self] error in
-                guard let self else { return }
-                Task { await self.application.displayError(error) }
-            }
             .store(in: &cancellables)
     }
 
--- a/OBAKit/Recent/RecentStopsViewController.swift
+++ b/OBAKit/Recent/RecentStopsViewController.swift
@@ -56,9 +56,9 @@ public class RecentStopsViewController: UIViewController,
     }
 
     private func bindViewModel() {
-        // Deliver on the main queue so the sink runs *after* @Published commits its new
-        // value (it publishes in willSet). Otherwise items(for:) would read stale
-        // alarms / recentStops and the list would render one update behind.
+        // `@Published` fires from `willSet`, so a synchronous sink would read the *old*
+        // stored values via `items(for:)` (alarms / recentStops). The main-queue hop
+        // defers the closure until after the property writes complete.
         Publishers.CombineLatest(viewModel.$alarms, viewModel.$recentStops)
             .receive(on: DispatchQueue.main)
             .sink { [weak self] _ in self?.listView.applyData(animated: true) }
--- a/OBAKit/Recent/RecentStopsViewModel.swift
+++ b/OBAKit/Recent/RecentStopsViewModel.swift
@@ -1,0 +1,57 @@
+//
+//  RecentStopsViewModel.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import Foundation
+import OBAKitCore
+
+@MainActor
+final class RecentStopsViewModel: ObservableObject {
+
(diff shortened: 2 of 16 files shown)


aaronbrethorst:
@mosliem please fix the merge conflict on this PR!

aaronbrethorst (changes requested):
Hi Mohamed,

Tier 2 of the VM series, and the diff continues to look like the right shape — thin Combine binding shells in the VCs, business logic in `@MainActor ObservableObject` types, plus three meaningful drive-by fixes worth more than the line counts suggest: the `UserDataStore.delete(alarm:)` precision bug, the 3-second debug `Task.sleep` that had been quietly punishing every Nearby Stops cold launch, and the `selectVehicle` concurrent-tap guard. The `CountingDataLoader` helper is exactly the right approach to actually exercising the concurrent-call guards.

One blocking issue I want fixed before merge — and a handful of items I'd nudge you on but won't gate on.

## Critical Issues (1 found)

- **`RecentStopsViewModel.delete(alarm:)` ties the remote DELETE to the VM's lifetime** [`OBAKit/Recent/RecentStopsViewModel.swift:52-59`]. The `Task { [weak self] in guard let self else { return } … }` pattern means: if a user taps delete and immediately pops the view (which is a *common* interaction since delete is often the last thing someone does on a screen), the `guard` aborts before `obacoService?.deleteAlarm(url:)` ever runs. The alarm is gone locally, lives on at the Obaco server, and the user gets the push notification anyway with no idea why. This is exactly the failure mode Obaco-managed alarms are supposed to prevent.

  The fix is one line — `self` isn't actually used inside the closure body; you're going through it only to reach `application.obacoService`. Capture what you need directly and drop `[weak self]`:

  ```swift
  let service = application.obacoService
  Task {
      do { try await service?.deleteAlarm(url: alarm.url) }
      catch { Logger.error("Failed to delete alarm remotely: \(error.localizedDescription)") }
  }
  ```

  This also makes the comment above truthful — currently it claims fire-and-log semantics but the code is fire-and-maybe-log-or-silently-drop. Worth a regression test too (`test_delete_alarm_removesItLocally` at line 165 doesn't catch this; you'd need a test that nils out the VM after the local delete and asserts the obaco mock still received the DELETE).

## Important Issues (4 found)

- **Stale-modal trap on the error sinks** [`OBAKit/Stops/NearbyStopsViewController.swift:87-93`, `OBAKit/Search/SearchResultsController.swift:68-74`]. Both use `.compactMap { $0 }` on the error publisher then call `application.displayError(error)`. When `loadStops()` / `selectVehicle()` reset the error to nil at the start of a retry, `compactMap` filters the nil out — so the alert from the previous failure is never told to dismiss. A user who taps "Retry" while an alert is up gets the new content underneath the old alert, or potentially a stacked second alert if the retry also fails. Drop the `compactMap`, sink on the full optional, and explicitly dismiss any presented controller when the value goes nil. (Or model errors as a `PassthroughSubject<Error, Never>` so there's no stale state to reconcile in the first place — likely the cleaner option.)

- **`DecodingError.keyNotFound` mapping is too broad** [`OBAKit/Search/SearchViewModel.swift:50-51`]. The `catch DecodingError.keyNotFound` matches *any* missing key from anywhere in `VehicleStatus` decoding, then maps it to `SearchError.noTripsAvailable`. If the OBA server renames a field, drops a key, or returns a malformed payload, the user sees "This vehicle is not currently on any trips" — a flatly wrong message — and the real backend regression is invisible to developers because we never log the actual DecodingError. Pattern-match the associated key so the special case stays narrow, and `Logger.error` the original error in the broader catch:

  ```swift
  } catch let DecodingError.keyNotFound(key, ctx) where key.stringValue == "trip" {
      vehicleError = SearchError.noTripsAvailable
  } catch {
      Logger.error("selectVehicle decode failure: \(error)")
      vehicleError = error
  }
  ```

  (Confirm `"trip"` is the actual key the server omits — I haven't traced the VehicleStatus codable.)

- **`UserDataStore.delete(alarm:)` papers over a broken `Alarm.isEqual`** [`OBAKitCore/Models/UserData/UserDataStore.swift:721-725`]. The fix and the comment are both right — `Alarm.isEqual` compares `tripDate`/`alarmDate`, which lose sub-microsecond precision through the `TimeInterval` round-trip on encode/decode (`OBAKitCore/Models/Obaco/Alarm.swift:28,32,43,47`). But this means anything else that depends on `Alarm` equality — `Set<Alarm>` membership, `Array.firstIndex(of:)`, diffable datasource snapshots — is silently broken too. Fix the root cause: normalize the dates inside `isEqual` (or override `==` to compare on `url` if that's the actual identity) and leave a `TODO` referencing the underlying ticket. Either way, file a follow-up issue tracking the broader equality problem; otherwise the comment becomes the only thing standing between us and someone confidently re-introducing `==` later.

- **Combine `.receive(on:)` comments are technically backwards** [`OBAKit/Recent/RecentStopsViewController.swift:59-61`, `OBAKit/Stops/NearbyStopsViewController.swift:76-79`]. Both say "Deliver on the main queue so the sink runs after `@Published` commits its new value (it publishes in `willSet`)". But `@Published` does emit *from* `willSet` — the subscriber sees the new value before the stored property has been updated. What actually saves us is the *async hop*: by the time the main-queue closure runs, `willSet` has returned and the stored property is now up to date, so `items(for:)` reads fresh state. The current wording is plausible enough that a future maintainer might reasonably remove `.receive(on:)` ("we're already on main") and re-introduce the stale-read bug. Suggest:

  > `@Published` fires from `willSet`, so a synchronous sink would read the *old* stored value via `items(for:)`. The main-queue hop defers the closure until after the property write completes.

## Suggestions

- **Distinguish "already loading" from "no API service" in the guard** [`NearbyStopsViewModel.swift:34`, `SearchViewModel.swift:42`]. The current `guard !isLoading, let apiService else { return }` collapses two distinct conditions — one is a legitimate no-op, the other is a configuration error that leaves the screen permanently dead. Split them and surface the misconfiguration via `operationError`/`vehicleError` so the existing error sink can display it.

- **Make the nil-region branch observable** [`RecentStopsViewModel.swift:28-31`]. The guard is correct and the comment about the "nil == nil" filter trap is real, but right now a user with no current region (mid-region-change, first launch race, location-permission denial) sees an indistinguishable "you have no recent stops" empty state. Either expose a `loadState` enum so the empty state can carry context, or `Logger.warn` so the condition shows up in logs.

- **Two tests promised but missing.** The PR description lists "delete alarm (local + remote success + remote failure paths)" under coverage, but only `test_delete_alarm_removesItLocally` exists. Add tests that stub `obacoService.deleteAlarm` to succeed and fail, and assert (a) local state is unaffected by the remote outcome, (b) the failure path logs. The fire-and-forget Task makes this slightly annoying — use an `XCTestExpectation` with timeout. Also worth adding a `UserDefaultsStoreTests.test_alarms_delete_afterUserDefaultsRoundTrip` that explicitly persists and reloads before deleting, so the precision-loss fix has a regression test by name.

- **The Null Island test is load-bearing** [`OBAKitTests/Recent/RecentStopsViewModelTests.swift:182-222`]. It's clever, but the geographic guarantee is implicit — if anyone ever adds a worldwide stub region (or a region whose bounding box happens to wrap the antimeridian), `currentRegion` becomes non-nil and the assertion passes for the wrong reason. Drop in an `expect(app.currentRegion).to(beNil())` before calling `loadData()`. Same outcome on green, fails loudly on regression.

- **"`nil != nil` guard would swallow the refresh"** [`NearbyStopsViewController.swift:78-79`] — the actual guard is `oldValue != searchFilter`. The shorthand is cute, but a future grep for `nil != nil` finds nothing. Spell it out: "the `oldValue != searchFilter` guard would swallow the refresh on first load when both are nil."

- **`UserDataStore.delete(alarm:)` could warn on duplicates.** The URL-based match silently deletes *all* alarms sharing a URL. Today that's a single-row no-op, but if a race ever produces two, one tap wipes both with no signal. A one-line `if matches.count > 1 { Logger.warn(...) }` would surface the invariant violation if it ever happens.

- **`SearchError` gaining `Equatable`** [`OBAKit/Search/SearchRequest.swift:17`] is fine — the enum has no associated values, so the conformance is trivial. Worth noting in the PR description that it was added for the test assertion at `SearchViewModelTests.swift:135,235`, not for production use.

## Strengths

- The `Task.sleep(nanoseconds: 3_000_000_000)` removal is silent in the diff but worth calling out explicitly — every Nearby Stops launch was paying that cost.
- The `UserDataStore.delete(alarm:)` comment is exactly the right level of specificity: it names the failing equality method, the field that loses precision, and the encode/decode mechanism that strips it. The fix would still be plausible without the comment; the comment makes it *defensible*.
- `CountingDataLoader` is a thoughtful test helper — the `Task.yield()` interleaving and the doc explaining why it's there both earn their keep.
- The error-then-success state transition test at `SearchViewModelTests.swift:264-281` verifies `vehicleError` is cleared on retry — exactly the kind of stateful contract that's easy to break and hard to debug.
- `makeKeyNotFoundLoader` synthesizes the malformed envelope inline with an explanatory comment — far better than a fixture file.
- The fix at `NearbyStopsViewController.swift:80-85` (apply directly rather than route through `searchFilter`) is a real subtle wiring bug caught between commits — visible in the rewording from the f629a0d6 → 1593e87b sequence.

## Recommended Action

1. Fix the critical alarm-delete capture bug. Add a test that asserts the Task survives VM nilling.
2. Decide whether you want to tackle the four important issues in this PR or as immediate follow-ups. I'd at least take the `keyNotFound` narrowing and the receive-on comment correction here, since they're tiny and they touch the new code anyway.
3. File an issue for the broader `Alarm.isEqual` precision bug — leave the workaround in place but make the root cause findable.
4. Squash and merge.

**Verdict: Request changes.**

Thanks,
Aaron

aaronbrethorst (approved):
Great work. This is exactly the kind of MVVM extraction I was hoping to see continue after the Tier-2 ViewModel work — and the quality bar is high. Three view controllers (`RecentStopsViewController`, `NearbyStopsViewController`, `SearchResultsController`) shed their business logic into focused `@MainActor ObservableObject` view models, each independently testable with no UIKit in the way. What stands out most is that you didn't just *move* code — you found and fixed a real persistence bug along the way, and the comments explaining the tricky Combine timing are genuinely excellent.

I built it locally (`** BUILD SUCCEEDED **`, 0 errors) and ran the three new VM suites plus `UserDefaultsStoreTests` — `** TEST SUCCEEDED **`, 68 tests, 0 failures.

## Agent Applicability Assessment

| Agent | Confidence | Rationale | Use? |
|-------|------------|-----------|------|
| code-reviewer | 95 | Substantial new Swift across UI + core | Yes |
| silent-failure-hunter | 90 | Error handling reworked (try?/catch → published errors) | Yes |
| pr-test-analyzer | 95 | 4 new test files, ~40 new tests | Yes |
| comment-analyzer | 85 | Several dense explanatory comments to verify | Yes |
| type-design-analyzer | 70 | New VMs are straightforward; below threshold | No |
| code-simplifier | 65 | Code is already clean; low marginal value | No |

## Critical Issues (0 found)

None. Builds clean, tests pass.

## Important Issues (0 found)

None blocking.

## Suggestions (3 found — all optional, none block merge)

- **`RecentStopsViewModel.loadData()` triggers two list reloads.** `Publishers.CombineLatest($alarms, $recentStops)` fires once per published property, so one `loadData()` (which writes both) yields two back-to-back `applyData(animated: true)` calls. Harmless — just a redundant animation. A `.debounce` or publishing a single combined snapshot would collapse it. [OBAKit/Recent/RecentStopsViewController.swift:42]

- **No test observes `isLoading` flipping to `true` mid-load.** The `NearbyStops`/`Search` suites assert `isLoading` is false before and after, and the concurrent-guard tests prove `callCount == 1`, but nothing pins the `false → true → false` transition itself. Collecting the published `isLoading` sequence would close that small gap. [OBAKitTests/Stops/NearbyStopsViewModelTests.swift]

- **Error-property naming drift:** `NearbyStopsViewModel.operationError` vs `SearchViewModel.vehicleError` for the same "route to `displayError`" concept. Worth aligning as this layer grows. [OBAKit/Stops/NearbyStopsViewModel.swift, OBAKit/Search/SearchViewModel.swift]

## Strengths

- **A real bug fix, not just a move.** `Alarm.isEqual`/`hash` now compare dates via `timeIntervalSince1970`, fixing a precision-loss bug where an alarm wouldn't equal *itself* after a UserDefaults round-trip — which silently turned equality-based deletes into no-ops in production. The regression test (`test_alarms_delete_afterUserDefaultsRoundTrip`) anchors the fix, and `UserDataStore.delete(alarm:)` switching to URL-based identity (with a warn on duplicate URLs) is the right call.
- **The Combine timing comments are textbook.** Documenting that `@Published` fires in `willSet` (so a synchronous sink reads the *old* value via `items(for:)`, hence the main-queue hop), and the deliberate "sink on the full optional, not `.compactMap`, so an error reset-to-nil isn't swallowed" reasoning — that's the kind of note that saves the next person hours.
- **Error handling got *better*, not just relocated.** The old `NearbyStopsViewController.loadStops` silently `return`ed on a nil API service (and carried a leftover `Task.sleep(3s)`); the new VMs surface that misconfiguration through the same `displayError` channel with a localized string. And `DecodingError.keyNotFound` is now narrowed to specifically `tripId` rather than swallowing *any* missing key as "no trips available" — a meaningful precision win.
- **Thoughtful concurrency.** `delete(alarm:)` captures `service`/`url` directly rather than `[weak self]` so a mid-delete deallocation doesn't abort the remote DELETE, and it's `@discardableResult` so tests can await it. The in-flight `guard !isLoading` guards — proven by `CountingDataLoader` collapsing concurrent calls to a single request — are a nice touch.
- **Broad test coverage:** success / network-error / no-service / key-not-found / in-flight-collapse / region-nil edge cases are all covered, including the "Null Island" trick to force `currentRegion == nil`, and the proactive `MapViewModelTests` fix to re-register stubs after `removeMappedResponses()` so an in-flight region resolve doesn't fatal the mock loader.

## Recommended Action

1. Merge.
2. Optionally fold in the three nits above whenever convenient (or in your next VM PR) — none are blocking.

**Verdict: Merge.** Clean, well-documented, well-tested, and it even fixes a latent persistence bug. Solid work.

Issue #1141, opened 2026-05-27
ViewModel layer fit-and-finish follow-ups


aaronbrethorst:
Tracking remaining fit-and-finish items from the PR #1135 review that don't need to block merging:

## SwiftLint warnings introduced by the StopViewController shrink

- `StopViewController.swift:1274` — `bindListData()` is 52 lines, exceeding the 50-line `function_body_length` threshold. Could be split into `bindArrivalsSink()` + `bindStopSink()` etc.
- `StopViewController.swift:17` — `swiftlint:disable file_length type_body_length` now has a superfluous `type_body_length` (the body shrunk below the limit after the VM refactor). Drop `type_body_length` from the disable list.

## PanelDetent expansion

`OBAKit/ViewModels/MapPanelViewModel.swift:16-20`. The three-case enum (`tip`, `half`, `full`) is fine for the current UIKit FloatingPanel, but the doc-comment promises SwiftUI `presentationDetents` portability that the closed enum can't deliver. When the SwiftUI sheets land, either add `.fraction(Double)` / `.height(CGFloat)` cases or narrow the doc-comment to the UIKit shape only.

## `refreshSurveys()` is invoked on every `refresh()`

`OBAKit/ViewModels/StopViewModel.swift:167,213-225`. Every 30 s auto-refresh re-enters `refreshSurveys()`, which dispatches a `Task` that calls `surveyService.fetchSurveys()`. The fetch has its own \"already loading\" guard so this isn't broken — but it's redundant work that the old code only did once on stop entry. Same shape as the analytics double-fire that was fixed pre-merge; worth gating to first-fetch (or to a much longer cadence than the arrivals refresh).

---

Related: PR #1135.

Borisserz:
Taking this — ViewModel fit-and-finish: SwiftLint cleanup, survey refresh gating, PanelDetent docs.

PR #1142, opened 2026-05-31
#1135 fit-and-finish follow-ups.

mosliem:
Implementing the points mentioned in #1141 

## Summary by CodeRabbit

* **Improvements**
  * Survey information now refreshes during initial stop load instead of on every periodic refresh, reducing redundant updates.
  * Panel presentation sizing updated for more consistent expansion across UI contexts.

* **Bug Fixes**
  * Surveys now emit refresh exactly once across repeated fetches, preventing duplicate survey updates and rendering.

* **Tests**
  * Added regression test ensuring surveys refresh exactly once across repeated fetches.

### DIFF
--- a/OBAKit/Stops/StopViewController.swift
+++ b/OBAKit/Stops/StopViewController.swift
@@ -1294,21 +1294,6 @@ private extension StopViewController {
             .store(in: &cancellables)
     }
 
-    func seedCollapsedPastDepartureSections(for arrivals: StopArrivals) {
-        guard pastDeparturesCollapsed else { return }
-        if viewModel.stopPreferences.sortType == .time {
-            collapsedSections.insert(ListSections.pastArrivalDepartures(suffix: "all").sectionID)
-        } else {
-            let groups = arrivals.arrivalsAndDepartures.group(
-                preferences: viewModel.stopPreferences,
-                filter: viewModel.isListFiltered
-            )
-            for group in groups {
-                collapsedSections.insert(ListSections.pastArrivalDepartures(suffix: group.route.id).sectionID)
-            }
-        }
-    }
-
     func bindStopSink() {
         viewModel.$stop
             .sink { [weak self] stop in
@@ -1344,6 +1329,21 @@ private extension StopViewController {
                 }
             }
             .store(in: &cancellables)
+    }
+
+    func seedCollapsedPastDepartureSections(for arrivals: StopArrivals) {
+        guard pastDeparturesCollapsed else { return }
+        if viewModel.stopPreferences.sortType == .time {
+            collapsedSections.insert(ListSections.pastArrivalDepartures(suffix: "all").sectionID)
+        } else {
+            let groups = arrivals.arrivalsAndDepartures.group(
+                preferences: viewModel.stopPreferences,
+                filter: viewModel.isListFiltered
+            )
+            for group in groups {
+                collapsedSections.insert(ListSections.pastArrivalDepartures(suffix: group.route.id).sectionID)
+            }
+        }
     }
 
     func bindLoadingState() {
--- a/OBAKit/Stops/StopViewController.swift
+++ b/OBAKit/Stops/StopViewController.swift
@@ -14,7 +14,7 @@ import Combine
 import SwiftUI
 import SafariServices
 
-// swiftlint:disable file_length type_body_length
+// swiftlint:disable file_length
 
 /// This is the core view controller for displaying information about a transit stop.
 ///
@@ -1272,23 +1272,18 @@ public class StopViewController: UIViewController,
 
 private extension StopViewController {
     func bindListData() {
+        bindArrivalsSink()
+        bindStopSink()
+        bindSurveysSink()
+        bindPreferencesSinks()
+    }
+
+    func bindArrivalsSink() {
         viewModel.$stopArrivals
             .sink { [weak self] arrivals in
                 guard let self, let arrivals else { return }
                 if firstLoad {
-                    if pastDeparturesCollapsed {
-                        if viewModel.stopPreferences.sortType == .time {
-                            collapsedSections.insert(ListSections.pastArrivalDepartures(suffix: "all").sectionID)
-                        } else {
-                            let groups = arrivals.arrivalsAndDepartures.group(
-                                preferences: viewModel.stopPreferences,
-                                filter: viewModel.isListFiltered
-                            )
-                            for group in groups {
-                                collapsedSections.insert(ListSections.pastArrivalDepartures(suffix: group.route.id).sectionID)
-                            }
-                        }
-                    }
+                    seedCollapsedPastDepartureSections(for: arrivals)
                     firstLoad = false
                     dataLoadFeedbackGenerator.dataLoad(.success)
                 }
@@ -1297,7 +1292,24 @@ private extension StopViewController {
                 beginUserActivity()
             }
             .store(in: &cancellables)
+    }
 
+    func seedCollapsedPastDepartureSections(for arrivals: StopArrivals) {
+        guard pastDeparturesCollapsed else { return }
+        if viewModel.stopPreferences.sortType == .time {
+            collapsedSections.insert(ListSections.pastArrivalDepartures(suffix: "all").sectionID)
+        } else {
+            let groups = arrivals.arrivalsAndDepartures.group(
+                preferences: viewModel.stopPreferences,
+                filter: viewModel.isListFiltered
+            )
+            for group in groups {
+                collapsedSections.insert(ListSections.pastArrivalDepartures(suffix: group.route.id).sectionID)
+            }
+        }
+    }
+
+    func bindStopSink() {
         viewModel.$stop
             .sink { [weak self] stop in
                 guard let self else { return }
@@ -1306,11 +1318,15 @@ private extension StopViewController {
                 configureTabBarButtons()
             }
             .store(in: &cancellables)
+    }
 
+    func bindSurveysSink() {
         viewModel.surveysDidRefresh
             .sink { [weak self] _ in self?.listView.applyData(animated: false) }
             .store(in: &cancellables)
+    }
 
+    func bindPreferencesSinks() {
         viewModel.$stopPreferences
             .sink { [weak self] _ in
                 Task { @MainActor [weak self] in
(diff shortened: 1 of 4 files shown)


aaronbrethorst (changes requested):
Hi Mohamed — thanks for picking up the loose ends from #1135. This is exactly the kind of methodical follow-up I like to see: you took the three items from #1141 one by one and addressed each. The `bindListData()` split is clean and behavior-preserving, the swiftlint narrowing is correctly justified, and moving `refreshSurveys()` off the per-refresh path is the right call. The branch builds and all `StopViewModelTests` / `MapPanelViewModelTests` pass (10/10).

I'm going to ask for one more round, though. The `PanelDetent` change resolves the *letter* of issue item #2 but introduces a new inconsistency, and a couple of small things deserve tightening before this lands.

## Critical Issues (0 found)

None. The refactor is behavior-preserving and the survey-timing change is correct (`performInitialStopSetupIfNeeded` is gated by `hasPerformedInitialStopSetup`, so surveys now fetch once per VM lifetime).

## Important Issues (3 found)

1. **`PanelDetent.value`'s `.tip` height (`120`) contradicts the actual layout constants.** [OBAKit/ViewModels/MapPanelViewModel.swift:35] The doc-comment promises this enum is the "single shared" source of truth so UIKit and SwiftUI "stay in sync," but the real UIKit tip geometry lives in `OBAKit/Controls/FloatingPanel/Layouts.swift` as `EstimatedDrawerTipStateHeight = 72` (portrait) and the landscape `.tip` anchor `absoluteInset: 69.0`. A SwiftUI sheet built from `.tip.value == .height(120)` would render ~50pt taller than the UIKit panel — the exact desync the comment claims to prevent. If the mapping lands now, source the constant from the layout layer (e.g. reference `MapPanelLayout.EstimatedDrawerTipStateHeight`, or hoist one shared `static let` both files consume) rather than a fresh `120`.

2. **`PanelDetent.value` / `Value` are dead code with an over-promising doc-comment.** [OBAKit/ViewModels/MapPanelViewModel.swift:18-40] Nothing references `.value` anywhere — the only `PanelDetent` consumer (`MapFloatingPanelController.swift:385-389`) still switches over the cases directly to map to `FloatingPanelState`, and there's no SwiftUI `presentationDetents` consumer yet. The doc-comment states the two-layer sync as present fact ("SwiftUI maps each case... so the two layers stay in sync"), but no such SwiftUI code exists. Note issue #1141 explicitly offered the cheaper resolution: *"or narrow the doc-comment to the UIKit shape only."* My preference is to do exactly that for now — keep the `Equatable` conformance (it's earning its place in the tests), drop `value`/`Value`, and reintroduce a detent→`PresentationDetent` mapping in the same commit as the SwiftUI sheet that consumes it, where it can be tested against real layout values. If you'd rather keep the mapping, then fix the constant (item 1), add a test (item 3 below), and reword the comment to mark it forward-looking — matching how the existing comments in this same file correctly hedge with "future `HomeSheetView`" (lines 44-52).

3. **The new survey comment misstates the refresh cadence.** [OBAKit/ViewModels/StopViewModel.swift:209] "Surveys don't change per **30s** refresh" — the auto-refresh timer is actually 15s (`timerInterval: TimeInterval = 15`, line 84). The `30` is `refreshThreshold` (line 85), an unrelated staleness threshold. The "fetch once per stop entry" half is accurate; just drop the wrong number: *"Surveys don't change between auto-refreshes; fetch once per stop entry, alongside the other first-fetch-only side effects above."*

## Suggestions (1 found)

1. **Add a test for the survey-timing change — it's the headline of this PR and currently unprotected.** [OBAKitTests/ViewModels/StopViewModelTests.swift] You already have two perfect templates: `test_analyticsFiresExactlyOnceAcrossRefreshes` (line 120) and `test_recentStop_recordedExactlyOnce` (line 144). A mirror like `test_surveys_refreshedExactlyOnceAcrossRefreshes` would pin the behavior so a future refactor can't silently move `refreshSurveys()` back into the per-refresh path. One caveat: `refreshSurveys()` emits from a detached `Task`, so assert with Nimble's `toEventually` rather than a synchronous `expect`, or it'll be flaky. A one-line table test for `PanelDetent.value` (`expect(PanelDetent.tip.value) == .height(...)`) would also pin the constants — but only worth adding if you keep the mapping per item 2.

## Strengths

- The `bindListData()` split registers the four sinks in the identical order to the original, all into the same `cancellables` — a genuinely behavior-preserving extraction.
- `seedCollapsedPastDepartureSections(for:)` preserves the original logic exactly; `guard pastDeparturesCollapsed else { return }` is equivalent to the old `if` wrapper.
- Dropping `type_body_length` from the swiftlint disable is correctly reasoned — the bind helpers now live in the `private extension`, whose body doesn't count toward the type body length.
- The `StopViewModel` comment earns its place by explaining *why* the fetch moved out of the per-refresh path; only the "30s" figure needs fixing.
- Adding `Equatable` to `PanelDetent` is a clean win, already exercised by `test_searchMode_drivesRequestedPanelDetent`.

## Recommended Action

1. Decide the `PanelDetent` direction (item 2): I recommend dropping `value`/`Value` and narrowing the doc-comment to the UIKit shape, keeping `Equatable`. If you keep the mapping, fix the `120` constant (item 1), reword the comment, and add the table test.
2. Fix the "30s" → "15s" comment (item 3).
3. Add the survey once-per-lifetime test (suggestion 1).
4. Re-request review.

---

**Verdict: Request Changes.**

aaronbrethorst (changes requested):
Hi Mohamed — this is a tidy, surgical follow-up that knocks out all three items from #1141 without scope creep. The `bindListData()` split reads like a table of contents, dropping the now-unnecessary `type_body_length` exemption is a satisfying signal that the file genuinely shrank, and relocating `refreshSurveys()` into the once-per-entry block is the right correctness fix. I built it, ran SwiftLint clean on all three changed files, and all 7 `StopViewModelTests` pass. There's one thing worth tightening before this goes in, plus a couple of smaller notes.

## Critical Issues (0 found)

None.

## Important Issues (1 found)

- **The new survey test can pass even if the behavior it guards regresses.** [OBAKitTests/ViewModels/StopViewModelTests.swift:179] — `await expect(emissions).toEventually(equal(1))` stops polling the instant `emissions` first hits `1`. It proves "emitted *at least* once," not "emitted *exactly* once." If a future change reintroduced per-refresh survey fetching, `emissions` would pass *through* `1` on its way to `2`/`3`, and the matcher would latch onto that transient and pass — the very regression this test exists to catch would slip through. This matters more than usual here because the `SurveyService` cooldown does *not* back you up in the test: the cooldown branch (`SurveyService.swift`) only short-circuits when `!allSurveys.isEmpty`, and the stub returns an empty list, so the `hasPerformedInitialStopSetup` gate is the *only* thing suppressing extra emissions — i.e. exactly the code under test. Please add a settle assertion after the `toEventually`, e.g. `await expect(emissions).toNever(beGreaterThan(1))` (or `toAlways(equal(1), until:)`), to turn this into a real "exactly once" guard. The sibling synchronous tests (`test_analyticsFiresExactlyOnceAcrossRefreshes`, `test_recentStop_recordedExactlyOnce`) get this for free because their side effects are synchronous; the async survey path needs the explicit settle.

## Suggestions (3 found)

- **`PanelDetent` doc-comment trades one forward-looking promise for another.** [OBAKit/ViewModels/MapPanelViewModel.swift:14-18] — #1141 asked you to "narrow the doc-comment to the UIKit shape only," and the first two sentences do exactly that and are accurate (`MapFloatingPanelController.bindPanelDetent()` does switch each case to a `FloatingPanelState`). But the new comment then promises a SwiftUI `PresentationDetent` mapping whose detent values will be "pinned against the real layout constants in `Layouts.swift`." `Layouts.swift` exists, but its constants are FloatingPanel-specific (`EstimatedDrawerTipStateHeight`, anchor insets, `WidthSize`) and don't cleanly translate to SwiftUI detent fractions — the portrait layout exposes no per-detent height constants at all. This is the kind of speculative prose that rots silently if the SwiftUI sheet lands differently. Consider trimming to just the verifiable UIKit mapping, and recording the future intent as a `// TODO:` rather than asserting unbuilt behavior as settled. (The companion line at `MapPanelViewModel.swift:35` makes the same unfulfilled promise.)

- **Stale helper comment (pre-existing, low effort to fix while you're here).** [OBAKitTests/ViewModels/StopViewModelTests.swift:78] — `stubSurveys`'s doc says the stub exists "so `refreshSurveys()` doesn't fatal-error," but the survey path routes failures to `lastError` rather than fatal-erroring. Predates this PR, but you're already touching this method (the `region` envelope addition is correct and necessary, since `StudyResponse.region` is non-optional).

- **Method ordering nit.** [OBAKit/Stops/StopViewController.swift:1297] — `seedCollapsedPastDepartureSections(for:)` sits *between* `bindArrivalsSink` and `bindStopSink`, interrupting the run of `bind*Sink` methods. Moving the helper below the four binders would let them read as a contiguous group. Purely cosmetic — only if you're already in the file.

## Strengths

- The `bindListData()` decomposition is behavior-preserving: all four sinks register in the same order with identical closures, and converting the `if pastDeparturesCollapsed { … }` body into an early-`guard` helper is a faithful, well-named translation.
- The `refreshSurveys()` relocation correctly handles the edge cases: a failed/stopless first fetch still defers surveys to the first *successful* fetch (both old and new code gate on `result.stop != nil`), and the empty-arrivals auto-extension recursion fires surveys exactly once.
- The `stubSurveys` payload fix is a genuinely necessary change, not cosmetic — without the `region` envelope the decode would fail, set `lastError`, and suppress the emission, making the new test assert `0` instead of `1`. Good catch.
- The inline comment justifying *why* surveys moved (lines 209-211) is accurate and earns its place.

## Recommended Action

1. Strengthen the survey test with a settle assertion so it actually asserts "exactly once."
2. Consider trimming the `PanelDetent` doc-comment's forward-looking SwiftUI claims (and the stale `stubSurveys` comment) while you're in there.
3. Re-run `StopViewModelTests` and push.

---

**Verdict: Request Changes.** The code is correct and everything builds and passes, but the one new test this PR adds — its primary contribution beyond the refactor — can't currently catch the regression it's meant to guard. That's a one-line fix, and once the assertion settles at exactly-once, this is ready to merge.

aaronbrethorst:
Merged. I picked up the rest of the review notes in #1165: the two new TODO comments were tripping orphaned_doc_comment so I folded them back into the doc blocks, added removeDuplicates() on the panel detent binding so the new Equatable conformance actually does something, and added a test for the failed survey fetch to lock down the lastError gate. Thanks Mohamed!

PR #1143, opened 2026-05-31
Make CI test failures diagnosable

mosliem:
## Summary

  Two small changes to `.github/workflows/obakittests.yml` so failing CI runs surface the failing test name + assertion message instead of just `exit code 65`.
  
  - Replace kishikawakatsumi/xcresulttool@v1.7.0 with slidoapp/xcresulttool@v3.1.1.
  The upstream action is unmaintained and fails on Xcode 16+ runners (exit code 64) due to missing --legacy support for xcresulttool. The Slido fork includes the fix, maintains compatibility with the existing inputs (path, show-passed-tests), and adds support for Node 24 ahead of GitHub’s Node 20 deprecation on June 16, 2026.
  - **Remove `-quiet` from the test step.** 
  When the xcresulttool action also breaks, the raw workflow log is the only fallback — and `-quiet` was suppressing every `Test Case … failed`  line and assertion message. `-quiet` stays on the build step where it isn't a diagnostic pain point.

  ## Motivation

  Current state when CI fails:
  **TEST EXECUTE FAILED**
`[error]Process completed with exit code 65`
    
  No test name, no file: line, no assertion message. Multiple PRs this week have spent debugging cycles just figuring out *what* failed before being able to start fixing it.

  After this change:
  - Workflow log contains lines like `Test Case '-[OBAKitTests.X testY]' failed (0.012s) — XCTAssertEqual failed: …`.
  - `slidoapp/xcresulttool` posts inline PR annotations on the failing assertion line, restoring the`kishikawakatsumi/xcresulttool` used to provide before it broke.

## Summary by CodeRabbit

* **Chores**
  * CI workflow updated to declare explicit job-level permissions for clearer, safer runs.
  * Test-result upload step improved: uploader replaced with a more reliable tool and its placement refined for consistent test reporting.

### DIFF
--- a/.github/workflows/obakittests.yml
+++ b/.github/workflows/obakittests.yml
@@ -11,6 +11,10 @@ on:
 # Installed versions of Xcode on macOS 15 are documented at:
 # https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md
 
+permissions:
+  checks: write
+  contents: read
+  
 jobs:
   build:
     runs-on: macos-15
--- a/.github/workflows/obakittests.yml
+++ b/.github/workflows/obakittests.yml
@@ -10,15 +10,14 @@ on:
 # https://github.com/actions/runner-images?tab=readme-ov-file
 # Installed versions of Xcode on macOS 15 are documented at:
 # https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md
-
-permissions:
-  checks: write
-  contents: read
   
 jobs:
   build:
     runs-on: macos-15
-
+    permissions:
+      checks: write
+      contents: read
+      
     steps:
     - uses: actions/checkout@v4
 
@@ -76,8 +75,8 @@ jobs:
     # Upload results
     - uses: slidoapp/xcresulttool@v3.1.1
       continue-on-error: true
-      with:
-        show-passed-tests: false    # Avoid truncation of annotations by GitHub by omitting succeeding tests.
-        path: |
-          OBAKitTests.xcresult
       if: success() || failure()
+      with:
+         path: OBAKitTests.xcresult
+         show-passed-tests: false
+         show-code-coverage: false
--- a/.github/workflows/obakittests.yml
+++ b/.github/workflows/obakittests.yml
@@ -68,10 +68,9 @@ jobs:
         -derivedDataPath DerivedData
         -enableCodeCoverage NO
         -resultBundlePath OBAKitTests.xcresult
-        -quiet
 
     # Upload results
-    - uses: kishikawakatsumi/xcresulttool@v1.7.0
+    - uses: slidoapp/xcresulttool@v3.1.1
       continue-on-error: true
       with:
         show-passed-tests: false    # Avoid truncation of annotations by GitHub by omitting succeeding tests.


aaronbrethorst (approved):
Hi Mohamed — this is a small change with an outsized payoff. "Exit code 65, good luck" is exactly the kind of CI papercut that quietly taxes everyone, and you've gone after both the root cause (the unmaintained reporter action that breaks on Xcode 16+) and the fallback (the `-quiet` flag that was swallowing the raw failure lines). I verified the YAML parses cleanly, confirmed `slidoapp/xcresulttool@v3.1.1` is a real, current release (published 2026-01-26) whose `path` / `show-passed-tests` / `show-code-coverage` inputs all match your usage, and confirmed `checks: write` is genuinely the permission that action needs to post its check run.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None. The substantive changes are all correct:

- **The job-level `permissions` block is a real security improvement, not just a requirement.** [.github/workflows/obakittests.yml:17-19] Scoping the token down to `checks: write` + `contents: read` (least privilege) is the right move, and `checks: write` is exactly what `slidoapp/xcresulttool` needs to post the check run. `contents: read` covers `actions/checkout`. Nothing else in the job needs more.
- **`continue-on-error: true` + `if: success() || failure()` is the correct pairing** so the reporter runs after a failed test step and the job's red/green status comes from the test step, not the reporter.
- **Dropping `-quiet` from the test step while keeping it on the build step** matches the stated intent precisely — surface failing test names + assertion messages in the raw log, without re-flooding the (rarely-diagnostic) build output.

## Suggestions (2 found)

- **Pin the third-party action to a commit SHA (repo-wide follow-up, not blocking).** [.github/workflows/obakittests.yml:76] `slidoapp/xcresulttool@v3.1.1` is referenced by a mutable tag, which an upstream maintainer can force-move to different code — and this action runs with `checks: write` and access to `github.token`. GitHub's hardening guidance recommends pinning third-party actions to a full commit SHA (`v3.1.1` currently resolves to `0930e06ae335d8c5e2d2f311ee98d495889a6103`). I'm calling this a suggestion rather than a blocker because every other action in this file is already tag-pinned (`actions/checkout@v4`, `actions/cache@v4`, and the `kishikawakatsumi` action you're replacing), so SHA-pinning *only* this one would be inconsistent — it's better done as a repo-wide sweep with Dependabot keeping the SHAs fresh. Tracked in #1146.
- **Trim the two trailing-whitespace lines you introduced.** [.github/workflows/obakittests.yml:13,20] Lines 13 (after the comment block) and 20 (after `contents: read`) are whitespace-only. Harmless, but since you're already in the file it's a two-second cleanup, and a `yamllint` rule would otherwise start flagging it. Also a minor style note: the new `with:` keys on lines 80-82 use 9-space indentation where the rest of the file uses 8 — valid YAML, just slightly off the house style.

## Strengths

- Tight, well-scoped diff: it does exactly what the title says and nothing more.
- The PR description is genuinely excellent — it explains the failure mode (exit 64/65), why the upstream action broke, why the fork is the right replacement (Node 24 ahead of the June 2026 Node 20 deprecation), and shows the before/after log output. That's the kind of write-up that makes a reviewer's job easy.
- The change is self-justifying and reversible, and the `continue-on-error` guard means that even if the new reporter misbehaves, it can't take CI down with it.

## Recommended Action

1. (Optional, in-scope) Trim the two trailing-whitespace lines while you're here.
2. (Follow-up, out-of-scope) Repo-wide action SHA-pinning + Dependabot tracked in #1146.
3. Merge.

---

**Verdict: Merge.** The change is correct, well-justified, and strictly improves CI diagnosability with no behavioral risk to the build. The two suggestions are genuinely optional — the SHA-pinning is a repo-wide hardening item best handled as its own pass, and the whitespace is cosmetic. Nice work making everyone's failing-CI debugging cheaper.

PR #1144, opened 2026-05-31
Surveys - Routes View Models

mosliem:
## Summary

  Continues the Tier 2 ViewModel migration. Lands two extractions on one branch:

  1. `SurveyViewModel` and `RoutePickerViewModel` pull form and route-picker state out of `SurveyViewController` and `RoutePickerViewController`.
  2. **Survey-flow extraction** — `SurveyOrchestrator` owns the survey primitives that were duplicated between `MapViewController` and `StopViewController` (gate, hero submit + outcome,
  mark-completed, dismiss-with-reminder, reminder advance). `MapViewModel` and `StopViewModel` gain thin screen-specific survey surfaces that delegate into the orchestrator.

  After this, both Map prompt and Stop inline-hero flows have a single VM consumer to bind against — UIKit today, SwiftUI tomorrow.

  ## What changes

  **New files**
  - `OBAKit/ViewModels/SurveyViewModel.swift` — Eureka form state + two-stage submission outcome publisher
  - `OBAKit/ViewModels/RoutePickerViewModel.swift` — cache-first + API-fallback route loading, dedup, search filtering
  - `OBAKit/ViewModels/SurveyOrchestrator.swift` — shared `@MainActor` coordinator over `SurveyService`

  **`MapViewModel`** gains `surveyToPresent` publisher, `checkForSurveyPrompt() async`, `didPresentSurveyPrompt(_:)`, `resetSurveySession()`. The reminder advances only on confirmed
  presentation (the VC checks `SurveyDisplayManager.showSurvey(...)`'s `Bool` and reports back).

  **`StopViewModel`** gains `@Published currentSurvey`, `submitHeroAnswer(_:stopLocation:) async`, `dismissCurrentSurvey()`, `presentFullSurvey` + `surveySubmissionError` publishers.
  `recomputeCurrentSurvey()` hooks into the existing `applySuccessfulFetch` and `refreshSurveys` paths so the card updates from either source.

  **VC rewires**
  - `SurveyViewController` holds `SurveyViewModel`; convenience init preserved for the two existing call sites.
  - `RoutePickerViewController` sinks `CombineLatest3($filteredRoutes, $didFinishLoading, $loadError)` on the main queue.
  - `MapViewController` deletes `checkForMapSurvey`, the session flag, and the retained `SurveyDisplayManager`; binds the new VM publisher.
  - `StopViewController` deletes `surveySection`'s eligibility logic, `handleSurveyAnswer`, and `handleSurveyDismiss`; binds the three new VM publishers.
  `showFullSurvey(_:heroResponseID:)` retained as the UIKit presentation bridge.

  `SurveyDisplayManager`, `SurveyBottomSheetController`, `SurveyStopListItem` are untouched — they stay as UIKit-only presentation helpers.

## Summary by CodeRabbit

* **New Features**
  * One-shot map survey prompt with session-aware reminder handling
  * Inline "hero" survey card on stop screens with smooth handoff to full survey
  * Route picker now loads and filters routes via a dedicated view-model

* **Bug Fixes**
  * More reliable survey submission, dismissal, validation, and error reporting
  * Improved route-loading error/empty-state handling and search behavior

* **Tests**
  * Expanded unit tests covering survey flows, orchestrator, route picker, and stop view model


aaronbrethorst (changes requested):
Mohamed, this is a strong continuation of the Tier 2 VM migration. The extraction is clean, the public surfaces are small, and a couple of the new types — `HeroSubmissionOutcome` and `FullSurveyPresentation` with its non-optional `heroResponseID` — make illegal states genuinely unrepresentable, which is exactly what you want before a SwiftUI consumer comes along. The orchestrator and route-picker tests are excellent: behavior-focused, side-effect assertions on the real `UserDefaultsStore`, and that "route through `apiService: nil` so any network reach throws" idiom is a clever way to *prove* a guard ran.

CLAUDE.md compliance is good: everything new lives in OBAKit (UIKit/Combine allowed), OBAKitCore was untouched, and the VMs themselves stay UIKit/MapKit-free. `@MainActor` isolation, the `nonisolated init`, and the delegate-hop pattern are all correct, and I found no retain cycles — `[weak self]` is used in every sink and closure.

I'm landing on **Request Changes**, almost entirely on test coverage. The two most behaviorally complex things this PR adds have their happy paths untested, and since the whole point of this migration is to lock behavior in *before* the SwiftUI rewrite, that's the gap I'd most want closed now. A couple of small correctness/naming items ride along.

> Note: I reviewed statically and did not run the simulator build/test suite locally.

---

## Critical Issues (0 found)

None. No swallowed errors, no empty catch blocks, no compile-obvious problems.

---

## Important Issues (3 found)

### 1. The inline-hero success paths in `StopViewModel` are untested
`OBAKitTests/ViewModels/StopViewModelTests.swift`

Every survey test here runs against a VM whose `currentSurvey` is always `nil` (the `stubSurveys` helper returns `{"surveys":[]}`). The three inline-hero tests only assert nil/no-op branches:
- `currentSurvey` is nil before fetch
- `submitHeroAnswer` no-ops with no current survey
- `dismissCurrentSurvey` no-ops with no current survey

None of the real machinery added in this PR is exercised:
- `recomputeCurrentSurvey()` driving `currentSurvey` non-nil after a fetch
- `submitHeroAnswer` → `.completed` clearing the card
- `submitHeroAnswer` → `.needsRemainingQuestions` clearing the card **and** emitting `presentFullSurvey` with the right `heroResponseID` + `stopLocation`
- the error path emitting on `surveySubmissionError`
- the `heroSubmitInFlight` re-entrancy guard (the comment at `StopViewModel.swift:298-301` calls card-clearing a re-tap-protection *invariant* — that invariant has no test)

The orchestrator tests cover the outcome *decision* in isolation, but the StopViewModel-specific wiring on top of it — the part most likely to regress — is uncovered. To drive `currentSurvey` non-nil you'll want a non-empty `surveys.json` stub plus `alwaysShowSurveysOnStops=true`; `SurveyServicePrioritizationTests` has the survey fixtures and `SurveyOrchestratorTests.buildLiveSurveyService` shows the live-service pattern.

### 2. The two-stage `submit()` success + retry-skips-hero paths in `SurveyViewModel` are untested
`OBAKitTests/ViewModels/SurveyViewModelTests.swift`

The validation coverage here is genuinely thorough — but because every test uses `apiService: nil`, no test ever produces `.success(())`. So these branches in `SurveyViewModel.submit()` are never verified:
- fresh path: submit hero → capture `submissionResponse.id` into `heroResponseID` → submit the *remaining* responses (the second leg, plus the `remainingResponses.isEmpty` skip)
- retry path: with `heroResponseID` preset, the hero submit is skipped and only `submitAdditionalQuestions` fires
- `markSurveyCompleted` + `.success` emission

This ordering isn't covered at the service layer either (`SurveyServiceStateTests` only tests the nil-apiService throw; `SurveyOrchestratorTests` covers only the single-stage hero). A regression that submitted `responses` instead of `remainingResponses`, or dropped the second leg entirely, would pass the whole suite today. `survey_submission_response.json` (already used by the orchestrator tests) is the fixture you need.

### 3. `SubmissionError.network` mislabels every non-validation failure
`OBAKit/ViewModels/SurveyViewModel.swift:25, 146-148`

The `submit()` catch wraps *any* thrown error — encoding, `APIError.surveyServiceNotConfigured`, decoding, cancellation — as `.network(error)`. The two-case split (validation vs. everything-else) is the right granularity because the VC routes them to different UI, so keep the enum; just rename the case to something honest like `.submissionFailed(Error)`. As written, a future reader (and any logging keyed off the case) will believe a configuration or decode failure was a network blip.

Related, smaller: `submit()` uses `.validationFailed` for two different situations — a real missing-required-answer (`:107`) and the internal "no hero question / no hero response" invariant violation (`:119-122`). The second isn't user error; it surfaces the generic "please answer all required questions" alert for what is actually malformed survey data. Worth a distinct case so the messages don't mislead.

---

## Suggestions (5 found)

- **`submit()` has no in-flight guard.** `StopViewModel.submitHeroAnswer` got a `heroSubmitInFlight` gate in this PR, but `SurveyViewModel.submit()` (called from `submitTapped`) has none, so a fast double-tap fires two concurrent hero submissions. Pre-existing, but now that its sibling is guarded, `SurveyViewModel` is the odd one out. [`SurveyViewModel.swift:106`]

- **Map prompt can still double-present in a narrow window.** `checkForSurveyPrompt()` never flips `hasShownSurveyThisSession` itself — the flag is set later in `didPresentSurveyPrompt`, invoked from a sink with `.receive(on: DispatchQueue.main)`, which always delivers asynchronously even when already on main. So two overlapping `checkForSurveyPrompt()` calls (e.g. `viewDidAppear` firing twice during the `await refreshSurveys()`) can both pass the post-await re-check at `:152` (flag still false) and both `send`. This is the same race the old `checkForMapSurvey` had — not a regression — and the re-check closes most of it, but the cleanest fix is to flip the flag synchronously at the `send` site and have `didPresentSurveyPrompt` only advance the reminder (rolling back if `presented == false`). [`MapViewModel.swift:143-164`, `MapViewController.swift:172-177`]

- **`refreshSurveys()`'s `lastError` gate can act on a stale error — but it's not the bug it looks like.** `fetchSurveys()` only clears `lastError` on an actual fetch; a cooldown-skip after a prior *failure* leaves it set, so the `guard surveyService.lastError == nil` skips that recompute. In practice the card still appears because `applySuccessfulFetch` calls `recomputeCurrentSurvey()` unconditionally on every arrivals refresh, so I'm not treating this as a defect. I'd just tighten the comment at `:246-248` to say the "still emits after a prior success" claim holds *because the prior success cleared `lastError`* — a cooldown-skip after a failure does not. [`StopViewModel.swift:244-257`]

- **`preconditionFailure` on server-shaped data.** `SurveyOrchestrator.submitHero` crashes if `survey.heroQuestion == nil` (`:62-64`). Both callers guard first, so it can't fire today — but `Survey` is decoded from the network, so a hero-less survey is data-shaped, not impossible. Consider `throws` over a hard crash for defense-in-depth. [`SurveyOrchestrator.swift:62`]

- **Two publishers, one redundant reload.** On a survey refresh, `refreshSurveys()` both recomputes `$currentSurvey` (which the VC sinks → `applyData`) and sends `surveysDidRefresh` (also → `applyData`). When the survey actually changes, the list reloads twice in a tick. Harmless, but worth checking whether `surveysDidRefresh` still earns its keep now that `$currentSurvey` drives the section. [`StopViewController.swift:1269-1276`]

Minor polish, take or leave: the `SurveyOrchestrator` "stateless" doc reads as "no consequences" when `submitHero`/`dismiss` mutate persistent reminder state — "stateless façade over `SurveyService`" is more honest; `RoutePickerViewModel.loadError` is pre-flattened to `String?` where the other VMs keep `Error`, which the planned SwiftUI consumer may want back; and `RoutePickerViewModelTests` doesn't cover the cache-first branch or the cancellation branch (both narrow).

---

## Strengths

- **`HeroSubmissionOutcome` and `FullSurveyPresentation`** are model-the-domain wins. The non-optional `heroResponseID` on the presentation payload means "you only get here when there's a hero response to skip" is enforced by construction.
- **The `try!` in `toggleCheckbox` is the right call and well-documented** — it's a deliberate improvement over the old silent `try?` that could leave a checked row with no stored answer. (I'd still phrase the comment as "crashes rather than silently drops" since `try!` doesn't *ensure* correctness, but the risk assessment is sound: encoding a `[String]` can't realistically fail.)
- **`RoutePickerViewModel` finalizes `didFinishLoading` on every exit path**, including `CancellationError` without a spurious error — that fixes the "stuck on Loading routes…" case cleanly.
- **The `@Published` re-emit guard** (`applySuccessfulFetch`, `:218-224`) and its test are a thoughtful touch that avoids re-running the VC's title/tab-bar work every 30 s.
- **`MapBaseType`** keeps MapKit out of the VM exactly the way `PanelDetent` does; the doc cross-reference checks out.
- Comments are unusually accurate across the board — I verified the `MapBaseType` MKMapType mapping, the cancellation rationale, and the re-emit guard against the code and they all hold.

---

## Recommended Action

1. Add success-path coverage for `StopViewModel` inline hero (`.completed`, `.needsRemainingQuestions` payload, error emission, in-flight guard) — Important #1.
2. Add success-path coverage for `SurveyViewModel.submit()` two-stage + retry-skips-hero — Important #2.
3. Rename `SubmissionError.network` and split the "no hero question" case out of `.validationFailed` — Important #3.
4. Consider the `submit()` in-flight guard and the map double-present tightening while you're in here.
5. Re-run the review after the test additions.

**Verdict: Request Changes** — primarily to land tests for the two flows this migration is meant to make safe to rewrite.

aaronbrethorst:
@mosliem this has a merge conflict in stopviewcontroller. please fix it and then I'll do a re-review.

mosliem:
> @mosliem this has a merge conflict in stopviewcontroller. please fix it and then I'll do a re-review.

Done

aaronbrethorst (changes requested):
Mohamed — thanks for the quick turnaround on the merge conflict, and more importantly for the test work. This is a genuinely strong follow-up. Every one of the three Important items from the last round is resolved, and resolved well:

- **`SubmissionError` is fixed and then some.** `.network` is gone; you now have a three-way split — `.validationFailed` (user-recoverable), `.malformedSurveyData` (server-data defect), `.submissionFailed(Error)` (everything else) — and `SurveyViewController` routes each to a distinct alert (`SurveyViewController.swift:69-74`). That's exactly the honest granularity I was after, and the doc comments justify each case.
- **The `StopViewModel` inline-hero success paths are now covered for real.** `test_submitHeroAnswer_completedOutcome_clearsCard`, `_needsRemainingOutcome_clearsCardAndEmitsPresent` (asserting the forwarded `heroResponseID` and `stopLocation`), `_errorPath_emitsErrorAndKeepsCard`, and the re-entrancy guard all stand up a real `SurveyService` over `MockDataLoader`, drive `currentSurvey` non-nil, and assert side effects. No more empty-stub no-ops.
- **The two-stage `SurveyViewModel.submit()` flow is covered with request counting.** The `HitCounter` POST/PUT discrimination pins exactly which legs fire on the fresh, fresh+follow-up, retry, and empty-remaining paths — the precise regression class I was worried about.

I built and ran the five VM test classes locally: **75 tests, 0 failures.** I also confirmed no retain cycles, correct `@MainActor`/`nonisolated init` isolation, and that OBAKitCore stays untouched.

I'm landing on **Request Changes**, on a single real correctness bug in the retry branch of `submit()`. It's a one-line fix, but it silently duplicates a survey response server-side and no test currently catches it — which is exactly the kind of thing this migration exists to prevent.

> Note: I reviewed statically and ran the unit tests locally; I did not exercise the survey flows in the simulator.

---

## Critical Issues (0 found)

None — no swallowed errors, no crashes on the happy paths, no compile problems.

---

## Important Issues (1 found)

### 1. The retry path in `submit()` re-submits the hero answer as an additional response
`OBAKit/ViewModels/SurveyViewModel.swift:145-149` (compare the fresh path at `:167`)

The two submission paths treat the hero response inconsistently:

- **Fresh path (`:167`)** deliberately filters the hero out before the additional-questions leg:
  `let remainingResponses = responses.filter { $0.questionId != heroQuestion.id }`
- **Retry path (`:146`)** passes the entire `responses` array unfiltered:
  `submitAdditionalQuestions(responseID: heroResponseID, additionalResponses: responses)`

This is reachable through the map-prompt full-form flow, which constructs the VC with `heroResponseID: nil` and the full form including the hero (`MapViewController.swift`). Sequence:

1. Fresh submit — hero POST succeeds, so `self.heroResponseID = submissionResponse.id` is set at `:165`.
2. The additional-questions PUT then throws (partial network failure between the two sequential calls). The catch emits `.submissionFailed`; `responses` is **not** cleared and still contains the hero answer.
3. The user taps Submit again. Now `heroResponseID != nil`, so the retry branch fires and sends the full `responses` — hero answer included — into `submitAdditionalQuestions`, duplicating the hero response server-side.

The fix mirrors the fresh path:

```swift
if let heroResponseID = heroResponseID {
    let additional = responses.filter { $0.questionId != survey.heroQuestion?.id }
    _ = try await surveyService.submitAdditionalQuestions(
        responseID: heroResponseID,
        additionalResponses: additional
    )
}
```

Worth noting: the existing retry test (`SurveyViewModelTests.swift:567`, `test_submit_retryPath_skipsHeroSubmit`) starts from a fresh VM whose `responses` never held the hero answer, so it passes regardless of this bug. A test that primes `responses` with the hero answer *and* a preset `heroResponseID` would pin it — and would have caught it.

---

## Suggestions (take or leave)

- **`MapViewModel` double-present race is untested.** `MapViewModel.swift:162-167` flips `hasShownSurveyThisSession = true` synchronously before `send` specifically so two concurrent `checkForSurveyPrompt()` calls emit once. Every map test calls it sequentially. This is the direct analogue of the `StopViewModel` re-entrancy test you *did* write — an `async let a/b` test asserting `received.count == 1` would lock the race you intentionally closed. [`MapViewModelTests.swift`]

- **`OrchestratorError.missingHeroQuestion` throw is untested.** `SurveyOrchestrator.swift:82-84` throws on a hero-less survey (good — `Survey` is network-decoded, so throwing beats crashing). No test passes a hero-less survey to `submitHero`; `test_submitHero_throwsWithoutAPIService` throws for a different reason. A one-liner with a `position != 1`-only survey would cover it. [`SurveyOrchestratorTests.swift`]

- **`test_loadRoutes_retryClearsPriorLoadError` can't actually fail on its own contract.** It builds two separate VMs (`vm` fails, `vm2` succeeds) rather than calling `loadRoutes()` twice on one VM, so the `loadError = nil` clear at `RoutePickerViewModel.swift:82` is only asserted on a VM that never had an error — the test would still pass if that line were deleted. A single-VM error→retry→assert-nil sequence would test the real thing. [`RoutePickerViewModelTests.swift:348`]

- **`StopViewModel.surveysDidRefresh` looks like dead code now.** The VC switched its survey-row driver to `$currentSurvey.dropFirst()` (`StopViewController.swift:1303-1305`), and `surveysDidRefreshSubject` no longer has subscribers. Not a regression — `recomputeCurrentSurvey()` always assigns `currentSurvey` and `@Published` re-emits even on nil→nil — but the publisher, subject, and `send()` are now unused surface. Remove the three, or re-point a subscriber if it was meant to stay. [`StopViewModel.swift:37-40, 263`]

- **`SurveyViewModel.submit()` gives no in-flight UI feedback.** The `submitInFlight` guard correctly drops a concurrent submit, but the button isn't disabled and there's no spinner, so during a slow network submit a second tap is *silently* dropped and the user sees nothing happen. Consider an `@Published isSubmitting` (mirroring `StopViewModel`'s `isLoading`) so the VC can disable/spin. [`SurveyViewModel.swift:134`]

- **Comment polish, two spots.** (a) The `SurveyOrchestrator` class doc opens with "Carries no per-screen state of its own" before walking it back — scope that lead to *instance* state, since `submitHero`/`dismiss`/`noteReminderAndAdvanceSession` all mutate persistent reminder state (`SurveyOrchestrator.swift:23`). (b) The `toggleCheckbox` `try!` comment says it "ensures … can never leave the UI [in a bad state]" — `try!` doesn't ensure correctness, it traps rather than silently desync. The risk assessment is sound; just phrase it as "we'd rather crash than desync" (`SurveyViewModel.swift:96-100`).

- **Type-design parity, for when SwiftUI arrives (not now).** `RoutePickerViewModel` spreads one state machine across `filteredRoutes` + `didFinishLoading` + `loadError` (three `@Published`), where the illegal combo `didFinishLoading == false && loadError != nil` is representable — the same PR shows the better pattern with `HeroSubmissionOutcome`. A single `@Published loadState` enum would collapse it. Similarly, `StopViewModel.surveySubmissionError` emits a raw `Error` while `SurveyViewModel` emits the typed `SubmissionError`; typing the stream would let both submission paths speak the same vocabulary. Both are future-facing, neither blocks this PR. (Also confirmed: `loadError` is `Error?` on this branch, which is the right call — keep it.)

---

## Strengths

- **`HeroSubmissionOutcome` and `FullSurveyPresentation` still make illegal states unrepresentable** — the non-optional `heroResponseID` flows end-to-end from the orchestrator's `.needsRemainingQuestions` payload into the presentation struct, verified intact this round.
- **`RoutePickerViewModel`'s three-shape cancellation match** (`Task.isCancelled` / `CancellationError` / `NSURLErrorCancelled`) finalizes `didFinishLoading` on every exit and never reports a spurious error — the "stuck on Loading routes…" fix is clean, and the cache-first / cancellation branches *are* tested.
- **Replacing the old `preconditionFailure` on `survey.heroQuestion == nil` with a thrown `OrchestratorError`** is the right defensive move for network-decoded data, and it propagates correctly to `surveySubmissionError`.
- **The test fixtures are load-bearing** — the asserted hero id `808d3a515daa39f4c15a` comes from the real `survey_submission_response.json`, so the success assertions can't drift into hardcoded guesses.
- Comments remain unusually accurate; I spot-checked the `MapBaseType`→`MKMapType` mapping, the `@Published` per-cycle re-emit, and the `lastError` gate against the code and they hold.

---

## Recommended Action

1. Filter the hero out of the retry branch in `submit()` (Important #1), and add the `responses`-primed retry test that pins it.
2. Optionally close the two narrow test gaps (map double-present race, orchestrator missing-hero throw) while you're in here.
3. Re-run the suite and re-request review.

**Verdict: Request Changes** — one real correctness bug in the retry path; everything else is suggestion-grade, and the test coverage that drove the last round is now genuinely solid.

aaronbrethorst (approved):
Mohamed — this is exactly the close I was hoping for. The one real correctness bug from the last round is fixed, and fixed with the precise test I asked for: `test_submit_retryPath_filtersHeroFromAdditionalResponses` primes `responses` with *both* the hero and follow-up answers, presets `heroResponseID`, drives the retry PUT, and then **decodes the actual wire payload** to assert `question_id` `[follow.id]` with the hero absent. That test fails the instant anyone reverts the filter — which is the whole point of this migration. Every one of the optional gaps I flagged got closed too. I built the five VM suites and ran them locally: **79 tests, 0 failures.**

> Note: I reviewed statically and ran the unit tests locally (`SurveyViewModel`, `StopViewModel`, `MapViewModel`, `SurveyOrchestrator`, `RoutePickerViewModel` — 79 tests, TEST SUCCEEDED on iPhone 17 Pro / iOS 27). I did not exercise the survey flows in the simulator.

---

## Critical Issues (0 found)

None.

---

## Important Issues (0 found)

The retry-path duplication bug is resolved. `submit()` now filters the hero out before the retry PUT (`SurveyViewModel.swift:157`), mirroring the fresh path:

```swift
if let heroResponseID = heroResponseID {
    let additional = responses.filter { $0.questionId != survey.heroQuestion?.id }
    _ = try await surveyService.submitAdditionalQuestions(
        responseID: heroResponseID,
        additionalResponses: additional
    )
}
```

I verified the filter is correct in every reachable state: the map-prompt full-form flow (hero in `responses`, removed precisely), and the StopViewModel full-survey flow (hero never enters `responses`, filter is a harmless no-op). The `Int != Int?` promotion means a hero-less survey leaves the filter a no-op, which is right.

The other three follow-ups from last round all landed cleanly:
- **`isSubmitting`** added and wired — `SurveyViewController.updateSubmitRow` disables the `"submit"` `ButtonRow` and swaps its title to "Submitting…" while keeping Cancel enabled, so a second tap during a slow submit is no longer silently dropped (`SurveyViewController.swift:88-99`).
- **`surveysDidRefresh`** dead publisher/subject/`send()` removed; no dangling references remain, and the test was migrated to count actual `/surveys.json` wire hits.
- **Comment polish** on the orchestrator class doc and the `toggleCheckbox` `try!` rationale, both now accurate.

---

## Suggestions (take or leave — all deferrable to a follow-up)

- **Checkbox answer serialization is non-deterministic.** `toggleCheckbox` converts a `Set<String>` straight to `Array` (`SurveyViewModel.swift:115`); sorting before encoding gives stable payloads. CodeRabbit flagged this too.
- **Fresh-path wire-level test parity.** The retry test inspects the actual PUT body; the fresh two-leg path (`SurveyViewModel.swift:179`) is pinned by POST/PUT request *counting* but not a body-level hero-exclusion assertion. Same mechanism, so low risk — add the parallel assertion for symmetry.
- **`updateSubmitRow` behavior is untested.** `test_isSubmitting_togglesAroundSubmit` covers the published `[false, true, false]` sequence but not the row's disable/title swap. UI-level; fine to defer.
- **Type-design parity for the SwiftUI consumer (future-facing, as last round).** A single `@Published loadState` enum on `RoutePickerViewModel` would collapse the representable-but-illegal `didFinishLoading == false && loadError != nil` combo; and typing `StopViewModel.surveySubmissionError` to `SubmissionError` would let both submission paths speak the same vocabulary.

These are tracked in #1169.

---

## Strengths

- **The retry fix is pinned at the wire level.** Decoding the real PUT body — not just counting requests — is the assertion that actually prevents the duplicate-response regression from sneaking back.
- **Every optional gap from the last review was closed:** the map double-present race (`async let a/b`, emit-once), the orchestrator missing-hero throw, and the RoutePicker retry-clear test rewritten to a single-VM error→retry→assert-nil sequence that would now fail if the `loadError = nil` clear were deleted.
- **`isSubmitting` removes a silent failure rather than adding surface** — the in-flight guard now has visible UI feedback.
- **Illegal states stay unrepresentable** — non-optional `heroResponseID` flows end-to-end from the orchestrator's `.needsRemainingQuestions` payload into `FullSurveyPresentation`, verified intact.

---

## Verdict

Merge. The deferrable polish items above are tracked in #1169 and none block this PR.

PR #1151, opened 2026-06-04
Speed up and instrument the OBAKitTests CI workflow

aaronbrethorst:
## Summary

The `OBAKitTests` workflow routinely ran **20–55 min** (one main run hit 55m). Investigation (local repro + parsing two real CI job logs) showed the **suite itself runs in ~5s locally / ~2–3 min in CI** — the wall-clock was almost entirely CI infrastructure:

- **Simulator runtime prep** — a silent dead gap between destination resolution and the app launching in the sim. In run `26969522096` this was **~19 min** (17:59:56 → 18:19:15) with zero log output; in another run only ~4 min. Classic on-demand iOS runtime download when the image lags the targeted Xcode point release. **Highly variable → the main cause of the run-to-run spread.**
- **`xcresulttool` upload** — a consistent **~9 min**, even on green runs.
- **`build-for-testing`** — ~5 min.

This PR attacks each and makes the rest visible/time-boxed so a slow run is obvious instead of a mystery gap.

### Workflow (`.github/workflows/obakittests.yml`)
- Pin Xcode via `maxim-lobanov/setup-xcode`; add an explicit, **time-boxed** `xcodebuild -downloadPlatform iOS` step so a runtime download is visible and isolated rather than a silent hang inside the test step.
- Add a simulator-diagnostics preflight (`simctl runtime list` / `-showsdks`).
- Select the simulator by **UDID (`id=`)** instead of `name=iPhone 16` (avoids the "multiple matching destinations" ambiguity; the old hardcoded name isn't a default device on the 26.2 image), and pre-boot it to isolate first-boot cost.
- **Gate the `slidoapp/xcresulttool` upload to `if: failure()`** — saves ~9 min on every green run.
- Add `timeout-minutes` to build (20) and test (15) so a hang fails fast.
- Surface the **slowest 20 tests** in the job summary so a future *test*-time regression (vs. infra time) is easy to spot.
- Narrow the over-broad DerivedData cache key (the old `**/*.yml` glob matched vendored yml under `.build/checkouts`).

### Test fix
- `DataLoadFeedbackGeneratorTests.test_init_withApplication` used `AppConfig(appBundle:userDefaults:analytics:)`, whose convenience init defaults `dataLoader: URLSession.shared`. `Application.init` fires `regionsService.updateRegionsList()`, so the test made a **live request to the production regions server**. It now injects a `MockDataLoader` with stubbed regions — fully offline. (Audited the whole suite: this was the only live-network test.)

## Test plan

- [x] Generated project, built, ran full suite locally: **904 tests, 0 failures in 4.8s**.
- [x] Verified `test_init_withApplication` passes and is offline (`MockDataLoader`, no `URLSession.shared`).
- [x] Exercised the new workflow shell logic locally: UDID picker, `simctl runtime list`, `-showsdks`, and the slowest-tests grep/sed/sort parser against a real test log.
- [x] Ran `xcodebuild` with the new `id=<UDID>` destination form — works.
- [x] `python -c yaml.safe_load` validates the workflow (13 steps).
- [ ] **CI itself** — can only be verified by this PR's own run. Watch whether the `Download iOS platform` step is a fast no-op (runtime preinstalled) or a long download (runtime missing); the diagnostics step's `simctl runtime list` answers that.

## Review notes (non-blocking decisions for the reviewer)

- **Green-run GitHub Checks annotations:** gating the upload to `if: failure()` means green runs no longer publish the per-test Checks summary (that step was the only consumer of the `checks: write` permission). The new "Report slowest tests" step writes a lighter summary to `$GITHUB_STEP_SUMMARY` instead. If you'd rather keep full green-run annotations, say so and I'll revisit the gate. Documented inline.
- **Future test retries:** if `-retry-tests-on-failure` is ever added, broaden the upload gate to `always()` so flaky-but-passing runs still report which tests retried (noted in a comment).
- **Follow-ups (not in this PR):** a shared `createTestApplication()` factory in `OBATestCase` would prevent the live-network footgun class from recurring; and `test_init_withApplication`'s `toNot(beNil())` assertion is tautological (pre-existing).

## Summary by CodeRabbit

* **Chores**
  * Improved macOS CI for more deterministic test runs, caching accuracy, and safer Xcode/runtime setup.
  * Restrained artifact uploads to failures and added simulator/SDK diagnostics.

* **Tests**
  * Added timeouts, structured test logging, and job-summary reporting of slowest tests.
  * Improved test isolation to avoid external network calls during initialization.

* **Style**
  * Reduced noise from per-request network logs by lowering verbosity.

### DIFF
--- a/.github/workflows/obakittests.yml
+++ b/.github/workflows/obakittests.yml
@@ -10,14 +10,26 @@ on:
 # https://github.com/actions/runner-images?tab=readme-ov-file
 # Installed versions of Xcode on macOS 15 are documented at:
 # https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md
-  
+#
+# Performance note (figures observed 2026-06 — a dated snapshot, not an invariant;
+# they will drift as the suite, runner images, and Xcode change): the OBAKitTests
+# suite itself executes in ~5s locally / ~2-3 min in CI. Everything beyond that is
+# CI infrastructure, not the tests:
+#   - iOS Simulator runtime prep / first boot (can silently download a multi-GB
+#     runtime if the image lags the targeted Xcode point release)
+#   - build-for-testing (~5 min)
+#   - the test-results upload step (~9 min; see the gated upload step below)
+# The steps below make each of those visible and time-boxed so a slow run is
+# obvious rather than a mystery 20-minute gap. If you're investigating "slow
+# tests", check the step timings first — it is almost never the tests.
+
 jobs:
   build:
     runs-on: macos-15
     permissions:
       checks: write
       contents: read
-      
+
     steps:
     - uses: actions/checkout@v4
 
@@ -28,16 +40,23 @@ jobs:
         restore-keys: |
           ${{ runner.os }}-spm-
 
+    # DerivedData cache. The key is intentionally narrow: it busts only when the
+    # resolved dependency graph, the XcodeGen project specs, or this workflow
+    # change. The previous `**/*.yml` glob also matched vendored yml under
+    # .build/checkouts and caused needless cache misses (→ full rebuilds).
     - uses: actions/cache@v4
       with:
         path: DerivedData
-        key: ${{ runner.os }}-deriveddata-${{ hashFiles('**/Package.resolved', '**/*.yml') }}
+        key: ${{ runner.os }}-deriveddata-${{ hashFiles('**/Package.resolved', 'project.yml', 'Apps/**/project.yml', '.github/workflows/obakittests.yml') }}
         restore-keys: |
           ${{ runner.os }}-deriveddata-
 
-    - name: Switch Xcode
-      # FIX: Updated from 26.0 to 26.2 because 26.0 was removed in the Jan 27 update
-      run: sudo xcode-select -switch /Applications/Xcode_26.2.app
+    # Pin Xcode deterministically so the set of bundled simulator runtimes is
+    # known, instead of relying on `xcode-select` against whatever the image ships.
+    - name: Setup Xcode
+      uses: maxim-lobanov/setup-xcode@v1
+      with:
+        xcode-version: '26.2'
 
     - name: Install xcodegen
       env:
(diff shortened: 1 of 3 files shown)


aaronbrethorst:
Follow-up: the "hundreds of live API calls" in the CI log are **not live** — they're mock-served, just logged loudly.

Evidence from this PR's own run (`26972769402`):
- **Median begin→finish latency 0.27 ms** (min 0.04 ms) across 1,465 paired requests — physically impossible for real network (DNS+TCP+TLS+round-trip); that's in-memory mock serving.
- **478 requests target `*.example.com`** (non-resolvable) yet all "succeed" in ~2 ms → served by `MockDataLoader`, which `fatalError`s on any unmatched request (and the suite passes).
- The `api.pugetsound.onebusaway.org` / `onebusaway.co` hostnames come from the **Puget Sound region fixture** the tests load; the requests are still intercepted by the mock.
- The 6 "errors" are URLErrors `-1001/-1009/-1011` **injected** by `ErrorClassifierTests`/weather-failure tests, not real failures.

Added commit `5bcdcf9`: downgrade the per-request begin/finish logs in `APIService.data(for:)` from `.info` to `.debug` so ~3,200 mock-request lines stop flooding the CI log and masquerading as live traffic. Failure logs stay at `.error`.

aaronbrethorst:
✅ Green and fast. Run `26974769322`: **6m31s total** (was 20–55 min).

Step breakdown:
- Download iOS platform: **5s** (now skips — the macOS image already ships iOS 26.2/26.1/26.0.1/18.x; `-downloadPlatform` was needlessly fetching 26.3.1, ~2.5 min, and flaked with exit 70 in run `26973663197`)
- Build: 2m42s
- **OBAKit Unit Test: 57s** — 921 tests, 0 failures
- xcresulttool upload: **skipped** (`if: failure()`, green run → ~9 min saved)

And on the original concern: the per-request network log lines dropped from **~3,200 → 0** (downgraded to `.debug`), so the log no longer looks like live API traffic. Combined with the latency evidence (0.27 ms median), confirmed there are **no live network calls** — they were always mock-served.

Minor pre-existing heads-up (not from this PR): GitHub warns `actions/checkout@v4`/`actions/cache@v4` run on Node 20, forced to Node 24 on 2026-06-16. Worth a follow-up bump.

PR #1153, opened 2026-06-05
Alerts - Agencies View Models

mosliem:
## Summary

  Continues the VC → ViewModel migration for the alerts and agencies screens. Extracts three `@MainActor
  ObservableObject` view models; view controllers keep only UIKit wiring.

  ## Changes

  - **`AgenciesViewModel`** — owns agency fetch + sorting and `isLoading` / `loadError`. VC reads
  `viewModel.agencies`.
  - **`AgencyAlertsViewModel`** — adopts `AgencyAlertsDelegate`, owns deduped `alerts`, `isLoading`, and
  `collapsedSections`. VC binds via Combine; share-item logic moved into `shareActivityItems(for:)`.
  - **`ServiceAlertViewModel`** — owns background HTML build, `markRead` side effect, and `@Published
  renderedHTML`. Removes ~165 lines from the VC, including a stale commented-out "Consequences" block.

  Unit tests added under `OBAKitTests/ViewModels/`.

## Summary by CodeRabbit

* **Improvements**
  * Faster, more responsive loading and UI updates for agency lists, agency alerts, and service alert views.
  * Agency lists now present sorted results when available.
  * Alerts are deduplicated and collapse/expand state is preserved across refreshes.
  * Service alert content now renders reliably into styled HTML and marking alerts as read is idempotent.

* **Tests**
  * Added unit tests covering agencies, agency alerts, and service alert rendering/behaviors.

### DIFF
--- a/OBAKit/Agencies/AgenciesViewController.swift
+++ b/OBAKit/Agencies/AgenciesViewController.swift
@@ -14,6 +14,14 @@ import OBAKitCore
 /// Loads and displays a list of agencies in the current region.
 class AgenciesViewController: TaskController<[AgencyWithCoverage]>, OBAListViewDataSource {
     let listView = OBAListView()
+    private let viewModel: AgenciesViewModel
+
+    override init(application: Application) {
+        self.viewModel = AgenciesViewModel(application: application)
+        super.init(application: application)
+    }
+
+    required init?(coder: NSCoder) { fatalError("init(coder:) has not been implemented") }
 
     override func viewDidLoad() {
         super.viewDidLoad()
@@ -28,10 +36,6 @@ class AgenciesViewController: TaskController<[AgencyWithCoverage]>, OBAListViewD
     }
 
     override func loadData() async throws -> [AgencyWithCoverage] {
-        guard let apiService = application.apiService else {
-            throw UnstructuredError("No API Service")
-        }
-
         ProgressHUD.show()
         defer {
             Task { @MainActor in
@@ -39,7 +43,7 @@ class AgenciesViewController: TaskController<[AgencyWithCoverage]>, OBAListViewD
             }
         }
 
-        return try await apiService.getAgenciesWithCoverage().list
+        return try await viewModel.loadData()
     }
 
     @MainActor
@@ -49,10 +53,10 @@ class AgenciesViewController: TaskController<[AgencyWithCoverage]>, OBAListViewD
 
     // MARK: - OBAListKit
     func items(for listView: OBAListView) -> [OBAListViewSection] {
-        guard let agencies = data else { return [] }
+        let agencies = viewModel.agencies
+        guard !agencies.isEmpty else { return [] }
 
         let rows = agencies
-            .sorted(by: { $0.agency.name < $1.agency.name })
             .map { agency -> OBAListRowView.DefaultViewModel in
                 OBAListRowView.DefaultViewModel(
                     title: agency.agency.name,
--- a/OBAKit/Alerts/AgencyAlertsViewController.swift
+++ b/OBAKit/Alerts/AgencyAlertsViewController.swift
@@ -8,12 +8,12 @@
 //
 
 import UIKit
+import Combine
 import OBAKitCore
 import SafariServices
 
 /// Displays `AgencyAlert` objects loaded from a Protobuf feed.
 class AgencyAlertsViewController: UIViewController,
-    AgencyAlertsDelegate,
     AgencyAlertListViewConverters,
     AppContext,
     OBAListViewCollapsibleSectionsDelegate,
@@ -22,11 +22,17 @@ class AgencyAlertsViewController: UIViewController,
 
     // MARK: - Stores
     public let application: Application
-    private let alertsStore: AgencyAlertsStore
+    private let viewModel: AgencyAlertsViewModel
+    private var cancellables = Set<AnyCancellable>()
 
     // MARK: - UI state
     let selectionFeedbackGenerator: UISelectionFeedbackGenerator? = UISelectionFeedbackGenerator()
-    var collapsedSections: Set<String> = []
+
+    /// Forwards to the VM. Required by `OBAListViewCollapsibleSectionsDelegate`.
+    var collapsedSections: Set<String> {
+        get { viewModel.collapsedSections }
+        set { viewModel.collapsedSections = newValue }
+    }
 
     // MARK: - UI elements
     let listView = OBAListView()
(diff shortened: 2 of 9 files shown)


aaronbrethorst (approved):
Hi Mohamed — this is a clean continuation of the VC → ViewModel migration, and it shows. The view controllers genuinely shed responsibility here: `ServiceAlertViewController` drops ~165 lines, the HTML builder gets a sensible value-type redesign, and the three view models share a consistent `@MainActor final class … ObservableObject` shape that reads like a deliberate template. I traced every behavior path against `main` and the refactor is faithful — no retain cycles, no concurrency regressions, and a couple of quiet *improvements* (more on those below). The test build compiles cleanly.

Nothing here blocks the merge. The loose ends are all in the test layer — two docstrings claim coverage that doesn't exist, the share helper they advertise has no test, and a `try?` hides failure causes — but none of that affects shipping behavior, and the untested logic (dedup, share helper) was already untested in the VC before this PR, so coverage isn't regressed. Worth cleaning up as a fast-follow.

## Critical Issues (0 found)

None. The shipping code is correct and safe.

## Fast-Follow (3 found)

1. **Test docstring claims coverage that doesn't exist — share helper** [OBAKitTests/ViewModels/AgencyAlertsViewModelTests.swift:16-18]
   The type docstring says it "Verifies the share-activity helper, `collapsedSections` round-trip, and the loading flag transitions," but there is no test for `shareActivityItems(for:)` — the class only covers init/loading/collapsed-sections/displayError. The helper has two branches ([AgencyAlertsViewModel.swift:60-65]: `localizedURL` present → `[url]`, absent → `[title, body]`) and is pure, deterministic, network-free. Please add `test_shareActivityItems_withURL` / `test_shareActivityItems_withoutURL`, or drop the clause. Adding the test is the better fix since it closes a real gap the docstring already promises.

2. **Test docstring describes a property the type doesn't have** [OBAKitTests/ViewModels/AgenciesViewModelTests.swift:16-17]
   The docstring claims the tests verify "that loading state resets to `false` after completion." `AgenciesViewModel` has no loading state at all — its only published property is `agencies`; the `ProgressHUD` handling stayed in the VC. Trim the clause to match reality ("Verifies the success path sorts by name").

3. **`try?` in the success test swallows the real failure** [OBAKitTests/ViewModels/AgenciesViewModelTests.swift:81]
   `_ = try? await viewModel.loadData()` means that if `loadData()` ever throws, the test falls through to `expect(agencies).toNot(beEmpty())` and fails with a useless "expected to not be empty" instead of the actual error. The test is already `async`; make it `async throws` and use `try` so the cause surfaces.

## Suggestions (6 found)

1. **The dedup logic — a headline feature of this PR — is never exercised against real data** [OBAKitTests/ViewModels/AgencyAlertsViewModelTests.swift]. Every test runs against an empty store, so `dedupedAlerts()` ([AgencyAlertsViewModel.swift:80-83]) never filters anything. There's a purpose-built test seam for this — `AgencyAlertsStore.insertAlerts(_:)` ([OBAKitCore/Models/AgencyAlertsStore.swift:178]) is `@MainActor`/internal specifically for `@testable` fixture population, and `MapPanelViewModelTests.swift:138` shows the `AgencyAlert(feedEntity:agency:)` construction pattern. Inserting two alerts sharing an `id` plus one unique, then asserting `alerts.count == 2`, would actually prove the dedup works. Worth doing, but fine as a fast-follow.

2. **`AgenciesViewModel.loadData()`'s nil-`apiService` throw is untested** [OBAKit/ViewModels/AgenciesViewModel.swift:40-42]. `ApplicationTests.swift:130` shows how to build an Application where `apiService` stays nil; a `test_loadData_noAPIService_throws` would pin the error contract the VC routes through `TaskController`.

3. **`AgenciesViewModel.loadData()` is a dual source of truth** [OBAKit/ViewModels/AgenciesViewModel.swift:39-47]. It both assigns `self.agencies` *and* returns the sorted array, because `TaskController` requires a return value — so the same list lives in both `viewModel.agencies` (what `items(for:)` renders) and `TaskController.data` (now unread). It works because both come from one `loadData()` call, but it's a latent trap and the one VM that diverges from the "publish only" pattern of the other two. At minimum, a one-line comment noting `data` exists only to drive `updateUI()`; ideally, pick one channel so future migrations have a clean template.

4. **`viewDidAppear()` leaks a UIKit lifecycle name into a UIKit-free VM** [OBAKit/ViewModels/ServiceAlertViewModel.swift:49]. Since the three VMs are meant as a template, a caller-agnostic name (`onAppear()` / `prepare()`) would keep the VM independent of the VC's vocabulary. Naming across the trio is also ad hoc (`loadData` / `reloadServerData` / `viewDidAppear`) — worth a convention.

5. **`isLoading` initial-emission ordering quirk** [OBAKit/Alerts/AgencyAlertsViewController.swift:77-90]. The `$isLoading` sink uses `.receive(on: .main)`, so its initial `false` is delivered asynchronously while `reloadServerData()` runs synchronously and sets `isLoading = true` + `beginRefreshing()`. The buffered values then arrive (`false` → `endRefreshing()`, `true` → spinner). It converges correctly within the same runloop, so no visible bug — but a transient `endRefreshing()` right after `beginRefreshing()` is fragile under future edits.

6. **`ServiceAlertViewModel` spreads "build once" across three flags** (`hasMarkedRead`, `isBuilding`, `renderedHTML == nil`) [ServiceAlertViewModel.swift:49-55]. A single `enum State { case idle, building, rendered(String) }` would collapse `isBuilding` + `renderedHTML` into one source of truth and make the legal transitions self-evident. Nice-to-have for the template VM that owns an async-rendered artifact. Also, `buildQueue` is `.concurrent` but only ever runs one job — `.serial` would express intent better.

## Strengths

- **Genuine VC slimming.** `ServiceAlertViewController` is now pure WebKit wiring; the HTML assembly is a `static` pure function ([ServiceAlertViewModel.swift:81]) with no `self` capture — naturally concurrency-safe and the best-encapsulated of the three.
- **`HTMLBuilder` class → struct is a real improvement.** Value semantics + `mutating`/`inout` eliminate the reference-aliasing footgun the old class allowed, and the nested-builder ergonomics survive intact.
- **A behavior *improvement*, not just a port.** The new `AgencyAlertsViewModel.agencyAlertsStore(_:displayError:)` clears `isLoading` on error — the original VC never implemented `displayError`, so a failed refresh spun the activity indicator forever. (User-facing error surfacing is still handled, unchanged, by `Application` as a separate store delegate — so nothing regressed there. A one-line comment on that empty-looking handler explaining *why* it only clears loading would prevent a future "this swallows errors" misread.)
- **Lifecycle is correct.** The VC strongly holds the VM; `AgencyAlertsStore` holds delegates in an `NSHashTable.weakObjects()`, so dropping the old `deinit`/delegate removal is safe.
- **The sort test is designed right** — it asserts against an intentionally-unsorted fixture, so it would actually catch a regression. That's a mistake many people make and you avoided it.

---

**Verdict: Merge.** Clean, behavior-preserving refactor with no code defects. The test-doc accuracy and coverage gaps are real but minor and don't regress anything — fold them into a follow-up.

PR #1154, opened 2026-06-05
Current Trip View Model

mosliem:
## Summary

Moves state ownership for the **"My Trip"** screen out of the view controller and into a new `@MainActor ObservableObject` under `OBAKit/ViewModels/`.

The view model now owns:

- `State` enum
- `matchResults`
- `findVehicle()` task
- Settable `pendingNavigation: ArrivalDeparture?` output
- Refresh and idle timers

The view controller retains UIKit-specific responsibilities:
- `OBAListView` management
- Navigation push behavior
- A small Combine binding layer

---

### New: `OBAKit/ViewModels/CurrentTripViewModel.swift`

The new view model owns:

- `State`
- `matchResults`
- `pendingNavigation`
- `findVehicle()`

Additional internal seams were introduced to improve testability:

- `handle(results:)`
- `handle(error:)`
- `pendingNavigationUnavailable()`

These allow tests to cover every state transition without requiring the full matcher round-trip.

### Refactored: `OBAKit/Trip/CurrentTripViewController.swift`
- `$state` sink now drives:
  - `listView.applyData()`
  - Haptic feedback
- `$pendingNavigation` sink now:
  - Performs the navigation stack swap
  - Falls back through `pendingNavigationUnavailable()` when no navigation controller is available

### New Tests: `OBAKitTests/ViewModels/CurrentTripViewModelTests.swift`

Added **10 tests** covering:

- Initial state
- All `handle(results:)` result-count branches
- Both `handle(error:)` branches
- `noStopsNearby` fall-through
- No-location flow
- Missing navigation controller fallback

---

## SwiftUI Readiness

`pendingNavigation` is implemented as:

```swift
@Published var pendingNavigation: ArrivalDeparture?
```

This matches the shape expected by:

- `.navigationDestination(item:)`
- `.sheet(item:)`

A future SwiftUI consumer can bind directly to this property without requiring any view model changes.

---

## Behavior Preserved

- `noResults` continues to trigger a **success** haptic, since an empty match array represents a successful fetch.
- **Failed** haptics remain reserved for thrown errors.
- The single-match navigation-controller fallback introduced in `51a917db` is preserved through the new `pendingNavigationUnavailable()` intent.
- `MatchError.noRealtimeData` continues to map to `.noRealtime`.
- All other errors continue to map to `.error(error)`.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced the current trip-finding experience with improved state transitions and clearer error feedback across scenarios including no location, no results, no real-time data, and multiple matches.
  * Added periodic refresh capability for trip availability monitoring.

* **Tests**
  * Added comprehensive test coverage for trip-finding state transitions and error handling.

### DIFF
--- a/OBAKit/Trip/CurrentTripViewController.swift
+++ b/OBAKit/Trip/CurrentTripViewController.swift
@@ -8,13 +8,13 @@
 //
 
 import UIKit
+import Combine
 import OBAKitCore
-import CoreLocation
 
 /// Finds and displays the user's current trip on a selected route.
 ///
 /// After the user picks a route in `RoutePickerViewController`, this controller:
-/// 1. Queries `NearbyTripMatcher` for nearby active vehicles on that route.
+/// 1. Asks `CurrentTripViewModel` to query nearby active vehicles on that route.
 /// 2. If one match → navigates directly to `TripViewController`.
 /// 3. If multiple → shows a disambiguation list.
 /// 4. If none → shows an appropriate error state.
@@ -25,24 +25,10 @@ class CurrentTripViewController: UIViewController,
 
     let application: Application
 
-    private let route: Route
+    private let viewModel: CurrentTripViewModel
     private let listView = OBAListView()
     private lazy var dataLoadFeedbackGenerator = DataLoadFeedbackGenerator(application: application)
-
-    /// Matching results found near the user.
-    private var matchResults = [NearbyTripMatcher.MatchResult]()
-
-    /// Current loading/error state.
-    private enum State {
-        case loading
-        case noLocation
-        case noResults
-        case noRealtime
-        case multipleResults
-        case error(Error)
-    }
-
-    private var state: State = .loading
+    private var cancellables = Set<AnyCancellable>()
 
     // MARK: - Timer
 
@@ -53,7 +39,7 @@ class CurrentTripViewController: UIViewController,
 
     init(application: Application, route: Route) {
         self.application = application
-        self.route = route
+        self.viewModel = CurrentTripViewModel(application: application, route: route)
         super.init(nibName: nil, bundle: nil)
 
         title = OBALoc(
@@ -67,7 +53,6 @@ class CurrentTripViewController: UIViewController,
 
     deinit {
         refreshTimer?.invalidate()
-        findVehicleTask?.cancel()
     }
 
     // MARK: - UIViewController Lifecycle
@@ -81,7 +66,8 @@ class CurrentTripViewController: UIViewController,
         view.addSubview(listView)
         listView.pinToSuperview(.edges)
 
-        findVehicle()
+        bindViewModel()
+        viewModel.findVehicle()
     }
 
     override func viewWillAppear(_ animated: Bool) {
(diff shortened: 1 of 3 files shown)


aaronbrethorst (approved):
Hi Mohamed — this is a really clean continuation of the MVVM migration. The extraction has good bones: the view model imports no UIKit, the `State` enum gives the controller a compile-time-exhaustive `switch`, the concurrency is handled carefully (single-flight `findVehicleTask`, `Task.isCancelled` guards before every publish, timer hop back onto `@MainActor`, and a `deinit` that only touches `invalidate()`/`cancel()`), and the testable seams (`handle(results:)`, `handle(error:)`, `pendingNavigationUnavailable()`) are exactly the right boundaries to pull out. I verified the 10 unit tests pass locally (`Executed 10 tests, with 0 failures` on an iOS 26.3.1 simulator), and I confirmed there are no retain cycles in the Combine bindings.

A few things to weigh — none of them block the merge.

## Critical Issues (0 found)

None.

## Important Issues (0 found)

None that must be fixed before merge. The items below are worth a look, but they're design/UX refinements rather than correctness bugs, so I've filed them as a tracking issue (see Recommended Action) rather than gating this PR on them.

## Suggestions (5 found)

1. **`pendingNavigation` is publicly settable, which is broader than the actual need** [`CurrentTripViewModel.swift:52`]. The single-match outcome lives in `pendingNavigation` while `state` stays `.loading` — a deliberate, tested choice (`CurrentTripViewModelTests.swift:146-147`) with a legitimate SwiftUI rationale (`.navigationDestination(item:)` wants a settable optional). But the only write the consumer ever performs is `= nil` (`CurrentTripViewController.swift:114`). A fully open `@Published var` also permits setting an *arbitrary* non-nil value while `state == .multipleResults`, producing a model the type can't reject. Since this PR is explicitly about SwiftUI-readiness, it's worth getting the contract tight now: either fold the single match into the enum (`case singleMatch(ArrivalDeparture)`, with `pendingNavigation` derived) or keep `pendingNavigation` as `private(set)` and expose an `acknowledge`/`clear` method (plus a `Binding` whose setter only accepts `nil`) for SwiftUI.

2. **`findVehicle()` now runs on every `viewWillAppear`, not once** [`CurrentTripViewController.swift:67`]. Previously the initial find happened in `viewDidLoad` (once) and `start()` only managed the timer. Now `start()` does both. In the multiple-results flow, tapping a row *pushes* `TripViewController` (`:211-214`), so returning to the disambiguation list re-fires `viewWillAppear` → `start()` → a fresh network find and a fresh success/failure haptic. This is arguably an improvement (fresh data on return), but it's an unflagged behavior change — confirm it's intended, especially the haptic on a screen the user is merely navigating back to.

3. **The `apiService == nil` path now plays a failure haptic where the original was silent** [`CurrentTripViewModel.swift:113-115` → `CurrentTripViewController.swift:94-95`]. In the old controller this branch set `.error` but deliberately skipped `dataLoad(.failed)`; routing all `.error` states through the `$state` sink means the "no region / no service" case now buzzes. Reasonable as an improvement, but it isn't in the PR's "Behavior Preserved" list, so flagging it as a conscious decision.

4. **The `apiService == nil` → `noServiceError()` path has no test** [`CurrentTripViewModel.swift:113-116`, `:191-201`]. This is the entire reason `noServiceError()` exists, and it's a real user-facing scenario (no region selected). The sibling `.noLocation` early-return *is* tested (`:270`); this one isn't, and `noServiceError()` has zero coverage, so a typo in its localization key or domain would slip through. It's testable by building the `Application` without `fixedRegionName` so `apiService` stays nil — though that needs a variant of the `createApplication` helper, which currently always supplies a region.

5. **`noStopsNearby` renders as a generic error with a dead "Try Again" button** [`CurrentTripViewModel.swift:154-161`]. `handle(error:)` special-cases only `noRealtimeData`; `MatchError.noStopsNearby` (a benign "no stops near you" condition) falls to `.error`, which fires a failure haptic, logs `Logger.error`, and shows a retry button that will deterministically fail again. This is **pre-existing** behavior, lifted verbatim from the old controller and even pinned by a test (`:206`), so it's explicitly out of scope here — but since the mapping just moved into a clean seam, it's a cheap future win to give it a dedicated benign state (`.noResults`-like) so it stops logging as an error.

## Strengths

- Disciplined Swift concurrency: single-flight task, `Task.isCancelled` checks at every publish point, `CancellationError` correctly swallowed as non-failure, `@MainActor` isolation, and clean `deinit` teardown.
- The `State` enum drives an exhaustive `switch` in the controller, so adding a case forces the UI to handle it — compile-time safety for free.
- Tests are honest and target real contracts, not implementation: `test_handleResults_single_setsPendingNavigation` asserts `state` *stays* `.loading` on a single match; `test_handleError_noStopsNearby_fallsThroughToError` guards an easy-to-get-wrong branch; `test_start_invokesFindVehicle` guards against `start()` silently becoming a no-op. Every branch of `handle(results:)` and `handle(error:)` is covered, and `pendingNavigationUnavailable()` is tested down to the "`matchResults` survives the fallback" detail.
- `shouldSkipProgrammaticRefresh` is a tidy way to keep the VoiceOver/UIKit concern out of the view model while preserving the original accessibility behavior.
- I checked the "silent stale list" theory on the no-nav-controller refresh path and it doesn't hold: `@Published` republishes on every assignment (and `State` isn't `Equatable`, so no `removeDuplicates` can swallow it), so `pendingNavigationUnavailable()` setting `state = .multipleResults` always re-fires the sink and `applyData()`. No redraw gap.

## Recommended Action

This is ready to merge. I've opened [#1168](https://github.com/OneBusAway/onebusaway-ios/issues/1168) to track the refinements above so they don't get lost:

1. Merge as-is.
2. Address the follow-ups (pendingNavigation encapsulation, the no-service test, and the two behavior-change confirmations) in the tracked issue, ideally as the SwiftUI consumer lands.

PR #1156, opened 2026-06-13
Fix region-wide alerts never showing as modal bulletins

aaronbrethorst:
## Summary

Region-wide transit alerts — high-severity alerts that apply to the whole region rather than a single agency, served by `sidecar.onebusaway.org` with an **empty** GTFS-RT `agency_id` — were being fetched successfully but never appeared as the modal BulletinBoard alert they're meant to trigger. There were two independent root causes:

- **Parsing silently dropped them.** `AgencyAlert`'s initializer rejected any alert whose `informed_entity` `agency_id` didn't match a known agency. An empty `agency_id` matches nothing, so every region-wide alert threw `AlertError.unknownAgency` and was swallowed by `try?` in both the REST and Obaco fetch paths. An empty `agency_id` is now a first-class **region-wide** alert (`agency == nil`, new `isRegionWide`); a *non-empty* unknown `agency_id` still throws. Both parse sites now **log** dropped entities instead of swallowing them, resolving a pre-existing `// TODO: Don't swallow error`.
- **The bulletin window was never displayed.** BulletinBoard's `showBulletin(in:)` creates a `UIWindow` with no `windowScene`, which iOS does not display in a scene-based app — so even when an alert *was* present, nothing showed. A new `BLTNItemManager.show(in:)` extension presents above the key window's top view controller, and all four bulletin call sites (agency alert, error, reachability, region mismatch) now route through it.

## Test plan

- [x] Unit: region-wide (empty `agency_id`) parses with `agency == nil`; non-empty unknown agency still throws `.unknownAgency`; missing `agency_id` (presence bit unset) throws `.invalidAlert`.
- [x] Unit: `ObacoAPIService.getAlerts` includes region-wide alerts, and one malformed entity in a feed doesn't drop the valid siblings (guards the new per-entity `do/catch`).
- [x] Unit: full `AgencyAlertsStore.update()` pipeline surfaces a region-wide alert into `recentUnreadHighSeverityAlerts` (the exact predicate the app uses to present the bulletin).
- [x] UI: `test_regionWideAlertBulletin_appears_andIsDismissible` drives launch -> bulletin appears above the main UI -> Dismiss -> bulletin gone, via a DEBUG-only alert-injection hook (mirrors the existing `TEST_SHOW_ALL_TIPS` pattern).
- [x] Full `OBAKitTests` unit suite green; `MainFlowUITests` green in isolation (the two pre-existing live-network tests are flaky only under simulator contention, not from this change).
- [x] SwiftLint clean on changed files.

## Notes for the reviewer

- **Base is `release-ready`, not `main`**, because this work builds on the `OBAKitUITests` target added earlier on that branch. It's a focused diff of just the alert fix.
- **DEBUG test scaffolding** (`seedRegionWideAlertForTesting`, `TEST_INJECT_REGION_WIDE_ALERT`, `TEST_SUPPRESS_ALERT_BULLETINS`, `suppressLiveFetchesForTesting`) is all `#if DEBUG`-gated and follows the existing `TEST_SHOW_ALL_TIPS` convention.
- **Non-blocking judgment call from review:** the two `getAlerts` parse sites use a catch-all `catch` (log + drop one entity) rather than `catch let error as AgencyAlert.AlertError`. Catch-all keeps the two paths consistent (the REST one is non-throwing) and the initializer's only throw surface today is `AlertError`. Happy to narrow it to a typed catch if you'd prefer unexpected errors to propagate.

## Release Notes

* **New Features**
  * Added support for region-wide alerts that apply to the entire transit system, not just specific agencies.

* **Improvements**
  * Enhanced alert bulletin presentation for improved reliability and consistency.
  * Strengthened error handling and logging when processing alerts to prevent silent failures.

### DIFF
--- a/OBAKit/Alerts/AgencyAlertBulletin.swift
+++ b/OBAKit/Alerts/AgencyAlertBulletin.swift
@@ -50,10 +50,6 @@ class AgencyAlertBulletin: NSObject {
     }
 
     func show(in application: UIApplication) {
-        guard !bulletinManager.isShowingBulletin else {
-            return
-        }
-
-        bulletinManager.showBulletin(in: application)
+        bulletinManager.show(in: application)
     }
 }
--- a/OBAKit/BLTNBoard/OBABulletinPage.swift
+++ b/OBAKit/BLTNBoard/OBABulletinPage.swift
@@ -23,3 +23,21 @@ class ThemedBulletinPage: BLTNPageItem {
         appearance.imageViewTintColor = ThemeColors.shared.brand
     }
 }
+
+extension BLTNItemManager {
+    /// Presents the bulletin above the topmost view controller of the application's key window.
+    ///
+    /// Use this instead of `showBulletin(in:)`: that method presents inside a `UIWindow`
+    /// it creates without a `windowScene`, and iOS never displays such a window in a
+    /// scene-based app, so the bulletin silently fails to appear.
+    func show(in application: UIApplication) {
+        guard
+            !isShowingBulletin,
+            let topViewController = application.keyWindowFromScene?.topViewController
+        else {
+            return
+        }
+
+        showBulletin(above: topViewController)
+    }
+}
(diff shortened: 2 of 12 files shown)


aaronbrethorst:
@coderabbitai review

PR #1163, opened 2026-06-21
Floating sheet system — foundation

mosliem:
## Summary

  Introduces the foundation for a SwiftUI-based **map panel experience** that replaces the classic UIKit tab bar with a full-screen map and a floating bottom sheet. The
  new experience is gated behind an experimental settings toggle and ships alongside the existing classic UI — no behavior change for users by default.

  ## What's included

  ### Sheet coordinator foundation
  - `SheetRoute` — typed enum describing every sheet destination (home, search, stop details, trip planner, etc.).
  - `SheetCoordinator` — `@MainActor` navigation stack driving sheet presentation.
  - `FloatingSheetContainer` — SwiftUI container that hosts the bottom sheet over the map.

  ### Content & DI
  - `HomeSheetView` + `HomeSheetViewModel` — initial home sheet content.
  - `AppSheetViewFactory` — single DI seam mapping `SheetRoute` cases to views, with compiler-enforced exhaustiveness.
  - `MapPanelRootView` — SwiftUI root that composes the map and floating sheet.

  ### Integration
  - `ApplicationRootControllerFactory` (`@objc`) reads the feature flag and returns either `ClassicApplicationRootController` (unchanged) or `MapPanelRootController` (hosts `MapPanelRootView`). AppDelegates call the factory directly.
  - `BulletinOverlayWindow` presents `BLTNItemManager` in a dedicated `.alert`-level `UIWindow` so bulletins aren't clamped by the floating sheet's detent.
  - `SettingsViewController` adds an **Experimental** section with the "Use map panel experience" toggle (requires app restart).

  ### Tests
  - `AppSheetRouteTests` — covers route identity and equality.
  - `SheetCoordinatorTests` — covers push/pop/replace/reset behavior.

  ## Rollout

  - Default OFF. Opt-in via Settings → Experimental → "Use map panel experience".
  - Restart required to switch experiences (root controller is constructed once at launch).
  
  ### Video 

https://github.com/user-attachments/assets/df44d2b2-be11-4dfc-8e4d-1295e3e918fd

## Release Notes

* **New Features**
  * Added an experimental **Map Panel Experience** toggle in Preferences.
  * Introduced a new **Map Panel** UI mode (map-first experience) alongside the classic tabbed interface.
  * Added a persistent **floating sheet** system with expandable Home search, per-destination detents, and optional stacked-sheet navigation.

* **Bug Fixes**
  * Updated map tab tapping and page navigation behavior to respect the active UI mode.

* **Tests**
  * Added unit tests for sheet routing IDs, detent configuration, and coordinator navigation (push/pop/truncate/pop-to-root).

### DIFF
--- a/Apps/KiedyBus/AppDelegate.m
+++ b/Apps/KiedyBus/AppDelegate.m
@@ -13,7 +13,7 @@
 
 @interface AppDelegate ()<OBAApplicationDelegate>
 @property(nonatomic,strong) NSUserDefaults *userDefaults;
-@property(nonatomic,strong) OBAClassicApplicationRootController *rootController;
+@property(nonatomic,strong) UIViewController *rootController;
 @end
 
 @implementation AppDelegate
@@ -73,7 +73,7 @@
 }
 
 - (void)applicationReloadRootInterface:(OBAApplication*)application {
-    self.rootController = [[OBAClassicApplicationRootController alloc] initWithApplication:application];
+    self.rootController = [OBAApplicationRootControllerFactory makeWithApplication:application];
     self.window.rootViewController = self.rootController;
 }
 
--- a/Apps/OneBusAway/AppDelegate.m
+++ b/Apps/OneBusAway/AppDelegate.m
@@ -15,7 +15,7 @@
 
 @interface AppDelegate ()<OBAApplicationDelegate>
 @property(nonatomic,strong) NSUserDefaults *userDefaults;
-@property(nonatomic,strong) OBAClassicApplicationRootController *rootController;
+@property(nonatomic,strong) UIViewController *rootController;
 @property(nonatomic,strong) OBAAnalyticsOrchestrator *analyticsClient;
 @property(nonatomic,strong) OBACloudPushService *pushService;
 @end
@@ -99,7 +99,7 @@
 
 - (void)applicationReloadRootInterface:(OBAApplication*)application {
     void(^showRootController)(void) = ^{
-        self.rootController = [[OBAClassicApplicationRootController alloc] initWithApplication:application];
+        self.rootController = [OBAApplicationRootControllerFactory makeWithApplication:application];
         self.window.rootViewController = self.rootController;
     };
 
--- a/OBAKit/Alerts/AgencyAlertBulletin.swift
+++ b/OBAKit/Alerts/AgencyAlertBulletin.swift
@@ -49,6 +49,7 @@ class AgencyAlertBulletin: NSObject {
         }
     }
 
+    @MainActor
     func show(in application: UIApplication) {
         bulletinManager.show(in: application, rootItem: alertPage)
     }
--- a/OBAKit/Alerts/AgencyAlertBulletin.swift
+++ b/OBAKit/Alerts/AgencyAlertBulletin.swift
@@ -50,6 +50,6 @@ class AgencyAlertBulletin: NSObject {
     }
 
     func show(in application: UIApplication) {
-        bulletinManager.show(in: application)
+        bulletinManager.show(in: application, rootItem: alertPage)
     }
 }
--- a/OBAKit/BLTNBoard/BulletinOverlayWindow.swift
+++ b/OBAKit/BLTNBoard/BulletinOverlayWindow.swift
@@ -26,6 +26,7 @@ import BLTNBoard
 /// Single-page bulletins only: BLTN fires the dismissal handler on the
 /// `currentItem`, which equals `rootItem` until something is pushed. OBA's
 /// bulletins are all single-page today; revisit if that changes.
+@MainActor
 final class BulletinOverlayWindow {
 
     static let shared = BulletinOverlayWindow()
--- a/OBAKit/BLTNBoard/BulletinOverlayWindow.swift
+++ b/OBAKit/BLTNBoard/BulletinOverlayWindow.swift
@@ -1,0 +1,83 @@
+//
+//  BulletinOverlayWindow.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import UIKit
+import BLTNBoard
+
+/// Owns a dedicated `UIWindow` for hosting `BLTNItemManager` presentations.
+///
+/// Bulletins originally rode `keyWindowFromScene?.topViewController`. That walk
+/// lands inside the SwiftUI map-panel home sheet's hosting controller, which is
+/// constrained by `UISheetPresentationController` to its current detent (~80pt
+/// when collapsed) — so a `present(_:animated:)` from there squeezes the bulletin
+/// into the sheet's height and wedges the floating sheet.
+///
+/// A dedicated window at `.alert` level dodges the sheet hierarchy entirely.
+/// Teardown is wired into BLTN's per-item `dismissalHandler` (its official
+/// dismissal hook), so the window retires exactly when the bulletin finishes
+/// animating out — no timers, no `dismiss(_:animated:)` overrides.
+///
+/// Single-page bulletins only: BLTN fires the dismissal handler on the
+/// `currentItem`, which equals `rootItem` until something is pushed. OBA's
+/// bulletins are all single-page today; revisit if that changes.
+final class BulletinOverlayWindow {
+
+    static let shared = BulletinOverlayWindow()
+
+    private var window: UIWindow?
+
+    private init() {}
+
+    /// Installs the overlay window (if not already up) and returns its host
+    /// controller for `showBulletin(above:)`. Hooks `rootItem.dismissalHandler`
+    /// to tear the window down once the bulletin dismisses, chaining any
+    /// handler the caller already set.
+    func install(in scene: UIWindowScene, rootItem: BLTNItem) -> UIViewController {
+        if let host = window?.rootViewController {
+            return host
+        }
+
+        let window = UIWindow(windowScene: scene)
+        window.windowLevel = .alert
+        window.backgroundColor = .clear
+
+        let host = UIViewController()
+        host.view.backgroundColor = .clear
+        window.rootViewController = host
+        // Non-key intentionally: iOS hit-tests top-down by window level, so the
+        // bulletin's buttons still receive touches without us hijacking key
+        // status from the main app window.
+        window.isHidden = false
+
+        self.window = window
+
+        let originalDismissal = rootItem.dismissalHandler
+        rootItem.dismissalHandler = { [weak self] item in
+            originalDismissal?(item)
+            self?.teardown()
+        }
+
+        return host
+    }
+
+    private func teardown() {
+        let scene = window?.windowScene
+        window?.rootViewController = nil
+        window?.isHidden = true
+        window?.windowScene = nil
+        window = nil
+
+        // iOS doesn't always rebind key status when an alert-level window
+        // hides — force the main app window to reclaim it so the visible UI
+        // resumes receiving touches.
+        scene?.windows
+            .first(where: { !$0.isHidden && $0.windowLevel == .normal })?
+            .makeKey()
+    }
+}
--- a/OBAKit/BLTNBoard/BulletinOverlayWindow.swift
+++ b/OBAKit/BLTNBoard/BulletinOverlayWindow.swift
@@ -35,12 +35,24 @@ final class BulletinOverlayWindow {
 
     private init() {}
 
-    /// Installs the overlay window (if not already up) and returns its host
-    /// controller for `showBulletin(above:)`. Hooks `rootItem.dismissalHandler`
-    /// to tear the window down once the bulletin dismisses, chaining any
-    /// handler the caller already set.
+    /// Installs the overlay window and returns its host controller for
+    /// `showBulletin(above:)`. Hooks `rootItem.dismissalHandler` to tear the
+    /// window down once the bulletin dismisses, chaining any handler the
+    /// caller already set.
+    ///
+    /// One bulletin at a time. Each `BLTNItemManager` already guards on
+    /// `isShowingBulletin`, but those guards are per-manager; this singleton is
+    /// shared across every OBA bulletin manager, so two managers can call
+    /// `install` while a third is mid-presentation. If that ever happens, the
+    /// second caller's `dismissalHandler` chain would never be installed (the
+    /// early-return path used to silently skip it) and teardown would fire on
+    /// the first dismissal, yanking the window out from under the second
+    /// bulletin. The DEBUG assert flags the violation at the point it occurs;
+    /// release builds still return the existing host so the second bulletin at
+    /// worst piggybacks on the first's window rather than crashing.
     func install(in scene: UIWindowScene, rootItem: BLTNItem) -> UIViewController {
         if let host = window?.rootViewController {
+            assertionFailure("BulletinOverlayWindow already in use — concurrent bulletin presentations aren't supported (shared singleton, single-bulletin-at-a-time).")
             return host
         }
 
--- a/OBAKit/BLTNBoard/BulletinOverlayWindow.swift
+++ b/OBAKit/BLTNBoard/BulletinOverlayWindow.swift
@@ -51,6 +51,15 @@ final class BulletinOverlayWindow {
     /// release builds still return the existing host so the second bulletin at
     /// worst piggybacks on the first's window rather than crashing.
     func install(in scene: UIWindowScene, rootItem: BLTNItem) -> UIViewController {
+        // Single-page only: teardown rides `rootItem.dismissalHandler`, but
+        // BLTN fires the dismissal handler on `currentItem` — which only
+        // equals `rootItem` until something gets pushed. A multi-page item
+        // would tear the window down on the first page's dismissal rather
+        // than the flow's final dismissal. Catch the violation at install
+        // time so a future contributor sees the contract instead of debugging
+        // a teardown-mid-flow weirdness in the wild.
+        assert(rootItem.next == nil, "BulletinOverlayWindow only supports single-page bulletins — multi-page flows would tear down on the first page's dismissal. See type docstring.")
+
         if let host = window?.rootViewController {
             assertionFailure("BulletinOverlayWindow already in use — concurrent bulletin presentations aren't supported (shared singleton, single-bulletin-at-a-time).")
             return host
--- a/OBAKit/BLTNBoard/OBABulletinPage.swift
+++ b/OBAKit/BLTNBoard/OBABulletinPage.swift
@@ -25,19 +25,30 @@ class ThemedBulletinPage: BLTNPageItem {
 }
 
 extension BLTNItemManager {
-    /// Presents the bulletin above the topmost view controller of the application's key window.
+    /// Presents the bulletin in a dedicated overlay `UIWindow` attached to the
+    /// active scene.
     ///
-    /// Use this instead of `showBulletin(in:)`: that method presents inside a `UIWindow`
-    /// it creates without a `windowScene`, and iOS never displays such a window in a
-    /// scene-based app, so the bulletin silently fails to appear.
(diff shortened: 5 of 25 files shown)


aaronbrethorst (changes requested):
Hi Mohamed — this is a genuinely impressive piece of foundation work. The two-layer sheet coordinator, the compiler-enforced exhaustive view factory, and the care you took to gate the whole thing behind a restart-required experimental flag all add up to a clean, low-risk scaffold for the map-panel experience. The doc comments are some of the best I've read in this codebase — the `BulletinOverlayWindow` header that traces the "rode the key window's top VC → landed in the sheet's hosting controller → got clamped to 80pt" bug is exactly the kind of *why* comment that saves the next person an afternoon. Build is clean and all 32 new tests pass locally (iPhone 17 Pro, iOS 27).

A heads-up before the findings: the CodeRabbit review on this PR was run against an older commit (`76ed702d`) with a different architecture (branching inside `ClassicApplicationRootController`). You've since refactored to the `ApplicationRootControllerFactory` seam, so most of its inline comments are stale and I've disregarded them. 

My one architectural caveat is that the bulletin rework is **not** gated behind the flag — it changes bulletin presentation for *every* user, classic mode included. That raises the bar on getting the overlay-window lifecycle right, and it's where most of my "important" notes land.

---

## Critical Issues (0 found)

None. Nothing here is a confirmed crash or data-loss path in the shipping (classic) experience.

## Important Issues (3 found)

- **The overlay-window reuse path drops the new bulletin's teardown hook.** [OBAKit/BLTNBoard/BulletinOverlayWindow.swift:43-45] `install(in:rootItem:)` early-returns the existing host when a window is already up, never reaching the `dismissalHandler` chaining at lines 61-65. `BulletinOverlayWindow.shared` is a singleton shared across four managers (`ErrorBulletin`, `AgencyAlertBulletin`, `ReachabilityBulletin`, `RegionMismatchBulletin`), each with its *own* `isShowingBulletin` guard — so two of them can overlap (e.g. a connectivity drop while an error bulletin is up). When that happens, the second bulletin's lifecycle is tied to the first's teardown: the first dismissing nils the window out from under the second. This is all in the non-gated path, so it affects classic-mode users. Either chain the new `dismissalHandler` on the reuse path, or add a DEBUG `assert(window == nil)` and document the single-bulletin-at-a-time assumption explicitly. **Relatedly:** please confirm you manually verified classic-mode bulletins (error, region mismatch, connectivity) still present correctly in the new `.alert`-level window — the PR video shows map-panel mode, but this path now governs the default experience too.

- **The cross-layer invariant is held together by convention, not the type.** [OBAKit/Sheet/Coordinator/SheetRoute.swift:120 vs :146] A `prefersStacking == true` route must have `isDismissDisabled == false`, otherwise the OS owns drag-down dismissal but `truncateStacked` never hears about it and `stackedEntries` points at a sheet that's no longer on screen. Today every case is consistent and your tests assert it ([AppSheetRouteTests.swift:93-104, 114-132]) — but `prefersStacking` and `detentConfiguration` are two independent switches 25 lines apart, and the tests only guard the cases that exist *now*, not the next one someone adds. A `precondition(!prefersStacking || !config.isDismissDisabled)` (or deriving one fact from the other) would convert this from a documented convention into a guarantee, and it's much cheaper to add now than after more routes land.

- **A swallowed bulletin leaves no trace when no scene is found.** [OBAKit/BLTNBoard/OBABulletinPage.swift:44-46] The guard returns silently in two cases. `isShowingBulletin` is a fine silent no-op; the **no-`bulletinTargetScene`** case is not — it drops the bulletin entirely with nothing logged. That matters most for the error bulletin: `displayError` logs the error and then calls `show(in:rootItem:)` ([Application.swift:622-625]), so if the scene lookup comes up empty the user never sees the error UI and there's zero diagnostic evidence. Split the guard arms and `Logger.error` (ideally report) on the no-scene path.

## Suggestions (7 found)

- **`setStackedDetent` is the one untested coordinator method.** [OBAKit/Sheet/Coordinator/SheetCoordinator.swift:92-95] Every sibling (`push`/`pop`/`truncateStacked`/`popToRoot`/`canPop`) is thoroughly covered, and `truncateStacked`'s identical bounds-guard *is* tested ([SheetCoordinatorTests.swift:142]) — so this omission is conspicuous rather than principled. Two lines close it: set a detent at depth 0 and assert it persists; call with an out-of-range depth and assert no change.

- **Dropped root navigation degrades silently in map-panel mode.** [OBAKit/ViewRouting/Router.swift:107-116] Logging instead of crashing is the right instinct, but `Logger.error` isn't user-visible, and the two real call sites — the "Find Stops on Maps" empty-state button ([RecentStopsViewController.swift:172]) and the `addRegion` deep link ([Application.swift:515], which still runs its region-add side effects) — become inert / half-done. Consider a DEBUG `assertionFailure` here to match the pattern you already use in `AppSheetViewFactory.unimplementedView`, so the gap trips during development rather than in the field.

- **Release-build `EmptyView()` fallback is a silent blank sheet.** [OBAKit/Sheet/DI/AppSheetViewFactory.swift:70-72] The DEBUG side (visible label + `assertionFailure`) is great; the release side renders nothing with no log. It's unreachable today (your TODOs correctly flag `.search` as a trap), but it's rot-prone — the next person to wire up a push won't see the assertion. A `Logger.error` before the fallback would leave a breadcrumb.

- **Search affordance loses accessibility action semantics.** [OBAKit/Sheet/Content/Home/HomeSheetView.swift:53] `.onTapGesture` on a non-button container doesn't expose action semantics to VoiceOver or Switch Control. Wrapping the row in a `Button(action: expandSheet)` with `.buttonStyle(.plain)` keeps the exact appearance while restoring proper semantics.

- **`SheetDetentConfiguration` has no construction-time validation.** [OBAKit/Sheet/Coordinator/SheetRoute.swift:30-44] Nothing prevents `initialDetent` (or `fullScreenDetent`) from sitting outside `detents`, which would be dead config — `fullScreenDetent` is matched by `==` against the current selection, so an out-of-set value never fires. A `precondition(detents.contains(initialDetent))` would catch authoring slips at the point of definition.

- **Two small simplifications in the sheet plumbing.** The bounds-checked `stackedRoutes[depth]` / `stackedEntries[depth].detent` indexing in `routeBinding`/`detentBinding` ([FloatingSheetContainer.swift:59-81]) duplicates logic the coordinator already owns — small `stackedRoute(at:)` / `stackedDetent(at:fallback:)` accessors would centralize it alongside `setStackedDetent`/`truncateStacked`. And the reachability sink ([Application.swift:261-274]) reads cleaner with an early-return on the `isConnected` branch (flattening the disconnected path) plus the house-style `} else {` rather than `else` on its own line.

- **Tighten the `assumeIsolated` comment.** [OBAKit/Orchestration/Application.swift:259] "guarantees main" is operationally correct but conflates thread with actor isolation — what makes `assumeIsolated` safe is the main *thread* delivery from `.receive(on: DispatchQueue.main)` (line 255). Since removing that upstream would silently turn this into a crash-on-mismatch trap, it's worth naming the real invariant: "...don't remove the `.receive(on:)` upstream."

## Strengths

- The coordinator's `root` + `additionalRoutes` split makes "always at least one route" *unrepresentable otherwise* — `currentRoute` never force-unwraps. That's the strongest design decision in the PR.
- Compiler-enforced exhaustive dispatch in `AppSheetViewFactory`, backed by the `exhaustivenessGuard` test that turns "did you remember to cover the new route?" into a compile error.
- `SheetCoordinatorTests` is exemplary behavioral testing — it nails the subtle invariants (pushing a stacked route must not touch `currentDetents`; pop-at-root is a detent-preserving no-op; OS-driven `truncateStacked` removes everything at-and-above a depth).
- The `@objc` `ApplicationRootControllerFactory` cleanly removes feature-flag knowledge from the Obj-C AppDelegates, keeping each concrete root controller single-purpose.
- Doc-comment quality throughout, especially the `let _` vs `_ =` `@ViewBuilder` note and the `.fraction(0.99)`-vs-`.large` detent-identity rationale — both document non-obvious traps that a well-meaning "cleanup" would reintroduce.

## Recommended Action

1. Harden the `BulletinOverlayWindow` reuse path and confirm classic-mode bulletins were verified (Important #1).
2. Enforce the stacking ⇒ `!isDismissDisabled` invariant with a `precondition` (Important #2).
3. Log the no-scene bulletin drop (Important #3).
4. Add the two-line `setStackedDetent` test and consider the DEBUG assertions for the dropped-navigation / unimplemented-route paths.
5. Pick up the accessibility and fit-and-finish suggestions as you like.
6. Re-run the suite and I'll take another look.

**Verdict: Request Changes** — the three Important items are small and well within reach, but two of them touch the default (non-gated) bulletin experience and the third locks in a foundational design contract while it's still cheap to enforce. None of this detracts from what is otherwise a really well-built foundation.

aaronbrethorst (approved):
Hi Mohamed — this is a genuinely impressive piece of foundation work. The two-layer sheet coordinator is a clean abstraction, the `@MainActor` isolation is handled correctly throughout, and the comment density is some of the best I've seen in this codebase — most of it explains *why* and encodes real, hard-won SwiftUI/UIKit/BLTNBoard knowledge. The `SheetCoordinator`/`AppSheetRoute` core is thoroughly unit-tested (32 tests, all passing locally), and gating the whole experience behind a default-off flag is exactly the right rollout posture.

There's one thing I want handled before this graduates, but not something that needs to block the merge: one piece of this PR — the bulletin presentation overhaul — ships to **production** users (not just experimental-flag testers) and carries a couple of real defects in a path transit riders hit regularly. I've filed **[#1170](https://github.com/OneBusAway/onebusaway-ios/issues/1170)** to track it as a fast-follow. Since the bulletin change is the only un-gated part and everything else is behind a default-off flag, I'm comfortable merging now and fixing the bulletin path in a follow-up rather than holding the foundation hostage to it. Details below so the fast-follow has full context.

Verified locally: `build-for-testing` succeeded (iPhone 17), and `AppSheetRouteTests` + `SheetCoordinatorTests` pass (32 tests, 0 failures). I fact-checked the load-bearing SwiftUI/concurrency claims against Apple's documentation; results noted inline.

---

## Critical Issues (0 found)

No crashes or data-loss issues on the happy path.

---

## Important Issues (4 found)

> **Items 1–3 are tracked as a fast-follow in [#1170](https://github.com/OneBusAway/onebusaway-ios/issues/1170)** and do not block this merge. They're documented here in full so the follow-up has context. Item 4 is a small config fix worth folding in now if it's quick.

**1. The bulletin presentation overhaul is NOT behind the feature flag — it ships to classic (production) users.** _(tracked: #1170)_
[OBAKit/BLTNBoard/OBABulletinPage.swift:38] + the four call sites (`AgencyAlertBulletin.swift`, `ErrorBulletin.swift`, `Reachability.swift`, `RegionMismatchBulletin.swift`)

Every bulletin now routes through the new `BulletinOverlayWindow` regardless of `useMapPanelExperienceKey`. The problem it solves — `UISheetPresentationController` clamping the bulletin to the home sheet's detent — only exists in map-panel mode, but the fix is global. That means any regression in the new window/scene-targeting logic hits *all* production users, not just experimental testers. The PR is described as "behind a feature flag," but this foundational change is un-gated. Either confirm the widened blast radius is intentional, or scope the new window to map-panel mode only (keep the old `keyWindowFromScene?.topViewController` path for classic mode). The two defects below are why this matters.

**2. `dismissalHandler` chaining accumulates across re-presentations of a reused page item.** _(tracked: #1170)_
[OBAKit/BLTNBoard/BulletinOverlayWindow.swift:82] + [OBAKit/Reachability/Reachability.swift:86]

`install` does `originalDismissal = rootItem.dismissalHandler; rootItem.dismissalHandler = { originalDismissal?(item); self?.teardown() }`. For a bulletin whose page is a long-lived stored property re-shown repeatedly — `ReachabilityBulletin.connectivityPage` is the clear case, since connectivity flapping in tunnels/subways re-calls `show(in:rootItem:)` — BLTN never resets `dismissalHandler` between presentations, so each cycle wraps the previous closure. Over N flaps the chain nests N deep and `teardown()` fires N times. It's idempotent and bounded by flap count, so not a crash — but it's an unbounded-in-principle closure accumulation on a shared singleton, in production, in exactly the scenario a transit app sees most. Snapshot the original handler once (or reset it) rather than chaining onto the persisted item.

**3. Concurrent bulletin managers can orphan a live bulletin in release builds.** _(tracked: #1170)_
[OBAKit/BLTNBoard/BulletinOverlayWindow.swift:63]

The `isShowingBulletin` guard is per-`BLTNItemManager`, but `BulletinOverlayWindow.shared` is one window across all managers. Realistic trigger: you go offline → `ReachabilityBulletin` shows → a network error fires → `Application.displayError` → `ErrorBulletin` (a *different* manager, so its `isShowingBulletin` is false) → `install` hits the in-use branch. In DEBUG that's an `assertionFailure` (good — it catches the case in dev). In release, it returns the *first* bulletin's host and tells the error to present above it; when the reachability bulletin later dismisses, `teardown()` nils the window out from under the still-live error bulletin. A genuine error the rider needs to see can be visually corrupted or stranded, and the stale `.alert`-level window may linger. Given offline + network-error commonly co-occur, this isn't exotic.

**4. `routePicker` is a base-layer route but is configured to allow interactive dismissal.**
[OBAKit/Sheet/Coordinator/SheetRoute.swift:133] + [SheetRoute.swift:194-199]

`.routePicker` returns `prefersStacking == false` (base layer), yet its detent config falls into the shared group with `isDismissDisabled: false`. The base sheet is presented with `.sheet(isPresented: .constant(true))` ([FloatingSheetContainer.swift:26]), so a dismissable base sheet is contradictory — the binding can never go false. `.home` and `.search` correctly set `isDismissDisabled: true`; `routePicker` is the lone outlier. Note the asymmetry in `push` ([SheetCoordinator.swift:69]): there's a `precondition` that *stacked* routes must be dismissable, but nothing enforces the converse for base routes. Not user-reachable today (it resolves to the placeholder and nothing pushes it), so it's latent — but it's wrong config, and a converse precondition would have caught it. Cheap to fix now.

---

## Suggestions (6 found)

**5. The `// Base layer` / `// Stacked layer` comments are already inaccurate — don't trust them as documentation.**
[OBAKit/Sheet/Coordinator/SheetRoute.swift:72,79]

`nearbyAll`, `recentStopsAll`, `bookmarksAll` are grouped under `// Base layer`, but `prefersStacking` returns **true** (stacked) for all three ([:131]). `routePicker` sits under `// Stacked layer` but returns **false** (base) ([:133]). So two groups of routes are mislabeled relative to their actual behavior. This is the drift risk made concrete: layer membership currently lives in three places that must agree (the comments, `prefersStacking`, and `detentConfiguration`). Consider deriving `prefersStacking` and the dismissal rule from a single `SheetLayer` discriminator so they can't disagree — that would also make #4 unrepresentable rather than asserted, and would have caught these mis-groupings at compile time.

**6. `presentationCompactAdaptation(.none)` rationale is inverted.**
[OBAKit/Sheet/Coordinator/FloatingSheetContainer.swift:119-122]

The comment says this acts "On iPad / regular size class … instead of adapting to a popover." Per Apple's docs, `presentationCompactAdaptation(_:)` affects **compact** size classes (e.g. landscape iPhone), not iPad/regular — and sheets adapt to a full-screen-cover, not a popover (it's popovers that adapt *into* sheets). The `.none` call itself is fine; just the rationale is wrong, and would mislead anyone reasoning about iPad layout from it. Re-anchor it to "keep the sheet presentation in compact size classes."

**7. The `.large` "different detent identity that `==` can't catch" claim is unsupported.**
[OBAKit/Sheet/Coordinator/SheetRoute.swift:144-146]

The visual rationale for `.fraction(0.99)` (keeps a sliver of map → floating-card look) is legitimate and worth keeping. But `.large` is a canonical `Equatable`/`Hashable` value — `selection == .large` matches fine; if anything the historical SwiftUI wrinkle runs the other way (custom `.fraction`/`.height` detents are the unreliable ones). Drop or empirically re-verify the identity assertion, and pin it to a tested iOS version if you keep it. Same note for the repeated `upThrough:`-with-custom-`.height` claim ([:26-27], [:158-161]) — plausible and internally consistent with the `fullScreenDetent` workaround, but undated; pin it to an iOS version so it can be re-validated.

**8. `AppSheetRoute.id` reinvents what the compiler gives you for free.**
[OBAKit/Sheet/Coordinator/SheetRoute.swift:95-124]

The enum is already `Hashable` with `Hashable` associated values, so `var id: Self { self }` would be collision-proof and eliminate the 30-line hand-maintained switch that has to stay in sync with every new case (and risks silent string collisions). The human-readable strings are genuinely useful for the factory's logging/placeholder copy — but that's a `description` concern, so consider splitting identity (`Self`) from a display string.

**9. The on/off switch itself is untested, and it's cheap to cover.**
[OBAKit/Orchestration/ApplicationRootControllerFactory.swift:24-30]

`make(application:)` is the single branch that decides the entire experience, and it's the most testable untested piece in the PR: set/clear the flag in `userDefaults`, call `make`, assert the returned type. `ApplicationTests` already constructs an `Application`, so the harness cost is near-zero. A regression here silently strands users on the wrong root.

**10. New "we logged it" paths use `Logger.error`, which is console-only.**
[OBAKit/ViewRouting/Router.swift:108], [OBABulletinPage.swift:46], [AppSheetViewFactory.swift:79]

`Logger.error` writes to the unified log, not Sentry/analytics — so dropped deep links (map-panel mode), dropped bulletins (no foreground scene), and stray-route placeholders are invisible in aggregate. For an experimental surface you specifically want field telemetry on, consider routing these through `analytics?.reportError`. The `Router.rootNavigateTo` drop is acknowledged with a `TODO(mosliem)` and is benign today (current callers target `.map`, ~a no-op in map-panel mode), so it's fine to defer — just wire it to something visible before the flag graduates. The three TODOs in this PR ([AppSheetViewFactory.swift:42], [HomeSheetViewModel.swift:21], [HomeSheetView.swift:34]) would also benefit from issue links so they don't become permanent.

---

## Strengths

- **`SheetCoordinator` is well-encapsulated**: `@Published private(set)` state, the `root`/`additionalRoutes` split that avoids force-unwraps, bounds-checked accessors that correctly absorb SwiftUI's transient out-of-bounds reads during sheet teardown. The `truncateStacked`/`setStackedDetent` invariants match their comments and tests.
- **The tests assert behavior, not internals**, and the exhaustiveness guard ([AppSheetRouteTests.swift:186]) forces future cases to be covered. The two-layer split (configs pinned to literals; coordinator verified to read configs) is sound.
- **Concurrency is correct.** I verified `MainActor.assumeIsolated` ([Application.swift:263]) is genuinely safe — it's guarded by `.receive(on: DispatchQueue.main)` upstream ([:255]), and the inline comment correctly warns against removing it. All `show(in:rootItem:)` call sites reach it from `@MainActor` contexts.
- **The `unimplementedView` placeholder is the right shape** ([AppSheetViewFactory.swift:60]): DEBUG assertion to catch stray pushes, release log + a *visible*, localized "coming soon" card surfacing `route.id` for tester reports — far better than a blank `EmptyView()`.
- **The `let _` vs `_ =` `@ViewBuilder` explanation and the VoiceOver/`HStack` rationale are both technically accurate** — nice attention to detail.
- The ObjC factory seam and `UIViewController*` widening in both AppDelegates line up correctly.

---

## Recommended Action

**Verdict: Merge.** The sheet system is excellent and well-gated, and the one un-gated concern — the bulletin overhaul — is captured as a fast-follow in [#1170](https://github.com/OneBusAway/onebusaway-ios/issues/1170). Merging now and fixing the bulletin path in a follow-up is the right call rather than holding the foundation hostage to it.

Follow-ups, none blocking this merge:
1. The bulletin work — blast radius (#1) and the two defects (#2, #3) — is tracked in #1170.
2. `routePicker`'s config (#4) — set `isDismissDisabled: true`, since it's a base-layer route. Optionally add the converse precondition or a `SheetLayer` discriminator (#5) so this class of mismatch can't recur.
3. The inaccurate comments (#5, #6, #7), and the test (#9) and telemetry (#10) suggestions, whenever you're next in this code.

Really nice work on this, Mohamed — the foundation is solid and a pleasure to read.

PR #1166, opened 2026-06-23
Weather overlay - Map Floating Panel

mosliem:
### Summary 

Adds the first overlay built on the floating sheet system from the base branch: a glassy weather card that replaces `MapViewController`'s `UIAlertController` and establishes the corner-button + popup pattern future overlays will follow.

#### Weather UI (SwiftUI, on `MapPanelRootView`)

- **`WeatherButton`** — floating temperature pill in the top-leading overlay slot.
- **`WeatherDetailPopup`** — tap-to-reveal card:
  - Header: icon, region, condition, chance of rain, current temp, 24-hour Hi/Lo.
  - Horizontally scrollable 24-hour strip; "Now" highlighted; past hours filtered.
  - Stats row: wind, precipitation %, feels-like — each VoiceOver-labelled.
  - Floating `xmark` button; backdrop tap also dismisses. Card body is non-tappable so the hourly strip stays scrollable.
  - Liquid Glass surface on iOS 26+; solid themed fallback pre-26 — no half-glass middle ground.

#### Shared formatting & view model

- **`WeatherFormatter`** (`OBAKitCore`, pure Foundation) — single source of truth for icon → SF Symbol, condition text, locale-aware temp (°F/°C) and wind (mph/kmh), hi/lo.
- **`WeatherDisplay`** (`OBAKit` ViewModels) — typed view-ready model with nested `Header` / `Stats` / `LegacyAlert`. Both the UIKit `MapViewController` button and the new SwiftUI card read from it, so they can never drift.
- **`MapViewModel.weatherDisplay`** — `@Published` and rebuilt once per `loadWeather()` (no per-body-read formatting cost). Raw `WeatherForecast` is no longer published — was unused outside tests.

#### iOS 26 / pre-26 glass helpers

- **`regularGlassEffectIfAvailable` / `clearGlassEffectIfAvailable`** view modifiers — Liquid Glass on iOS 26 with `.regularMaterial` / `.secondarySystemBackground` fallbacks. Self-contained surface fills (call sites don't add a background).

#### UIKit consumer migration

- `MapViewController.showWeather()` / `bindWeather()` now read through `WeatherDisplay.legacyAlert` and `$weatherDisplay`. ~30 lines of inline formatting removed; behaviour unchanged.

### Architecture notes

- `HourlyEntry.id` is the hour's `Date`, not an array index — keeps `ForEach`/`LazyHStack` cells stable across forecast refreshes.
- `WeatherDetailPopup` lives in `MapPanelRootView`'s overlay layer, not a UIKit modal — coexists with the persistent floating sheet that would otherwise fight a `.alert(...)` presenter.
- `WeatherFormatter` is fully `nonisolated`; `formatTime` allocates per call rather than mutating a shared `DateFormatter` (no actor isolation, no data race).

### Tests
  - WeatherFormatterTests — covers icon mapping, condition grouping, locale-aware temp/wind, high/low edge cases.
  - WeatherDisplayTests — covers past-hour filter, "Now" labelling, 24-entry cap, Date-based identity, empty inputs.
  - MapViewModelTests — covers the published `weatherDisplay` after the raw model was dropped.
  
  
### Notes
**This PR should be merged after #1163** 

#### iOS 26 Demo
  
https://github.com/user-attachments/assets/51d41cc2-6f39-407e-bcac-45ee03cbc2d4

#### iOS 18 Demo 

https://github.com/user-attachments/assets/4d0517f6-f002-4a05-8fb6-4a908eb5a2cf

## Summary by CodeRabbit

* **New Features**
  * Added a route-based floating sheet experience with support for stacked layered sheets.
  * Introduced map-panel weather UI: weather button and a dismissible weather detail popup.
  * Added an Experimental settings toggle to enable the map panel experience.
  * Added optional glass-styled SwiftUI helpers for supported system versions.
* **Bug Fixes**
  * Improved weather loading behavior to avoid stale data and prevent UI flicker on transient failures.
  * Weather UI now cleanly reflects availability and updates from the prepared display model.
* **Tests**
  * Expanded coverage for sheet routing and weather formatting/display behavior.


Vedd-Patel:
@mosliem hey, i was actually working on this earlier(#1073 ) and i thought my PR had already been merged, but after seeing that you opened the PR, i went through the open PRs and realized mine is still open
@aaronbrethorst should i close it from my side?
also, apologies for my unavailability and for not keeping proper track of my own PR

aaronbrethorst (changes requested):
Hi Mohamed — this is a great PR. The `WeatherDisplay`/`WeatherFormatter` split is the kind of thing I wish more of our codebase looked like: one source of truth that both the UIKit button and the SwiftUI card read from, so the two surfaces *can't* drift. The `SheetCoordinator` two-layer model is clean, the `AppSheetRoute` exhaustiveness compile-guard in the tests is a lovely touch, and you kept `OBAKitCore` Foundation-only so extension safety holds. The diff is big but it reads like it was designed, not accreted.

There are two real defects I want fixed before this merges, both of which hide well (one behind `#if DEBUG`, one behind an unused local that Swift won't warn about). Everything below the line is polish.

## Critical Issues (1 found)

- **`fullScreenDetent` background-interaction override is dead code** [OBAKit/Sheet/Coordinator/FloatingSheetContainer.swift:111-122]. `applyDetentConfig` computes `effectiveBackgroundInteraction` — correctly forcing `.disabled` when the sheet is parked at `fullScreenDetent` — and then never uses it. Line 122 passes `config.backgroundInteraction` instead:

  ```swift
  .presentationBackgroundInteraction(config.backgroundInteraction)   // should be effectiveBackgroundInteraction
  ```

  The entire `fullScreenDetent` field on `SheetDetentConfiguration`, the detailed rationale in `SheetRoute.swift:22-28` and `:143-147`, and the `test_homeDetent_flipsToFullScreenAtLargeDetent` test all describe behavior that never executes. For `.home` parked at `largeDetent` (`.fraction(0.99)`), the map behind the full-coverage sheet stays interactive instead of going inert — exactly the iPhone-landscape case the comments call out. One-line fix; please also add a test that actually exercises the override (see below), since that gap is why this slipped through.

## Important Issues (1 found)

- **Release builds render a blank, non-dismissable sheet for `.search`** [OBAKit/Sheet/DI/AppSheetViewFactory.swift:55-65]. `.search` is reachable in production — `HomeSheetView.expandThenPushSearch()` (HomeSheetView.swift:71) calls `coordinator.push(.search)` on a real search-bar tap, and the experimental toggle ships in release builds (SettingsViewController.swift:178-181). In release, `unimplementedView` falls to `EmptyView()`, and `.search`'s detent config (SheetRoute.swift:155-163) is `detents: [.large]`, `isDismissDisabled: true`, `backgroundInteraction: .disabled`. So a user who enables the experiment and taps search gets a full-screen, non-dismissable, blank sheet over a disabled map — a dead-end requiring an app restart. The `#if DEBUG` assertion masks this entirely during development and logs nothing in release. Until `.search` has a real view, either don't `push(.search)`, or make the release fallback a visible, dismissable placeholder that calls `Logger.error(...)`.

## Suggestions (fit-and-finish)

**Lint (both newly introduced by this PR):**
- `redundant_discardable_let` — `let _ = assertionFailure(...)` should be `_ = assertionFailure(...)` [OBAKit/Sheet/DI/AppSheetViewFactory.swift:57].
- `cyclomatic_complexity` — `saveFormValues` is now complexity 11 (was 10 on `main`; the `useMapPanel` block tipped it) [OBAKit/Settings/SettingsViewController.swift:79]. A tiny `saveExperimentalValues(_:)` helper clears it.

**Correctness / behavior:**
- `loadWeather()`'s `guard let apiService = application.obacoService else { return }` leaves `weatherDisplay` stale, while the `catch` branch nils it [OBAKit/ViewModels/MapViewModel/MapViewModel.swift:92-99]. If the service goes away after a successful fetch (region change), the old forecast keeps rendering. Set `weatherDisplay = nil` in the guard's `else` so both unavailability paths agree.
- `HourlyEntry.list` selects "Now" via `upcoming.first?.time`, trusting the API to deliver hourly entries pre-sorted [OBAKit/ViewModels/MapViewModel/WeatherDisplay.swift:142-145]. Out-of-order input would silently mislabel the "Now" cell. A `.sorted(by: \.time)` before `prefix(24)` would make it robust; at minimum worth a comment documenting the precondition.

**Comment accuracy:**
- `formatTime` docstring says "Hour-only, localized: e.g. `9 AM`", but `setLocalizedDateFormatFromTemplate("Hmm")` requests hour **and minutes** and `H` is the 24-hour field — actual output is like `9:00 AM` [OBAKitCore/Weather/WeatherFormatter.swift:87-96]. Either switch to `"j"`/`"H"` for hour-only or fix the comment. (Minutes in the narrow hourly cells may also be a layout concern — worth a look in the simulator.)
- `MapViewModel`'s header comment references a `NewMapView` that doesn't exist and the unrelated `MapPanelViewModel`/`PanelDetent` pattern [OBAKit/ViewModels/MapViewModel/MapViewModel.swift:17, :25-26]. The actual SwiftUI consumer is `MapPanelRootView`.
- `WeatherDisplay.Header` is documented as "today's hi/lo" but `WeatherFormatter.highLow` computes over a rolling next-24-hours window, not the calendar day [WeatherDisplay.swift:25-26, WeatherFormatter.swift:100-108].
- `clearGlassEffectIfAvailable`'s shared docstring "call sites do not need to add a background" is misleading for the clear variant — every call site (WeatherDetailPopup.swift:64-65, :97-98) layers its own `.ultraThinMaterial` underneath, and on iOS 26 the clear glass needs it [OBAKit/Extensions/SwiftUIExtensions.swift:39-41]. Scope that promise to the `regular` variant.

**Test gaps** (current suite is strong — these are the holes):
- No test for `WeatherFormatter.formatTime` at all — directly related to the docstring drift above.
- The full `WeatherDisplay.init` (buttonTitle, `Header.chanceOfRainText`, populated `Header.highLowText`, `Stats`, and the entire `LegacyAlert` message) is only indirectly touched via one `regionName` assertion. Given `LegacyAlert` is documented to stay "in lockstep" with the card, assert it from a fixture.
- The `loadWeather` no-`obacoService` guard and `isWeatherFeatureAvailable` are untested.
- `highLow`'s 24-entry cap is unverified (the `HourlyEntry.list` cap is well-covered — nice asymmetry to close).
- A test for the `fullScreenDetent` → `.disabled` override (would have caught the critical bug).

**Type design (optional, low-cost):**
- `SheetDetentConfiguration.init` lets you construct `initialDetent`/`fullScreenDetent` outside `detents` [SheetRoute.swift:30]. An `assert(detents.contains(initialDetent))` makes that invariant real.
- `AppSheetRoute.id` is compiler-safe against a *missing* case but not against a *duplicate/typo'd* string [SheetRoute.swift:86-115]. Deriving the case-name prefix mechanically (`String(describing:)`) and hand-writing only the associated-value suffix removes that risk.

## Strengths

- `WeatherDisplay` + `WeatherFormatter`: single derive-from-source initializer, all `let`, both surfaces pre-rendered from shared data. The "two UIs can't drift" goal is structurally enforced, not just hoped for.
- `SheetCoordinator`'s non-optional `root` eliminates the empty-stack force-unwrap; the content-swap vs. stacked split is well-modeled and well-commented.
- `AppSheetRouteTests` exhaustiveness compile-guard forces future cases to extend the tests.
- `MapViewModelTests.test_loadWeather_errorClearsDisplay` loads successfully *first* to prove the error path clears, and documents the atomic mock-swap to dodge the `MockDataLoader` race — careful work.
- The `MapViewController` migration onto `WeatherDisplay.legacyAlert` is a faithful, behavior-preserving swap that deletes ~30 lines of inline formatting.
- `ClassicApplicationRootController`'s transitional optional-tab shape is well-guarded and carries a clear TODO for the eventual split.

## Recommended Action

1. Fix the critical dead-code bug at FloatingSheetContainer.swift:122 (and add the override test).
2. Resolve the release-build `.search` dead-end — block the push or ship a dismissable fallback.
3. Clear the two new SwiftLint warnings.
4. Pick up the comment/correctness/test-gap items as you see fit (the stale-state and "Now"-sort ones are the most substantive).
5. Re-request review after the critical + important fixes.

---

**Verdict: Request Changes.** The architecture is sound and I'd happily merge it once the two defects above are addressed — they're both real, both reachable in shipped builds (the experimental flag ships), and both currently invisible to the compiler and the test suite.

aaronbrethorst (changes requested):
Mohamed, this is a really clean piece of work. The single decision at the heart of the PR — routing every weather string through one immutable `WeatherDisplay` built from `(WeatherForecast, Locale)` so the UIKit alert and the SwiftUI card *cannot* drift — is exactly the right invariant, and you backed it up with a fixture-driven test that pins the `LegacyAlert` copy end-to-end. The OBAKitCore split (`WeatherFormatter` as pure Foundation), the per-call `DateFormatter` to keep the helpers `nonisolated` and race-free, and the `[weak self]` discipline in every `Task` all hold up under inspection. `TEST BUILD SUCCEEDED` and the 25 new tests pass locally.

A couple of real items keep this from a clean merge, both small and concrete.

## Critical Issues (0 found)

None.

## Important Issues (3 found)

- **The branch has merge conflicts and won't merge as-is.** GitHub reports it as `CONFLICTING` / `DIRTY` against its base. Rebase on the latest base branch and resolve the conflicts so the PR is mergeable and CI exercises the real merged state. (Worth doing before the fixes below so you're not resolving the same conflicts twice.)

- **`highLow` uses a different window than the hourly strip — a stale past hour can skew the Hi/Lo.** `WeatherFormatter.highLow` takes `hourly.prefix(24)` of the *raw* array [`OBAKitCore/Weather/WeatherFormatter.swift:105-110`], and the caller passes the unfiltered `forecast.hourlyForecasts` [`WeatherDisplay.swift:81`]. You explicitly documented that "the Obaco API includes the previous full hour in `hourly_forecast` for context" and filtered + sorted it out in `HourlyEntry.list` (`$0.time >= currentHourStart`, then sort ascending) [`WeatherDisplay.swift:142-147`]. `highLow` does neither, so the min/max is computed over `[previous full hour … +23h]` instead of the documented "rolling next-24-hour" window [`WeatherDisplay.swift:25-27`], and is order-dependent on however the API returns the array. Real-world impact is small (one extra/missing sample at the edge), but it's the same quirk you guarded against in one place and missed in the other. Factor the filter+sort into a shared helper so both paths see the same window.

- **New `weather.*` localization keys were never extracted into the in-repo `.strings` files.** The PR adds a batch of new keys — `weather.condition.*`, `weather.chance_of_rain_format`, `weather.now`, `weather.stat.*`, `weather.high_low_format`, `weather.alert.*_line_format` — but no `.strings` file is touched in the diff, and none of these keys exist in `OBAKit/Strings/en.lproj/Localizable.strings` (or any of the other 12 locales). Because `OBALoc(..., value:)` falls back to the inline English, nothing crashes — but every non-English locale ships English weather copy with no entry to translate. This codebase keeps localizations in-repo; please run `scripts/extract_strings` and commit the resulting `.strings` deltas. (`map_controller.show_weather_button`, which `WeatherButton` reuses, already exists — that one's fine.)

## Suggestions (found)

- **Unknown icon keys degrade silently with no breadcrumb.** `systemImageName(for:)` falls back to `"cloud.fill"` and `conditionText(for:)` to `"—"` [`WeatherFormatter.swift:22, 61`]. The visual fallback is the correct UX — the card should never blank — but an unmapped key is the *only* signal you'll ever get that Obaco added a new condition (`thunderstorm`, `hail`, …), and right now it's dropped everywhere. Worse, the two fallbacks disagree (a legit-looking cloud icon paired with "—" text). Consider a single `Logger.warn` at `WeatherDisplay.Header.init` (not inside the formatter — those helpers run ~24× per popup open) when the key misses the table.
- **Log the raw error, not `localizedDescription`.** `MapViewModel.swift:104` logs `error.localizedDescription`, which flattens the domain/code you'd actually want months later. The rest of the repo logs `\(error)` (e.g. `CoreApplication.swift`).
- **`precipText` hardcodes and doesn't localize `%`.** `WeatherDisplay.swift:106` builds `"\(Int(...*100))%"` inline while the header's chance-of-rain goes through `OBALoc("weather.chance_of_rain_format", …)`. Minor inconsistency in the stats row.
- **`HourlyEntry`'s `Identifiable`-by-`Date` relies on the feed never repeating an hour.** `list(...)` filters and sorts but never de-dupes by `time` [`WeatherDisplay.swift:145-147`]. Two entries sharing a `time` would hand `ForEach` duplicate ids (undefined behavior) and mark both as `isNow`. Realistically unique, but a `dedupe-by-hour-bucket` after the sort would make the contract hold rather than trust upstream.
- **`LegacyAlert` is eagerly built on every path, including the SwiftUI one that discards it.** It's stored on `WeatherDisplay` [`WeatherDisplay.swift:48`], so the panel path pays for three `String(format:)` calls it never reads, and the eventual flag-removal touches the primary initializer. A computed/`func`-based shim in a separate extension file would isolate its transitional nature and make the cleanup a single-file delete. (If you keep it stored, fine — just note in the doc comment that it's eagerly built.)
- **Test gaps worth a quick fill** (all currently green, but they'd pass through real bugs):
  - The Puget Sound fixture has `precip_probability == 0`, so `Int(precipProbability * 100)` truncation is never exercised at a non-zero value [`WeatherDisplay.swift:65, 106`]. A case with e.g. `0.45 → "45%"` would lock it down.
  - `highLowText` is asserted only `.toNot(beNil())` [`WeatherDisplayTests.swift:171`] — the `"H:%@  L:%@"` join itself (the one bit of formatting `Header` owns) is unverified.
  - `test_conditionText_groupsDayAndNightVariants` only checks `day == night`, never that `snow` actually returns "Snow" etc. — a mis-mapped condition wouldn't be caught.
- **A few UX papercuts to consider (not blocking):** a transient fetch failure clears `weatherDisplay` and so vanishes the floating button rather than keeping the last-known forecast; the open popup captures `presentedWeather` at tap time and won't refresh under it on `onAppBecameActive`; and the dimming backdrop sits below the floating sheet (`.floatingSheet` is applied after the overlay in `MapPanelRootView`), so the sheet stays bright/interactive while the card is up.

## Strengths

- The "two surfaces can't drift" invariant is the standout — immutable, construction-time-derived, and pinned by `test_init_populatesHeaderStatsAndLegacyAlertFromFixture`.
- The `obacoService == nil` (silent clear) vs. `catch` (clear + log) split is precisely the right distinction between "weather isn't offered here" and "the fetch broke" — and the new clear-on-guard fixes a latent stale-forecast bug from the pre-PR code.
- `HourlyEntry.list` coverage is exemplary: a pinned UTC `now`/calendar kills wall-clock flakiness, and every documented rule (drop-past-hour, "Now", 24-cap, Date identity, out-of-order sort, all-past→empty) has a dedicated test.
- `MapBaseType` and `WeatherFormatter` are textbook small, host-agnostic types — MapKit-free VM, `enum`-namespace formatter in OBAKitCore. Extension-safety confirmed.
- Liquid Glass usage checks out against the iOS 26 SDK (`Glass.regular` / `Glass.clear` are both valid `glassEffect(_:in:)` variants), with sensible pre-26 material fallbacks.

## Recommended Action

1. Rebase on the latest base branch and resolve the merge conflicts so the PR is mergeable.
2. Share one filter+sort helper between `highLow` and `HourlyEntry.list` so the Hi/Lo window matches the strip.
3. Run `scripts/extract_strings` and commit the `.strings` deltas for the new `weather.*` keys.
4. Optionally pick up the cheap wins: log unknown icon keys, log raw `error`, and add the non-zero-precip / highLowText assertions.
5. Re-run the review after fixes.

Note: this PR is marked to merge after #1163, which is still open — so the base needs to land first regardless.

---

**Verdict: Request Changes** — the merge conflicts, the Hi/Lo window inconsistency, and the un-extracted localization strings should all be resolved before this merges. Everything else is sound and the architecture is genuinely good.

aaronbrethorst (approved):
nice work, this is ready to merge. I filed #1174 as a follow-up.

PR #1171, opened 2026-06-30
Feature/panel routes picker

mosliem:
## Summary

Adds the SwiftUI Route Picker and Current Trip surfaces over the existing view models and wires them into the floating sheet routing layer.

  ## What's in this PR

  ### Route Picker & Current Trip SwiftUI surfaces
  - **`RoutePickerView`** — SwiftUI view over the existing route-picker view model, surfaced through the sheet route stack.
  - **`CurrentTripView`** — SwiftUI view over `CurrentTripViewModel`, with a Close button and proper empty/loading/error states.
  - `CurrentTripViewModel`:
    - `State` moved to an extension and made `Equatable`.
    - `findVehicle` now resets state to `.loading` at the start so retries clear stale errors.

  ### Trip presentation hand-off
  - New **`TripPresentationBridge`** carries an `onPresentTrip` closure from the SwiftUI host down to where a single matched route needs to push the legacy
  `TripViewController`.
  - Threaded `onPresentTrip` through `AppSheetViewFactory` (initially a no-op, then wired up).
  - Bridge is retained on the SwiftUI side so the single-match navigation actually fires instead of being deallocated mid-flight.
  - `TripViewController` is presented via a modal-chain walk so the SwiftUI host can hand off cleanly to UIKit.

  ### Sheet routing
  - `.currentTrip` sheet case now carries a `Route` so `CurrentTripViewModel` can be initialized with the right input.
  - Added `Strings+Sheet` entries and localizable strings for the new route-picker and current-trip surfaces.

  ### Floating "My Trip" button
  - New **`MyTripButton`** floating overlay on `MapPanelRootView`.
  - Button slides and fades along the sheet drag so it gets out of the way as the sheet expands.
  - Extracted **`liquidGlassButtonStyle`** view modifier (with pre-iOS 26 fallback) and applied it to `MyTripButton`.

  ### Shared SwiftUI primitives
  - **`EmptyStateView`** — reusable status component (icon + title + message + optional action) used by both new views.
  - **`AppSymbol`** — shared SF Symbol vocabulary so SwiftUI views stop hard-coding symbol names.
  - Adopted in both `RoutePickerView` and `CurrentTripView`.
  
 ## Trade-off: `MapPanelRootView` is not wired to `MapRegionManager`

  The SwiftUI `Map` in `MapPanelRootView` is intentionally **not** connected to `application.mapRegionManager` in this PR. That's a known gap.
  
## Effect on the Route Picker

  Because `MapPanelRootView`'s SwiftUI `Map` doesn't feed `MapRegionManager`, the Route Picker behaves differently when launched from the new SwiftUI surface than it
  does from the existing UIKit `MapViewController` flow.
  
### Follow-up
  A follow-up PR should:
  1. Observe `cameraPosition` on the SwiftUI `Map`.
  2. Convert it to an `MKCoordinateRegion` and feed it into `MapRegionManager` (matching what `MapViewController` does today).
  3. Render the cached stops as annotations on the SwiftUI map.
  
  This should be merged after #1166 

### Video
https://github.com/user-attachments/assets/93209bf7-7ce8-4934-b49a-4e8d983bf215
 

## Summary by CodeRabbit

* **New Features**
  * Added a new route picker with search, loading, and empty states.
  * Added a “My Trip” floating button and trip-finding flow with multiple-results support.
  * Introduced a refreshed empty-state experience with clearer actions and icons.
  * Improved button styling for a more polished glass-like appearance on supported devices.

* **Bug Fixes**
  * Prevented repeated trip presentations and improved refresh behavior while keeping the current screen stable.
  * Fixed loading/cancellation handling so canceled route loads no longer show completion errors.

### DIFF
--- a/OBAKit/Controls/SwiftUI/AppSymbol.swift
+++ b/OBAKit/Controls/SwiftUI/AppSymbol.swift
@@ -1,0 +1,54 @@
+//
+//  AppSymbol.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import Foundation
+
+/// Centralised SF Symbol names used across the SwiftUI surfaces.
+///
+/// Use this for symbols whose *semantic role* recurs across views (loading,
+/// error, retry, search, etc.). View-private symbols that name a single
+/// concrete state in one screen can still live in a file-local `Symbol`
+/// enum next to that view; promote them here once a second consumer
+/// appears. The UIKit equivalent for `UIImage`-based icons is `Icons`.
+enum AppSymbol {
+
+    // MARK: - Generic status
+
+    /// "Something went wrong." Use for error-state placeholders.
+    static let error = "exclamationmark.triangle"
+
+    /// Generic loading / in-progress indicator. Also doubles as the canonical
+    /// refresh/retry glyph because Apple uses the same symbol for both.
+    static let loading = "arrow.clockwise"
+
+    /// "Try again." Same glyph as `loading` by design — both communicate
+    /// "this will fire a fresh fetch."
+    static let retry = "arrow.clockwise"
+
+    // MARK: - Search
+
+    /// Search field icon / "no search results" placeholder.
+    static let search = "magnifyingglass"
+
+    // MARK: - Location
+
+    /// "Location services unavailable" / permission-denied state.
+    static let locationUnavailable = "location.slash"
+
+    /// "Finding your location" / acquiring a fix.
+    static let locationFinding = "location.viewfinder"
+
+    // MARK: - Transit
+
+    /// Bus icon. Used for transit-related empty states.
+    static let bus = "bus"
+
+    /// "No real-time tracking available." Crossed-out broadcast antenna.
+    static let noRealtime = "antenna.radiowaves.left.and.right.slash"
+}
--- a/OBAKit/Controls/SwiftUI/EmptyStateView.swift
+++ b/OBAKit/Controls/SwiftUI/EmptyStateView.swift
@@ -1,0 +1,54 @@
+//
+//  EmptyStateView.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import SwiftUI
+import OBAKitCore
+
+/// Reusable empty / loading / informational state for SwiftUI views.
+///
+/// Wraps `ContentUnavailableView` so call sites declare *what* to display
+/// (title, SF Symbol, optional action) instead of re-spelling the underlying
+/// initializer each time. Use `ErrorView` for richer error rendering with
+/// classification; use this for plain status messages (loading, empty list,
+/// "no results", info).
+struct EmptyStateView<Actions: View>: View {
+    let title: String
+    let systemImage: String
+    @ViewBuilder let actions: () -> Actions
+
+    var body: some View {
+        ContentUnavailableView {
+            Label(title, systemImage: systemImage)
+        } actions: {
+            actions()
+        }
+    }
+}
+
+// MARK: - No-actions convenience
+
+extension EmptyStateView where Actions == EmptyView {
+    init(title: String, systemImage: String) {
+        self.init(title: title, systemImage: systemImage, actions: { EmptyView() })
+    }
+}
+
+// MARK: - Preview
+
+#if DEBUG
+#Preview("Informational") {
+    EmptyStateView(title: "No routes found nearby.", systemImage: "magnifyingglass")
+}
+
+#Preview("With Action") {
+    EmptyStateView(title: "No active vehicle found on this route near you", systemImage: "bus") {
+        Button("Try Again") { }
+    }
+}
+#endif
--- a/OBAKit/Extensions/SwiftUIExtensions.swift
+++ b/OBAKit/Extensions/SwiftUIExtensions.swift
@@ -67,3 +67,28 @@ public extension View {
         }
     }
 }
+
+// MARK: - liquidGlassButtonStyle
+
+public extension View {
+    /// Apply to a `Button` to give it Apple's interactive Liquid Glass surface
+    /// on iOS 26+ (the press/morph "grab" response that comes with
+    /// `.buttonStyle(.glass)`), with a `.plain` + `.regularMaterial` fallback
+    /// on older systems so the button still reads as floating.
+    ///
+    /// Two shape parameters because the two surfaces use different APIs:
+    /// `borderShape` drives the iOS 26 glass morphing, `fallbackShape` fills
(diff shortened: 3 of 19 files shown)


aaronbrethorst (changes requested):
Hi Mohamed — this is a lot of careful work, and it shows. The `TripPresentationBridge` in particular is the kind of thing that's easy to get subtly wrong, and you nailed the memory model (controller owns the bridge strongly, factory closure captures `[weak bridge]`, `host`/`application` weak) *and* documented exactly why the modal-chain walk is necessary. The `AppSheetRoute.currentTrip(routeID:)` → `currentTrip(route:)` migration is clean, the tests were updated consistently across every affected case, and `Fixtures.createRoute` builds through the real decoding path instead of hand-mocking. The comments throughout are genuinely excellent.

There's one issue in the refresh path that I think breaks the feature's primary flow, and it needs to be fixed before this merges.

## Critical Issues (1 found)

### The 20-second refresh timer resets `state` to `.loading` on every tick, which breaks both the single-match and multi-match flows

Root cause: `CurrentTripViewModel.findVehicle()` now unconditionally sets `state = .loading` (`CurrentTripViewModel.swift:88`), and `startRefreshTimer()` calls `findVehicle()` every 20 seconds (`CurrentTripViewModel.swift:170`) — the *same* entry point as the manual retry button. Your comment at `:84-87` justifies the reset for the user-initiated retry case, but it applies equally to the background auto-refresh, which is exactly when you want to *preserve* what's on screen. Three consequences:

1. **Single match (the common case) traps the user.** For one match, `handle(results:)` sets `pendingNavigation` but deliberately leaves `state == .loading` (`CurrentTripViewModel.swift:134`). `CurrentTripView` presents the trip and clears `pendingNavigation` (`CurrentTripView.swift:236-242`), but the underlying view stays on the "Finding your vehicle…" spinner forever. As you note in the bridge comment (`MapPanelRootController.swift:596-603`), `CurrentTripView` stays mounted and its timer keeps firing under the modal. The dedup guard (`MapPanelRootController.swift:100-108`) correctly prevents *stacking* while the trip is up — but it does nothing once the user taps **Done**. On dismiss, `CurrentTripView` reappears frozen on the spinner, and within 20 seconds the next tick re-fires `pendingNavigation` and re-presents the trip the user just dismissed. The only escape is tapping **Close** on the spinner before the tick lands. For "find my current trip on this route," a single match is the expected case, so this is the happy path.

2. **Multi-match list flickers every 20 seconds.** When parked on `.multipleResults`, each tick swaps `content` (`CurrentTripView.swift:272-287`) from the `List` to the full-screen `EmptyStateView`, then back — destroying and rebuilding the `List`, losing scroll position, and interrupting an in-progress row tap.

3. **Haptic spam.** `.onChange(of: state)` fires `feedback.dataLoad(.failed)` on every `.loading → .error`/`.noRealtime` transition (`CurrentTripView.swift:243-252`), so a persistently-failing route buzzes every 20s; and the single-match path re-fires `feedback.dataLoad(.success)` behind the modal every 20s.

**Suggested fix:** split the two entry points. Keep the `.loading` reset on the user-initiated retry/`start()` path, but give the timer a variant that doesn't reset state — e.g. `findVehicle(resetState: Bool = true)` and pass `false` from `startRefreshTimer`, so background refreshes update results in place. Separately, the single-match path should transition to a terminal state on hand-off rather than sitting on `.loading`, so the underlying view isn't a frozen spinner and doesn't auto-re-present after a user-initiated dismiss.

## Important Issues (2 found)

### RoutePicker error state has no retry affordance
On load failure, `RoutePickerView` renders `EmptyStateView(title: error.localizedDescription, systemImage: AppSymbol.error)` with no action (`RoutePickerView.swift:400-401`), unlike `CurrentTripView`, which offers a `retryButton` on `.error`. Because `loadRoutes()` only re-runs via `onAppear` after `loadTask` is nilled on disappear, a transient network error forces the user to fully close and reopen the picker to retry. Add a retry action that re-invokes `loadRoutes()`, mirroring `CurrentTripView`.

### The `.loading`-reset behavior has no test
This is the PR's central behavioral change and it's currently unguarded — deleting `CurrentTripViewModel.swift:88` leaves every existing test green. It's synchronously testable since `state = .loading` is set before the `Task` is spawned:
```swift
viewModel.handle(results: [])          // drive to .noResults
viewModel.findVehicle()                // no await
expect(viewModel.state) == .loading
```
Whatever the fix for the critical issue above, please pin the intended retry-vs-background-refresh behavior with a test. Worth adding a couple of lines for the new hand-written `State.==` too (two `.error`s equal iff `localizedDescription`s match) — a typo there silently breaks SwiftUI's `.onChange` detection.

## Suggestions (4 found)

- **`scenePhase` `.inactive` churn** (`CurrentTripView.swift:223-235`): the `else` branch runs on `.inactive` as well as `.background`, so transient events (Control Center, banners, system alerts) trigger a `deactivate()`+`start()` and re-issue a find on return. Gating on `.background` only would avoid the redundant network calls.
- **`AppSheetRoute.id` vs `Hashable` divergence** (`SheetRoute.swift:121`): `id` derives from `route.id` only, while synthesized `Equatable`/`Hashable` compares the whole `Route`. Not a live bug today (the coordinator keys off `Identifiable.id`), but it's a latent trap if a route is ever stored in a `Set`/dictionary. `test_hash_consistency_forValueEqualRoutes` passes only because the fixtures are fully value-equal, so it doesn't actually pin the id-based contract it implies.
- **Silent no-ops want a log line**: the bridge's `guard let host, let application else { return }` (`MapPanelRootController.swift:69`) and the `.noLocation` / no-service state assignments (`CurrentTripViewModel.swift:98,103`) leave no trace. The network path already logs via `Logger.error` — extending that to these paths closes a telemetry gap for region/location misconfiguration.
- **Cancelled route load looks like "no routes"** (`RoutePickerViewModel.swift:121-124` → `RoutePickerView.swift:402`): on cancellation, `didFinishLoading = true` with no routes and no `loadError` falls through to the `allRoutes.isEmpty` branch, rendering "No routes found nearby." identically to a genuinely-empty region. Usually masked by the onAppear re-run, but the two states shouldn't render the same.

## Strengths

- `TripPresentationBridge` lifecycle and the topmost-presenter walk + already-presented guard are correct and thoroughly explained.
- Idle-timer disable/re-enable is balanced and ownership-guarded (`wasIdleTimerDisabledByUs`), and correctly re-acquired on `scenePhase == .active`.
- Cancellation handling in both view models distinguishes `CancellationError` from real errors and logs the latter with context.
- CLAUDE.md compliance is clean: strings through `OBALoc`, UIKit-only APIs kept out of OBAKitCore, `liquidGlassButtonStyle` with a proper pre-iOS 26 fallback.
- The `MapRegionManager` gap is clearly called out in both the PR description and an in-code `TODO`, so it's a known, scoped follow-up rather than a surprise.

## Recommended Action

1. Fix the critical refresh-timer / `.loading`-reset issue — split the retry vs background-refresh paths and give the single-match hand-off a terminal state.
2. Add a retry affordance to the RoutePicker error state.
3. Add a test pinning the retry/refresh state behavior (and the `State.==` error semantics).
4. Consider the suggestions.

**Verdict: Request Changes.**

aaronbrethorst (approved):
let's get this in. Nice work 👍

PR #1173, opened 2026-07-02
Tighten ViewModel test coverage and docstrings

### LINKED ISSUE #1167: Tighten ViewModel test coverage & docstrings (follow-up to #1153)
Follow-up to #1153 (Alerts/Agencies VM migration). The refactor is sound; these are test-layer cleanups deferred from review:

- [ ] **Fix two inaccurate test docstrings.** `AgencyAlertsViewModelTests.swift:16-18` claims it "Verifies the share-activity helper" (no such test exists); `AgenciesViewModelTests.swift:16-17` claims it verifies a "loading state" that `AgenciesViewModel` doesn't have.
- [ ] **Test `AgencyAlertsViewModel.shareActivityItems(for:)`** — both branches (URL present → `[url]`; absent → `[title, body]`). Pure, deterministic.
- [ ] **Test the dedup path.** `dedupedAlerts()` is never exercised against real data (all tests run an empty store). Use the `AgencyAlertsStore.insertAlerts(_:)` test seam + `AgencyAlert(feedEntity:agency:)` (see `MapPanelViewModelTests.swift:138`) with two alerts sharing an `id`.
- [ ] **Test `AgenciesViewModel.loadData()` nil-`apiService` throw** (`ApplicationTests.swift:130` shows how to build an app where `apiService` stays nil).
- [ ] **`try?` → `try`** in `AgenciesViewModelTests.test_loadData_success` so a thrown error surfaces with its real cause.

Optional polish (template consistency for future VM migrations):
- [ ] Resolve `AgenciesViewModel.loadData()` dual source of truth (returns array *and* assigns `self.agencies`).
- [ ] Rename `ServiceAlertViewModel.viewDidAppear()` to a caller-agnostic name (`onAppear()`/`prepare()`); the VM is UIKit-free.
- [ ] Drive `AgencyAlertsViewController`'s `isLoading` synchronously to avoid the Combine initial-emission ordering quirk.
- [ ] Collapse `ServiceAlertViewModel`'s `isBuilding`/`renderedHTML` guards into one state enum.

sushanttiwari-14:
## Summary

This PR completes the required test-layer follow-up work from #1167.

### Changes

- Fix inaccurate class-level docstrings in `AgencyAlertsViewModelTests` and `AgenciesViewModelTests`.
- Replace `try?` with `try` in the `loadData()` success test so failures surface their underlying error.
- Add deterministic tests covering both branches of `AgencyAlertsViewModel.shareActivityItems(for:)`.
- Add a real-data test covering the `dedupedAlerts()` path using `AgencyAlertsStore.insertAlerts(_:)`.
- Add a test covering the `nil` `apiService` error path in `AgenciesViewModel.loadData()`.
- Refactor private test helpers to remove duplicated fixture construction while preserving behavior.

### Verification

- [x] Project builds successfully.
- [x] `AgencyAlertsViewModelTests` pass.
- [x] `AgenciesViewModelTests` pass.
- [x] No production code changed.

Closes #1167

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for agency loading, including successful name-sorted results and failure when no API service is available.
  * Updated agency loading assertions to use async/throws and improved setup for scenario-specific configurations.
  * Expanded alert view model tests to cover loading-state transitions, error-delegate handling, collapsed section persistence, alert sharing with/without URLs, and alert deduplication behavior.
* **Documentation**
  * Improved test descriptions and scenario clarity to reflect the newly covered behaviors.

### DIFF
--- a/OBAKitTests/ViewModels/AgenciesViewModelTests.swift
+++ b/OBAKitTests/ViewModels/AgenciesViewModelTests.swift
@@ -13,8 +13,8 @@ import Combine
 @testable import OBAKit
 @testable import OBAKitCore
 
-/// Tests for `AgenciesViewModel`. Verifies the success path sorts by name,
-/// and that loading state resets to `false` after completion.
+/// Tests for `AgenciesViewModel`. Covers the success path in `loadData()` (agencies sorted by name)
+/// and the nil-`apiService` error path.
 @Suite(.serialized)
 final class AgenciesViewModelTests: OBATestCase {
     var queue: OperationQueue!
@@ -69,7 +69,7 @@ final class AgenciesViewModelTests: OBATestCase {
     }
 
     @Test @MainActor
-    func `Load data success populates agencies sorted by name`() async {
+    func `Load data success populates agencies sorted by name`() async throws {
         let dataLoader = MockDataLoader(testName: name)
         let app = createApplication(dataLoader: dataLoader)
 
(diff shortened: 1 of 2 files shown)


sushanttiwari-14:
@aaronbrethorst I'd appreciate a review whenever you have a chance. Thanks!

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Approved on the merits.

This does what #1167 asked and does it carefully. The parts I checked:

- The new assertions exercise real production behavior rather than the test's own setup. The two setup-shaped `#expect`s (`app.apiService == nil`, `agencyAlerts.count == 2`) are genuine preconditions followed by behavior assertions, not substitutes for them.
- The dedup test is the strongest addition. Two alerts sharing an `id` but differing in proto content really do both survive `Set<AgencyAlert>` insertion, since equality and hashing combine the whole proto — so `dedupedAlerts()` collapsing them to one is a real assertion that would fail if the filter were removed. That is the kind of test that earns its place.
- Swapping `try?` for `try` in the agencies success test is a small change that turns a silent skip into a real failure.

One correction, which is a little ironic in a PR about fixing inaccurate docstrings: the comment saying "`agencyAlertsUpdated()` is the only caller of `dedupedAlerts()`" is not right. `AgencyAlertsViewModel.init` calls it too — `AgencyAlertsViewModel.swift:47`, alongside `:70`. Harmless to the test, but worth correcting since accuracy is the point here.

Two smaller ones you can fold into the same push:

- `makeAlertVM`'s "Uses a UUID-based id so each call is unique" does not buy anything in this context — the returned `TransitAlertDataListViewModel` already mints its own `id: UUID = UUID()`, and these alerts never enter a set or store.
- The nil-`apiService` comment attributes the nil region to `LocationManagerMock` being unauthorized with no `fixedRegionName`, but omits the third condition it actually depends on: `RegionsService`'s single-active-region auto-select not firing. That holds today (the fixture has 10 active regions) and `ApplicationTests` already relies on the same premise, so it is not flaky — just an incomplete explanation.

I am not merging this yet, and the reason is on my end rather than yours. The `OBAKitTests` workflow runs for this branch are sitting at `action_required` waiting on my approval, so this branch has never actually been compiled by CI — `gh pr checks` reports pending. Given that a sibling PR just turned out not to compile for exactly this reason, I want a real green build before landing anything. I will get the workflow approved; once it goes green this merges as-is.

PR #1177, opened 2026-07-03
New Weather Card integration in UIKit

mosliem:
### Summary

Brings the new SwiftUI weather card (already shipped on `MapPanelRootView`) into the classic UIKit `MapViewController`. Tapping the `HoverBar` weather button now presents the same glassy `WeatherDetailPopup` — header, hourly strip, stats row — via `UIHostingController` instead of the bare `UIAlertController` that read `WeatherDisplay.legacyAlert`. The legacy alert path (and its `todaySummary` mirror field) is deleted with the same PR: nothing referenced it after the swap.

#### SwiftUI wrapper for UIKit hosting

- **`WeatherDetailPopupHost`** (new, `OBAKit/Sheet/Root/`) — thin bridge between `MapViewModel` and the existing `WeatherDetailPopup`. Holds the view model as `@ObservedObject` so a forecast refresh that lands while the card is open updates the display in place (same "refresh under the user" contract as `MapPanelRootView`).
- Dismissal is fully owned by SwiftUI: the card animates its own exit via `isPresented`, then the wrapper calls `@Environment(\.dismiss)` — no `WeakBox`, no closures reaching into UIKit, no double-animated tear-down.
- Enter animation is driven by an `.onAppear` state flip (`isPresented: false → true`) so SwiftUI sees a real change and runs the scale/opacity transition, rather than swallowing it as an initial value.

#### UIKit consumer rewrite

- `MapViewController.showWeather()` now instantiates `UIHostingController<WeatherDetailPopupHost>` and presents it `.overFullScreen` with `.crossDissolve` and a clear background so the card's own dim overlay owns the visual experience. The `HoverBar` `UIButton` and the `@objc private` selector stay unchanged — just the popup contents change.

#### Legacy alert path removed

- `WeatherDisplay+LegacyAlert.swift` deleted; `WeatherDisplay.todaySummary` (the OBAKit-side mirror that only fed the alert title) removed with it.
- `WeatherForecast.todaySummary` in `OBAKitCore` — the actual Obaco API field — is untouched. Only the UI mirror goes.
- `WeatherDisplayTests` fixture test renamed to `test_init_populatesHeaderStatsAndHourlyFromFixture`; the four `LegacyAlert` assertions and their block comment are gone. Header, Stats, `HourlyEntry`, and `buttonTitle` coverage all survive.

### Architecture notes

- The wrapper stores no view-owned data other than `isPresented`. `MapViewModel` is owned by `MapViewController`, so `@ObservedObject` is correct — `@StateObject` would incorrectly duplicate the model.
- `.overFullScreen` + `host.view.backgroundColor = .clear` keeps the map beneath the popup visible without adding a background layer; the card supplies its own `Color.black.opacity(0.25)` dim.
- The `@Environment(\.dismiss)` chain fires only after SwiftUI's exit animation completes, so UIKit tears down an already-transparent view — no cross-dissolve on top of the scale/opacity fade.
- The `WeatherDetailPopup` guard against `display == nil` (region change while popup open) already resets `isPresented` to `false`, which flows through the wrapper and dismisses the host — no extra plumbing.

### Video

https://github.com/user-attachments/assets/01bb7bee-0e8b-4e40-aa5d-debbe674fdab

## Summary by CodeRabbit

* **New Features**
  * Weather details now open in a richer full-screen popup with smoother cross-dissolve transitions instead of a standard alert.
  * The weather card now uses a dedicated SwiftUI-based detail experience.

* **Bug Fixes**
  * Improved popup dismissal so the close animation completes cleanly before the popup fully disappears.
  * Removed the legacy alert-based weather formatting, improving consistency in how weather information is presented.


aaronbrethorst (approved):
cool, nice work 👍

PR #1178, opened 2026-07-04
WeatherFormatter: derive icon fallbacks from a single metadata table

### LINKED ISSUE #1174: WeatherFormatter: derive icon → symbol/condition-text/color from one source
Follow-up from PR #1166 (weather overlay).

`WeatherFormatter` currently maintains three parallel tables keyed by the same Obaco icon strings:
- `iconToSymbol` (icon → SF Symbol) in `OBAKitCore/Weather/WeatherFormatter.swift`
- the `conditionText(for:)` switch (icon → localized condition word), same file
- `WeatherIcon.colors` (icon → palette) in `OBAKit/Sheet/Root/WeatherDetailPopup.swift`

They agree today, but adding a new condition (e.g. `thunderstorm`, `hail`) means editing all three, and the drift-warning gate (`isKnownIconKey`) only checks `iconToSymbol` — so a key added to the symbol table but missing from `conditionText` would pair a real icon with the "—" fallback and log nothing.

Low priority. Consider a single `[String: (symbol, conditionKey)]` metadata table in OBAKitCore that both fallbacks and the warning gate derive from, with the SwiftUI color palette layered on top by key. Not blocking — the fallbacks keep the UI usable.

mosliem:
Fixes #1174.

`WeatherFormatter` had three parallel tables keyed by the same Obaco icon strings (symbol map, condition-text switch, and the SwiftUI color palette), and the drift-warning gate only checked one of them. Collapses the OBAKitCore side into a single `iconMetadata: [String: IconMetadata]` (icon → symbol + condition family) that `systemImageName`, `conditionText`, and `isKnownIconKey` all read from — adding a condition is now one edit.

- New `ConditionKey` enum + `IconMetadata` struct back the table; day/night variants share one `ConditionKey` case.
- New public `WeatherFormatter.knownIconKeys` exposes the modeled set so downstream tables can assert coverage.
- The SwiftUI color palette in `WeatherDetailPopup` moves out of the private `WeatherIcon` view into an internal `WeatherIconPalette` so it can be verified from tests.

## Tests
- `test_isKnownIconKey_impliesConditionText` — every known key produces a real localized string, never the `—` fallback.
- `test_weatherIconPalette_coversAllKnownIconKeys` — the SwiftUI palette has an entry for every `knownIconKeys` key.

Full weather test suite (26 tests) passes.

## Summary by CodeRabbit

* **Refactor**
  * Simplified weather icon and condition handling into a more consistent, table-driven setup.
  * Separated icon color palette data into a dedicated place for easier maintenance.

* **Bug Fixes**
  * Improved consistency between weather symbols, condition text, and icon colors.
  * Reduced the chance of new weather icons falling back to unintended default styling.

* **Tests**
  * Added coverage to ensure all known weather icons have matching palette colors.
  * Added checks that known icon keys always return valid condition text.

### DIFF
--- a/OBAKit/Sheet/Root/WeatherDetailPopup.swift
+++ b/OBAKit/Sheet/Root/WeatherDetailPopup.swift
@@ -232,21 +232,26 @@ private struct StatsRow: View {
 // MARK: - Weather Icon
 
 /// Pulls the SF Symbol name from `WeatherFormatter` (single source of truth) and
-/// pairs it with a SwiftUI-only color palette.
+/// pairs it with a SwiftUI-only color palette. The palette lives here rather
+/// than in OBAKitCore because `Color` isn't reachable from the extension-safe
+/// core; `WeatherIconPalette.colors` is `internal` so tests can assert every
+/// key in `WeatherFormatter.knownIconKeys` also has a palette entry.
 private struct WeatherIcon: View {
     let iconName: String
     let size: CGFloat
 
     var body: some View {
         let symbol = WeatherFormatter.systemImageName(for: iconName)
-        let palette = Self.colors[iconName] ?? (.gray, .clear)
+        let palette = WeatherIconPalette.colors[iconName] ?? (.gray, .clear)
         Image(systemName: symbol)
             .symbolRenderingMode(.palette)
             .foregroundStyle(palette.primary, palette.secondary)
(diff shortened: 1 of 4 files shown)


aaronbrethorst (approved):
this looks like a good refactor. nice work.

PR #1183, opened 2026-07-11
Map Controls Floating Buttons

mosliem:
## Summary

Introduces SwiftUI-native floating map controls on `MapPanelRootView` — a top-center status pill (zoom/permission) and a bottom-trailing cluster with a map-type toggle and a "center on user" button — plus the `MapViewModel` state, permission dialog, and Liquid Glass styling to back them. Brings the SwiftUI composition root to functional parity with the UIKit `MapViewController` for these surfaces.

## What's new

**Controls (new SwiftUI views in `OBAKit/Sheet/Root/Controls/`)**
- `MapStatusPill` — top-center pill with two modes: "Zoom in for stops" and location-permission prompts. Display + tap only; presentation of the permission dialog is delegated to the host.
- `MapTypeButton` — standard/hybrid toggle, mirrors the UIKit `toggleMapTypeButton` icons, VM call, and accessibility strings.
- `CurrentLocationButton` — floating "center on my location" pill; hidden when location isn't authorized, matching UIKit behavior.
- `MapControlsCluster` — bottom-trailing vertical grouping of the map-type toggle above the current-location button.
- `MapPermissionDialog` — a `confirmationDialog` view modifier driven by an optional `TopPillState`. Attached inside the floating sheet's content builder so it presents on top of the sheet's `UISheetPresentationController` instead of stealing the map layer's presentation context (fixes the base sheet being dismissed when the dialog appeared).

**`MapPanelRootView` wiring**
- Overlays for the pill (top) and cluster (bottom-trailing, padded above `AppSheetRoute.homeCollapsedHeight`).
- Tracks `mapSize` via `onGeometryChange` and `visibleRegion` via `onMapCameraChange` so `centerOnUser` and `zoomInForStops` can build proper `MKCoordinateRegion`s for the SwiftUI `Map`.
- Applies `.mapStyle(...)` from `MapViewModel.mapType` and seeds the initial map type from `application.mapRegionManager.userSelectedMapType`.
- Nudges the weather button down when the pill is visible so they don't overlap.

**`MapViewModel` additions**
- `TopPillState` enum (`hidden`, `zoomInForStops`, `notDetermined`, `locationServicesOff`, `impreciseLocation`) with a computed `topPillState` that mirrors `MapStatusView.configure(for:zoomInStatus:)` — zoom warning wins over permission state.
- Extracted shared zoom constants: `zoomInForStopsSpan` (0.01) and `zoomLevelForCurrentLocation()` (17 full-accuracy, 11 reduced), replacing the inline values in `MapViewController`.
- Thin permission helpers: `requestLocationAuthorization()` and `requestTemporaryFullAccuracy(purposeKey:)`.

**Extensions**
- `MKCoordinateRegion.init(centeredOn:zoomLevel:mapSize:)` so the SwiftUI `Map` recenters with the same math as `MKMapView.setCenterCoordinate(zoomLevel:)`.
- `View.liquidGlassButtonStyle(borderShape:fallbackShape:)` — `.glass` + `buttonBorderShape` on iOS 26+, with a `.plain` + `.regularMaterial` fallback for older systems.

**Tests**
- `MapViewModelTests` coverage for `zoomInForStopsSpan`, `zoomLevelForCurrentLocation()`, and the `TopPillState` transitions (zoom-warning precedence, hidden-when-authorized, `.denied` → `.locationServicesOff`, `.notDetermined` → `.notDetermined`).

## Notable fixes on the branch
- Map-status permission dialog was dismissing the floating sheet — resolved by attaching the dialog inside the sheet's content builder.
- "Keep Location Off" buttons were losing their labels when given the `.cancel` role — dropped the role so labels render.
- Weather button now shifts down when the status pill is visible so the two don't overlap.

## Video

https://github.com/user-attachments/assets/d08414b6-ab9d-4160-9d26-5f6e6d32072b

## Summary by CodeRabbit

- **New Features**
  - Added floating map controls to toggle map type and center on the current location.
  - Added a top status pill for zoom guidance and permission/precision states.
  - Added permission alerts with actions to enable access, opt out, or open Settings.
  - Added Liquid Glass button styling with automatic fallbacks for older iOS versions.
- **Improvements**
  - Map type is now persisted and kept in sync automatically; updated zooming and accuracy-based behavior.
  - Refreshed map panel layout and live geometry/alert handling.
- **Tests**
  - Expanded MapViewModel and permission/zoom/persistence coverage; improved async timing in CI tests.


aaronbrethorst (changes requested):
This has merge conflicts and, even after fixing them, I'm still seeing some build errors. Get them all sorted out and we'll get moving on this again.

OBAKit/Sheet/Root/Controls/MapPermissionAlert.swift
```diff
@@ -0,0 +1,161 @@
...
+        switch state {
+        case .notDetermined:
+            return OBALoc(
+                "locationservices_alert_notdetermined.title",
+                value: "Enable Location Services",
+                comment: "Title of the alert asking the user to enable location services."
+            )
+        case .locationServicesOff:
+            return OBALoc(
+                "locationservices_alert_off.title",
```
aaronbrethorst: **Alert title renders a literal `%@`.** `title(for:)` passes the keys `locationservices_alert_off.title` and `locationservices_alert_imprecise.title` with fallbacks "Location Services Off" / "Precise Location Off". But those exact keys already exist in `Localizable.strings` as format strings — e.g. `"%@ works best with your location."` (used by `MapStatusView.alert` with a `String(format:…, appName)` arg). Since `OBALoc` returns the `.strings` value when the key is present, the title renders `"%@ works best with your location."` — including a stray `%@` — instead of "Location Services Off". Use distinct keys for the SwiftUI alert titles.

_Verified CONFIRMED — high-effort review._

OBAKit/ViewModels/MapViewModel/MapViewModel.swift
```diff
@@ -131,12 +165,71 @@ class MapViewModel: NSObject, ObservableObject, LocationServiceDelegate {
...
+        case locationServicesOff
+        case impreciseLocation
+    }
+
+    var topPillState: TopPillState {
+        if showZoomWarning { return .zoomInForStops }
+        switch locationAuthStatus {
+        case .notDetermined:
+            return .notDetermined
+        case .denied, .restricted:
```
aaronbrethorst: **MDM/restricted users get a dead-end Settings prompt.** `topPillState` folds `CLAuthorizationStatus.restricted` into the same `.locationServicesOff` branch as `.denied`, so `handlePermissionTap` presents an "Open Settings" alert. The old `MapStatusView` modeled `.restricted` as `.locationServicesUnavailable`, whose `alert(for:)` returned nil — a restricted user cannot lift the restriction in Settings, so the prompt is a dead end. Consider restoring a distinct non-actionable state for `.restricted`.

_Verified CONFIRMED — high-effort review._

OBAKit/Sheet/Root/MapPanelRootView.swift
```diff
@@ -78,15 +95,42 @@ struct MapPanelRootView: View {
...
+        } action: { _, newValue in
+            mapSize = newValue
+        }
+        .onMapCameraChange(frequency: .onEnd) { context in
+            visibleRegion = context.region
+        }
+        .overlay(alignment: .top) {
+            MapStatusPill(
+                state: mapViewModel.topPillState,
+                onZoomInForStops: zoomInForStops,
```
aaronbrethorst, line 119: **The "Zoom in for stops" pill can never appear in the SwiftUI panel.** Nothing calls `MapViewModel.updateZoomWarning(_:)` in the SwiftUI root — only the UIKit `MapViewController` does — so `showZoomWarning` stays false and `topPillState` never becomes `.zoomInForStops`. The `MapStatusPill` `onZoomInForStops` handler, `zoomInForStops()`, and the new `visibleRegion` tracking added in this PR are unreachable here. Either wire `updateZoomWarning` into the SwiftUI map, or drop the dead plumbing.

_Verified CONFIRMED — high-effort review._

OBAKit/Mapping/MapViewController.swift
```diff
@@ -316,13 +316,11 @@ class MapViewController: UIViewController,
...
-        var zoomLevel = 17
-
-        if application.locationService.accuracyAuthorization == .reducedAccuracy {
-            zoomLevel = 11
-        }
-
-        mapRegionManager.mapView.setCenterCoordinate(centerCoordinate: userLocation.coordinate, zoomLevel: zoomLevel, animated: true)
+        mapRegionManager.mapView.setCenterCoordinate(
+            centerCoordinate: userLocation.coordinate,
+            zoomLevel: viewModel.zoomLevelForCurrentLocation(),
```
aaronbrethorst, line 321: **Coarse zoom after a one-shot precise-location grant.** `centerMapOnUserLocation()` replaced a live read of `locationService.accuracyAuthorization` with the cached `viewModel.zoomLevelForCurrentLocation()` (backed by `@Published accuracyAuthorization`). iOS does not reliably deliver `locationManagerDidChangeAuthorization` for temporary full-accuracy grants, so after a user grants precise location via `requestTemporaryFullAccuracy` and taps "center on my location", the cache is still `.reducedAccuracy` and the map zooms to level 11 instead of 17.

_Verified PLAUSIBLE — high-effort review._

OBAKit/ViewModels/MapViewModel/MapViewModel.swift
```diff
@@ -131,12 +165,71 @@ class MapViewModel: NSObject, ObservableObject, LocationServiceDelegate {
...
+            return .locationServicesOff
+        case .authorizedAlways, .authorizedWhenInUse:
+            return accuracyAuthorization == .reducedAccuracy ? .impreciseLocation : .hidden
```
aaronbrethorst, line 221: **Imprecise-location pill can get stuck.** `topPillState` derives the imprecise state from the cached `@Published accuracyAuthorization` rather than reading `locationService.accuracyAuthorization` live (as the old `MapStatusView.configure(with:)` did). If iOS does not deliver `locationManagerDidChangeAuthorization` after a temporary full-accuracy grant, the pill remains stuck on "Precise location unavailable" even though precise location is active. The regression test calls the delegate directly, so it will not catch a missing OS callback.

_Verified PLAUSIBLE — high-effort review._

OBAKit/Sheet/Root/MapPanelRootView.swift
```diff
@@ -134,26 +181,119 @@ struct MapPanelRootView: View {
...
+        }
+    }
+
+}
+
+extension MapPanelRootView {
+
+    // MARK: - Actions
+
+    private func centerOnUser() {
```
aaronbrethorst, line 250: **Zero-span region on a cold-launch tap.** `centerOnUser()` builds the region from `mapSize`, which is `.zero` until `.onGeometryChange` fires. If the user taps current-location before layout reports a size, `MKCoordinateRegion(centeredOn:zoomLevel:mapSize:)` yields a zero-span `MKCoordinateSpan` and the Map animates to a degenerate zoom. The old UIKit path used `bounds.size`, which is always laid out — so this is a SwiftUI-only regression. Consider guarding against `mapSize == .zero`.

_Verified PLAUSIBLE — high-effort review._

OBAKit/Sheet/Root/MapPanelRootView.swift
```diff
@@ -97,11 +141,14 @@ struct MapPanelRootView: View {
                     )
                     .presentationBackground(.clear)
                 }
+                .mapPermissionAlert(state: $permissionAlertState, onAction: handleAlertAction)
```
aaronbrethorst: **Permission alert is attached per stacked-sheet layer.** The `.mapPermissionAlert(state: $permissionAlertState, ...)` modifier lives inside the `floatingSheet` content builder, which `FloatingSheetContainer`/`StackedSheetLayer` re-invoke at every stack depth. Each presented sheet attaches its own alert modifier bound to the single shared `permissionAlertState`, so setting it non-nil fires multiple concurrent `.alert(isPresented:)` presentations — the alert may fail to present, present on the wrong sheet, or double-present.

_Verified PLAUSIBLE — high-effort review._

OBAKit/ViewModels/MapViewModel/MapViewModel.swift
```diff
@@ -131,12 +165,71 @@ class MapViewModel: NSObject, ObservableObject, LocationServiceDelegate {
...
+        @unknown default:
+            return .hidden
+        }
```
aaronbrethorst, line 226: **Future auth status silently hides the pill.** `topPillState`'s `@unknown default` returns `.hidden`, whereas the old `MapStatusView.LocationState.init` mapped an unknown `CLAuthorizationStatus` to `.locationServicesUnavailable` (a visible warning). If Apple introduces a future denied-like status, the map shows no location-status pill and no tap target to fix it.

_Verified PLAUSIBLE — high-effort review._

OBAKit/Sheet/Root/Controls/MapStatusPill.swift
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+}
+
+// MARK: - Display
+
+extension MapStatusPill {
+    /// Presentation model for a single pill state. Split so the empty case
+    /// (`.hidden`) simply returns `nil` and the view collapses to `EmptyView`.
+    fileprivate struct Display {
```
aaronbrethorst, line 69: **Duplicated state→symbol→label mapping.** `MapStatusPill.Display` re-encodes the SF Symbol names (`location.slash`, `location.circle`, `plus.magnifyingglass`) and OBALoc keys already centralized in `MapStatusView.configure(for:zoomInStatus:)`. A future copy change must be edited in both the UIKit and SwiftUI surfaces or the two pills drift out of sync. Consider a shared source of truth.

_Verified CONFIRMED — high-effort review._

OBAKit/Extensions/MapKitExtensions.swift
```diff
@@ -159,6 +159,40 @@ public extension MKMapView {
...
+        switch self {
+        case .standard: return .mutedStandard
+        case .hybrid: return .hybrid
+        }
+    }
+
+    /// Inverse of `mkMapType`. Any MapKit value other than `.hybrid` collapses
+    /// to `.standard` so a stray persisted `.satellite` (or future addition)
+    /// never boots the app into an unhandled render mode.
+    init(_ mkMapType: MKMapType) {
```
aaronbrethorst, line 191: **This new shared init diverges from the still-hand-rolled UIKit conversion.** `MapBaseType(_ mkMapType:)` collapses any non-`.hybrid` `MKMapType` to `.standard`. But `MapViewController.swift:78` (unchanged in this PR) still hand-rolls the inverse mapping — `userSelectedMapType == .mutedStandard ? .standard : .hybrid` — which collapses any non-`.mutedStandard` value to `.hybrid`, the opposite default. `MapPanelRootView` uses this new init; the UIKit map still uses the old ternary. If a stray/legacy `MKMapType` (e.g. `.satellite`) is ever persisted, the UIKit map boots into hybrid while the SwiftUI root boots into standard for the same stored value. Consider routing `MapViewController:78` through this shared init so the two surfaces can't drift.

_Verified CONFIRMED — high-effort review. (Anchored here because MapViewController.swift:78 is unchanged context and not commentable inline.)_

aaronbrethorst (changes requested):
Get the items noted fixed and we'll merge this PR.

OBAKit/ViewModels/MapViewModel/MapViewModel.swift
mosliem: `topPillState` drives a SwiftUI pill, so it has to update reactively — a live read wouldn't re-render on its own, and it'd break `test_topPillState_accuracyElevationClearsImprecisePill`, which relies on the `@Published` accuracy update. So I've kept the cache here.

I did fix the actionable half — the imperative path: `zoomLevelForCurrentLocation()` now reads accuracy live (see the "coarse zoom" thread), where a one-shot read can't go stale.

The robust fix for a truly-missed OS callback would be plumbing `requestTemporaryFullAccuracyAuthorization(…, completion:)` through `LocationService` into `@Published accuracyAuthorization` — keeps reactivity and updates on grant. That touches `OBAKitCore`, so I left it for a follow-up unless you'd rather I do it here.

OBAKit/Sheet/Root/Controls/MapPermissionAlert.swift
mosliem: Fixed. These reused `locationservices_alert_off.title` / `…imprecise.title`, which are `%@`-format strings consumed by `MapStatusView.alert(for:)`, so `OBALoc` returned the raw format string. Now using distinct keys `map_status_alert.location_services_off.title` / `…precise_location_off.title` (added to en + all 12 locales).

OBAKit/ViewModels/MapViewModel/MapViewModel.swift
mosliem: Fixed. Added a distinct non-actionable `.locationServicesUnavailable` state. `.restricted` now maps there — visible pill, no "Turn On in Settings" alert (tapping is a no-op) — separated from `.denied`. Mirrors the old `MapStatusView.LocationState.locationServicesUnavailable`. Covered by `test_topPillState_restrictedMapsToLocationServicesUnavailable`.

OBAKit/Sheet/Root/MapPanelRootView.swift
mosliem: Wired it. `onMapCameraChange` now calls `updateZoomWarning` using a new shared `MapRegionManager.shouldShowZoomInWarning(forVisibleMapRectHeight:)` (same 40,000-pt threshold the UIKit `zoomInStatus` uses, via `context.rect.height`). The pill, its `onZoomInForStops` handler, and `visibleRegion` tracking are now reachable. Threshold predicate covered by `test_shouldShowZoomInWarning_thresholdBehavior`.

OBAKit/Mapping/MapViewController.swift
mosliem: Fixed. `zoomLevelForCurrentLocation()` now reads `accuracyAuthorization` live from `locationService` instead of the cached `@Published` value. Its an imperative one-shot read at tap time, so a temporary full-accuracy grant that never delivered `locationManagerDidChangeAuthorization` no longer zooms to 11. Existing zoom-level tests still pass.

OBAKit/Sheet/Root/MapPanelRootView.swift
mosliem: Fixed. `centerOnUser()` now guards `mapSize != .zero`, so a tap before `.onGeometryChange` reports a size is a no-op rather than animating to a degenerate zero-span region.

OBAKit/Sheet/Root/MapPanelRootView.swift
mosliem: Fixed. The live `permissionAlertState` binding is now attached only on the base (non-stacking) sheet layer; stacked layers get `.constant(nil)`. The base sheet shows exactly one route at a time, so this yields a single alert host instead of one per stack depth.

OBAKit/ViewModels/MapViewModel/MapViewModel.swift
mosliem: Fixed. `@unknown default` now returns `.locationServicesUnavailable` (a visible, non-actionable pill) instead of `.hidden`, so a future denied-like status still surfaces a status indicator.

OBAKit/Sheet/Root/Controls/MapStatusPill.swift
mosliem: Fixed. Extracted a shared `MapStatusIndicator` (single source of truth for SF Symbol names + `OBALoc` copy). Both the UIKit `MapStatusView.configure` and the SwiftUI `MapStatusPill.Display` now map their state onto it, so the two surfaces cant drift.

OBAKit/Extensions/MapKitExtensions.swift
mosliem: Fixed. `MapViewController` init now routes through the shared `MapBaseType(_ mkMapType:)` init instead of its own `== .mutedStandard ? .standard : .hybrid` ternary, so both surfaces resolve a stray/legacy `MKMapType` the same way.

mosliem:
Done

PR #1186, opened 2026-07-13
fix: harden BulletinOverlayWindow before it ships to production (#1170)

### LINKED ISSUE #1170: Harden BulletinOverlayWindow before it ships to production (fast-follow to #1163)
## Background

PR #1163 (floating sheet system foundation) introduced `BulletinOverlayWindow` — a dedicated `.alert`-level `UIWindow` for hosting `BLTNItemManager` presentations — and routed every bulletin (`AgencyAlertBulletin`, `ErrorBulletin`, `ReachabilityBulletin`, `RegionMismatchBulletin`) through it.

The motivating problem (`UISheetPresentationController` clamping a bulletin to the map-panel home sheet's detent) **only exists in map-panel mode**, but the fix was applied **globally** — so it ships to classic (production) users, not just experimental-flag testers. This is a fast-follow to either scope the change down or harden it before it can regress the shipping app.

### 1. Decide and document the blast radius
The new window path is not behind `OBAUseMapPanelExperience`. Either:
- **Scope it to map-panel mode** — keep the old `keyWindowFromScene?.topViewController` presentation for classic mode and only use `BulletinOverlayWindow` when the flag is on; **or**
- **Consciously accept** shipping it to everyone — in which case items 2 and 3 must be fixed first.

Files: `OBAKit/BLTNBoard/OBABulletinPage.swift` and the four bulletin call sites.

### 2. `dismissalHandler` chaining accumulates on reused page items
`BulletinOverlayWindow.install` snapshots `rootItem.dismissalHandler` and wraps it:
```swift
let originalDismissal = rootItem.dismissalHandler
rootItem.dismissalHandler = { [weak self] item in
    originalDismissal?(item)
    self?.teardown()
}
```
For a bulletin whose page is a long-lived stored property re-shown repeatedly — `ReachabilityBulletin.connectivityPage` is the clear case, since connectivity flapping (tunnels/subways) re-calls `show(in:rootItem:)` — BLTN never resets `dismissalHandler` between presentations, so each cycle wraps the previous closure. Over N flaps the chain nests N deep and `teardown()` fires N times. Idempotent and bounded by flap count, but unbounded-in-principle closure accumulation on a shared singleton, in production, in the scenario a transit app sees most.

**Fix:** snapshot/reset the original handler once rather than chaining onto the persisted item.

Files: `OBAKit/BLTNBoard/BulletinOverlayWindow.swift` (install), `OBAKit/Reachability/Reachability.swift`.

### 3. Concurrent bulletin managers can orphan a live bulletin in release
The `isShowingBulletin` guard is per-`BLTNItemManager`, but `BulletinOverlayWindow.shared` is one window across all managers. Realistic trigger: go offline → `ReachabilityBulletin` shows → a network error fires → `Application.displayError` → `ErrorBulletin` (a *different* manager, so its `isShowingBulletin` is false) → `install` hits the in-use branch. DEBUG asserts (good); release returns the first bulletin's host and presents the error above it, then the reachability bulletin's dismissal runs `teardown()` and nils the window out from under the still-live error bulletin. Offline + network-error commonly co-occur, so this isn't exotic.

**Fix:** introduce cross-manager coordination/queueing (or a single shared manager) so a second bulletin can't tear the window out from under a live one.

Files: `OBAKit/BLTNBoard/BulletinOverlayWindow.swift`.

## Acceptance criteria
- [ ] Blast radius decision made and reflected in code (gated to map-panel, or items 2 & 3 fixed)
- [ ] Reachability flapping no longer accumulates nested `dismissalHandler` closures
- [ ] Concurrent bulletins from different managers can't orphan/corrupt each other or leak the overlay window
- [ ] Optionally: bulletin-drop paths (`no foreground scene`) route through `analytics?.reportError` instead of console-only `Logger.error`

Fast-follow to #1163.

mosliem:
Fast-follow to #1163. Closes #1170.

## Summary

- **Gates the dedicated `BulletinOverlayWindow` presentation path behind `OBAUseMapPanelExperience`.** Classic (production) mode reverts to the pre-#1163 `keyWindowFromScene?.topViewController` presentation. The `UISheetPresentationController` sheet-clamping bug the overlay window was created to fix only exists when the map-panel flag is on, so classic users no longer take on the new window path (and its singleton lifecycle) for a bug they don't have.
- **Reads the flag from the same app-group suite Settings writes to** (`Bundle.main.appGroup` → `UserDefaults(suiteName:)`), matching `SettingsViewController`, `ApplicationRootControllerFactory`, and `CoreApplicationKey.defaultValue`. Reading `UserDefaults.standard` would have silently kept the overlay-window path off for map-panel testers.
- **Replaces `dismissalHandler` closure wrapping in `BulletinOverlayWindow.install` with snapshot-and-restore semantics.** Reused BLTN page items (`ReachabilityBulletin.connectivityPage` re-shown across connectivity flaps) no longer accumulate nested handlers across presentations.

## Notes

The four bulletin call sites (`AgencyAlertBulletin`, `ErrorBulletin`, `ReachabilityBulletin`, `RegionMismatchBulletin`) are unchanged — all logic changes live in `OBABulletinPage.swift` and `BulletinOverlayWindow.swift`.

## Summary by CodeRabbit

* **New Features**
  * Added flag-controlled bulletin presentation for the map-panel experience, with an automatic fallback to the classic presentation when disabled.

* **Bug Fixes**
  * Hardened overlay bulletin lifecycles by snapshotting and restoring dismissal behavior, preventing duplicated/accumulated dismissal handling across repeated presentations.
  * Improved resilience when the required presentation scene or top view controller isn’t available.

* **Documentation**
  * Updated the bulletin overlay hardening design spec and clarified testing expectations for the overlay path.

* **Tests**
  * Added regression coverage for repeated presentations and correct one-time dismissal invocation.

### DIFF
--- a/OBAKit/BLTNBoard/BulletinOverlayWindow.swift
+++ b/OBAKit/BLTNBoard/BulletinOverlayWindow.swift
@@ -33,23 +33,35 @@ final class BulletinOverlayWindow {
 
     private var window: UIWindow?
 
+    /// The item currently being presented, held weakly so we can restore
+    /// its original `dismissalHandler` on teardown without extending its
+    /// lifetime. Items outlive this window in practice (they're stored
+    /// properties on bulletin classes like `ReachabilityBulletin`), but the
+    /// weak reference is correct regardless.
+    private weak var savedRootItem: BLTNItem?
+
+    /// The `dismissalHandler` the caller had on `rootItem` before `install`
+    /// swapped in the overlay's wrapper. Restored verbatim on teardown so
+    /// the item goes back to exactly the state we found it in — no matter
+    /// how many prior presentations occurred.
+    private var savedDismissalHandler: ((BLTNItem) -> Void)?
+
     private init() {}
 
     /// Installs the overlay window and returns its host controller for
-    /// `showBulletin(above:)`. Hooks `rootItem.dismissalHandler` to tear the
-    /// window down once the bulletin dismisses, chaining any handler the
-    /// caller already set.
+    /// `showBulletin(above:)`. Snapshots `rootItem.dismissalHandler` and
+    /// replaces it with a wrapper that dispatches to the original then
+    /// tears the window down. On teardown the original handler is restored,
+    /// so a reused item (e.g. `ReachabilityBulletin.connectivityPage` re-shown
+    /// across connectivity flaps) never accumulates nested wrappers.
     ///
     /// One bulletin at a time. Each `BLTNItemManager` already guards on
     /// `isShowingBulletin`, but those guards are per-manager; this singleton is
     /// shared across every OBA bulletin manager, so two managers can call
-    /// `install` while a third is mid-presentation. If that ever happens, the
-    /// second caller's `dismissalHandler` chain would never be installed (the
-    /// early-return path used to silently skip it) and teardown would fire on
-    /// the first dismissal, yanking the window out from under the second
-    /// bulletin. The DEBUG assert flags the violation at the point it occurs;
-    /// release builds still return the existing host so the second bulletin at
-    /// worst piggybacks on the first's window rather than crashing.
+    /// `install` while a third is mid-presentation. The DEBUG assert flags the
+    /// violation at the point it occurs; release builds still return the
+    /// existing host so the second bulletin at worst piggybacks on the first's
+    /// window rather than crashing.
     func install(in scene: UIWindowScene, rootItem: BLTNItem) -> UIViewController {
         // Single-page only: teardown rides `rootItem.dismissalHandler`, but
         // BLTN fires the dismissal handler on `currentItem` — which only
@@ -79,16 +91,35 @@ final class BulletinOverlayWindow {
 
         self.window = window
 
-        let originalDismissal = rootItem.dismissalHandler
-        rootItem.dismissalHandler = { [weak self] item in
-            originalDismissal?(item)
-            self?.teardown()
-        }
+        swapDismissalHandler(on: rootItem)
 
         return host
     }
 
+    /// Snapshots the item's current `dismissalHandler` and swaps in a wrapper
+    /// that fires the original handler then tears down. Extracted from
+    /// `install` so tests can exercise the handler-management contract without
+    /// synthesizing a `UIWindowScene`.
+    internal func swapDismissalHandler(on rootItem: BLTNItem) {
+        savedRootItem = rootItem
+        savedDismissalHandler = rootItem.dismissalHandler
+        rootItem.dismissalHandler = { [weak self] item in
+            self?.savedDismissalHandler?(item)
+            self?.teardown()
+        }
+    }
+
+    /// Restores the previously-saved `dismissalHandler` onto the tracked item
+    /// and clears the saved state. Extracted from `teardown` for testability.
+    internal func restoreDismissalHandler() {
+        savedRootItem?.dismissalHandler = savedDismissalHandler
+        savedRootItem = nil
+        savedDismissalHandler = nil
+    }
+
     private func teardown() {
+        restoreDismissalHandler()
+
         let scene = window?.windowScene
         window?.rootViewController = nil
         window?.isHidden = true
--- a/OBAKit/BLTNBoard/OBABulletinPage.swift
+++ b/OBAKit/BLTNBoard/OBABulletinPage.swift
@@ -60,7 +60,17 @@ extension BLTNItemManager {
         // intended behavior — callers (e.g. reachability flapping) lean on this.
         guard !isShowingBulletin else { return }
 
-        if UserDefaults.standard.bool(forKey: FeatureFlags.useMapPanelExperienceKey) {
+        // Settings writes this flag to `application.userDefaults`, which the
+        // OBA app initializes as an app-group suite (see `AppDelegate.m` and
+        // `CoreApplicationKey.defaultValue`). `UserDefaults.standard` reads a
+        // different suite and would always see the flag as OFF. Resolve the
+        // same suite here; if it can't be resolved (unlikely outside tests),
+        // treat the flag as OFF — the classic path is the safe production
+        // default anyway.
+        let appGroupDefaults = Bundle.main.appGroup.flatMap { UserDefaults(suiteName: $0) }
+        let useMapPanelExperience = appGroupDefaults?.bool(forKey: FeatureFlags.useMapPanelExperienceKey) ?? false
+
+        if useMapPanelExperience {
             // No usable scene means the bulletin disappears with no UI trace,
             // including for the error path (`Application.displayError`). Log so
             // there's at least a diagnostic breadcrumb instead of silent loss.
--- a/OBAKit/BLTNBoard/OBABulletinPage.swift
+++ b/OBAKit/BLTNBoard/OBABulletinPage.swift
@@ -32,36 +32,55 @@ class ThemedBulletinPage: BLTNPageItem {
 }
 
 extension BLTNItemManager {
-    /// Presents the bulletin in a dedicated overlay `UIWindow` attached to the
-    /// active scene.
+    /// Presents the bulletin in the appropriate host for the current experience.
     ///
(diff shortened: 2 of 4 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The new regression tests pass unchanged against the buggy pre-PR implementation, so they don't cover the fix they're named for. Both assert only how many times the *original* handler fires, and that number is identical under chaining and under snapshot-restore. `BLTNItem.dismissalHandler` is a plain stored `((BLTNItem) -> Void)?` and `onDismiss()` just calls `dismissalHandler?(self)`, so the old `let originalDismissal = rootItem.dismissalHandler; rootItem.dismissalHandler = { originalDismissal?(item); self?.teardown() }` shape yields exactly the same counts: in `test_reused_item_does_not_accumulate_dismissal_handlers` each wrapper `Wk` still calls `H0` exactly once, so five loop fires plus the final direct fire is 6 either way; in `test_teardown_restores_original_handler` the second `page.dismissalHandler?(page)` hits `W1 -> H0` under the old code and sets `originalFired` just the same. The signal that actually distinguishes them is the number of `teardown()` invocations (5 with the fix vs. 20 with chaining) or the wrapper-chain depth, and neither is asserted (bug due to `OBAKitTests/BLTNBoard/BulletinOverlayWindowTests.swift` asserting `originalCallCount`/`originalFired` rather than teardown count or handler identity).

https://github.com/OneBusAway/onebusaway-ios/blob/f4b3ac9542f0661d1f0696e37c4d1df5df136581/OBAKitTests/BLTNBoard/BulletinOverlayWindowTests.swift#L45-L49

https://github.com/OneBusAway/onebusaway-ios/blob/f4b3ac9542f0661d1f0696e37c4d1df5df136581/OBAKitTests/BLTNBoard/BulletinOverlayWindowTests.swift#L68-L71

aaronbrethorst (changes requested):
The production code here is right, and I checked the parts that worried me most:

- The app-group suite resolution is correct. `AppDelegate.m:36` builds the app's defaults with `initWithSuiteName:appGroup`, and `SettingsViewController` / `ApplicationRootControllerFactory` both read and write the flag through `application.userDefaults` — so `UserDefaults.standard` genuinely would have read the wrong store and always seen the flag as OFF. Falling back to the classic path when the suite can't be resolved is the safe default.
- The classic branch faithfully restores the pre-#1163 presentation, and it keeps the drop-logging that the #1163 review round asked for rather than reverting to a silent return.
- Snapshot-and-restore is the right shape for the handler fix.

One thing blocks the merge, and it's in the tests rather than the code.

**`test_reused_item_does_not_accumulate_dismissal_handlers` does not detect the bug it's named for.** Both new tests assert only how many times the *original* handler fires, and that count is identical under the old chaining implementation and the new snapshot-restore one. Under chaining, the original sits at the tail of the wrapper chain, so a dismissal still reaches it exactly once — five cycles plus one direct call is 6 either way. The accumulation shows up as `teardown()` running once per accumulated wrapper (5 vs 20 across that loop), and nothing asserts on that. `test_teardown_restores_original_handler` has the same problem: it only checks that the original fires, which chaining also satisfies.

That matters more than a normal coverage gap because the file's docstring and the design doc both present this as proof the flap loop no longer accumulates handlers — so a future change that reintroduced chaining would leave a green test with exactly that name sitting next to it.

The fix is small: make the teardown observable and assert on it — an internal invocation counter on `BulletinOverlayWindow`, or an injectable teardown hook, either is fine. Then the loop can assert one teardown per cycle instead of one original-handler call per cycle.

Two non-blocking things while you're in here:

- The design doc contradicts the shipped code. `2026-07-05-bulletin-overlay-window-hardening-design.md` says the flag is read from `UserDefaults.standard` in its Design section and Risks section, while its own "Which `UserDefaults` here" paragraph and the actual code use the app-group suite. Since the doc is committed, it'll mislead whoever reads it next. The test names in the doc don't match the shipped ones either.
- The docstring rewrite dropped the paragraph explaining that a concurrent `install` early-returns without installing the handler chain, so teardown can fire on the first dismissal and yank the window out from under a second bulletin. That behavior is unchanged and deliberately deferred to a follow-up, so removing the only in-code description of it loses the warning. Worth keeping a sentence, even with the blast radius now flag-gated.

Fix the test assertion and I'll re-review promptly — the rest of this is ready to go.

aaronbrethorst:
### Code review

Found 1 issue:

1. The rewritten `install` docstring drops the paragraph describing what actually happens on a concurrent `install`, and the sentence left in its place understates it: "release builds still return the existing host so the second bulletin at worst piggybacks on the first's window rather than crashing." Piggybacking isn't the worst case. The early return skips `swapDismissalHandler`, so the second caller's item never gets a teardown wrapper — and then the *first* bulletin's dismissal runs `teardown()` and nils the window out from under the still-live second bulletin. That is work-item 3 of #1170 verbatim (offline -> `ReachabilityBulletin` -> network error -> `ErrorBulletin`, which the issue calls "not exotic"), and this PR deliberately defers it, so the removed paragraph was the only in-code description of the hazard. This was also raised in the standing review round and is unchanged since — `BulletinOverlayWindow.swift` is byte-identical between f4b3ac9 and the current head (bug due to `OBAKit/BLTNBoard/BulletinOverlayWindow.swift` deleting "the second caller's `dismissalHandler` chain would never be installed ... yanking the window out from under the second bulletin" while keeping the "at worst piggybacks" claim).

https://github.com/OneBusAway/onebusaway-ios/blob/373358d256be8d60965272ef141180391530f3af/OBAKit/BLTNBoard/BulletinOverlayWindow.swift#L57-L65

aaronbrethorst (approved):
The blocker from last round is properly fixed. I re-read the rewritten suite and the new assertion does the thing the old one couldn't:

```swift
page.dismissalHandler?(page)
#expect(page.dismissalHandler == nil, "cycle \(cycle): dismissal should restore the item to the caller's handler, not leave a wrapper behind")
```

Under the old chaining implementation the item is left holding wrapper 1 after the very first pass, so this fails immediately and buries itself one level deeper each cycle. Under snapshot-and-restore it's back to the caller's value. That genuinely separates fix from bug, which the call-count version did not.

You took a different route than the teardown counter I suggested and it's the better one — no new production surface, and the assertion is about observable item state rather than an instrumented side effect. The suite docstring explaining *why* a call count can't distinguish the two implementations is exactly the note that stops someone from "simplifying" this back into a counter later. Starting from an item whose handler is `nil` to sidestep closure-identity comparison is a nice touch, and `isolated deinit { overlay.restoreDismissalHandler() }` correctly handles the singleton being shared across the bundle.

The design doc now matches the shipped code — app-group suite in both the Design and Risks sections, and the four test names line up with the real `@Test` names.

Merging. One thing you missed, which I'd flagged as non-blocking last time and still is:

**The `install` docstring paragraph is still gone.** `BulletinOverlayWindow.swift` is byte-identical to the SHA I reviewed, so the description of a concurrent `install` early-returning without installing the handler chain never came back. The surviving sentence — "at worst piggybacks on the first's window rather than crashing" — understates it: teardown from the first dismissal can yank the window out from under a second bulletin, which is what #1170 work-item 3 calls out as not exotic. The behavior is unchanged and deliberately deferred, so this is purely about not losing the only in-code warning. Worth a sentence in your next PR touching this file.

Two smaller things for the follow-up, no action needed now:

- `ApplicationRootControllerFactory` reads `OBAUseMapPanelExperience` once at launch while `show(in:rootItem:)` now reads it live. A tester who flips the flag *off* mid-session with a map-panel root on screen gets the classic branch presenting into the sheet hierarchy — the clamping bug this PR exists to avoid. The Settings footer says to restart, and it's experimental-flag-only, so the blast radius is tiny. But the Risks bullet claiming "both paths present a valid bulletin — not a broken state" isn't quite true in that direction.
- Still no scene-disconnect teardown: if the hosting `UIWindowScene` disconnects with a bulletin up, the singleton keeps an orphaned window and every later `install` hits the in-use early return, killing bulletins for the rest of the process. Pre-existing from #1163 and now flag-gated, so I'm content to leave it — but it belongs on the #1170 follow-up list.

Thanks for the careful second pass on the tests.

PR #1188, opened 2026-07-14
Refresh all localizations; fix white-label brand leaks and plural rules

aaronbrethorst:
## Summary

Every locale was behind the code, and bringing them up to date surfaced three source-string bugs worth fixing properly.

- **Localization refresh.** `en.lproj` had itself drifted 23 keys behind the Swift source (Live Activities, occupancy chips, VoiceOver labels) and carried 3 dead keys — so a naive key-parity check understated the real gap. Regenerated it, then translated the resulting **147-key gap into all 12 locales** (1,764 translations). Each language was translated twice independently and adjudicated by a third pass against the already-shipped strings, so terminology and register match what's live. Also added the missing Italian `InfoPlist.strings` — Italian users were getting English system permission prompts.

- **iOS says "Settings", not "System Preferences".** `live_activity.error.message` used macOS terminology. Not hypothetical: Arabic, Polish and Vietnamese had faithfully translated the *wrong* term, so this was already user-visible.

- **White-label brand leaks.** OBAKit/OBAKitCore also ship as **KiedyBus**, but 7 user-facing strings hardcoded "OneBusAway" — including the Health permission prompt, which said "OneBusAway" to KiedyBus users. These now take the app name via `%@` + `Bundle.main.appName`. The translation side needed real grammar work rather than a find-and-replace: an arbitrary app name can't be inflected, agreed with, or take a particle. That surfaced **18 further latent bugs across 8 locales** (pt-BR articles, it/fr participle agreement, ru/pl case endings, ko particles, fil ligatures). A new SwiftLint rule bans the literal in the frameworks.

- **Plural rules.** Count-bearing strings used a bare `%d`, which is ungrammatical for most counts in Arabic, Russian and Polish. Added `Localizable.stringsdict` for 8 such keys across 13 locales with real CLDR categories — Arabic uses all six, including the dual. This let me delete three hand-rolled `if count == 1` branches that only ever encoded English's one/other split.

## Test plan

- [x] `xcodebuild build` — clean, both OneBusAway and KiedyBus targets.
- [x] Full suite: **1,268 tests, 0 failures** (1,264 existing + 4 new).
- [x] **Plural selection verified on-device** via `xcodebuild -testLanguage ru|pl|ar`, which is the only way to exercise real CLDR rules:
  - ru: `1 остановка` · `2 остановки` · `5 остановок` · `21 остановка` — the one/few/many cycle, including the 21→*one* wraparound.
  - pl: `1 przystanek` · `2 przystanki` · `5 przystanków` — correct 2-4 vs 5+ split.
  - ar: `موقف واحد` · `موقفان` (dual) · `5 مواقف` — all six categories.
- [x] Confirmed `String(format:)` and `localizedStringWithFormat` agree in every locale at every count (a review pass claimed the stringsdict entries were inert; they are not — CFBundle bakes the source localization's plural rules into the returned format).
- [x] Launched the app in the simulator in Russian; localization loads and renders.
- [x] Key parity + format-specifier parity across all 645 keys × 12 locales: zero mismatches.
- [x] All 26 `.strings` and 13 `.stringsdict` files lint as valid plists, UTF-8, no duplicate keys.
- [x] Verified the new `testEveryLocaleStringsdictIsWellFormed` guard actually **fails** when a `.stringsdict` is removed — it isn't decorative.
- [x] SwiftLint clean, and the new `hardcoded_app_name` rule fires on a real violation.

## Review notes

Non-blocking, deliberately left out of scope — happy to file follow-ups:

1. **`DonationLearnMoreView` is structurally un-localizable.** Every string is a bare SwiftUI `Text` literal, and `Text(LocalizedStringKey)` resolves against `Bundle.main` while OBAKit's strings live in the *framework* bundle — so they never resolve and ship in English in all 13 locales. I fixed the brand leak there and documented the trap, but actually localizing it needs an `OBALoc`/`Text(_:bundle:)` migration affecting ~30 call sites.
2. **`donations_dismiss_alert.message`** reads "%@ is a volunteer-run organization…" — the placeholder is the *app* name but the sentence claims something about the *organization* (OTSF). Under KiedyBus it asserts KiedyBus is a volunteer-run org. Fixing means rewriting the English copy and re-translating in 12 locales; it's a product/copy call, so I left it.
3. **SwiftLint is skipped in CI** (`scripts/swiftlint.sh` early-exits on `$CI`, and no workflow invokes it), so the new rule is advisory locally rather than a merge gate. The new `LocalizationTests` *do* run in CI and cover the parity risk.
4. **~44 other `%d` keys still lack plural rules** (`formatters.deviation.*`, the minute strings). `formatters.time_fmt` renders "1 minutes" even in English.
5. **Health usage descriptions aren't localized in any locale** (pre-existing) — they can be now that they're per-app.
6. **`Bundle.main.appName` force-unwraps `CFBundleDisplayName`.** Safe for every current target, but `NetworkOperation` builds an *error message* with it, so a future extension missing the key would trap while reporting an error.
7. **`UmamiAnalytics`** hardcodes `"OneBusAway/"` as the user agent, so KiedyBus traffic reports as OneBusAway. Left alone because changing it affects analytics continuity.

## Summary by CodeRabbit

* **New Features**
  * Expanded localized UI for Live Activities, surveys, weather, route selection, transfer-arrival messaging, and richer stop/details.
  * Added Health app walking-speed permission descriptions.

* **Improvements**
  * The app name now appears dynamically across donation, feedback, and location-services alerts.
  * Enhanced stop-page and accessibility labels, including improved “show earlier departures” and service-alert phrasing.
  * Updated Live Activities guidance and refined permission/alert messaging.

* **Bug Fixes**
  * Fixed/updated Italian localization issues (including regional URL text) and improved localization consistency across locales.


aaronbrethorst:
Pushed two follow-up commits:

**`Incorporate PR #1188 review feedback`** — applied the actionable CodeRabbit findings: es `dónde tu localization` → `su ubicación`; fil `%dm` → `%d min` (was ambiguous with meters); it `Mostra testo avvisi` → `Mostra avvisi di test`; zh-Hans dropped the hardcoded 英里/分钟 units from `walk_time_view` (the `%$@` placeholders already carry unit-bearing values, matching en/zh-Hant); and the `force_try` SwiftLint suppression on the test regex. Left `NSHealthUpdateUsageDescription` in place (declined the removal suggestion).

**`Localize HealthKit Info.plist purpose strings`** — the HealthKit purpose strings are now localized in every supported language. Extracted the two keys out of both `project.yml` files into a new shared include `Apps/Shared/info_plist.yml`, gave the base value an **app-neutral** body (iOS puts the app name in the dialog title, matching the existing `NSLocation*` convention), and added translations to each `*.lproj/InfoPlist.strings` (OneBusAway: 12 locales; KiedyBus: Polish). Each translation reuses the localized Health-app name / walking-speed terminology already shipped in `settings_controller.walking_speed.*`.

Verified: both OneBusAway and KiedyBus build clean; the built bundles carry the localized keys in every `.lproj` (spot-checked fr/ru/zh-Hans/pl); all `.strings` lint as valid plists; LocalizationTests pass.

**Non-blocking design note:** shared Info.plist purpose strings now live in two files — `app_shared.yml` (`NSLocation*`) and the new `info_plist.yml` (`NSHealth*`). This was a deliberate split (a dedicated `info_plist.yml` was requested); a breadcrumb comment in `app_shared.yml` points at it. If preferred, the `NSLocation*`/ATS keys could later move into `info_plist.yml` so the split is by concern rather than leaving two files that both hold Info.plist properties.

PR #1191, opened 2026-07-16
Rewrite Bookmarks tab in SwiftUI; improve custom regions

aaronbrethorst:
## Summary
- **Bookmarks tab rewritten in SwiftUI**: `BookmarksListView` + `BookmarkCardView` replace the `OBAListView`-based screen. Trip bookmarks render as cards matching the Live Activity lock-screen card (route badge, name, time · adherence, countdown, upcoming-departure chips); whole-stop bookmarks get a map-pin row with served routes. Collapsible sections keep the legacy UserDefaults key/encoding and section-ID vocabulary, so existing users' collapse state survives. `BookmarksViewController` shrinks to a hosting shell that owns UIKit chrome, modals, and the Live Activity lifecycle.
- **Custom regions improvements** (first commit): collapsed map preview on Choose Region, current-location centering, no more Contact Email field, Base URL clarification + `https://` assumption + live server validation via `RegionsService.fetchAgenciesWithCoverage(baseURL:)`, plain Name field with agency-based fallback naming, and a fix for editing a region prefilling only the URL path. Review follow-ups: `/api/where` suffix stripping (help text promised it), `@MainActor` on `doSave`/`doDelete` (they mutate `@State`), orphaned code/strings removal.
- **Xcode 27 toolchain fixes**: bumped `swift-http-types` (1.5.1→1.6.0), `swift-openapi-urlsession` (1.2.0→1.3.1), and `swift-openapi-runtime` (1.9.0→1.12.0) — the old pins fail to compile under Swift 6.4.
- **Robustness fixes found in review**: `BookmarkDataLoader.cancelUpdates()` now closes out the retired batch so `isLoading` can't stay stuck true (would hang pull-to-refresh awaiters); the Track context-menu item requires loaded arrival data instead of silently no-oping; card change-flashes survive unrelated per-bookmark rebuilds; new shared `Error.isCancellation` helper replaces duplicated cancellation checks and `TripViewModel` clears its spinner on out-of-band URLSession cancellation.

## Test plan
- [x] Built with Xcode 27 beta 3 (Swift 6.4) for iPhone 17 Pro simulator
- [x] Full unit suite: **1279 tests, 0 failures** (12 new tests: section building/legacy section-ID pinning, region filtering, collapse-state persistence round-trip, row equality contract, stop-bookmark arrivals clamp, `Error.isCancellation`, Base URL normalization)
- [x] On-simulator verification of the Bookmarks tab: empty state, bookmark creation from a stop page, card rendering (badge/countdown/chips/adherence), section collapse + persistence across backgrounding, context menu (Track/Edit/Delete + stop-page preview), row tap → stop page → back, pull-to-refresh (spinner, live data update, no crash)
- [x] Re-verified after simplification/review fixes: rebuild, full suite, render smoke test, tap-through navigation

## Review notes (non-blocking, surfaced by the multi-agent review)
- **Duplication debt**: the card header, departure pill, and badge silhouette now exist in three places (`BookmarkCardView`, `GroupedListView`, `TripLiveActivityCardView`), compounding the pre-existing `RouteBadgeView`/`CountdownView` duplication between OBAKit and OBAKitCore. Worth a dedicated extraction pass.
- **Custom-region validation** uses an empty API key when no region is active yet (first-run onboarding), so key-enforcing servers would fail validation for a correct URL; it also accepts a server whose agency list is empty. Both are product-level calls.
- The deep-link `add-region` path still duplicates region validation/creation logic with different rigor than the new form; consolidating into one `RegionsService` operation would close the gap.
- Per-fetch `dataLoaderDidUpdate` fan-out means N section rebuilds per refresh batch (progressive fill-in is arguably intentional; the equality-gated publish now suppresses the redundant SwiftUI diffs).
- New `OBALoc` keys aren't yet extracted into the in-repo `.strings` catalogs (`scripts/extract_strings`); English fallbacks ship in code, so runtime is unaffected.

## Summary by CodeRabbit

* **New Features**
  * Bookmarks now use a refreshed SwiftUI experience with collapsible/sortable sections, pull-to-refresh, improved previews, and optional Live Activity tracking for trips.
  * Custom regions now use a Base URL builder with automatic URL normalization and server validation, plus location-aware setup.
  * Empty-state screens can now show an optional description.

* **Bug Fixes**
  * Improved refresh/cancellation handling to avoid stuck loading states.
  * Live Activity matching/reconciliation is more reliable, and bookmark highlighting/countdowns are more accessible.

* **Refactor**
  * Migrated the Bookmarks screen to a SwiftUI-hosting structure and simplified legacy list/navigation behaviors.

* **Tests**
  * Added coverage for cancellation/error handling, region URL normalization, and bookmarks section/collapse logic.


aaronbrethorst:
Addressed all six actionable comments in 6220c8a1:

- **Perpetual "Loading..." on empty results**: `BookmarkDataLoader` now tracks which stops have fetched successfully; cards render "No upcoming departures" once data lands.
- **Batch-scoped pull-to-refresh**: new `loadDataAndWait()` awaits the specific batch it started (retired batches resume their awaiters), replacing the `$isLoading` subscription.
- **Empty state for other-region-only bookmarks**: `emptyState` no longer consults the global bookmark collection.
- **Distance sort after late location fix**: the view model now observes `LocationServiceDelegate.locationChanged` and rebuilds (equality-gated).
- **Live Activity reconciliation**: creation and matching now share one normalized key helper.
- **`routeShortName` restored to snapshot equality.**

One deliberate decline: the `[weak self]` suggestion in `cancelUpdates()` — the teardown task must run even if the owner released the loader, because it resumes any suspended `loadDataAndWait()` continuations; a weak capture could leak them. Noted in a comment at the capture site.

Full suite: 1281 tests, 0 failures.

PR #1200, opened 2026-07-18
Rebuild onboarding as a versioned step registry with a push-notifications step

aaronbrethorst:
## Summary

- Replaces the per-app `OnboardingNavigationController` (hardcoded page transitions, duplicated in OneBusAway and KiedyBus) with a shared, registry-driven onboarding flow in OBAKit: each step declares an id, weight, version, and eligibility predicate, and the flow shown to any user is just the registry filtered to steps they haven't seen, sorted by weight. Adding a future step is one registry entry + one view case.
- Adds five redesigned SwiftUI screens (Welcome, Location priming, Region with auto-detected card + live map preview, **Push notifications**, All set recap) sharing one scaffold (progress segments, hero circle, button dock), light/dark aware, white-label safe (`Bundle.main.appName`, theme accent color). The existing `DataMigrationView` and full `RegionPickerView` are reused, not rebuilt.
- **The headline behavior:** existing users are backfilled as having seen the legacy steps — but never the notifications step — so anyone who has never seen the push pitch gets exactly that one screen at next launch (NEW badge, "Maybe Later"), once, ever, whether they've used the app 0 days or 18 years. Declining counts as seen. "Turn On Notifications" requests OS authorization + APNs registration only (no new subscription logic; deliberately calls `UNUserNotificationCenter` directly because `PushService.pushID()`'s callback never fires on denial).
- Launch glue is now shared (`OBARootInterfaceLauncher` in CommonClient); KiedyBus gains onboarding for the first time (its old copy was dead code). Cold-launch stop deep links are stashed in a unified `pendingStopID` and drained on activation instead of being dropped while the async flow evaluation installs the root.
- Design doc + implementation plan committed under `docs/superpowers/`; design source: Claude Design "OBA Onboarding Rethink".

## Test plan

- [x] Unit suite 1306/1306 green (54 new onboarding tests: registry math, eligibility gates, backfill semantics, version bumps, Allow-Once reversion, store round-trip/corruption, registry uniqueness, weight sort, `evaluate()` nil-for-existing-user / controller-for-new-user, pendingStopID stash)
- [x] SwiftLint 0 violations; zero new warnings (Swift 6 concurrency ratchet enforcing); OneBusAway + KiedyBus both build
- [x] Simulator (iPhone 17 Pro): full forced flow (`TEST_ONBOARDING=1`) through all six steps in weight order, light + dark mode, hand-off to map with flip transition
- [x] Simulator: existing-user path — seen-store backfilled to exactly `{welcome, location, region, done}`, app goes straight to the map; smoke launch re-verified after the launch-glue refactor
- [ ] Device pass: the real single-step notifications launch for existing users (push provider isn't configurable on Simulator) and an end-to-end push-token registration

## Review notes

Non-blocking findings from the multi-agent review, deliberately deferred:

- **Priming-screen HIG deviation (deliberate):** location and notifications keep two buttons ("Not Now"/"Maybe Later") against the HIG's one-button pre-alert guidance; declining-marks-seen-forever was judged more user-respecting. Low-but-nonzero App Review risk, recorded in the spec.
- **pendingStopID drain timing:** the drain runs on `sceneDidBecomeActive`, which on a cold launch usually fires before the async root install completes — a cold-launch widget tap may not navigate until the next foreground cycle. Same latent shape as the pre-existing `pendingAlarmStopID` pattern; a proper fix (drain on root-install) is a follow-up.
- **Rootless-window watchdog:** the window briefly has no root VC while `evaluate` awaits notification settings (skipped entirely when no push provider exists). If the XPC call ever wedged, recovery is only a relaunch; a placeholder-root or timeout fallback is a possible hardening follow-up (a breadcrumb log of the computed flow was added for field diagnosis).
- **Machine translations:** the 45 new strings were machine-translated into all 12 non-English locales to satisfy the key-parity test — needs a native-speaker pass (especially ar/ko/vi/zh; fr/pl have gender-agreement nits).
- **Migration-before-region ordering:** a restored-from-backup fresh install can hit the migrator's "No current region is set" error two steps before the region screen — pre-existing ordering, worth revisiting.
- **Region screen retry:** if the initial regions refresh fails, the error is alerted + logged but there's no explicit retry affordance (the bundled-regions fallback list keeps the screen usable).
- CLAUDE.md still references an `Apps/MTA` variant that doesn't exist in the tree.

## Summary by CodeRabbit

* **New Features**
  * Added a refreshed onboarding experience (welcome → region selection → location access → service alerts) with a final recap screen.
  * Onboarding now adapts to user status and remembers progress across launches.
  * Improved consistency of onboarding start and root screen presentation across apps.
  * Enhanced custom-region setup guidance and validation messaging.
* **Bug Fixes**
  * Stop “view” deep links received during cold launch are preserved and opened once the interface is ready.
  * Bookmarks now clearly explain when there are no upcoming departures.
* **Localization**
  * Added/updated onboarding and custom-region strings across supported languages.


aaronbrethorst:
Addressed the CodeRabbit review in 0dab19b0:

- **Cold-launch drain (Application.swift)** — implemented the suggested shape: pending presentations (stop navigation, donation prompt, add-region alert) now drain from a new `Application.rootUserInterfaceDidLoad()`, invoked by the shared `OBARootInterfaceLauncher` on both the onboarding-finished path (in the flip transition's completion) and the no-onboarding path. The `applicationDidBecomeActive` drain is retained for later foregrounds but now skips while the onboarding flow is the window's root, so a deferred navigation can never yank the user out of onboarding. One correction to the finding's framing: the stashes were deferred to the next foreground cycle, not permanently lost — the drain guard left them intact — but the deterministic post-root drain is the right fix either way.
- **Flip transition (OBARootInterfaceLauncher.m)** — `showRootController()` now runs inside the `animations:` block per the suggestion.
- **Spec fenced code block** — language annotation added.
- **`hasCurrentRegion` "compile error" (OnboardingRegistryTests.swift)** — skipped as stale: that parameter was deliberately removed from `OnboardingEnvironment` in 6569551f (it was dead — no predicate read it), and the test helper matches the current five-parameter initializer. The suite builds and passes at HEAD (1306/1306).

Shapy-Sees:
Heads-up that af566d1 fixed more than cold launch: on the released 26.2.2, **every** `view-stop` link is silently dropped — URL delivery happens while the scene is still `.foregroundInactive`, so `topViewController` resolves nil and the guard eats the link. Widget bookmark taps included.

Verified on-device (iPhone 17 Pro, iOS 26.5.2): on 26.2.2, warm, cold, and widget taps all fail; on the 26.3.0 TestFlight build all three navigate correctly.

Related: #1219 fixes the URL-scheme doc comment, whose example shows parameter names the parser rejects.

PR #1209, opened 2026-07-20
Fix main-thread hang warning from CLLocationManager.locationServicesEnabled()

mosliem:
## Problem

The debugger surfaced this warning:

> This method can cause UI unresponsiveness if invoked on the main thread. Instead, consider waiting for the `-locationManagerDidChangeAuthorization:` callback and checking `authorizationStatus` first.

It traced to `LocationManager.isLocationServicesEnabled`, which called `CLLocationManager.locationServicesEnabled()` — a **synchronous, blocking XPC round-trip to `locationd`**. That call was hit on the main thread every time `LocationService.isLocationUseAuthorized` was evaluated (~10 UI call sites: `MapViewController`, `MapRegionManager`, `TripViewController`, `MapPanelRootView`, `OnboardingFlowView`, `Application`, plus internally on every authorization change).

## Fix

Follow Apple's recommended pattern: stop proactively polling the global toggle and rely on `authorizationStatus`, letting a failed location start report the "services off" case via the delegate.

1. **Removed** `isLocationServicesEnabled` from the `LocationManager` protocol and its `CLLocationManager.locationServicesEnabled()` implementation.
2. **`isLocationUseAuthorized`** is now derived from `authorizationStatus`.
3. **Handle `CLError.denied`** — because iOS keeps a per-app status of `.authorizedWhenInUse` even when Location Services is off system-wide, the service now latches a `locationServicesDenied` flag when Core Location reports a `denied` error, folds it into `isLocationUseAuthorized`, stops updates (per Apple guidance), and re-notifies delegates so the UI hides the locate button and user dot. The latch clears on the next authorization callback, so access recovers automatically.

## UI behavior (no consumer changes required)

- `MapRegionManager` turns off `showsUserLocation` on the auth-change notification.
- `MapViewModel`'s `@Published locationAuthStatus` re-fires (no dedupe), driving `MapViewController.renderMapStatus()` to hide the locate button.

## Tests

Added to `LocationServiceTests`:
- `test_deniedError_marksLocationUnavailable` — denied error → unavailable, delegates re-notified, updates stopped, raw error still forwarded.
- `test_deniedError_clearedByAuthorizationChange` — a later auth callback restores availability.
- `test_nonDeniedError_doesNotMarkUnavailable` — transient `locationUnknown` does not disable location.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved location permission handling using an effective authorization state that respects system-wide Location Services being off, including a persisted “denied” latch across launches.
  * Location updates now stop reliably when moving to the background.
  * Added a foreground retry so location availability can automatically recover after Location Services are re-enabled.
* **Tests**
  * Expanded coverage for denied-latch persistence, relaunch/foreground recovery, and correct behavior for non-denial errors.
  * Updated mocks and scenarios to match the new effective authorization behavior.

### DIFF
--- a/OBAKitCore/Location/Location/LocationService.swift
+++ b/OBAKitCore/Location/Location/LocationService.swift
@@ -148,6 +148,21 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
         }
     }
 
+    /// Latched when Core Location reports a `denied` error. This can happen even
+    /// while `authorizationStatus` still reads as authorized — most commonly when
+    /// the user switches Location Services off system-wide, which does not change
+    /// the app's per-app authorization. While set, `isLocationUseAuthorized`
+    /// reports `false` so the UI hides location affordances (locate button, user
+    /// dot) that would otherwise appear active but never produce a fix. It is
+    /// cleared on the next authorization callback, when the user may have
+    /// re-enabled access.
+    private var locationServicesDenied = false {
+        didSet {
+            guard locationServicesDenied != oldValue else { return }
+            notifyDelegatesAuthorizationChanged(authorizationStatus)
+        }
+    }
+
     /// This is true when the app is in a state such that the user can/should be
     /// prompted for location services authorization. In other words: the app has
     /// not been denied or approved, and the user also has not generally restricted
@@ -181,9 +196,16 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
     }
 
     /// Answers the question of whether the device GPS can be consulted for location data.
+    ///
+    /// This is derived from `authorizationStatus` plus the `locationServicesDenied`
+    /// latch. We deliberately avoid `CLLocationManager.locationServicesEnabled()`
+    /// because it performs a blocking, synchronous XPC call that Apple warns can
+    /// hang the main thread. When location services are disabled system-wide,
+    /// attempts to start updates fail via `locationManager(_:didFailWithError:)`
+    /// with a `denied` error, which we latch here — the pattern Apple recommends.
     public var isLocationUseAuthorized: Bool {
-        return locationManager.isLocationServicesEnabled &&
-            (authorizationStatus == .authorizedWhenInUse || authorizationStatus == .authorizedAlways)
+        guard !locationServicesDenied else { return false }
+        return authorizationStatus == .authorizedWhenInUse || authorizationStatus == .authorizedAlways
     }
 
     @available(iOS 14, *)
@@ -249,11 +271,17 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
 
     @available(iOS, introduced: 4.2, deprecated: 14.0)
     public func locationManager(_ manager: CLLocationManager, didChangeAuthorization status: CLAuthorizationStatus) {
+        // A fresh authorization callback means access was re-evaluated, so clear
+        // any latched denial before recomputing derived state.
+        locationServicesDenied = false
         authorizationStatus = status
     }
 
     @available(iOS 14, *)
     public func locationManagerDidChangeAuthorization(_ manager: CLLocationManager) {
+        // A fresh authorization callback means access was re-evaluated, so clear
+        // any latched denial before recomputing derived state.
+        locationServicesDenied = false
         authorizationStatus = manager.authorizationStatus
 
         // Accuracy can change while the coarse status stays put (e.g. "Allow
@@ -297,6 +325,16 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
     }
 
     public func locationManager(_ manager: CLLocationManager, didFailWithError error: Error) {
+        // A `denied` error means location is currently unavailable even if
+        // `authorizationStatus` still reads as authorized (e.g. Location Services
+        // switched off system-wide). Apple recommends stopping updates in this
+        // case; we also latch the state so `isLocationUseAuthorized` reports the
+        // location as unusable and the UI can hide its location affordances.
+        if let clError = error as? CLError, clError.code == .denied {
+            locationManager.stopUpdatingLocation()
+            locationServicesDenied = true
+        }
+
         notifyDelegatesErrorReceived(error)
     }
 
--- a/OBAKitCore/Location/Location/LocationService.swift
+++ b/OBAKitCore/Location/Location/LocationService.swift
@@ -160,6 +160,16 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
         didSet {
             guard locationServicesDenied != oldValue else { return }
             notifyDelegatesAuthorizationChanged(authorizationStatus)
+
+            // When the latch clears while the app is still authorized, resume
+            // updates ourselves. The `authorizationStatus` didSet only restarts
+            // them when the coarse status *value* changes, which it often does
+            // not on recovery (e.g. it stays `.authorizedWhenInUse`), so relying
+            // on it alone would leave the stream stopped until the next app
+            // foreground.
+            if !locationServicesDenied, isLocationUseAuthorized {
+                startUpdates()
+            }
         }
     }
 
@@ -269,19 +279,23 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
 
     // MARK: - Delegate
 
+    /// A fresh authorization callback means access was re-evaluated by the user
+    /// or system, so any latched `denied` state is stale. Clearing it before we
+    /// recompute `authorizationStatus` lets derived state (and any resumed
+    /// updates) reflect the new reality.
+    private func resetDeniedLatchForAuthorizationChange() {
+        locationServicesDenied = false
+    }
+
     @available(iOS, introduced: 4.2, deprecated: 14.0)
     public func locationManager(_ manager: CLLocationManager, didChangeAuthorization status: CLAuthorizationStatus) {
-        // A fresh authorization callback means access was re-evaluated, so clear
-        // any latched denial before recomputing derived state.
-        locationServicesDenied = false
+        resetDeniedLatchForAuthorizationChange()
         authorizationStatus = status
     }
 
     @available(iOS 14, *)
     public func locationManagerDidChangeAuthorization(_ manager: CLLocationManager) {
-        // A fresh authorization callback means access was re-evaluated, so clear
-        // any latched denial before recomputing derived state.
-        locationServicesDenied = false
+        resetDeniedLatchForAuthorizationChange()
         authorizationStatus = manager.authorizationStatus
 
         // Accuracy can change while the coarse status stays put (e.g. "Allow
@@ -330,8 +344,11 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
         // switched off system-wide). Apple recommends stopping updates in this
         // case; we also latch the state so `isLocationUseAuthorized` reports the
         // location as unusable and the UI can hide its location affordances.
+        //
+        // Stop first (while still authorized, so the guards in `stopUpdates()`
+        // pass and both location and heading are torn down), then latch.
         if let clError = error as? CLError, clError.code == .denied {
-            locationManager.stopUpdatingLocation()
+            stopUpdates()
             locationServicesDenied = true
         }
 
--- a/OBAKitCore/Location/Location/LocationService.swift
+++ b/OBAKitCore/Location/Location/LocationService.swift
@@ -304,8 +304,15 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
 
     // MARK: - Heading
 
+    /// Guarded on authorization as well as availability. `isHeadingAvailable`
+    /// wraps `CLLocationManager.headingAvailable()`, which reports whether the
+    /// *device* has a magnetometer — it says nothing about whether we may use it.
+    /// Without the authorization guard, `startUpdates()` would start heading even
+    /// when starting location had just synchronously latched a `denied` error and
+    /// torn the manager down, leaving the magnetometer powered for the rest of
+    /// the process.
     public func startUpdatingHeading() {
-        guard locationManager.isHeadingAvailable else {
+        guard isLocationUseAuthorized, locationManager.isHeadingAvailable else {
             return
         }
 
@@ -336,16 +343,32 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
     /// The read is a blocking XPC call, hence off the main thread; the result is
     /// applied back on the main actor. It reconciles both directions, so a single
     /// call covers "services came back on" and "services are off."
+    /// The in-flight probe, if any.
+    ///
+    /// Probes are single-flight by cancel-and-replace. Three call sites can start
+    /// one — the authorization callback, `retryIfLocationServicesDenied()`, and
+    /// foregrounding by way of the latter — and unstructured tasks have no
+    /// ordering guarantee at the `await`, so an older read could land after a
+    /// newer one and clobber the latch with stale state. Cancelling the previous
+    /// probe makes the newest read the only one that applies.
+    private var servicesProbeTask: Task<Void, Never>?
+
     private func refreshLocationServicesEnabled() {
         // The latch is only meaningful while the app is itself authorized; when it
         // isn't, `authorizationStatus` reports the raw status directly and there is
         // nothing to reconcile.
         guard isPerAppAuthorized else { return }
 
-        Task { [weak self] in
+        servicesProbeTask?.cancel()
+        servicesProbeTask = Task { [weak self] in
             guard let self else { return }
             let enabled = await self.locationManager.locationServicesEnabled()
-            guard self.isPerAppAuthorized else { return }
+
+            // Cancellation cannot interrupt the blocking read itself, so a
+            // superseded probe resumes here with an answer it must not apply.
+            guard !Task.isCancelled, self.isPerAppAuthorized else { return }
+
+            self.servicesProbeTask = nil
             self.applyAuthorizationState(rawStatus: self.rawAuthorizationStatus, servicesDenied: !enabled)
         }
     }
--- a/OBAKitCore/Location/Location/LocationService.swift
+++ b/OBAKitCore/Location/Location/LocationService.swift
@@ -164,7 +164,7 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
     }
 
     private var isPerAppAuthorized: Bool {
-        rawAuthorizationStatus == .authorizedWhenInUse || rawAuthorizationStatus == .authorizedAlways
+        rawAuthorizationStatus.isAuthorized
     }
 
     /// The current *effective* authorization state of the app.
@@ -182,32 +182,40 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
         return .denied
     }
 
-    /// The single funnel for both inputs to `authorizationStatus`. It applies
-    /// the new values and *then* reacts once to the effective transition.
+    /// The single funnel for both inputs to `authorizationStatus`. It applies the
+    /// new values, reconciles the manager's running state to match, and notifies
+    /// delegates once on an effective transition.
     ///
     /// Property observers on the two inputs can't do this correctly: whichever
-    /// one is assigned first sees the other's stale value, and both firing
-    /// means delegates get notified twice for one logical change.
+    /// one is assigned first sees the other's stale value, and both firing means
+    /// delegates get notified twice for one logical change.
+    ///
+    /// Reconciliation is unconditional (not gated on the transition) and relies on
+    /// `startUpdates()`/`stopUpdates()` being idempotent. That is deliberate: Core
+    /// Location fires an authorization callback carrying an *unchanged* status the
+    /// moment the delegate is assigned, and an already-authorized app must start
+    /// its first fix from it. Reconciling here means there is a single place that
+    /// starts updates — no separate post-funnel start path that would fire a
+    /// second time on a real grant (and, with services off, produce two failed
+    /// probes and two error callbacks).
     private func applyAuthorizationState(rawStatus: CLAuthorizationStatus, servicesDenied: Bool) {
         let oldStatus = authorizationStatus
-        let wasAuthorized = isLocationUseAuthorized
 
         rawAuthorizationStatus = rawStatus
         locationServicesDenied = servicesDenied
 
-        guard authorizationStatus != oldStatus else { return }
-
-        notifyDelegatesAuthorizationChanged(authorizationStatus)
-
+        // Reconcile running state to the effective authorization. When access is
+        // revoked this is what tears the manager down — nobody else will, and a
+        // `CLLocationManager` left running keeps the location-usage indicator lit
+        // and the magnetometer powered for the rest of the process.
         if isLocationUseAuthorized {
             startUpdates()
-        } else if wasAuthorized {
-            // Access was revoked. Tear the manager down here — nobody else will,
-            // and a `CLLocationManager` left running keeps the location-usage
-            // indicator lit and the magnetometer powered for the rest of the
-            // process.
+        } else {
             stopUpdates()
         }
+
+        guard authorizationStatus != oldStatus else { return }
+        notifyDelegatesAuthorizationChanged(authorizationStatus)
     }
 
     /// This is true when the app is in a state such that the user can/should be
@@ -251,7 +259,7 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
     /// updates fail via `locationManager(_:didFailWithError:)` with a `denied`
     /// error, which we latch — the pattern Apple recommends.
     public var isLocationUseAuthorized: Bool {
-        return authorizationStatus == .authorizedWhenInUse || authorizationStatus == .authorizedAlways
+        return authorizationStatus.isAuthorized
     }
 
     @available(iOS 14, *)
@@ -314,61 +322,79 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
 
     // MARK: - Delegate
 
-    /// Attempts to resume location updates after a latched `denied` error.
+    /// Reconciles the denied latch against the system-wide Location Services
+    /// switch, read off the main thread.
+    ///
+    /// This is the authoritative signal the latch approximates. `didFailWithError`
+    /// (a `denied` error) and `didUpdateLocations` (a fix) are the async proxies
+    /// we also honor between probes, but each has a blind spot: toggling Location
+    /// Services back on delivers no authorization callback and, indoors, no fix
+    /// either — so a latch cleared only by those signals could stay stuck for a
+    /// whole session. Reading the real switch recovers regardless, and latches a
+    /// services-off launch without waiting for the first `denied` error to arrive.
+    ///
+    /// The read is a blocking XPC call, hence off the main thread; the result is
+    /// applied back on the main actor. It reconciles both directions, so a single
+    /// call covers "services came back on" and "services are off."
+    private func refreshLocationServicesEnabled() {
+        // The latch is only meaningful while the app is itself authorized; when it
+        // isn't, `authorizationStatus` reports the raw status directly and there is
+        // nothing to reconcile.
+        guard isPerAppAuthorized else { return }
+
+        Task { [weak self] in
+            guard let self else { return }
+            let enabled = await self.locationManager.locationServicesEnabled()
+            guard self.isPerAppAuthorized else { return }
+            self.applyAuthorizationState(rawStatus: self.rawAuthorizationStatus, servicesDenied: !enabled)
+        }
+    }
+
+    /// Attempts to recover from a latched `denied` error, e.g. when the app
+    /// returns to the foreground after the user visited Settings.
     ///
     /// Toggling Location Services back on system-wide does not change the app's
     /// per-app authorization, so Core Location may deliver no authorization
     /// callback at all — nothing would otherwise clear the latch, and the locate
-    /// button and user dot would stay hidden until the app was force-quit. The
-    /// app calls this when it returns to the foreground, which is where the user
-    /// lands after visiting Settings.
+    /// button and user dot would stay hidden until the app was force-quit.
     ///
-    /// This probes rather than assumes: the latch stays set until we have an
-    /// answer, so the UI does not flash location affordances on and back off.
-    /// A fix clears the latch (see `locationManager(_:didUpdateLocations:)`);
-    /// another `denied` error leaves it exactly as it was.
+    /// Recovery goes through the off-main `locationServicesEnabled` probe rather
+    /// than optimistically restarting updates: the latch stays set until the probe
+    /// answers, so the UI never flashes location affordances on and back off, and
+    /// the probe clears the latch even when no GPS fix will arrive (indoors).
     public func retryIfLocationServicesDenied() {
         guard locationServicesDenied, isPerAppAuthorized else { return }
-        locationManager.startUpdatingLocation()
+        refreshLocationServicesEnabled()
     }
 
     @available(iOS 14, *)
     public func locationManagerDidChangeAuthorization(_ manager: CLLocationManager) {
-        // A *change* of per-app authorization means the user just made a fresh
-        // decision, so any latched `denied` state is stale evidence and we
-        // re-arm optimistically. A callback that leaves the status untouched
-        // carries no new information — Core Location fires one whenever a
-        // delegate is assigned — so it must not clear the latch, or every
-        // spurious callback would restart updates against a location subsystem
-        // that is still off.
-        //
         // Read the status from the injected manager, not from `manager`, so the
         // path under test is the path that ships.
+        //
+        // Clear the latch only when the transition crosses *into* authorization:
+        // a fresh grant (e.g. `.notDetermined`/`.denied` → `.authorizedWhenInUse`)
+        // is genuinely new evidence, so re-arm optimistically. A change between two
+        // authorized values (WhenInUse → Always) does *not* cross in, so it can't
+        // clear a latch while services are still off — that would flicker the
+        // locate button and user dot. The system-wide toggle, which delivers no
+        // authorization callback at all, is handled by the off-main probe below.
         let newStatus = locationManager.authorizationStatus
-        let statusChanged = newStatus != rawAuthorizationStatus
-        applyAuthorizationState(rawStatus: newStatus, servicesDenied: statusChanged ? false : locationServicesDenied)
+        let crossesIntoAuthorization = !rawAuthorizationStatus.isAuthorized && newStatus.isAuthorized
+        applyAuthorizationState(rawStatus: newStatus, servicesDenied: crossesIntoAuthorization ? false : locationServicesDenied)
 
-        // Core Location delivers a callback as soon as the delegate is assigned,
-        // carrying an *unchanged* status. That is the app's earliest signal that
-        // it is authorized, and it is where location updates have always been
-        // started from — early enough that the first fix lands before the map
-        // settles. `applyAuthorizationState` deliberately reacts only to
-        // transitions, so starting updates has to happen outside of it or a
-        // relaunch of an already-authorized app would wait for the next
-        // foreground. Both calls below are idempotent.
-        if isLocationUseAuthorized {
-            startUpdates()
-        } else {
-            // Same reasoning for a latch seeded from the previous launch: probe
-            // it now rather than making the user wait for a foreground event.
-            retryIfLocationServicesDenied()
-        }
+        // Reconcile the latch against the system-wide switch. Core Location fires
+        // this callback the moment the delegate is assigned, so this is also where
+        // a latch seeded from the previous launch — or a fresh, services-off
+        // authorization — is resolved, without waiting for a foreground event.
+        refreshLocationServicesEnabled()
 
         // Accuracy can change while the coarse status stays put (e.g. "Allow
         // Once" elevates a reduced-accuracy session to full). That leaves the
-        // `authorizationStatus` didSet silent, so detect and forward the
-        // accuracy transition on its own channel.
-        let newAccuracy = manager.accuracyAuthorization
+        // effective `authorizationStatus` unchanged, so detect and forward the
+        // accuracy transition on its own channel — reading from the injected
+        // manager, for the same testability reason as the status above.
+        let newAccuracy = locationManager.accuracyAuthorization
         if newAccuracy != lastAccuracyAuthorization {
             lastAccuracyAuthorization = newAccuracy
             notifyDelegatesAccuracyAuthorizationChanged(newAccuracy)
@@ -413,13 +439,17 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
         // A `denied` error means location is currently unavailable even if the
         // per-app authorization still reads as authorized (e.g. Location
         // Services switched off system-wide). Latching it flips the effective
-        // `authorizationStatus` to `.denied`, which stops updates as Apple
-        // recommends and lets the UI explain the situation.
-        if let clError = error as? CLError, clError.code == .denied {
-            // Unconditional, as Apple recommends: a probe that re-confirms an
-            // already-latched denial must still leave the manager stopped, and
-            // `applyAuthorizationState` only reacts to state *transitions*.
-            stopUpdates()
+        // `authorizationStatus` to `.denied`, which stops updates (the funnel
+        // reconciles the manager down, as Apple recommends) and lets the UI
+        // explain the situation.
+        //
+        // Only latch while the app is itself authorized. A `denied` error that
+        // arrives while the raw status is `.denied`/`.notDetermined` would be
+        // masked by `authorizationStatus` anyway, but persisting it there means a
+        // later grant (made while the app wasn't running) launches with a stale
+        // latch that no status-change callback clears — the app would wrongly show
+        // "Location Services Off" until a fix happened to arrive.
+        if let clError = error as? CLError, clError.code == .denied, isPerAppAuthorized {
             applyAuthorizationState(rawStatus: rawAuthorizationStatus, servicesDenied: true)
         }
 
--- a/OBAKitCore/Location/Location/LocationService.swift
+++ b/OBAKitCore/Location/Location/LocationService.swift
@@ -40,11 +40,18 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
 
     public init(userDefaults: UserDefaults, locationManager: LocationManager) {
         self.locationManager = locationManager
-        authorizationStatus = locationManager.authorizationStatus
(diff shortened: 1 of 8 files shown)


OBAKitCore/Location/Location/LocationService.swift
```diff
@@ -247,13 +279,23 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
 
     // MARK: - Delegate
 
+    /// A fresh authorization callback means access was re-evaluated by the user
+    /// or system, so any latched `denied` state is stale. Clearing it before we
+    /// recompute `authorizationStatus` lets derived state (and any resumed
+    /// updates) reflect the new reality.
+    private func resetDeniedLatchForAuthorizationChange() {
+        locationServicesDenied = false
```
aaronbrethorst: **Latch is cleared *before* `authorizationStatus` is reassigned, so the didSet starts the GPS at the moment access is revoked.** (confirmed)

Both authorization callbacks do `resetDeniedLatchForAuthorizationChange()` (L292/L298) and only then `authorizationStatus = ...` (L293/L299). The latch's `didSet` at L170 therefore evaluates `isLocationUseAuthorized` against the *stale, still-authorized* status.

Scenario: Location Services goes off system-wide (`CLError.denied` latches, updates stop). The user re-enables Location Services but sets this app to **Never**. iOS fires `locationManagerDidChangeAuthorization`; the latch clears first, L170 sees `authorizationStatus == .authorizedWhenInUse`, and calls `startUpdates()` → `startUpdatingLocation()` + `startUpdatingHeading()`. Only afterwards is `authorizationStatus = .denied` assigned, and that didSet has no stop branch.

The app has now started location *and heading* updates against a manager the user just denied, and can never stop them: every later `stopUpdates()` (including `Application.swift:551` on backgrounding, and `stopUpdatingLocation()` at L255) is guarded by `isLocationUseAuthorized`, which is now false. Heading/magnetometer updates run for the rest of the process and the location-usage indicator stays lit.

Assign the status first, then clear the latch.

OBAKitCore/Location/Location/LocationService.swift
```diff
@@ -148,6 +148,31 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
...
+    /// the user switches Location Services off system-wide, which does not change
+    /// the app's per-app authorization. While set, `isLocationUseAuthorized`
+    /// reports `false` so the UI hides location affordances (locate button, user
+    /// dot) that would otherwise appear active but never produce a fix. It is
+    /// cleared on the next authorization callback, when the user may have
+    /// re-enabled access.
+    private var locationServicesDenied = false {
+        didSet {
+            guard locationServicesDenied != oldValue else { return }
+            notifyDelegatesAuthorizationChanged(authorizationStatus)
```
aaronbrethorst: **The latch notifies delegates with the *unchanged* `authorizationStatus`, so the explanatory pill never appears.** (confirmed)

`notifyDelegatesAuthorizationChanged(authorizationStatus)` here fires with `.authorizedWhenInUse` — the denied path at L349-351 never touches `authorizationStatus`. Consumers that switch on the status *value* rather than `isLocationUseAuthorized` therefore conclude everything is fine:

- `MapViewModel.topPillState` (`OBAKit/ViewModels/MapViewModel/MapViewModel.swift:210`) maps `.authorizedWhenInUse` → `.hidden`
- `MapStatusView.state(for:)` (`OBAKit/Mapping/MapStatusView.swift:174-178`) derives purely from `service.authorizationStatus` → `.locationServicesOn`, so no indicator

Meanwhile `renderMapStatus()` hides the locate button and `MapRegionManager` removes the blue dot, both keyed off `isLocationUseAuthorized`. Net user-visible result: **the locate button and user dot silently vanish with no "Location Services Off / Turn on in Settings" pill and no path to recovery** — the opposite of the doc comment's stated intent at L154-156.

OBAKitCore/Location/Location/LocationService.swift
```diff
@@ -181,9 +206,16 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
...
 
     /// Answers the question of whether the device GPS can be consulted for location data.
+    ///
+    /// This is derived from `authorizationStatus` plus the `locationServicesDenied`
+    /// latch. We deliberately avoid `CLLocationManager.locationServicesEnabled()`
+    /// because it performs a blocking, synchronous XPC call that Apple warns can
+    /// hang the main thread. When location services are disabled system-wide,
+    /// attempts to start updates fail via `locationManager(_:didFailWithError:)`
+    /// with a `denied` error, which we latch here — the pattern Apple recommends.
     public var isLocationUseAuthorized: Bool {
```
aaronbrethorst, line 261: **Cold start with Location Services off system-wide now reports location as available.** (confirmed)

`isLocationUseAuthorized` no longer consults any system-wide signal at query time, and the latch starts `false` on every launch. If the user has Location Services off system-wide but the app's stored per-app authorization is still `.authorizedWhenInUse`, then on launch this returns `true` until (and unless) a denied error happens to arrive.

Consequences in that window: `applicationDidBecomeActive` calls `startUpdates()`, `MapRegionManager` sets `showsUserLocation = true` (`MapRegionManager.swift:210`), and the locate button renders enabled. The user taps "center on my location" and nothing happens — no fix, no error pill. Previously the `locationManager.isLocationServicesEnabled` conjunct made this `false` from the first frame. The same gap lets `startMonitoringProximity` (L367) register geofences that can never fire.

Worth seeding the latch pessimistically, or gating the UI on "have we successfully gotten a fix" rather than on the optimistic default.

OBAKitTests/Location/LocationServiceTests.swift
```diff
@@ -132,4 +132,91 @@ class LocationServiceTests: XCTestCase {
...
+        let service = LocationService(userDefaults: UserDefaults(), locationManager: locationManagerMock)
+
+        service.requestInUseAuthorization()
+        service.locationManager(CLLocationManager(), didFailWithError: CLError(.denied))
+        expect(service.isLocationUseAuthorized).to(beFalse())
+        expect(locationManagerMock.locationUpdatesStarted).to(beFalse())
+
+        // Status stays `.authorizedWhenInUse`, so recovery must come from the latch
+        // clearing rather than an `authorizationStatus` value change.
+        service.locationManager(CLLocationManager(), didChangeAuthorization: .authorizedWhenInUse)
```
aaronbrethorst: **The recovery test drives the iOS-14-deprecated callback, so the shipping recovery path is untested.** (confirmed)

`test_deniedError_clearedByAuthorizationChange` calls `service.locationManager(CLLocationManager(), didChangeAuthorization:)` — and `AuthorizableLocationManagerMock._authorizationStatus.didSet` (`OBAKitTests/Helpers/Mocks/LocationServiceMocks.swift:92`) also only calls the deprecated variant. With an iOS 18 deployment target and `locationManagerDidChangeAuthorization(_:)` implemented, Core Location **never** invokes the deprecated overload on device.

The production path (L296-310) is a materially different one: it reads `manager.authorizationStatus` from the *real* `CLLocationManager` argument rather than the injected `locationManager`, with different ordering side effects. A regression there (latch never clearing, or clearing against a mismatched status source) leaves this whole new suite green while users are stuck with location permanently disabled after a denied error.

OBAKitCore/Location/Location/LocationService.swift
```diff
@@ -148,6 +148,31 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
...
+            guard locationServicesDenied != oldValue else { return }
+            notifyDelegatesAuthorizationChanged(authorizationStatus)
+
+            // When the latch clears while the app is still authorized, resume
+            // updates ourselves. The `authorizationStatus` didSet only restarts
+            // them when the coarse status *value* changes, which it often does
+            // not on recovery (e.g. it stays `.authorizedWhenInUse`), so relying
+            // on it alone would leave the stream stopped until the next app
+            // foreground.
+            if !locationServicesDenied, isLocationUseAuthorized {
```
aaronbrethorst: **This `startUpdates()` is redundant, and the comment justifying it is factually wrong about Swift `didSet` semantics.** (confirmed)

The comment claims "The `authorizationStatus` didSet only restarts them when the coarse status *value* changes". Swift property observers fire on **every** assignment, including same-value ones. Both authorization callbacks unconditionally run `authorizationStatus = <status>` immediately after clearing the latch, so L143's notify + L146's `startUpdates()` always run anyway.

Result on recovery: `authorizationStatusChanged` is delivered **twice** and `startUpdates()` runs twice. `LocationPermissionBulletin.locationService(_:authorizationStatusChanged:)` calls `self.completion()` on each delivery, so the onboarding bulletin's completion can advance/dismiss two steps; `MapRegionManager` toggles `showsUserLocation` off/on twice, flickering the user dot.

```suggestion
        didSet {
            guard locationServicesDenied != oldValue else { return }
            notifyDelegatesAuthorizationChanged(authorizationStatus)
        }
```
(delete the block and its comment — the assignment that follows already restarts updates)

OBAKitTests/Location/LocationServiceTests.swift
```diff
@@ -132,4 +132,91 @@ class LocationServiceTests: XCTestCase {
...
+
+        service.locationManager(CLLocationManager(), didFailWithError: CLError(.denied))
+
+        // Per-app authorization is unchanged, but location is now reported unavailable.
+        expect(service.authorizationStatus) == .authorizedWhenInUse
+        expect(service.isLocationUseAuthorized).to(beFalse())
+        // Delegates were re-notified so UI (locate button, user dot) can update.
+        expect(del.status).toNot(beNil())
+        // The raw error is still forwarded to delegates.
+        expect((del.error as? CLError)?.code) == .denied
```
aaronbrethorst: **This assertion checks only non-nil, so it can't catch the notification carrying a misleading status value.** (confirmed)

`expect(del.status).toNot(beNil())` passes for *any* `CLAuthorizationStatus`. The value actually delivered is `.authorizedWhenInUse` — the one that makes `MapViewModel.topPillState` hide the "Location Services Off" pill. Because the test never asserts the value, a change to which status is broadcast (or a fix that correctly broadcasts a denied-like status) is indistinguishable from today's behavior, and the pill regression stays invisible to the suite.

Assert the concrete expected status instead.

OBAKitCore/Location/Location/LocationService.swift
```diff
@@ -148,6 +148,31 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
...
 
+    /// Latched when Core Location reports a `denied` error. This can happen even
+    /// while `authorizationStatus` still reads as authorized — most commonly when
+    /// the user switches Location Services off system-wide, which does not change
+    /// the app's per-app authorization. While set, `isLocationUseAuthorized`
+    /// reports `false` so the UI hides location affordances (locate button, user
+    /// dot) that would otherwise appear active but never produce a fix. It is
+    /// cleared on the next authorization callback, when the user may have
+    /// re-enabled access.
+    private var locationServicesDenied = false {
```
aaronbrethorst: **The latch has no self-healing reset path.** (plausible)

`locationServicesDenied` can only be cleared inside an authorization delegate callback. If the user turns Location Services off system-wide (latch sets), then turns it back on, the app's *per-app* authorization never changed — that's this PR's own stated premise. If Core Location does not deliver a `locationManagerDidChangeAuthorization` for the master-switch toggle, nothing clears the latch: the foreground handler at `OBAKit/Orchestration/Application.swift:461` is gated on `if locationService.isLocationUseAuthorized`, which is now `false`, so it neither restarts updates nor resets the latch.

The locate button (`MapViewController.swift:1206`), the blue dot (`MapRegionManager.swift:1108`), and the onboarding location row stay hidden until the user force-quits. The previous implementation recovered automatically because `isLocationServicesEnabled` was re-read on every access.

A cheap belt-and-braces fix: clear the latch (and retry `startUpdates()`) on `applicationDidBecomeActive`, before the `isLocationUseAuthorized` gate.

OBAKitCore/Location/Location/LocationService.swift
```diff
@@ -148,6 +148,31 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
         }
```
aaronbrethorst, line 163: **`authorizationStatus`'s didSet (L141-149) only ever starts updates; it never stops them.** (plausible)

When the user revokes permission in Settings, `authorizationStatus = .denied` notifies delegates but takes no `else` branch, so `stopUpdatingLocation()` / `stopUpdatingHeading()` are never called. From then on every explicit `stopUpdates()` — app backgrounding at `Application.swift:551`, or L254-260 here — short-circuits on `guard isLocationUseAuthorized`, leaving the `CLLocationManager` permanently started: battery drain and a lingering location-usage indicator after the user explicitly said no.

Notably, the `didFailWithError` path in this PR explicitly works around exactly this guard ordering ("Stop first ... then latch"); the authorization-revocation path wasn't given the same treatment.

OBAKitCore/Location/Location/LocationService.swift
```diff
@@ -247,13 +279,23 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
...
+
     @available(iOS, introduced: 4.2, deprecated: 14.0)
     public func locationManager(_ manager: CLLocationManager, didChangeAuthorization status: CLAuthorizationStatus) {
+        resetDeniedLatchForAuthorizationChange()
         authorizationStatus = status
     }
 
     @available(iOS 14, *)
     public func locationManagerDidChangeAuthorization(_ manager: CLLocationManager) {
+        resetDeniedLatchForAuthorizationChange()
```
aaronbrethorst: **Clearing the latch unconditionally on every authorization callback re-arms location optimistically, producing a visible start→fail→stop flicker.** (plausible)

With Location Services still off system-wide, any authorization callback — Core Location fires `locationManagerDidChangeAuthorization` once whenever a delegate is assigned, e.g. a second `LocationService`/`CLLocationManager` created during the session — clears the latch. That didSet notifies delegates (`MapRegionManager` sets `showsUserLocation = true`, the locate button un-hides) and calls `startUpdates()`. Core Location asynchronously replies with another `.denied`, so `didFailWithError` (L341) stops updates and re-latches, firing a third notification that hides everything again.

The user watches the locate button and blue dot appear then vanish, and the app repeatedly pokes a disabled location subsystem. Previously `isLocationServicesEnabled` was read fresh on every access, so the affordances simply stayed hidden.

aaronbrethorst (changes requested):
please address feedback and then we'll get this merged.

aaronbrethorst (changes requested):
please resolve the inline issues noted and we'll get this merged.

OBAKitCore/Location/Location/LocationService.swift
mosliem: Confirmed and fixed. The stale-read was structural, so rather than reorder the two assignments I removed the property observers entirely and replaced them with a single funnel:

```swift
private func applyAuthorizationState(rawStatus: CLAuthorizationStatus, servicesDenied: Bool) {
    let oldStatus = authorizationStatus
    let wasAuthorized = isLocationUseAuthorized

    rawAuthorizationStatus = rawStatus
    locationServicesDenied = servicesDenied

    guard authorizationStatus != oldStatus else { return }
    ...
}
```

Both inputs are applied *before* anything reacts, and every effect is computed from the post-assignment effective status — so there is no longer an ordering in which one observer can see the other's stale value. Your exact scenario (services re-enabled, app set to Never) now assigns `.denied` + latch-cleared together, and the funnel takes the revocation branch.

The "can never stop them" half is fixed too: `stopUpdatingLocation()` is no longer guarded on `isLocationUseAuthorized`, since stopping is always safe and gating it strands a manager started before revocation. See the reply on the L163 thread for the teardown branch.

OBAKitCore/Location/Location/LocationService.swift
mosliem: Confirmed and fixed. This was the most consequential of the nine — the latch was a side channel that value-switching consumers couldn't see. `authorizationStatus` is now computed, and reports the *effective* state:

```swift
public var authorizationStatus: CLAuthorizationStatus {
    guard locationServicesDenied, isPerAppAuthorized else { return rawAuthorizationStatus }
    return .denied
}
```

The per-app value Core Location gives us is preserved separately as `rawAuthorizationStatus`. So `MapViewModel.topPillState` and `MapStatusView.state(for:)` now see `.denied` and render the "Location Services Off / Turn On in Settings" pill, matching what `isLocationUseAuthorized`-keyed UI is already doing. The masking is deliberately skipped while the app isn't itself authorized, so `.notDetermined` survives and the app can still prompt.

`test_deniedError_marksLocationUnavailable` now asserts the concrete value — `expect(service.authorizationStatus) == .denied` and `expect(del.status) == .denied` — per your L225 comment.

OBAKitCore/Location/Location/LocationService.swift
mosliem: Confirmed and fixed by seeding the latch pessimistically, as you suggested. It's now persisted:

```swift
// Seed the latch from the last session. Location Services being off
// system-wide survives an app relaunch, but nothing tells us so at
// launch: the per-app authorization still reads as authorized and no
// `denied` error has arrived yet.
locationServicesDenied = userDefaults.bool(forKey: UserDefaultsKeys.locationServicesDenied)
```

written back only on change (new key `LocationService.locationServicesDenied`). So a launch that follows a denied session starts with location reported unavailable rather than optimistically available — no enabled locate button that does nothing, and `startMonitoringProximity` sees the correct state.

The seed is an assumption, not a verdict: the user may have re-enabled Location Services while the app was dead. It's resolved by probing rather than by re-reading a blocking API — `retryIfLocationServicesDenied()` starts the manager at the init-time authorization callback, and a fix clears the latch while another `denied` keeps it. `test_deniedLatch_persistsAcrossLaunches` and `test_seededLatch_isProbedOnAuthorizationCallback` cover both halves.

OBAKitTests/Location/LocationServiceTests.swift
mosliem: Confirmed and fixed. You're right that with an iOS 18 target the deprecated overload is dead code on device, so rather than have the tests drive it I deleted `locationManager(_:didChangeAuthorization:)` from `LocationService` outright. There's now exactly one authorization path, and it's the one that ships.

`AuthorizableLocationManagerMock._authorizationStatus.didSet` fires the modern callback to match:

```swift
delegate?.locationManagerDidChangeAuthorization?(CLLocationManager())
```

On the "different status source" point: the production callback now reads from the *injected* `locationManager`, not from the `manager` argument, so tests and device exercise the same read. The argument is still used for `accuracyAuthorization`, where the real manager is the correct source.

All the authorization tests in the suite now go through this single path.

OBAKitCore/Location/Location/LocationService.swift
mosliem: You're right about `didSet` semantics — property observers fire on every assignment, and the comment claiming otherwise was wrong. Both observers are gone now, replaced by the single `applyAuthorizationState(rawStatus:servicesDenied:)` funnel, which notifies **once** and only on an effective-status transition. The double `authorizationStatusChanged` you traced through `LocationPermissionBulletin.completion()` and the `showsUserLocation` dot flicker is fixed.

One deviation from the suggested diff I want to flag, because I hit it the hard way. I first took the "delete `startUpdates()` and let the assignment restart it" route, and it caused a regression: Core Location fires an authorization callback as soon as the delegate is assigned, carrying an *unchanged* status. That callback is where updates have always been started from at `init`. Once the funnel deduped on transitions it became a no-op, the first fix slipped to `applicationDidBecomeActive`, and the map sat at its zoomed-out fallback long enough to flash the "Zoom in for stops" pill.

So starting updates now lives outside the funnel:

```swift
applyAuthorizationState(rawStatus: newStatus, servicesDenied: statusChanged ? false : locationServicesDenied)

if isLocationUseAuthorized {
    startUpdates()
} else {
    retryIfLocationServicesDenied()
}
```

`startUpdatingLocation()` is idempotent, so this costs nothing on repeat callbacks — and the actual problem in your comment, duplicate *delegate notifications*, stays fixed, since that's inside the funnel's transition guard. Covered by `test_authorizationCallback_startsUpdates_whenAlreadyAuthorized` and `test_authorizationCallback_doesNotStartUpdates_whenUnauthorized`.

OBAKitTests/Location/LocationServiceTests.swift
mosliem: Confirmed and fixed. `test_deniedError_marksLocationUnavailable` now asserts the concrete values rather than non-nil:

```swift
expect(service.authorizationStatus) == .denied
expect(del.status) == .denied
```

Which is exactly the assertion that would have caught the pill regression you found on the L162 thread — before the effective-status change it delivered `.authorizedWhenInUse` and this test was blind to it. It now fails if the broadcast status stops being denied-like.

OBAKitCore/Location/Location/LocationService.swift
mosliem: Confirmed and fixed, taking your belt-and-braces suggestion. There's now a probe:

```swift
public func retryIfLocationServicesDenied() {
    guard locationServicesDenied, isPerAppAuthorized else { return }
    locationManager.startUpdatingLocation()
}
```

called from `applicationDidBecomeActive` **before** the `isLocationUseAuthorized` gate — which was the thing making the old handler unreachable in exactly this state:

```swift
// Runs before the gate below: the user may have re-enabled Location
// Services while they were away, which leaves `isLocationUseAuthorized`
// false until we probe for it.
locationService.retryIfLocationServicesDenied()

if locationService.isLocationUseAuthorized {
    locationService.startUpdates()
}
```

It probes rather than clearing the latch outright, so there's no start→fail→stop flicker (the L298 thread): the latch holds until there's an answer. A fix clears it via `didUpdateLocations`; another `denied` leaves it untouched. It also runs at the init-time authorization callback, so a latch seeded from the previous launch is resolved in the first frames instead of waiting for a foreground event. `test_deniedError_clearedByForegroundRetry` covers the recovery.

OBAKitCore/Location/Location/LocationService.swift
mosliem: Confirmed and fixed. The revocation branch now exists in the funnel:

```swift
if isLocationUseAuthorized {
    startUpdates()
} else if wasAuthorized {
    // Access was revoked. Tear the manager down here — nobody else will,
    // and a `CLLocationManager` left running keeps the location-usage
    // indicator lit and the magnetometer powered for the rest of the
    // process.
    stopUpdates()
}
```

And the guard-ordering trap you identified is gone at the source — `stopUpdatingLocation()` is no longer gated:

```swift
/// Unlike its `start` counterpart this is unguarded: stopping is always safe,
/// and gating it on authorization would make a manager we started before
/// access was revoked impossible to ever turn off.
public func stopUpdatingLocation() {
    locationManager.stopUpdatingLocation()
}
```

`Application.swift:551` on backgrounding is unconditional now for the same reason. Heading is torn down too — `stopUpdates()` stops both, so the magnetometer doesn't stay powered. `test_authorizationRevoked_stopsUpdates` asserts both `locationUpdatesStarted` and `headingUpdatesStarted` go false on revocation.

OBAKitCore/Location/Location/LocationService.swift
mosliem: Confirmed and fixed. The latch is now cleared only when the per-app authorization actually *changed* — a callback that leaves the status untouched carries no new information and is treated as such:

```swift
let newStatus = locationManager.authorizationStatus
let statusChanged = newStatus != rawAuthorizationStatus
applyAuthorizationState(rawStatus: newStatus, servicesDenied: statusChanged ? false : locationServicesDenied)
```

So the delegate-assignment callback you describe — a second `LocationService`/`CLLocationManager` created mid-session — no longer clears the latch, and the start→fail→stop notification storm doesn't happen. A real change (user made a fresh decision in Settings) still re-arms optimistically, which is correct: that's genuinely new evidence.

Recovery when the status *doesn't* change is handled by probing instead — `retryIfLocationServicesDenied()`, see the L159 thread. It starts the manager without touching the latch or notifying anyone, so nothing appears on screen until a fix actually arrives.

`test_deniedError_survivesRedundantAuthorizationCallback` covers it, asserting the latch holds, no delegate notification fires, and neither location nor heading updates start.

OBAKitCore/Location/Location/LocationService.swift
```diff
@@ -274,6 +382,11 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
             return
         }
 
+        // A fix is proof that location is usable again, whatever we last latched.
+        if locationServicesDenied {
```
aaronbrethorst, line 435: **Correctness (blocking):** The persisted denied latch can only be cleared by an actual location fix (or an observed authorization status *change*). After the user re-enables Location Services system-wide while the app isn't running, the raw status is unchanged, so the only recovery path is the probe producing a real GPS fix. Indoors or in a parking garage, Core Location delivers only `kCLErrorLocationUnknown` (ignored by `didFailWithError`) and never a fix — so the app stays stuck reporting `.denied` all session: the "Location Services Off" pill shows, the locate button and user dot stay hidden, and nearby-stops sorting is disabled even though everything is enabled. This is a regression vs. `main`, which re-read `locationServicesEnabled()` and recovered immediately. Consider re-checking `locationServicesEnabled()` off the main thread (e.g. on a background queue on foreground/retry) to clear the latch without needing a fix.

OBAKitCore/Location/Location/LocationService.swift
```diff
@@ -297,6 +410,19 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
     }
 
     public func locationManager(_ manager: CLLocationManager, didFailWithError error: Error) {
+        // A `denied` error means location is currently unavailable even if the
+        // per-app authorization still reads as authorized (e.g. Location
+        // Services switched off system-wide). Latching it flips the effective
+        // `authorizationStatus` to `.denied`, which stops updates as Apple
+        // recommends and lets the UI explain the situation.
+        if let clError = error as? CLError, clError.code == .denied {
```
aaronbrethorst: **Correctness (blocking):** This latches (and persists) `locationServicesDenied` even when the app is not per-app authorized. If a `CLError.denied` arrives while the raw status is `.denied`/`.notDetermined`, the latch is persisted while masked. When the user later grants "While Using" in Settings while the app isn't running, the next launch seeds an authorized raw status *plus* `locationServicesDenied == true` — and the delegate-assignment callback sees no status change, so it never clears the latch. The app wrongly shows "Location Services Off" despite the fresh grant, until a GPS fix happens to arrive. Suggest only latching when per-app authorized (`isPerAppAuthorized`).

OBAKitCore/Location/Location/LocationService.swift
```diff
@@ -181,9 +243,15 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
...
 
     /// Answers the question of whether the device GPS can be consulted for location data.
+    ///
+    /// Derived from the effective `authorizationStatus`. We deliberately avoid
+    /// `CLLocationManager.locationServicesEnabled()` because it performs a
+    /// blocking, synchronous XPC call that Apple warns can hang the main thread.
+    /// When location services are disabled system-wide, attempts to start
+    /// updates fail via `locationManager(_:didFailWithError:)` with a `denied`
+    /// error, which we latch — the pattern Apple recommends.
     public var isLocationUseAuthorized: Bool {
```
aaronbrethorst, line 261: **Correctness:** The old synchronous check guaranteed the UI never advertised location while services were off system-wide. The latch only re-establishes that *after* a first denied error has been persisted — so on the first run after upgrading to this build (latch key absent, services off, app authorized), `isLocationUseAuthorized` reads `true`: the locate button appears, `showsUserLocation` is enabled, `startUpdates()` runs, and then the async `CLError.denied` visibly retracts the UI to the "Location Services Off" pill. The previous release rendered the correct state from the first frame.

OBAKitCore/Location/Location/LocationService.swift
```diff
@@ -247,14 +314,55 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
...
+        // carries no new information — Core Location fires one whenever a
+        // delegate is assigned — so it must not clear the latch, or every
+        // spurious callback would restart updates against a location subsystem
+        // that is still off.
+        //
+        // Read the status from the injected manager, not from `manager`, so the
+        // path under test is the path that ships.
+        let newStatus = locationManager.authorizationStatus
+        let statusChanged = newStatus != rawAuthorizationStatus
+        applyAuthorizationState(rawStatus: newStatus, servicesDenied: statusChanged ? false : locationServicesDenied)
```
aaronbrethorst: **Correctness (likely):** A raw status change between two authorized values (`.authorizedWhenInUse` → `.authorizedAlways`) while Location Services are still off system-wide clears the latch optimistically: delegates get notified of an authorized status, the "Location Services Off" pill disappears and the locate button appears — then `startUpdates()` immediately fails with `CLError.denied`, re-latching and re-notifying `.denied`. That's exactly the on/off flicker the probing design (per `retryIfLocationServicesDenied`'s doc comment) exists to prevent. Consider only clearing the latch when the transition crosses into authorization, or letting the probe clear it silently.

OBAKitTests/Helpers/Mocks/LocationServiceMocks.swift
```diff
@@ -110,9 +111,18 @@ public class AuthorizableLocationManagerMock: LocationManagerMock {
...
 
+    /// Simulates Location Services being switched off system-wide: the app's own
+    /// authorization is untouched, but every attempt to start updates comes back
+    /// as `CLError.denied` instead of a fix.
+    public var simulatesLocationServicesOff = false
+
     public override func startUpdatingLocation() {
         super.startUpdatingLocation()
-        if authorizationStatus == .authorizedWhenInUse {
+        guard authorizationStatus == .authorizedWhenInUse else { return }
```
aaronbrethorst: **Test correctness:** `startUpdatingLocation()` only simulates fixes or denied errors when `authorizationStatus == .authorizedWhenInUse`, so the `.authorizedAlways` path — including the new `simulatesLocationServicesOff` mode — is silently inert: the guard returns before delivering anything. Any test exercising the Always-authorized flow passes vacuously. Suggest guarding on either authorized status.

OBAKitCore/Location/Location/LocationService.swift
```diff
@@ -247,14 +314,55 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
...
+
+        // Core Location delivers a callback as soon as the delegate is assigned,
+        // carrying an *unchanged* status. That is the app's earliest signal that
+        // it is authorized, and it is where location updates have always been
+        // started from — early enough that the first fix lands before the map
+        // settles. `applyAuthorizationState` deliberately reacts only to
+        // transitions, so starting updates has to happen outside of it or a
+        // relaunch of an already-authorized app would wait for the next
+        // foreground. Both calls below are idempotent.
+        if isLocationUseAuthorized {
```
aaronbrethorst: **Cleanup:** `startUpdates()` runs twice per real authorization grant — `applyAuthorizationState` already starts updates on the transition, then this post-funnel block starts them again. With Location Services off, that means two failed probes and two `errorReceived` delegate callbacks for a single grant (any error UI fires twice); on the happy path the start calls are redundantly repeated. A reconcile-style funnel (start when effectively authorized, stop when it just became unauthorized, notify only on transition) would remove this special case entirely.

OBAKitTests/Location/LocationServiceTests.swift
```diff
@@ -16,6 +16,12 @@ import Nimble
 
 @MainActor
 class LocationServiceTests: XCTestCase {
+    /// An isolated defaults suite. The denied latch is persisted, so these tests
+    /// must not read or write each other's (or the standard suite's) state.
+    private func freshDefaults() -> UserDefaults {
```
aaronbrethorst, line 25: **Test hygiene:** `freshDefaults()` re-implements `OBATestCase.buildUserDefaults(suiteName:)` but never calls `removePersistentDomain`, so every test run leaks ~15 UUID-named defaults suites as plist files in the test host's Library/Preferences — accumulating indefinitely on dev simulators and CI images. Either reuse the base-class helper with its tearDown, or add an `addTeardownBlock { UserDefaults().removePersistentDomain(forName: suiteName) }` inside `freshDefaults()`.

OBAKitTests/Location/LocationServiceTests.swift
```diff
@@ -132,4 +138,243 @@ class LocationServiceTests: XCTestCase {
...
         expect(delError) == err
     }
+
+    /// Core Location delivers an authorization callback as soon as the delegate
+    /// is assigned, carrying an unchanged status. An already-authorized app must
+    /// start updates from it: that is what gets the first fix in before the map
+    /// settles, and without it the map opens zoomed out and jumps once the fix
+    /// finally arrives at the next foreground event.
+    func test_authorizationCallback_startsUpdates_whenAlreadyAuthorized() {
+        let locationManagerMock = AuthorizableLocationManagerMock(updateLocation: TestData.mockSeattleLocation, updateHeading: TestData.mockHeading)
```
aaronbrethorst: **Cleanup:** This identical two-line mock+service construction is repeated ~15 times across the new tests. A small factory — e.g. `makeService(defaults: UserDefaults = freshDefaults()) -> (AuthorizableLocationManagerMock, LocationService)` — would collapse the duplication and let each test's actual arrange step stand out.

OBAKitCore/Location/Location/LocationService.swift
```diff
@@ -137,14 +145,68 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
...
-            notifyDelegatesAuthorizationChanged(authorizationStatus)
+            guard locationServicesDenied != oldValue else { return }
+            userDefaults.set(locationServicesDenied, forKey: UserDefaultsKeys.locationServicesDenied)
+        }
+    }
 
-            if isLocationUseAuthorized {
-                startUpdates()
-            }
+    private var isPerAppAuthorized: Bool {
```
aaronbrethorst, line 166: **Cleanup:** `isPerAppAuthorized` and `isLocationUseAuthorized` (line ~254) duplicate the same `== .authorizedWhenInUse || == .authorizedAlways` predicate against different status sources. An `isAuthorized` computed var in the existing `CLAuthorizationStatus` extension (CoreLocationExtensions.swift) would give both a single definition.

OBAKitCore/Location/Location/LocationService.swift
```diff
@@ -247,14 +314,55 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
...
+        if isLocationUseAuthorized {
+            startUpdates()
+        } else {
+            // Same reasoning for a latch seeded from the previous launch: probe
+            // it now rather than making the user wait for a foreground event.
+            retryIfLocationServicesDenied()
+        }
 
         // Accuracy can change while the coarse status stays put (e.g. "Allow
         // Once" elevates a reduced-accuracy session to full). That leaves the
```
aaronbrethorst, line 416: **Correctness (testability):** Just below this hunk, line 371 reads `manager.accuracyAuthorization` from the callback's throwaway parameter — while the comment above (line 345) mandates reading from the injected `locationManager` "so the path under test is the path that ships." Any test that sets a mock's `accuracyAuthorization` to `.reducedAccuracy` and fires `locationManagerDidChangeAuthorization(CLLocationManager())` will read the fresh real manager's `.fullAccuracy` instead, so `notifyDelegatesAccuracyAuthorizationChanged` never fires under test and the accuracy-change delegate channel is untestable. Should be `locationManager.accuracyAuthorization`, for the same reason as the status read.

aaronbrethorst (changes requested):
Thanks for iterating on this, @mosliem — the delegate-funnel rework and the new test coverage are a real improvement. The review did surface a few blocking recovery-path issues with the persisted denied latch (it can get stuck reporting `.denied` after the user re-enables Location Services, plus a couple of related edge cases), noted inline. Please fix the items noted in the feedback and we will get this merged!

OBAKitCore/Location/Location/LocationService.swift
mosliem: Fixed. Reintroduced `locationServicesEnabled()` — but `async`, read off the main thread inside a `Task.detached`, as you suggested. A new `refreshLocationServicesEnabled()` reconciles the latch against that authoritative switch on foreground/retry and on the init-time authorization callback, so recovery no longer needs a GPS fix and works indoors. It reconciles both directions, so it also latches a services-off launch without waiting for a denied error.

OBAKitCore/Location/Location/LocationService.swift
mosliem: Fixed. `didFailWithError` now latches only when `isPerAppAuthorized`, so a `denied` error arriving while `.denied`/`.notDetermined` isn't persisted and can't be resurrected after a later grant. Covered by `Denied error while unauthorized is not latched`.

OBAKitCore/Location/Location/LocationService.swift
mosliem: Mitigated. The init-time authorization callback now runs the off-main `locationServicesEnabled` probe, so a first launch with services off latches without waiting for the async `denied` error — shrinking the window to a single off-main hop, and it works even when no fix would arrive. A truly flash-free first frame needs the synchronous read this PR removes precisely to stop hanging the main thread, so this is the closest async equivalent; the persisted latch then makes every subsequent launch correct from frame one.

OBAKitCore/Location/Location/LocationService.swift
mosliem: Fixed. On an authorization callback the latch is now cleared only when the transition crosses *into* authorization (`!wasAuthorized && nowAuthorized`). WhenInUse → Always no longer clears it, so there's no flicker while services are off; the system-wide toggle (which delivers no callback) is handled by the probe clearing it silently. Covered by `Latch cleared when authorization crosses into authorized` and `Denied latch cleared by probe on authorization callback`.

OBAKitTests/Helpers/Mocks/LocationServiceMocks.swift
mosliem: Fixed. The mock's `startUpdatingLocation`/`startUpdatingHeading` now guard on `isAuthorized` (either authorized status) instead of `== .authorizedWhenInUse`, so the `.authorizedAlways` path actually delivers fixes/errors rather than returning vacuously.

OBAKitCore/Location/Location/LocationService.swift
mosliem: Fixed. `applyAuthorizationState` is now the single reconcile funnel: it starts/stops updates idempotently and notifies only on an effective transition. The separate post-funnel start is gone, so a real grant no longer starts twice — and no longer fires two failed probes / two `errorReceived` callbacks when services are off.

OBAKitTests/Location/LocationServiceTests.swift
mosliem: Fixed. `freshDefaults()` now records each suite name and `deinit` calls `removePersistentDomain(forName:)` for all of them, so runs no longer leak a plist per test.

OBAKitTests/Location/LocationServiceTests.swift
mosliem: Fixed. Added a `makeService(defaults:servicesOff:initialStatus:)` factory; the repeated mock + service construction across the new tests is collapsed onto it.

OBAKitCore/Location/Location/LocationService.swift
mosliem: Fixed. Added `CLAuthorizationStatus.isAuthorized` in `CoreLocationExtensions`; `isPerAppAuthorized`, `isLocationUseAuthorized`, and the mocks all share that single definition now.

OBAKitCore/Location/Location/LocationService.swift
mosliem: Fixed. Reads `locationManager.accuracyAuthorization` (the injected manager) rather than the throwaway `manager` argument, matching the status read — so the accuracy-change delegate channel is now testable.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
This came back in excellent shape. I re-read the whole of `LocationService` at head rather than diffing, and nine of the ten points from last round are fully addressed — the tenth as far as an async design allows.

The parts I checked hardest:

**The latch can no longer strand a session.** `refreshLocationServicesEnabled()` probes the real system switch off-main and `retryIfLocationServicesDenied()` runs it on foreground, so recovery no longer depends on a GPS fix arriving. That was my main worry — a user indoors who fixed Settings and got nothing back.

**First-launch prompting still works.** `authorizationStatus` guards on `locationServicesDenied, isPerAppAuthorized`, so the latch is masked whenever the app isn't itself authorized and `.notDetermined` survives to reach `canRequestAuthorization`. That's the subtle failure mode I was most afraid of introducing and it's handled correctly.

**No optimistic clearing on WhenInUse → Always.** `crossesIntoAuthorization = !rawAuthorizationStatus.isAuthorized && newStatus.isAuthorized` is exactly right — a transition between two authorized values can't clear a latch while services are still off, so no on/off flicker.

**One funnel, one notification.** `applyAuthorizationState` snapshots `oldStatus` before mutating, reconciles start/stop unconditionally, and notifies only on an effective transition. The comment explaining why property observers can't do this — whichever assigns first sees the other's stale value — is the right thing to leave behind.

**Single-flight probe.** Cancel-and-replace plus the post-`await` `guard !Task.isCancelled, self.isPerAppAuthorized` correctly handles a superseded probe resuming with an answer it must not apply. Cancellation can't interrupt the blocking read, and you say so in the comment.

The mocks now guard on `isAuthorized` rather than `== .authorizedWhenInUse`, so the `.authorizedAlways` tests aren't vacuous any more, and `freshDefaults()` cleans up after itself. Merging.

1. **Misattached doc comment.** Around `servicesProbeTask`, two doc blocks got merged: the first fourteen lines document `refreshLocationServicesEnabled()`, then `/// The in-flight probe, if any.` appears mid-paragraph, and `refreshLocationServicesEnabled()` itself ends up with no doc comment at all. Editing artifact from the last commit — worth splitting back apart.

2. **Foreground recovery is one-directional.** `retryIfLocationServicesDenied()` early-returns unless the latch is already set, so services being turned *off* while backgrounded isn't caught on foreground: the app starts updates, takes an async `denied`, and the UI retracts. Same class as the first-frame flash and it self-corrects, but the doc comment claims the probe "reconciles both directions," which is true of `refreshLocationServicesEnabled()` and not of the gate in front of it. Calling it unconditionally would make the comment honest.

3. **`Task.detached` for the probe** blocks a cooperative-pool thread on an XPC round-trip. Probe frequency is low and the main thread is genuinely freed, which is the point — but a `DispatchQueue` hop is the textbook form if you ever want to tidy it.

Thanks for the thorough second pass, especially on the mock guards and the defaults cleanup.

PR #1211, opened 2026-07-20
Wire the SwiftUI map panel to MapRegionManager

mosliem:
## Summary

Brings the experimental SwiftUI map surface (`MapPanelRootView`) up to functional parity with the classic UIKit map. Stops now load, render, and clear against the real `MapRegionManager` data path; tapping a stop opens its detail sheet; and location/zoom behavior matches the UIKit map. Shared thresholds are lifted into `MapRegionManager` so both surfaces agree on when stops appear.

All new logic is unit-tested and kept off the `MapViewModel`, which stays MapKit-free.

### 1. Stop loading in `MapRegionManager`

- New `requestStops(in:)` — loads stops for an **explicitly provided** region (SwiftUI hosts don't own the manager's `mapView`), reusing the existing fudge-factor, cache-save, and cache-fallback behavior. UIKit's `requestDataForMapRegion()` now delegates to it.
- New `scheduleStopsRequest(in:)` — debounced (0.25s, matching the UIKit timer), fire-and-forget entry point that coalesces rapid camera settles and cancels any in-flight request.
- `requiredHeightToShowStops` promoted to a shared `static` constant so both surfaces gate on the same zoom threshold.

### 2. `MapStopsObserver` (new)

Bridges `MapRegionManager`'s `stopsUpdated` delegate callback to a `@Published [Stop]` for SwiftUI. Kept separate from `MapViewModel`. Accumulates (unions) stops rather than replacing them—mirroring the UIKit `displayUniqueStopAnnotations` behavior—which prevents annotation teardown/re-add flicker when panning away and back. Bounded by `reset()` on zoom-out.

### 3. `StopIconFactory.buildSquircleIcon(for:)` (new)

Renders the squircle stop pin as a `UIImage` for use as a SwiftUI `Map` annotation, honoring bookmark styling and trait/appearance.

### 4. `MapPanelRootView` wiring

- Renders stop annotations from `MapStopsObserver`, suppressed above the zoom threshold (matching UIKit removing annotations when zoomed out).
- Camera settle drives `scheduleStopsRequest(in:)` and the "Zoom in for stops" pill using the shared threshold.
- Tapping a stop pushes `.stopDetails`; selection is cleared so re-tapping the same stop re-pushes.
- **Location parity:** Recenters on the user's first location fix after permission is granted (previously the camera stayed parked until the center-on-location button was tapped).

### 5. Stop-detail sheet

- `.stopDetails` route wired through `AppSheetViewFactory` to `StopDetailSheetHost`, which hosts the redesigned `StopPageViewController` in a navigation controller.
- Leading **Close** button (dismisses via the stacked-sheet binding, keeping `SheetCoordinator` in sync; title forced black over the app's global green bar-button tint).
- Sheet pinned to full height (`.large`).

### 6. `MapViewModel`

Implements the `locationChanged` delegate callback, latching a one-shot `didReceiveInitialLocation` flag that the view observes to perform the recenter described above.

## Tests

- **`MapRegionManagerTests`** — explicit-region load path and debounced scheduling.
- **`MapStopsObserverTests`** — accumulation/union semantics, seeding, and `reset()`.
- **`StopIconFactoryTests`** — squircle icon rendering.
- **`MapViewModelTests`** — `didReceiveInitialLocation` latches once on the first location fix.
- **`StopDetailSheetHostTests`** — navigation wrapping and leading **Close** button.
- **`AppSheetViewFactoryTests`** — `.stopDetails` route resolution.

## Video

https://github.com/user-attachments/assets/2a67dfd8-2918-4ee3-a13d-eebda71d4276

## Summary by CodeRabbit

- **New Features**
  - Added SwiftUI stop annotations with tap-to-select and proper stop-detail presentation.
  - Introduced debounced, cancelable stop loading that refreshes on map camera/zoom changes with cache-first behavior.
  - Upgraded stop icons to squircle rendering with improved caching, bookmark styling, and light/dark support.
- **Improvements**
  - Stop details sheet now opens directly at full height (single detent).
  - Stop visibility and loading are gated by zoom, with viewport-based pruning and improved initial recentering.
  - Expanded “extra stop data” visibility thresholds and public helper behavior.
- **Tests**
  - Expanded coverage for stop loading, icon caching/rendering, observer behavior, and sheet/view-model rules.


OBAKit/Sheet/Root/MapStopsObserver.swift
```diff
@@ -0,0 +1,63 @@
...
+        // each time — panning away and back would re-render the same pins (the
+        // reported flicker). By only appending genuinely new stops (and keeping
+        // existing instances), returning to a visited area is a no-op: nothing is
+        // republished, so the pins already on screen stay put. The accumulated set
+        // is bounded by `reset()` on zoom-out.
+        let existingIDs = Set(self.stops.map(\.id))
+        let newStops = stops.filter { !existingIDs.contains($0.id) }
+        guard !newStops.isEmpty else { return }
+
+        self.stops += newStops
```
aaronbrethorst: **Observer unions stops forever; stale pins accumulate** · correctness · CONFIRMED

`MapRegionManager.requestStops(in:)` **replaces** `manager.stops` for each settled region, but `mapRegionManager(_:stopsUpdated:)` here only appends new IDs and never removes stops that left the current region. `reset()` is only reached on zoom-out.

**Failure:** A user panning across a metro area while staying zoomed in accumulates hundreds of stop annotations from every region panned through — they all stay rendered simultaneously, cluttering the map with pins far outside the viewport. This set also diverges from `mapRegionManager.stops` (which holds only the last region) that RoutePicker/CurrentTrip read.

OBAKit/Sheet/Root/MapPanelRootView.swift
```diff
@@ -61,28 +72,54 @@ struct MapPanelRootView: View {
...
+        }
+        .onMapCameraChange(frequency: .onEnd) { context in
+            // Gate on the same threshold as the UIKit region-change path so both
+            // surfaces agree on when stops load, and so region-level zoom doesn't
+            // fire wasteful (server-throttled) full-region stop requests.
+            isZoomedInForStops = context.rect.height <= MapRegionManager.requiredHeightToShowStops
+            guard isZoomedInForStops else {
+                // Zoomed out: clear accumulated stops, matching the UIKit path
+                // which removes all stop annotations above the threshold.
+                stopsObserver.reset()
```
aaronbrethorst, line 155: **Zoom-out `reset()` doesn't cancel the pending stops request** · correctness · CONFIRMED

On zoom-out, `stopsObserver.reset()` clears stops but the last zoomed-in `scheduleStopsRequest` is never cancelled (the `guard isZoomedInForStops else { reset(); return }` early-return skips the cancel-and-replace at line 117). Its debounced request lands ~250ms later and repopulates the accumulated set.

**Failure:** User zooms in (a request is pending in its ~250ms debounce), then zooms out past the threshold in one gesture. `reset()` runs, but the pending request completes afterward, sets `mapRegionManager.stops`, and re-appends into the observer. When the user zooms back in, the previous region's stale pins flash on screen before the new region's stops load — undoing the clear `reset()` was meant to guarantee.

OBAKit/Sheet/Root/MapPanelRootView.swift
```diff
@@ -61,28 +72,54 @@ struct MapPanelRootView: View {
...
-        // the same on-screen viewport. Fix is to observe `cameraPosition`
-        // changes, convert to an `MKCoordinateRegion`, and feed it into
-        // `MapRegionManager` so both surfaces agree on the cached stop set
-        // (also unblocks rendering stop annotations on the SwiftUI map).
-        Map(position: $cameraPosition) {
+        Map(position: $cameraPosition, selection: $selectedStopID) {
             UserAnnotation()
+            // Suppressed at region-level zoom to match the UIKit path, which
+            // removes stop annotations above `requiredHeightToShowStops`.
+            if isZoomedInForStops {
```
aaronbrethorst, line 126: **Bookmark pins vanish on zoom-out; the UIKit map kept them** · correctness · CONFIRMED

UIKit `reloadStopAnnotations`, when zoomed out past `requiredHeightToShowStops`, removes only Stop annotations (`removeAnnotations(type: Stop.self)`) while leaving Bookmark annotations on screen at every zoom level. The SwiftUI path instead gates **all** pins behind `if isZoomedInForStops` and calls `stopsObserver.reset()` on zoom-out, so bookmarked stops are hidden too.

**Failure:** A user with a bookmarked stop who zooms out to a city/region overview sees their bookmark pin disappear, losing the orientation cue the old map preserved.

OBAKit/Sheet/Root/MapPanelRootView.swift
```diff
@@ -61,28 +72,54 @@ struct MapPanelRootView: View {
...
+                    // `.annotationTitles(.hidden)` on `Map`, so we route
+                    // accessibility through the icon's own `accessibilityLabel`
+                    // instead of a hidden title string).
+                    Annotation("", coordinate: stop.coordinate) {
+                        Image(uiImage: application.stopIconFactory.buildSquircleIcon(
+                            for: stop,
+                            isBookmarked: application.userDataStore.findBookmark(stopID: stop.id) != nil,
+                            traits: UITraitCollection(userInterfaceStyle: colorScheme == .dark ? .dark : .light)
+                        ))
+                        .accessibilityLabel(stop.name)
```
aaronbrethorst: **Stop annotation exposes only `stop.name` to VoiceOver** · correctness · CONFIRMED

The UIKit `StopAnnotationView` announced `Formatters.formattedAccessibilityLabel(stop:)` — `"name; direction; routes served; Stop ID: NNN"`. The replacement SwiftUI `Annotation` sets `.accessibilityLabel(stop.name)` only.

**Failure:** A VoiceOver user focusing a stop pin hears just `"Main St & 5th"` instead of `"Main St & 5th; North; Routes 40, 44; Stop ID: 12345"`, losing the direction, routes-served, and stop-ID information needed to distinguish and choose the correct stop. Consider reusing `Formatters.formattedAccessibilityLabel(stop:)` here.

OBAKit/Sheet/Root/MapPanelRootView.swift
```diff
@@ -61,28 +72,54 @@ struct MapPanelRootView: View {
...
+            // surfaces agree on when stops load, and so region-level zoom doesn't
+            // fire wasteful (server-throttled) full-region stop requests.
+            isZoomedInForStops = context.rect.height <= MapRegionManager.requiredHeightToShowStops
+            guard isZoomedInForStops else {
+                // Zoomed out: clear accumulated stops, matching the UIKit path
+                // which removes all stop annotations above the threshold.
+                stopsObserver.reset()
+                return
+            }
+            application.mapRegionManager.scheduleStopsRequest(in: context.region)
```
aaronbrethorst, line 160: **Drops the `searchResponseOverridesStopLoading()` guard** · correctness · PLAUSIBLE

UIKit `reloadStopAnnotations` begins with `if searchResponseOverridesStopLoading() { return }`, suppressing region stop-loading while a single search result is displayed. The SwiftUI `onMapCameraChange` path calls `scheduleStopsRequest(in:)` gated only on zoom, with no search-response check.

**Failure:** If a single-result search is active on the shared `MapRegionManager`, a camera settle overwrites `mapRegionManager.stops` with region stops, so RoutePicker/CurrentTrip show the whole region's routes rather than the searched stop's routes. Lower confidence: search is not yet wired into the SwiftUI panel, so this requires the shared manager to hold a search-response set from elsewhere — worth guarding before search is wired up.

OBAKit/Sheet/Root/MapPanelRootView.swift
```diff
@@ -189,6 +187,60 @@ struct MapPanelRootView: View {
                 mapViewModel.onAppBecameActive()
             }
         }
+        .onChange(of: mapViewModel.didReceiveInitialLocation) { _, received in
+            // Recenter once on the first fix, so granting permission after
+            // launch moves the camera off the last region the user set.
+            guard received else { return }
+            centerOnUser()
```
aaronbrethorst: **Recenter-on-first-fix lost if the fix precedes a non-zero map size** · correctness · PLAUSIBLE

`MapViewModel` latches `didReceiveInitialLocation` exactly once on the first fix, and `.onChange` responds by calling `centerOnUser()`. But `centerOnUser()` has `guard mapSize != .zero else { return }`.

**Failure:** On a cold launch (or right after granting permission) a cached/fast fix commonly arrives before `.onGeometryChange` has reported the Map's size, so `mapSize` is still `.zero` and `centerOnUser()` bails. Because `didReceiveInitialLocation` is a one-shot latch that never flips again, `.onChange` never re-fires and the map is never recentered — defeating the feature this PR adds. Consider also re-attempting the recenter when `mapSize` transitions to non-zero while the initial-location flag is set but not yet honored.

OBAKit/Sheet/Root/MapPanelRootView.swift
```diff
@@ -61,28 +72,54 @@ struct MapPanelRootView: View {
...
+                ForEach(stopsObserver.stops) { stop in
+                    // Empty title on the Annotation suppresses the visual label
+                    // (this project's SwiftUI SDK does not expose
+                    // `.annotationTitles(.hidden)` on `Map`, so we route
+                    // accessibility through the icon's own `accessibilityLabel`
+                    // instead of a hidden title string).
+                    Annotation("", coordinate: stop.coordinate) {
+                        Image(uiImage: application.stopIconFactory.buildSquircleIcon(
+                            for: stop,
+                            isBookmarked: application.userDataStore.findBookmark(stopID: stop.id) != nil,
```
aaronbrethorst: **Toggling a bookmark doesn't restyle an already-visible pin** · correctness · PLAUSIBLE

Pin bookmark styling is read from `application.userDataStore` (not an observed dependency of `MapPanelRootView`), and `mapRegionManager(_:stopsUpdated:)` returns early when the stop set is unchanged.

**Failure:** A stop is visible with the neutral squircle icon; the user bookmarks it. No body re-evaluation occurs for that annotation, so the pin keeps the plain (non-brand) icon instead of switching to the bookmarked brand-colored icon while both surfaces are visible.

aaronbrethorst: **`findBookmark(stopID:)` decodes all bookmarks per stop, per render** · cleanup · CONFIRMED

`UserDataStore.bookmarks` JSON-decodes the whole bookmark list out of UserDefaults on every get, and `findBookmark(stopID:)` calls it inside the annotation `ForEach`. That builder runs for every stop in `stopsObserver.stops` each time `body` re-evaluates (every pan settle, every colorScheme change).

**Impact:** With N accumulated stops that is N full UserDefaults reads + JSON decodes on the main thread per render, producing visible map hitching. The UIKit path builds a single `bookmarksHash` once (`displayUniqueStopAnnotations`); the SwiftUI path should likewise decode bookmarks once into a `Set<StopID>` per body eval.

OBAKit/Mapping/StopIconFactory.swift
```diff
@@ -131,6 +132,108 @@ class StopIconFactory: NSObject {
...
+            routeType: stop.prioritizedRouteTypeForDisplay,
+            direction: stop.direction,
+            isBookmarked: isBookmarked,
+            isDarkMode: traits.userInterfaceStyle == .dark
+        )
+    }
+
+    /// Draws a squircle stop icon with the specified route type, direction, and
+    /// bookmarked state.
+    func renderSquircleIcon(routeType: Route.RouteType, direction: Direction, isBookmarked: Bool, isDarkMode: Bool) -> UIImage {
```
aaronbrethorst: **`renderSquircleIcon` copy-pastes `renderIcon`'s scaffolding** · cleanup · CONFIRMED

`renderSquircleIcon` repeats `renderIcon`'s `imageBounds`/`rect`/`UIGraphicsImageRenderer` scaffolding and the same `drawIcon` + `drawArrowImage` calls; only the background and border draw differ.

**Impact:** A future change to the shared icon geometry (arrowTrackSize inset, renderer sizing, glyph/arrow ordering) must be made in two places or the two icon variants silently diverge in layout. Consider extracting the shared scaffolding and parameterizing the background/border.

OBAKit/Sheet/DI/AppSheetViewFactory.swift
```diff
@@ -74,6 +77,13 @@ final class AppSheetViewFactory {
         MoreSheetHost(application: application)
     }
 
+    /// Bridges `AppSheetRoute.stopDetails` to the existing UIKit `StopViewController`
```
aaronbrethorst: **Doc comment names the wrong controller class** · cleanup · CONFIRMED

The comment says `stopDetailView` "Bridges `AppSheetRoute.stopDetails` to the existing UIKit `StopViewController`", but `StopDetailSheetHost` actually hosts `StopPageViewController` (a SwiftUI hosting controller).

**Impact:** The comment misdirects a maintainer to the wrong controller class when tracing what the stop-detail sheet renders. Same stale reference at `StopDetailSheetHost.swift:14`.

OBAKit/Mapping/MapRegionManager.swift
```diff
@@ -280,49 +287,161 @@ public class MapRegionManager: NSObject,
...
+        var mapRegion = region
+        mapRegion.span.latitudeDelta *= factor
+        mapRegion.span.longitudeDelta *= factor
+        return mapRegion
+    }
+
+    /// Publishes cached stops for `region` immediately (instant revisits),
+    /// publishing only the latest region — neighborhood persistence lives in
+    /// `MapStopsObserver`. A cache miss or cancelled task is a no-op.
+    private func serveCachedStops(in region: MKCoordinateRegion) async {
```
aaronbrethorst: **SwiftUI map publishes stops only via the cache round-trip** (correctness)

`scheduleStopsRequest(in:)` publishes stops ONLY via `serveCachedStops` (cache read) and `refreshStopCache` (cache write) — it never calls `requestStops(in:)`, the only method that assigns `self.stops` directly from the network response.

`CoreApplication.stopCacheRepository` is a lazy optional that returns nil when the stop-cache SQLite database fails to open (disk full / corruption). With the repository nil, `saveStopsToCache` is a no-op and `cachedStops(in:)` returns `[]`, so `serveCachedStops` always early-returns on `!cachedStops.isEmpty`.

Net result: the SwiftUI map panel shows **no stop pins at any zoom** despite successful `getStops` calls, whereas the UIKit `MapViewController` shows them normally. Same root cause also at lines 346 and 404, and the API-failure branch drops the error alert the UIKit path surfaces via `application.displayError`.

OBAKit/Sheet/Root/MapPanelRootView.swift
```diff
@@ -61,28 +82,53 @@ struct MapPanelRootView: View {
...
+                }
+            }
+        }
+        .onMapCameraChange(frequency: .onEnd) { context in
+            // Same stop-loading zoom gate as the UIKit region-change path.
+            isZoomedInForStops = context.rect.height <= MapRegionManager.requiredHeightToShowStops
+            guard isZoomedInForStops else {
+                // Zoomed out: clear stops and cancel any pending request.
+                application.mapRegionManager.cancelScheduledStopsRequest()
+                stopsObserver.reset()
```
aaronbrethorst, line 155: **Zoomed-out branch returns without resetting `showStopLabels`** (correctness)

`showStopLabels` is set only inside the zoomed-in path (lines 122–124). When `context.rect.height > requiredHeightToShowStops` the handler cancels the request, calls `stopsObserver.reset()`, and returns early — leaving `showStopLabels` at its last value.

Bookmark annotations render at every zoom level and their under-pin label shows whenever `showStopLabels` is true. Pinching out in one motion from street level to region level fires `.onEnd` once at the final camera, takes the early return, and leaves bookmark labels painted over the zoomed-out map until a later settle lands in the 7k–40k band that flips `showStopLabels` false — a visible label-clutter glitch that diverges from the UIKit gate.

OBAKit/Mapping/StopIconFactory.swift
```diff
@@ -103,31 +104,135 @@ class StopIconFactory: NSObject {
...
+
+    /// Draws the gradient squircle background. Bookmarked stops use the brand
+    /// gradient (echoing `Icons.squircleTransportIcon`); regular stops use a
+    /// neutral fill so bookmarks read as distinct.
+    private func drawSquircleBackground(rect: CGRect, isBookmarked: Bool, context: CGContext) {
+        context.pushPop {
+            let cornerRadius = rect.width * squircleCornerRadiusRatio
+            UIBezierPath(roundedRect: rect, cornerRadius: cornerRadius).addClip()
+
+            let baseColor = isBookmarked ? themeColors.brand : fillColor
```
aaronbrethorst, line 206: **Squircle colors resolve against ambient traits, not the requested traits** (correctness)

`buildSquircleIcon`/`renderSquircleIcon` thread `traits`/`isDarkMode` only into the NSCache key. The actual gradient, glyph, and border colors are dynamic UIColors resolved via `.cgColor` / `blended`→`getRed` against the ambient `UITraitCollection.current`, not the requested traits.

When the SwiftUI `@Environment(\.colorScheme)` passed as `traits` diverges from `UITraitCollection.current` — e.g. an ancestor applies `.preferredColorScheme(.dark)` while the system appearance is light, or the icon is built before ambient traits settle — the squircle draws with wrong-appearance colors. Because the wrong-appearance image is then stored under the correct-appearance cache key, every stop pin for that route/direction stays mis-colored for the rest of the session. Consider `resolvedColor(with:)` or `UITraitCollection.performAsCurrent(_:)`.

OBAKit/Sheet/Root/MapStopsObserver.swift
```diff
@@ -0,0 +1,194 @@
...
+            publish()
+        }
+    }
+
+    // MARK: - Bookmarks
+
+    /// `.bookmarksDidChange` is posted from the main actor, so hop straight in.
+    @objc
+    private nonisolated func bookmarksDidChange() {
+        MainActor.assumeIsolated {
```
aaronbrethorst: **`bookmarksDidChange` handler may trap off the main thread** (correctness)

The `@objc nonisolated bookmarksDidChange()` handler runs synchronously on whatever thread posts `.bookmarksDidChange` and immediately calls `MainActor.assumeIsolated { reloadBookmarks() }`. The comment asserts the notification is always main-actor-posted, but `.bookmarksDidChange` is posted by `UserDefaultsStore`, a plain `NSObject` (not `@MainActor`).

If a bookmark add/update/delete ever runs off the main thread (background sync, import, or data migration invoking `UserDataStore.add/delete`), the synchronous post reaches this handler on that background thread and `MainActor.assumeIsolated` traps, crashing the app. A `Task { @MainActor in … }` hop would be safe regardless of poster thread.

OBAKit/Sheet/Root/MapPanelRootView.swift
```diff
@@ -61,28 +82,53 @@ struct MapPanelRootView: View {
...
+        Map(position: $cameraPosition, selection: $selectedStopID) {
             UserAnnotation()
+            // Bookmark pins render at every zoom level, like the UIKit map.
+            ForEach(stopsObserver.bookmarks) { bookmark in
+                stopAnnotation(for: bookmark.stop, isBookmarked: true, label: bookmark.name)
+            }
+            // Regular stops show only when zoomed in; bookmarked stops are
+            // excluded (they already render above as bookmark pins).
+            if isZoomedInForStops {
+                ForEach(stopsObserver.stops.filter { !stopsObserver.bookmarkedStopIDs.contains($0.id) }) { stop in
```
aaronbrethorst: **Stop filter recomputed on every body evaluation** (cleanup)

`stopsObserver.stops.filter { !bookmarkedStopIDs.contains($0.id) }` is recomputed inside the `Map` content builder on every body evaluation. On every re-evaluation the full accumulated set (up to `renderCap=400`) is filtered against `bookmarkedStopIDs` and each survivor runs `Formatters.formattedTitle`, redoing O(n) work per frame in dense metros. Publishing a pre-filtered array from `MapStopsObserver` would avoid the repeated allocation and Set lookups.

OBAKit/Sheet/Root/MapPanelRootView.swift
```diff
@@ -61,28 +82,53 @@ struct MapPanelRootView: View {
...
+            // Same stop-loading zoom gate as the UIKit region-change path.
+            isZoomedInForStops = context.rect.height <= MapRegionManager.requiredHeightToShowStops
+            guard isZoomedInForStops else {
+                // Zoomed out: clear stops and cancel any pending request.
+                application.mapRegionManager.cancelScheduledStopsRequest()
+                stopsObserver.reset()
+                return
+            }
+            // Labels show only zoomed in close, on the standard map, with the default on.
+            showStopLabels = MapRegionManager.shouldShowExtraStopData(forVisibleMapRectHeight: context.rect.height)
```
aaronbrethorst: **Inline label gate duplicates `shouldHideExtraStopAnnotationData`** (cleanup)

This SwiftUI label gate re-implements the map-type-standard and `mapViewShowsStopAnnotationLabels` UserDefaults checks that `MapRegionManager.shouldHideExtraStopAnnotationData` already bundles. Only `shouldShowExtraStopData(forVisibleMapRectHeight:)` was factored out, so a change to the UIKit label criteria won't propagate here and the two surfaces will show labels under different conditions.

OBAKit/Sheet/Root/MapStopsObserver.swift
```diff
@@ -0,0 +1,194 @@
...
+            reloadBookmarks()
+        }
+    }
+
+    private func reloadBookmarks() {
```
aaronbrethorst, line 141: **`reloadBookmarks` duplicates the last-wins dedup logic** (cleanup)

The `bookmarksByStopID` last-wins dedup loop is duplicated from `MapRegionManager.displayUniqueStopAnnotations` (the comment even says "matching displayUniqueStopAnnotations"). If the precedence rule for duplicate bookmarks changes in one place, the SwiftUI observer and the UIKit annotation path will select different winning bookmarks for the same stop.

OBAKit/Mapping/StopIconFactory.swift
```diff
@@ -103,31 +104,135 @@ class StopIconFactory: NSObject {
...
+        }
+
+        squircleIconCache.setObject(image, forKey: key)
+        return image
+    }
+
+    /// Draws the gradient squircle background. Bookmarked stops use the brand
+    /// gradient (echoing `Icons.squircleTransportIcon`); regular stops use a
+    /// neutral fill so bookmarks read as distinct.
+    private func drawSquircleBackground(rect: CGRect, isBookmarked: Bool, context: CGContext) {
```
aaronbrethorst, line 201: **`drawSquircleBackground` copies gradient constants from `Icons.squircleTransportIcon`** (cleanup)

The gradient blend amounts (0.18 / 0.12) and `squircleCornerRadiusRatio` (0.28) are copied from `Icons.squircleTransportIcon` into `StopIconFactory`. A future tweak to the recent-stops squircle appearance in `Icons.swift` leaves the map-annotation squircle visually out of sync, since the constants live in two files with no shared source.

OBAKit/Mapping/MapRegionManager.swift
```diff
@@ -280,49 +287,161 @@ public class MapRegionManager: NSObject,
...
+        let cachedStops = cachedStops(in: fudgedRegion(for: region, factor: preferredLoadDataRegionFudgeFactor))
+        guard !cachedStops.isEmpty, !Task.isCancelled else { return }
+
+        await MainActor.run {
+            self.stops = cachedStops
```
aaronbrethorst: **Cache-serve triggers a full annotation diff on the offscreen mapView** (cleanup)

This `self.stops = …` assignment fires the `stops` `didSet` (~line 537), which runs `displayUniqueStopAnnotations()`. `MapStopsObserver` only receives updates as the tail (`notifyDelegatesStopsChanged`) of that method. So on every SwiftUI camera settle, `serveCachedStops` forces a full add/remove annotation diff against the manager's `MKMapView` that the SwiftUI surface never displays — main-thread work proportional to stop count on a hot path, purely to notify the observer.

aaronbrethorst (changes requested):
Getting close! see the feedback that's been left and investigate why unit tests are failing.

aaronbrethorst:
### Code review

Found 1 issue:

1. `scheduleStopsRequest(in:)` makes cancellation a routine event — every camera settle calls `pendingStopsRequestTask?.cancel()`, and `cancelScheduledStopsRequest()` fires on every zoom-out past the threshold — but the error path only recognizes `CancellationError`. When a `getStops` call is cancelled mid-flight, `URLSession.data(for:)` resumes with `URLError(.cancelled)`, which `APIService.data(for:)` rethrows unchanged as an `NSError` in `NSURLErrorDomain` (code `NSURLErrorCancelled`). That misses `catch is CancellationError`, falls into the generic `catch`, and — whenever `servedFromCache` is false (cold/empty cache, a newly-visited area, or `stopCacheRepository == nil`) — calls `application.displayError(error)`, which `ErrorClassifier` maps to `APIError.networkFailure` and `Application.displayError` presents as a modal `ErrorBulletin`. Net effect: panning or zooming out while a stop request is in flight over an uncached area pops a network-error bulletin at the user. The repo already documents the three shapes a cancelled request can arrive in and handles all of them — see `RoutePickerViewModel.loadRoutes`: `Task.isCancelled || error is CancellationError || (nsError.domain == NSURLErrorDomain && nsError.code == NSURLErrorCancelled)`. The same check is needed here. (bug due to `OBAKit/Mapping/MapRegionManager.swift` catching only `CancellationError` while `OBAKitCore/Network/APIService/APIService+GetData.swift` rethrows the raw `NSError`)

https://github.com/OneBusAway/onebusaway-ios/blob/921bec576f6230fde8677676f2179f50932bf968/OBAKit/Mapping/MapRegionManager.swift#L445-L455

aaronbrethorst (changes requested):
There's a lot of good work in here, and most of my earlier round looks addressed. The cache-first serve-then-refresh sequencing is thoughtful — publishing the band on both sides of the network call so you never go band → narrow → band is exactly the right instinct, and the comment explaining why is the kind I want in this file.

One thing blocks the merge.

**Cancelled stop requests surface a modal error bulletin to the user.** In `scheduleStopsRequest(in:)`, the refresh is wrapped in `catch is CancellationError { return }` followed by a generic `catch` that calls `application.displayError(error)` whenever `servedFromCache` is false. But a cancelled `URLSession` request doesn't arrive as `CancellationError` — it comes through as an `NSError` in `NSURLErrorDomain` with code `NSURLErrorCancelled` (-999), which sails past the typed catch into the generic one.

Since `scheduleStopsRequest` cancels the previous task on every camera settle and `cancelScheduledStopsRequest()` fires on every zoom-out past the threshold, this is the hot path, not an edge case: pan across an uncached area twice in quick succession and the user gets an `ErrorBulletin` for a request we cancelled on purpose.

We already know about this shape and have a helper for it — `Error.isCancellation` in `OBAKitCore/Extensions/FoundationExtensions.swift` matches both forms and exists specifically because the typed catch isn't sufficient. `RoutePickerViewModel.loadRoutes` and `TripViewModel` both guard all the forms. Swapping the typed catch for that helper (plus a `Task.isCancelled` check, per the helper's own doc comment) should be it.

Also needs doing before this can land, independent of the review: the branch is in a conflicting state against `main`. Please merge or rebase `main` in — #1215, #1219 and #1220 went in today.

A handful of smaller things I noticed but am not blocking on; use your judgment:

- `showStopLabels` in `MapPanelRootView` is only recomputed inside `.onMapCameraChange`, so toggling map type or the `mapViewShowsStopAnnotationLabels` default leaves labels stale until the user next moves the map. The UIKit path re-evaluates per annotation view.
- `serveCachedStops` runs the GRDB bounding-box query before checking `Task.isCancelled`, so already-cancelled tasks still pay for it. With the re-serve and the cache write, each camera settle is up to two SQLite reads plus a write on the main actor.
- `MapLabelOutline` reuses the `ViewModifier.Content` proxy nine times in one body. That isn't a documented-supported use of `Content`, and it renders each label 9× — noticeable on a dense viewport.
- `MapStopsObserver.reloadBookmarks()` only re-runs on `.bookmarksDidChange` but filters by `application.currentRegion`, so a region change can leave stale bookmark pins until some unrelated bookmark edit reposts the notification.

Fix the cancellation handling and get it rebased, and I'll take another pass.

aaronbrethorst:
Merging in #1250

Issue #1189, opened 2026-07-14
Live Activities: tracking creates duplicate activities.


diveshpatil9104:
While testing the Live Activities from #1179, I ran into a couple of related issues with starting a Live Activity from the "Track" action. Filing them together since they seem connected.

### Problem 1 — Tracking the same bookmark twice makes two Live Activities

Steps:
1. Turn on Live Activities for the app.
2. Bookmarks tab → long-press a bookmark → tap **Track**.
3. Long-press the **same** bookmark again → tap **Track** again.

Now there are **two** Live Activities for the same stop, instead of one. Tapping "Track" again on something that's already being tracked doesn't seem to check if it's already running.

### Problem 2 — Tracking a second bookmark doesn't switch the Dynamic Island

Steps:
1. Track bookmark **A** → the Dynamic Island shows **A**.
2. Then track a different bookmark **B**.

The Dynamic Island **still shows A**, not the new one (B). Both activities stay active at the same time — the old one is never replaced — so B never shows up even though I just tracked it.

Related: #1179, #609

Borisserz:
Taking Problem 2 of this issue (Dynamic Island not switching to the newly tracked trip).

Problem 1 was already addressed in #1215. Plan for Problem 2 (app-side):
- Assign a monotonic `relevanceScore` when starting a Live Activity so the newest Track wins the Dynamic Island
- Demote other live peers' scores
- Preserve `relevanceScore` on local content updates (today's update path rebuilds `ActivityContent` with the default score of 0, which undoes prominence)

OBACloud push `relevance-score` alignment stays a follow-up if server pushes flatten scores — same caveat noted in #1215.

PR #1215, opened 2026-07-24
 Prevent duplicate Live Activities for the same stop and route

diveshpatil9104:
## Summary

Partially addresses #1189 (Problem 1).

Tapping **Track** on a bookmark that was already being tracked — or starting the same trip from both the bookmarks list and the stop page — created a second, independent Live Activity for the same stop. The result was two Lock Screen cards and two OBACloud push registrations for a single bus.

## Root cause

Neither start path checked whether an activity was already running before calling `Activity.request(...)`. Every request minted a fresh `activity.id`, and nothing downstream deduplicates by content — `LiveActivityTracker` and `LiveActivityRegistry` are both keyed on that id. This PR adds the missing check.

## What changed

- **`TripAttributes.StaticData.tracksSameTrip(as:)` (new)** — the single definition of Live Activity identity: matching `stopID`, `routeShortName`, and `routeHeadsign`. `routeColorHex` and `regionID` are deliberately excluded, since color is `nil` until arrivals load and including it would let a fast second tap slip through.
- **`Activity<TripAttributes>.running(matching:)` (new)** — finds an already-running activity for a given identity, filtered through `LiveActivityRegistry.isLive(_:)` so a *dismissed* activity can still be re-tracked.
- **Guard added to both start paths** — `BookmarksViewController` and `StopPageViewController` now check before calling `Activity.request(...)`. A duplicate tap re-shows the "Tracking on Lock Screen" confirmation rather than silently doing nothing.
- **`updateRunningLiveActivities()` now delegates to `tracksSameTrip(as:)`** instead of hand-rolling the same three-field comparison, so the reconciliation match and the new guard can't drift apart.
- **Toast/HUD dismissals are now cancellable** — both the `StopViewModel` dismiss task and the `ProgressHUD.showSuccessAndDismiss` timer. Previously, a second tap late in an existing toast's window inherited the first toast's imminent dismissal and vanished almost immediately: exactly the "looks like nothing happened" case this fix is about.

## Before / After

<table>
<tr>
<th width="50%">Before</th>
<th width="50%">After</th>
</tr>
<tr>
<td><video src="https://github.com/user-attachments/assets/2a644d22-1ef5-4834-85fb-6f13dd8f743c"></video></td>
<td><video src="https://github.com/user-attachments/assets/b392d497-0437-4a76-b793-9685fa474caa"></video></td>
</tr>
<tr>
<td>Tracking the same bookmark twice → two Live Activities on the Lock Screen.</td>
<td>The second tap is a no-op, one Live Activity remains, and the confirmation still shows.</td>
</tr>
</table>

## Out of scope

This PR covers Problem 1 of #1189 only. Problem 2 (the Dynamic Island not switching to the newest tracked trip) is a separate change — it needs `relevanceScore` on both the app and the OBACloud push side — so it is intentionally left out.

## Summary by CodeRabbit

- **New Features**
  - Improved Live Activity trip matching to prevent duplicate Lock Screen cards and registrations.
  - Matching now consistently identifies trips using stop, route, and destination details.

- **Bug Fixes**
  - Prevented earlier success notifications from dismissing newer ones prematurely.
  - Stabilized Live Activity confirmation messages when duplicate start requests occur.
  - Added coverage for trip identity matching, including route color, region, and destination edge cases.

### DIFF
--- a/OBAKit/Bookmarks/BookmarksViewController.swift
+++ b/OBAKit/Bookmarks/BookmarksViewController.swift
@@ -218,6 +218,16 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
             regionID: application.currentRegion?.regionIdentifier ?? 0
         )
 
+        // Tapping Track again on a bookmark that is already tracked — or tracking
+        // the same trip from the stop page — would otherwise mint a second
+        // activity for one stop, leaving the user with duplicate Lock Screen
+        // cards and duplicate OBACloud push registrations.
+        if Activity<TripAttributes>.running(matching: staticData) != nil {
+            Logger.info("Live Activity already running for stop \(staticData.stopID) route \(staticData.routeShortName); not starting a duplicate.")
+            showLiveActivityStartedToast()
+            return
+        }
+
         guard let contentState = buildContentState(from: arrivalDepartures) else {
             // Shouldn't happen — the context menu only offers Track once arrival
             // data has loaded — but if data was cleared between the menu render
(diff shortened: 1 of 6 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is a well-built fix. A few things I specifically checked and found solid:

- Excluding `routeColorHex` from identity is load-bearing, not an oversight — the color arrives with the first arrivals payload, so folding it in would let a fast second tap through in exactly the case the guard exists to prevent. Good that a test pins it.
- Filtering through `LiveActivityRegistry.isLive` is the right call. Dismissed and ended activities stay in `Activity.activities`, so a bare presence check would have refused to start an activity the user had already swiped away.
- Routing `updateRunningLiveActivities()` through the same `tracksSameTrip(as:)` rule is behavior-preserving, and it means the match and the guard can't drift apart later.
- Re-showing the confirmation on the short-circuit path instead of silently returning is the right product call.

Two notes for later, neither blocking:

1. `signalLiveActivityStarted()` still has a hairline race. If the previous task's `Task.sleep` has already resumed and its continuation is sitting on the MainActor queue, `cancel()` arrives too late and that continuation sets `liveActivityStarted = false` right after the new signal set it true. A `try Task.checkCancellation()` between the sleep and the assignment closes it. Sub-millisecond window, so I'm not holding the merge on it — but the doc comment's "supersedes the pending dismiss rather than racing it" is a touch stronger than what the code guarantees. The `ProgressHUD` sibling change doesn't have this problem, since `DispatchWorkItem.cancel()` before execution is reliable on a serial queue.

2. The stop-page duplicate path now returns early without refreshing or re-arming the existing activity. Still strictly better than minting a duplicate, and bookmarked stops get covered by `updateRunningLiveActivities()`, but there's no equivalent refresh loop on the stop page. Worth a follow-up issue rather than more work here.

Merging. Thanks — the comments explaining *why* each field is in or out of the identity rule are the part that'll keep this correct a year from now.

PR #1219, opened 2026-07-25
Fix URLSchemeRouter doc comment: parameters are stopID/regionID, not stop_id/region_id

Shapy-Sees:
The doc comment on `URLSchemeRouter` shows

    onebusaway://view-stop?region_id=1&stop_id=12345

but `decodeViewStop` only reads the camelCase names (`stopID`, `regionID`), so a link built from the comment's example fails to decode and the app just opens to wherever it was. I integrated the scheme in my own app from the comment's example and hit exactly that.

Comment-only change.

## Summary by CodeRabbit

* **Documentation**
  * Updated the custom URL scheme deep-link example to use the current `stopID` and `regionID` query parameters.

### DIFF
--- a/OBAKitCore/DeepLinks/URLSchemeRouter.swift
+++ b/OBAKitCore/DeepLinks/URLSchemeRouter.swift
@@ -57,7 +57,7 @@ public enum URLType {
 }
 /// Provides support for deep linking into the app by way of a custom URL scheme.
 ///
-/// Custom URL scheme deep linking (e.g. `onebusaway://view-stop?region_id=1&stop_id=12345`)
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Nice catch — that example URL has been wrong since it was written in 2020, in the same commit that introduced the camelCase query items.

I verified the correction against the code: `encodeViewStop` emits `stopID` and `regionID`, `decodeViewStop` reads those exact names, and the query-item lookup is a case-sensitive exact match — so the old snake_case example genuinely would not have decoded. The new example is right, and it even matches the order `encodeViewStop` emits.

Also confirmed nothing else in the repo still documents the old names. Thanks for fixing it.

PR #1220, opened 2026-07-25
Fix: Sort filter routes lexicographically with numeric awareness (#1218)

### LINKED ISSUE #1218: Filter Routes should be sorted lexicographically
That is to say: 3, 10, 11, 12, 49, Bellevue

This function ought to do the trick for sorting the list of routes: https://developer.apple.com/documentation/swift/sequence/lexicographicallyprecedes(_:)

Image

Borisserz:
## Summary
Fixes #1218.

This PR updates route list sorting (`localizedCaseInsensitiveSort()`) in `OBAKitCore` to use `localizedStandardCompare(_:)` instead of `localizedCaseInsensitiveCompare(_:)`.

### Behavior Change
- **Before:** Route numbers were sorted by alphabetical code-point comparison (`"10", "11", "12", "3", "49", "Bellevue"`).
- **After:** Route numbers are sorted with natural numeric awareness (`"3", "10", "11", "12", "49", "Bellevue"`).

### Changes
1. Updated `Sequence where Element == Route`'s `localizedCaseInsensitiveSort()` to use `localizedStandardCompare`.
2. Updated `Sequence where Element == String`'s `localizedCaseInsensitiveSort()` to use `localizedStandardCompare`.
3. Added unit test `test_arrayExtensionSort_numeric()` in `RouteTests.swift`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sorting for strings and route names by recognizing numeric values naturally.
  * Route lists now sort values such as “3” before “10” for more intuitive ordering.

* **Tests**
  * Added coverage to verify numeric-aware route sorting behavior.

### DIFF
--- a/OBAKitCore/Extensions/FoundationExtensions.swift
+++ b/OBAKitCore/Extensions/FoundationExtensions.swift
@@ -249,7 +249,7 @@ public extension Sequence where Element == String {
     /// - Returns: A localized, case-insensitive sorted Array.
     func localizedCaseInsensitiveSort() -> [Element] {
         return sorted { (s1, s2) -> Bool in
-            return s1.localizedCaseInsensitiveCompare(s2) == .orderedAscending
+            return s1.localizedStandardCompare(s2) == .orderedAscending
         }
     }
 }
--- a/OBAKitCore/Models/REST/References/Route.swift
+++ b/OBAKitCore/Models/REST/References/Route.swift
@@ -179,7 +179,7 @@ public extension Sequence where Element == Route {
     /// - Returns: A localized, case-insensitive sorted Array.
     func localizedCaseInsensitiveSort() -> [Element] {
         return sorted { (s1, s2) -> Bool in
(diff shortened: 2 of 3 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is the right fix. `localizedStandardCompare` is already what we use for user-facing ordering elsewhere in OBAKitCore (`Formatters`, `GroupedArrivalDeparture`, the stop page headers), so this removes an inconsistency rather than introducing a new convention — and it's no more locale-sensitive than the comparator it replaces.

I checked that the new test actually pins the behavior: under the old `localizedCaseInsensitiveCompare` that fixture sorts to `["10","11","12","3","49","Bellevue"]`, so it genuinely fails pre-change. The existing `test_arrayExtensionSort` still holds, since Finder-style collation is case-insensitive too.

One non-blocking follow-up for me, not you: `localizedCaseInsensitiveSort()` and its doc comment now under-describe what the helper does. That's public API in a white-label framework, so a rename is worth doing on its own.

Thanks for tracking this down to the right helper instead of patching the view.

PR #1224, opened 2026-07-26
Show the predicted arrival time, not just the scheduled one

### LINKED ISSUE #1214: make it clearer that times on the arrival times list are schedule rather than real-time
The list of arrivals shows the scheduled arrival time, rather than the predicted arrival time given the current real-time information, even when real-time information is available.  I just responded to a report from a rider who was confused by this.

A fix would be to show the scheduled time in strikethrough if it's late, followed by the actual predicted arrival time.  (I like showing the actual predicted time rather than the number of minutes late -- then the rider doesn't need to compute the time it arrives, and it should be much clearer which is the scheduled time and which is the predicted time.)

If it's on time, show the time and "on time", as in the current design.

aaronbrethorst:
Fixes #1214.

## Summary

- The new Stop page printed `scheduledDate` at all five of its clock-time sites while feeding the countdown `arrivalDepartureMinutes`, which is prediction-derived. A late bus rendered `7m` beside `10:42 AM` with nothing to say which to trust, so riders read the timetable time and missed the bus. Departures now show the corrected time with the scheduled time struck through beside it.
- The strikethrough decision compares the **formatted strings**, not the dates. That avoids rendering `10:42 AM 10:42 AM` for a prediction seconds off the timetable, and it also fixes a case a "strike through only when late" rule would miss: `ScheduleStatus` calls anything inside ±1.5 min "on time", but such a trip still lands on a different clock minute and still needs its time corrected.
- VoiceOver can't perceive a strikethrough, so the chronological rows and both grouped-card labels now speak "scheduled X, now expected Y". Two new strings, translated into all 12 non-English locales.
- Also fixes the schedule-deviation baseline in `ArrivalDeparture`: it measured against `scheduledDeparture` even for arrivals, reporting a late arrival as *early* by the length of the layover. Showing both times side by side would have made that disagreement visible on screen.

Only the new Stop page was affected. The old page, bookmarks, and `Formatters` already displayed `arrivalDepartureDate`.

## Test plan

- [x] `OBAKitTests` — 1452 tests, 0 failures
- [x] New `DepartureTimeDisplayTests` (10): late, early, same-minute, different-minute-inside-the-on-time-band, schedule-only, and a feed that carries predicted times while declaring `predicted: false`
- [x] New `ArrivalDepartureTests` deviation tests pin both the arriving and departing branch of the layover fix
- [x] Locale key parity across all 13 locales; `plutil -lint` clean on each
- [x] Rendered every state through `ImageRenderer` at default size and at `accessibility3`, confirming the two times wrap one-per-line at large type
- [x] SwiftLint clean (one pre-existing `file_length` violation in `StopPageViewController.swift`, untouched here)

## Review notes

Non-blocking findings from the cleanup pass that I deliberately left out of scope:

1. **`DepartureTimeDisplay` lives in OBAKit, so the widget and Live Activity can't adopt it.** They import only `OBAKitCore`. I left it in place because neither surface exhibits #1214 today — `BookmarkCardView` and `TripActivityPresenter` already render `arrivalDepartureDate` — and moving the two strings between framework bundles is a separate risk worth its own change. Tracked in #1225.
2. **`BookmarkCardView` still renders a bare time with no strikethrough.** Migrating it is a one-liner but it's a product decision about bookmark cards, not part of this issue.
3. **These values are rebuilt per render rather than per data refresh.** `StopPageListBuilder` is the natural home if it ever shows up in a profile; after the cleanup commit the render path costs one or two `DateFormatter` calls and no bundle lookups, close to what it cost before.
4. **Pre-existing, old stop page:** `Formatters.formattedScheduleDeviation` picks its wording from `deviationFromScheduleInMinutes == 0` while the color comes from `scheduleStatus`'s ±1.5 min band, so a 1-minute-late trip renders a green "arrives 1 min late".
5. **Pre-existing, localization:** `stop_page.status.late_fmt` / `early_fmt` have no `.stringsdict` plural rules, and OBAKitCore has no stringsdict at all, so "1 min late" is ungrammatical in several locales.

## Summary by CodeRabbit

* **New Features**
  * Improved stop-page departure time presentation to clearly show scheduled vs predicted (expected) times, including strikethrough when rescheduled.
  * Enhanced VoiceOver labeling for departure times, including localized, rescheduled-time phrasing.
* **Bug Fixes**
  * Schedule deviation calculations now correctly measure deviation for arriving vs departing services. 
* **Tests**
  * Added coverage for departure time display behavior and schedule deviation scenarios.

### DIFF
--- a/OBAKit/Stops/StopPage/Departures/DepartureRowView.swift
+++ b/OBAKit/Stops/StopPage/Departures/DepartureRowView.swift
@@ -42,8 +42,8 @@ struct DepartureRowView: View {
 
     private var showsAlarmAffordance: Bool { canAlarm || hasAlarm }
 
-    private var scheduledTimeText: String {
-        formatters.timeFormatter.string(from: departure.scheduledDate)
+    private var timeDisplay: DepartureTimeDisplay {
+        DepartureTimeDisplay(arrivalDeparture: departure, formatters: formatters)
     }
 
     var body: some View {
@@ -60,9 +60,8 @@ struct DepartureRowView: View {
                     countdown
                 }
                 headsignText
-                Text(scheduledTimeText)
+                DepartureTimeText(display: timeDisplay)
                     .font(.footnote)
-                    .monospacedDigit()
                     .foregroundStyle(.secondary)
                 statusText
                 occupancyBadge
@@ -72,9 +71,8 @@ struct DepartureRowView: View {
                     VStack(alignment: .leading, spacing: 3) {
                         headsignText
                         HStack(spacing: 6) {
-                            Text(scheduledTimeText)
+                            DepartureTimeText(display: timeDisplay)
                                 .font(.footnote)
-                                .monospacedDigit()
                                 .foregroundStyle(.secondary)
                             Text("·").foregroundStyle(.tertiary)
                             statusText
@@ -164,7 +162,9 @@ struct DepartureRowView: View {
     }
 
     private var accessibilityText: String {
-        var clauses = [baseAccessibilityText]
+        // The clock time follows the countdown: VoiceOver can't perceive the
+        // strikethrough that carries this on screen.
+        var clauses = [baseAccessibilityText, timeDisplay.accessibilityTimeDescription]
 
         if status.showsOccupancy, let occupancy = departure.occupancyStatus, occupancy != .unknown {
             clauses.append(OccupancyBadge.localizedDescription(occupancy))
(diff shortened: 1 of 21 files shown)


aaronbrethorst:
### Code review

Five independent passes (CLAUDE.md adherence, shallow bug scan, git history, prior PR review comments, in-file comment compliance) turned up nothing above the reporting threshold. Two things worth recording:

The deviation-baseline change is a 2019 oversight rather than a deliberate choice being undone. [`4bc72091`](https://github.com/OneBusAway/onebusaway-ios/commit/4bc72091) added the arrival/departure-aware `scheduledDate` property and, in the same patch, wrote `rawDeviationFromScheduleInMinutes` against `scheduledDeparture` a few lines below it. [`9db69120`](https://github.com/OneBusAway/onebusaway-ios/commit/9db69120) ("Fixes the late/early calculation for arrival departures") corrected a units bug on that line and left the mismatch in place.

https://github.com/OneBusAway/onebusaway-ios/blob/4d06e75061535a175bd2b327529e1ed825db5938/OBAKitCore/Models/REST/ArrivalDeparture.swift#L304-L312

The row's unary-root / List fast-path constraint is preserved: `DepartureTimeText` branches internally, but both arms return `Text`, and it sits well below each row's root.

https://github.com/OneBusAway/onebusaway-ios/blob/4d06e75061535a175bd2b327529e1ed825db5938/OBAKit/Stops/StopPage/Shared/DepartureTimeDisplay.swift#L108-L117

PR #1228, opened 2026-07-26
Fix natural sorting in route filters

### LINKED ISSUE #1218: Filter Routes should be sorted lexicographically

mitre88:
## Summary
- sort route short names with `localizedStandardCompare` so numeric route names use natural order
- add regression coverage for the reported `3, 10, 11, 12, 49, Bellevue` sequence

Closes #1218

## Validation
- `swift -e` assertion for the exact natural-sort sequence
- `git diff --check`
- XcodeGen project generation succeeded with XcodeGen 2.46.0

The focused XCTest could not run locally because this machine has no available iOS Simulator runtime; the project CI can execute `OBAKitTests/RouteTests`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved route sorting to use natural numeric ordering, producing more intuitive results for route names containing numbers.
  * Maintained localized, case-insensitive sorting behavior.

* **Tests**
  * Expanded route sorting coverage with numeric and named route examples.
  * Added validation for the complete sorted result.

### DIFF
--- a/OBAKitCore/Models/REST/References/Route.swift
+++ b/OBAKitCore/Models/REST/References/Route.swift
@@ -174,12 +174,12 @@ public class Route: NSObject, Identifiable, Codable, HasReferences {
 
 public extension Sequence where Element == Route {
 
-    /// Performs a localized case insensitive sort on the receiver.
+    /// Performs a localized, natural sort on route short names.
     ///
-    /// - Returns: A localized, case-insensitive sorted Array.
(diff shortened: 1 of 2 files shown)


mitre88:
Closing as a duplicate of #1220, which already implements this fix and includes broader coverage. Thanks for the earlier work.

PR #1231, opened 2026-07-27
ci: pin third-party GitHub Actions to full commit SHAs

### LINKED ISSUE #1146: Harden GitHub Actions supply chain: pin third-party actions to commit SHAs
Follow-up from PR #1143.

Our workflows reference third-party and first-party actions by mutable tag rather than immutable commit SHA, e.g. in `.github/workflows/obakittests.yml`:

- `actions/checkout@v4`
- `actions/cache@v4`
- `slidoapp/xcresulttool@v3.1.1` (runs with `checks: write` + access to `github.token`)

A version tag can be force-moved by the upstream maintainer to point at different code, so tag-pinning leaves us exposed to a compromised or malicious upstream. GitHub's [hardening guidance](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) recommends pinning third-party actions to a full-length commit SHA.

For reference, `slidoapp/xcresulttool@v3.1.1` currently resolves to commit `0930e06ae335d8c5e2d2f311ee98d495889a6103`.

This is a repo-wide concern that shouldn't block the scoped CI-diagnosability fix in #1143; tracking separately so we can pin all actions consistently (ideally with Dependabot's `github-actions` ecosystem keeping the SHAs current). Also worth folding in a `yamllint`/trailing-whitespace check so workflow edits stay clean.

Borisserz:
## Summary

Pins all third-party and first-party GitHub Actions to immutable full-length commit SHAs instead of mutable tags (e.g. `@v4`), as requested in #1146.

## Problem

Using mutable version tags (e.g. `actions/checkout@v4`) leaves the repository exposed to supply chain attacks if an action's maintainer is compromised and a tag is force-moved to a malicious commit. This is especially sensitive for actions like `slidoapp/xcresulttool` which run with `checks: write` access.

## Solution

Following [GitHub's security hardening guidelines](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions), all actions are now pinned to their current resolved commit SHAs. 

The version tags remain as comments for readability (e.g. `@11d5960a326750d5838078e36cf38b85af677262 # v4`).

Fixes #1146

## Summary by CodeRabbit

* **Chores**
  * Pinned automated build, test, caching, and deployment actions to immutable versions.
  * Preserved existing workflow behavior while improving reliability and consistency.

### DIFF
--- a/.github/workflows/testflight.yml
+++ b/.github/workflows/testflight.yml
@@ -18,10 +18,10 @@ jobs:
     timeout-minutes: 45
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
 
     # Cache Swift Package Manager dependencies.
-    - uses: actions/cache@v4
+    - uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
       with:
         path: .build
         key: ${{ runner.os }}-spm-${{ hashFiles('**/Package.resolved') }}
@@ -29,7 +29,7 @@ jobs:
           ${{ runner.os }}-spm-
 
     # Cache Ruby gems (fastlane and its dependencies).
-    - uses: actions/cache@v4
+    - uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
       with:
         path: vendor/bundle
         key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -31,9 +31,9 @@ jobs:
       contents: read
 
     steps:
-    - uses: actions/checkout@v7
+    - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
-    - uses: actions/cache@v6
+    - uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
       with:
         path: .build
         key: ${{ runner.os }}-spm-${{ hashFiles('**/Package.resolved') }}
@@ -44,7 +44,7 @@ jobs:
     # resolved dependency graph, the XcodeGen project specs, or this workflow
     # change. The previous `**/*.yml` glob also matched vendored yml under
     # .build/checkouts and caused needless cache misses (→ full rebuilds).
-    - uses: actions/cache@v6
+    - uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
       id: deriveddata-cache
       with:
         path: DerivedData
@@ -75,7 +75,7 @@ jobs:
     # (see a1244c48). The trade-off is that the toolchain is no longer pinned, so a
     # new beta landing in the image can move the concurrency warning count below.
     - name: Setup Xcode
-      uses: maxim-lobanov/setup-xcode@v1
+      uses: maxim-lobanov/setup-xcode@ed7a3b1fda3918c0306d1b724322adc0b8cc0a90 # v1.7.0
       with:
         xcode-version: 'latest'
 
@@ -238,7 +238,7 @@ jobs:
     # writes a lighter summary to $GITHUB_STEP_SUMMARY instead. If test-retry flags
     # are ever added, broaden this to `always()` so flaky-but-passing runs still
     # surface which tests retried.
-    - uses: slidoapp/xcresulttool@v3.1.1
+    - uses: slidoapp/xcresulttool@0930e06ae335d8c5e2d2f311ee98d495889a6103 # v3.1.1
       continue-on-error: true
       if: failure()
       with:


aaronbrethorst:
### Code review

aaronbrethorst (changes requested):
Good change to make, and I'm glad someone did it — mutable tags are exactly the supply-chain hole worth closing, and pinning all twelve `uses:` lines rather than a subset is the right scope.

I verified every pin independently rather than taking the comments at face value, since that's the entire value of this PR:

| Action | Pin | Result |
|---|---|---|
| `actions/checkout` | `11d5960a…` | commit, matches `refs/tags/v4` |
| `actions/cache` | `0057852b…` | commit, matches `refs/tags/v4` |
| `actions/upload-pages-artifact` | `56afc609…` | commit, matches `refs/tags/v3` |
| `actions/deploy-pages` | `d6db9016…` | commit, matches `refs/tags/v4` |
| `slidoapp/xcresulttool` | `0930e06a…` | commit — annotated tag `v3.1.1` correctly dereferenced |
| `maxim-lobanov/setup-xcode` | `1242409711ff…` | **not a commit** — see below |

Five of six are exactly right, including the `xcresulttool` one where you correctly walked the annotated tag through to its commit.

**`setup-xcode` is pinned to the tag object, not the commit.** `gh api repos/maxim-lobanov/setup-xcode/commits/1242409711ff…` returns 422 "No commit found for SHA"; that SHA is the annotated tag object for `refs/tags/v1`, which dereferences to commit `ed7a3b1fda3918c0306d1b724322adc0b8cc0a90` (v1.7.0). Actions does resolve tag-object SHAs, so CI won't break, and a git tag object is content-addressed so the immutability guarantee technically survives — but anyone auditing the pin against the action's commit history gets "commit not found," and it's inconsistent with how you handled `xcresulttool` two lines away. Worth making:

```yaml
uses: maxim-lobanov/setup-xcode@ed7a3b1fda3918c0306d1b724322adc0b8cc0a90 # v1.7.0
```

**This also needs a rebase.** `0b768cc5` ("Rearrange GitHub Actions") landed on `main` after you opened this and reshuffled all three workflow files, so the branch is conflicted. Related: there's no `build` check on this PR at all — only CodeRabbit and the CLA ran — so nothing has actually exercised these workflows. The rebase will fix that too.

One optional polish while you're in there: the version comments are major-only (`# v4`, `# v3`, `# v1`). Pinning to exact releases in the comment (`# v4.4.0`, `# v4.3.0`, `# v3.0.1`, `# v4.0.5`, `# v1.7.0`) makes future bumps much easier to reason about, since the SHA alone tells you nothing about how far behind you are.

Rebase, fix the one pin, and let's get a green build on it — then this is an easy merge.

Borisserz:
Addressed the review:

1. **Rebased** onto current `main` (the old branch conflicted after the workflow rearrange; `docs.yml` is gone upstream, so this PR only touches the workflows that still exist).
2. **`setup-xcode`** is now pinned to commit `ed7a3b1…` (v1.7.0), not the annotated tag object.
3. **Exact release comments** on every pin (`# v7.0.1`, `# v6.1.0`, `# v4.4.0`, `# v4.3.0`, `# v1.7.0`, `# v3.1.1`).
4. Pins match today's majors on `main` (`checkout@v7` / `cache@v6` in tests, `checkout@v4` / `cache@v4` in testflight).

Ready for re-review once CI is green.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Both blockers are cleared, and I re-verified every pin from scratch rather than diffing against my last pass — that independent check is the entire value of a PR like this, so it shouldn't get cheaper the second time.

| Action | Pin | Resolves to a real commit? | Tags |
|---|---|---|---|
| `actions/checkout` (tests.yml:34) | `3d3c42e5…` | yes | `v7`, `v7.0.1` |
| `actions/cache` (tests.yml:36, :47) | `55cc8345…` | yes | `v6`, `v6.1.0` |
| `maxim-lobanov/setup-xcode` (tests.yml:78) | `ed7a3b1f…` | yes | `v1.7.0` |
| `slidoapp/xcresulttool` (tests.yml:241) | `0930e06a…` | yes | annotated `v3.1.1` → tag obj `cb91005a…` → this commit |
| `actions/checkout` (testflight.yml:21) | `11d5960a…` | yes | `v4`, `v4.4.0` |
| `actions/cache` (testflight.yml:24, :32) | `0057852b…` | yes | `v4`, `v4.3.0` |

The `setup-xcode` fix is exactly right — `ed7a3b1fda3918c0306d1b724322adc0b8cc0a90` is a genuine commit carrying `refs/tags/v1.7.0`, where the old pin was the annotated tag object. And you kept the `xcresulttool` dereference correct through the rebase; I walked `v3.1.1` → `cb91005a…` → `0930e06a…` myself and it lands on your pin.

Coverage is complete: all 8 `uses:` lines across both workflows are SHA-pinned, with none missed. The rebase is clean, `docs.yml` is gone upstream so those two pins correctly disappeared, and `build` passed on the head SHA — which was the other thing missing last time, since nothing had ever exercised these workflows.

You also took the optional note about exact-version comments (`# v4.4.0`, `# v6.1.0`) rather than major-only. That'll make future bumps much easier to reason about, so thank you.

One follow-up before #1146 can close: there's no `.github/dependabot.yml`, and the issue asks for the `github-actions` ecosystem to keep these current. That matters more now than it did — with the mutable tags gone there's nothing left to auto-follow, so these SHAs will rot silently. Worth a small follow-up PR.

Two things I looked at and deliberately left alone: `persist-credentials: false` on the checkout steps is reasonable hardening but pre-existing posture and scope creep here, and the `checkout` major divergence (v7.0.1 in tests.yml vs v4.4.0 in testflight.yml) is faithfully preserved from `main` — worth a separate bump now that the SHA hides the major.

Merging. Good, careful work on a change where carefulness is the whole point.

PR #1232, opened 2026-07-27
fix: make date strategy symmetric in RegionsFileStorage

### LINKED ISSUE #1152: RegionsFileStorage: encode/decode date-strategy asymmetry is a latent round-trip footgun
Follow-up from PR #1103 (region storage migration to disk).

`RegionsFileStorage` saves with a plain `JSONEncoder()` (default `.deferredToDate` date strategy) but loads with `JSONDecoder.RESTDecoder()`, which sets `dateDecodingStrategy = .millisecondsSince1970`. See `OBAKitCore/Location/Regions/RegionsFileStorage.swift` (`saveDefaultRegions`/`saveCustomRegion` vs `loadDefaultRegions`/`loadCustomRegions`).

This is harmless today because `Region` has no `Date` properties, so no value round-trips through a date strategy. But if `Region` ever gains a `Date` field, saved files will silently fail to round-trip (encode as seconds-since-reference-date, decode as millis-since-1970).

Fix: use a matching encoder (e.g. a `RESTEncoder` with `.millisecondsSince1970`, or set the strategy explicitly on the `JSONEncoder`) so encode/decode are symmetric, and/or add a brief comment documenting the constraint.

Minor, non-blocking. Tracking so it isn't forgotten.

Borisserz:
## Summary

Fixes the asymmetry in `RegionsFileStorage`'s date encoding/decoding strategies by introducing a `JSONEncoder.RESTEncoder()` helper and using it for serialization, as requested in #1152.

## Problem

`RegionsFileStorage` was saving regions with the default `JSONEncoder()` (which defaults to `.deferredToDate` / seconds since reference date). However, it loads them using `JSONDecoder.RESTDecoder()`, which explicitly sets `dateDecodingStrategy = .millisecondsSince1970`. 

While harmless today (because `Region` lacks any `Date` properties), this would silently break serialization round-tripping if a `Date` field is ever added to `Region`.

## Solution

1. Added `JSONEncoder.RESTEncoder()` to `InternalTypes.swift` (matching the existing `JSONDecoder.RESTDecoder()`) which configures `.millisecondsSince1970`.
2. Updated `RegionsFileStorage.saveDefaultRegions` and `saveCustomRegion` to use this new encoder.

Fixes #1152

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when saving default and custom regions by using a consistent date format.
  * Region data is now encoded with millisecond-precision timestamps for more reliable persistence and retrieval.

### DIFF
--- a/OBAKitCore/Location/Regions/RegionsFileStorage.swift
+++ b/OBAKitCore/Location/Regions/RegionsFileStorage.swift
@@ -87,7 +87,7 @@ public final class RegionsFileStorage: RegionsFileStorageProtocol {
     public func saveDefaultRegions(_ regions: [Region]) throws {
         let url = try defaultRegionsFileURL()
         try createDirectoryIfNeeded(for: url)
-        let data = try JSONEncoder().encode(regions)
+        let data = try JSONEncoder.RESTEncoder().encode(regions)
         try data.write(to: url, options: .atomic)
     }
 
@@ -121,7 +121,7 @@ public final class RegionsFileStorage: RegionsFileStorageProtocol {
     public func saveCustomRegion(_ region: Region) throws {
         let url = try customRegionFileURL(identifier: region.regionIdentifier)
         try createDirectoryIfNeeded(for: url)
-        let data = try JSONEncoder().encode(region)
+        let data = try JSONEncoder.RESTEncoder().encode(region)
         try data.write(to: url, options: .atomic)
     }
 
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Verified the compatibility question myself before merging: `Region` has hand-written `init(from:)`/`encode(to:)` and there is no `Date` anywhere in its encoded object graph, so the bytes already on disk are identical to what the new encoder produces. Nothing to migrate.

That makes this a pure footgun removal — the encoder and decoder are now in lockstep, so the day `Region` gains a `Date` field it won't silently write in one format and read in another. Exactly the fix #1152 asked for. Thanks!

PR #1233, opened 2026-07-27
ui: improve arrival time badge contrast in light mode

### LINKED ISSUE #599: Arrival/departure times contrast is too low
See screenshot for details. The contrast ratio between the white text and its green background is too low.

I haven't checked the contrast ratio for these elements in light mode, but that's worth looking at too.

I think a good solution for this might be to properly respect the "Increase Contrast" and "Differentiate Without Color" accessibility settings, but it's also worth considering alternative ways to solve this without needing to set an accessibility flag.

image

Borisserz:
## Summary

Improves the contrast of the `departureOnTimeColor` (green) in Light Mode to meet WCAG AA standards, as requested in #599.

## Problem

The previous custom green color used for "on time" badges in Light Mode (`#129900`) had a contrast ratio of roughly `3.76:1` against white text (inside the badge). This is below the WCAG AA minimum contrast ratio of `4.5:1`, making it difficult to read for users without perfect vision.

## Solution

I've darkened the Light Mode green from `#129900` (red: 0.07, green: 0.60, blue: 0.0) to **`#007300`** (red: 0.00, green: 0.45, blue: 0.00).

* **Contrast Ratio against White:** ~6.1:1 (clears WCAG AA for normal-size text; also clears AA for the walk-pill tint use at ~4.9:1).
* The text is now comfortably legible without requiring the user to explicitly enable the "Increase Contrast" accessibility flag.
* Dark mode is unaffected (it continues to use `.systemGreen` which contrasts well against a dark background).

Fixes #599

## Review follow-up

* Corrected the comment/PR claims that previously overstated the ratio as >7:1 / AAA.
* Added `test_onTimeColor_meetsWCAGAAInLightMode` to lock the AA contrast win.
* Left the always-dark `StopPageHeaderView` colorScheme follow-up for a separate issue, per review.

## Summary by CodeRabbit

* **Accessibility**
  * Improved the visibility of on-time departure indicators in light mode with a darker green color meeting WCAG AA contrast requirements.

* **Bug Fixes**
  * Added regression coverage to help ensure the improved contrast remains available in future updates.

### DIFF
--- a/OBAKitCore/Theme/Theme.swift
+++ b/OBAKitCore/Theme/Theme.swift
@@ -132,14 +132,15 @@ public final class ThemeColors: NSObject, @unchecked Sendable {
         departureEarly = .systemRed
         departureEarlyBackground = .systemRed
 
-        // Hex #129900 is better visibility for small text in light mode.
+        // Hex #007300 (red: 0.00, green: 0.45, blue: 0.00) has a 6.1:1 contrast ratio against white in
+        // light mode, clearing the WCAG AA minimum of 4.5:1 for normal-size text.
         // UIColor.systemGreen is better visibility for small text in dark mode.
-        // See #506 and #508 for user feedback.
+        // See #506, #508, and #599 for user feedback.
         let departureOnTimeColor = UIColor { traitCollection in
             if traitCollection.userInterfaceStyle == .dark {
                 return UIColor.systemGreen
             } else {
-                return UIColor(red: 0.07, green: 0.60, blue: 0.00, alpha: 1.00)
+                return UIColor(red: 0.00, green: 0.45, blue: 0.00, alpha: 1.00)
             }
         }
         departureOnTime = departureOnTimeColor
--- a/OBAKitTests/Extensions/UIColorWCAGTests.swift
+++ b/OBAKitTests/Extensions/UIColorWCAGTests.swift
@@ -79,4 +79,17 @@ struct UIColorWCAGTests {
         let midGray = UIColor(red: 127.0/255.0, green: 127.0/255.0, blue: 127.0/255.0, alpha: 1.0)
         #expect(midGray.badgeTextColor(preferring: .white, minimumRatio: 7.0) == UIColor.black)
     }
+
+    // MARK: - ThemeColors on-time green (#599)
+
+    /// Light-mode `departureOnTime` (#007300) must clear WCAG AA (4.5:1) against
+    /// white — the solid capsule and tinted walk-pill both put white or green
+    /// text on this color. Identity checks on `DepartureStatus` would pass for
+    /// any ThemeColors value; this locks the actual contrast.
+    @Test func `On time color meets WCAG AA in light mode`() {
+        let light = ThemeColors.shared.departureOnTime.resolvedColor(
+            with: UITraitCollection(userInterfaceStyle: .light)
+        )
+        #expect(light.wcagContrastRatio(against: .white) >= 4.5)
+    }
 }


Borisserz:
@coderabbitai review

aaronbrethorst:
## Review

Nice, tightly-scoped change. Darkening the light-mode on-time green from `#129900` to `#007300` is the right call, and I verified it improves every place white text sits on this color. One factual fix needed in the new comment, plus one follow-up worth knowing about.

### Verified contrast math (WCAG 2.1 relative luminance)

| Site | Old `#129900` | New `#007300` |
|---|---|---|
| Solid capsule w/ white text (`StopPageHeaderView:93`, alarm chips in `GroupedListView`, `DepartureRowView`) | 3.76:1 ❌ | **6.10:1** ✅ |
| Walk pill: green text on 14% green tint (`StopPageSheetHeaderView:170–173`) | 3.17:1 ❌ | **4.93:1** ✅ |

Both were below AA before and clear it now. Dark mode is untouched, and the `accessibilityContrast == .high` border logic in `PaddingLabel.configure()` is orthogonal, so there's no double-darkening interaction.

### 1. The new comment overstates the ratio (please fix)

```swift
// Hex #007300 (red: 0.0, green: 0.45, blue: 0.0) provides a > 7:1 contrast ratio against white
```

It's **6.10:1**, not >7:1. That matters because 7:1 is precisely the WCAG **AAA** threshold for normal text, so the comment (and the PR description's "exceeds WCAG AAA requirements") claims a conformance level this color doesn't reach. It does clear AA (4.5:1) comfortably, and AAA for large text. Suggested wording:

```swift
// Hex #007300 (red: 0.0, green: 0.45, blue: 0.0) has a 6.1:1 contrast ratio against white in
// light mode, clearing the WCAG AA minimum of 4.5:1 for normal-size text.
```

Wrong numbers in comments outlive the PR that introduced them — this one will get cited the next time someone touches this color. (Minor, description-only: the old color was 3.76:1 against white, not 2.76:1.)

### 2. Follow-up — the always-dark header card resolves the *light* variant

`StopPageHeaderView` is a permanently dark surface (map snapshot under a 0.35→0.7 black scrim) but never overrides `colorScheme`. So on a light-mode device it draws the **light** green on a dark background — the opposite of what the per-appearance branch intends.

The walk pill there is fine (white text on the capsule, so it's in the 6.10:1 row above). The one that moves the wrong way is the pulsing on-time dot in `HeaderStatusLine` (`StopPageHeaderView.swift:226`): a 7pt decorative circle whose entire job is being visible against the dark scrim. Over typical light Apple Maps tiles at the 0.7 scrim it goes from ≈2.5:1 to ≈1.5:1, and it animates down to 0.35 opacity on top of that.

This is pre-existing — the dot already failed SC 1.4.11 non-text contrast (3:1) before this PR — so I don't think it should block merging. But this change nudges it further, and the fix is one line:

```swift
.environment(\.colorScheme, .dark)   // on the always-dark header card
```

which makes that card pick up `.systemGreen` and scores 4.2–8.1:1 against the scrim. Worth a separate issue rather than expanding this PR.

### 3. No regression test

`DepartureStatusTests` and `DepartureStatusBridgeTests` only assert color *identity* (`status.color == ThemeColors.shared.departureOnTime`), so they pass no matter what the value is — nothing here can catch a future regression. Optional, but a small test would lock the win in:

```swift
func testOnTimeColorMeetsWCAGAAInLightMode() {
    let light = ThemeColors.shared.departureOnTime.resolvedColor(
        with: UITraitCollection(userInterfaceStyle: .light))
    XCTAssertGreaterThanOrEqual(contrastRatio(light, .white), 4.5)
}
```

### Nit

`UIColor(red: 0.0, green: 0.45, blue: 0.0, alpha: 1.00)` mixes one- and two-decimal literals. The old line used `0.07, 0.60, 0.00, 1.00` throughout — worth matching for consistency.

---

**Verdict:** Approve once the `> 7:1` comment is corrected to 6.1:1. The color choice itself is well-reasoned and measurably better everywhere it's used.

Borisserz:
Addressed the review notes:

1. **Comment / PR description** — corrected to **6.1:1**, clearing WCAG **AA** (4.5:1). Removed the incorrect >7:1 / AAA claim. Old color ratio in the description is now 3.76:1.
2. **Regression test** — added `test_onTimeColor_meetsWCAGAAInLightMode` asserting light-mode `departureOnTime` vs white ≥ 4.5 via `wcagContrastRatio`.
3. **Nit** — RGB literals aligned to two decimal places (`0.00, 0.45, 0.00`).
4. **Header card `colorScheme` follow-up** — left out of this PR as suggested; happy to open a separate issue if useful.

aaronbrethorst:
### Code review

Found 1 issue:

1. The new regression test is written in XCTest form (`func test_…` + `XCTAssertGreaterThanOrEqual`), but this target is on Swift Testing (CLAUDE.md says tests are "Written with **Swift Testing** (`@Suite` / `@Test` / `#expect`), not XCTest. The single exception is `PolylinePerformanceTests`"). A bare `func test_…` isn't discovered by the Swift Testing runner, so the contrast assertion this PR adds to lock the win would silently never execute. It should be `@Test func` ... with `#expect(light.wcagContrastRatio(against: .white) >= 4.5)`.

https://github.com/OneBusAway/onebusaway-ios/blob/4df8363521f00131e565dc293d6ad9e542672a0c/OBAKitTests/Stops/StopPage/DepartureStatusTests.swift#L27-L36

The color change itself checks out: `#007300` measures 6.10:1 against white (WCAG 2.1 relative luminance), and every call site of `departureOnTime` is either green-on-light-background text or a solid fill under white foreground content, so all of them move in the right direction — including the 14%-tint walk pill in `StopPageSheetHeaderView` at ~4.93:1. The dark branch still returns `.systemGreen`, so there's no dark-mode regression, and the color stays a dynamic `UIColor` rather than a hardcoded one.

aaronbrethorst (changes requested):
Thanks for turning the earlier notes around so quickly — the comment now says 6.1:1 and cites AA rather than AAA, the RGB literals are aligned, and you added the regression test I asked for. The color change itself I've re-verified and it's good: `#007300` is 6.10:1 against white, the 14% walk-pill tint lands at ~4.93:1, dark mode is untouched, and it stays a dynamic `UIColor`.

Two mechanical things stand between this and merging, and they're the same root cause.

**The branch predates the Swift Testing migration.** `52f21b86` ("Migrate the standalone XCTestCase suites to Swift Testing") landed on `main` after you branched, and it rewrote `DepartureStatusTests.swift` — which is why this PR now shows conflicts. On `main` that file is a `@Suite(.serialized) final class` using `@Test` and `#expect`; on this branch it's still `final class DepartureStatusTests: XCTestCase`.

**The new test won't run once you rebase.** It's written as `func test_onTimeColor_meetsWCAGAAInLightMode()` with `XCTAssertGreaterThanOrEqual`. A bare `func test…` inside a Swift Testing `@Suite` is never discovered — so the regression test added specifically to lock this contrast win would silently not execute. That's worth catching now rather than discovering it the next time someone changes the color. (My earlier suggestion was written in XCTest form — that was my error, and the migration has since made it wrong. Sorry for the misdirection.)

So: rebase onto `main`, and land the test as

```swift
@Test func `On time color meets WCAG AA in light mode`() {
    let light = ThemeColors.shared.departureOnTime.resolvedColor(
        with: UITraitCollection(userInterfaceStyle: .light)
    )
    #expect(light.wcagContrastRatio(against: .white) >= 4.5)
}
```

`wcagContrastRatio(against:)` is already on `main` in `OBAKitCore/Extensions/UIColor+WCAG.swift`, so that part carries over unchanged.

One placement thought, take it or leave it: the assertion is about a `ThemeColors` value rather than `DepartureStatus` behavior, so `OBAKitTests/Extensions/UIColorWCAGTests.swift` may be a more natural home than the departure-status suite.

The header-card `colorScheme` follow-up — yes, please do open that as its own issue. It's a real pre-existing gap and I'd rather track it than let it evaporate.

Rebase plus the `@Test` conversion and this is good to go; no further review needed on the color itself.

Borisserz:
Addressed the review:

1. **Rebased** onto current `main` (resolves the Swift Testing migration conflict in `DepartureStatusTests`).
2. **Regression test** is now a real `@Test` with `#expect`, living in `UIColorWCAGTests` (ThemeColors contrast, not DepartureStatus identity).
3. Local: `UIColorWCAGTests` + `DepartureStatusTests` — 16/16 green, including `On time color meets WCAG AA in light mode`.
4. Header-card `colorScheme` follow-up filed as #1255.

Ready for re-review.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Both mechanical blockers are cleared. The branch is rebased onto `main` — single commit, `CLEAN`, and it no longer touches `DepartureStatusTests.swift` at all — and the regression test is now proper Swift Testing (`@Test` + `#expect`), landed in `UIColorWCAGTests.swift` where I suggested. Thanks for taking the placement note too.

I confirmed the test is load-bearing rather than decorative, since that was the whole worry. Recomputing against the shipped `wcagRelativeLuminance`:

- `#007300` → **6.10:1**
- the old `(0.07, 0.60, 0.00)` → **3.76:1**

So reverting the color drops it under the `>= 4.5` assertion and the test goes red. That's exactly what I wanted this to protect. I also checked the literal agrees with the documented hex (`0.45 × 255 ≈ 115 = 0x73`), and that reading `ThemeColors.shared` from this non-serialized suite is safe — every stored property is a `let` set in `init`, and `departureOnTime` is the dynamic closure, so `resolvedColor(with:)` deterministically takes the light branch.

Good call filing #1255 for the header-card `colorScheme` gap rather than folding it in here.

One thing worth knowing for next time: CodeRabbit's review on this PR embeds an "🤖 Prompt for all review comments with AI agents" block that argues for `#006600` against an AAA target. That's stale — it predates the correction from AAA to AA — and it's addressed to tooling rather than to you. Ignore it; AA at 6.1:1 is the right target here and you've hit it.

Merging.

PR #1192, opened 2026-07-17
Adopt the Swift 6 language mode across OBAKit

aaronbrethorst:
## Summary

- **Every framework and app target now builds in the Swift 6 language mode with zero data-race diagnostics** — OBAKitCore, OBAKit, OBAWidget, and both app flavors (OneBusAway and KiedyBus). The per-target isolation architecture is MainActor-by-default for the UI layer and nonisolated for OBAKitCore, chosen from measured diagnostic counts (full plan, measurements, and per-phase implementation notes in `docs/swift6-migration-plan.md`).
- A CI **concurrency-warning ratchet** (`scripts/concurrency_ratchet` + checked-in baseline) keeps the count from regressing; it went 999 → 39 across the branch, with all 39 remaining warnings in the test target.
- **OBAKitTests is deliberately pinned to the Swift 5 language mode**: Xcode 27 beta 3's XCTest declares `XCTestCase`'s designated initializers nonisolated, and Swift 6 mode rejects every `@MainActor` test class's synthesized init overrides — `init(invocation:)` takes `NSInvocation` (unavailable in Swift), so there is no source-level workaround (verified with standalone `swiftc` probes). Re-evaluate on the next Xcode; details in the plan doc's "Phase 4 status".
- Genuine bug fixes surfaced by the checking: `DataMigrationView` mutated SwiftUI `@State` from `NSItemProvider`'s background queue; `AgencyAlert`'s lazy translation caches and `AgencyAlertsStore`'s state were unsynchronized shared mutable state.
- A simplify + multi-agent review pass consolidated patterns (shared `Timer.scheduledMainActorTimer` helper, `NSCache` for icon caching, single source of truth for build settings) and added regression tests for the two real behavior changes (AgencyAlert translation fallback tiers, debounce suppression).

## Test plan

- [x] Full `build-for-testing` of the App scheme (Xcode 27 beta 3, iPhone 17 Pro simulator) — zero warnings/errors in Swift 6 targets
- [x] 1,286 unit tests pass, 0 failures (was 1,281 on `main`; +5 new tests)
- [x] KiedyBus flavor generated and built clean
- [x] App launched in the simulator: boots through `Application` init (regions, location, analytics), onboarding renders, process stable with timers/services running, no crash logs; deep-link URL opened without incident. *Full UI click-through was not possible in this environment (no interactive simulator access) — the map golden path deserves a manual smoke test, particularly Live Activity updates (see review notes).*
- [x] SwiftLint: 0 violations
[...]

PR #1235, opened 2026-07-28
test: injectable Plausible client seam and typed prop-value tests

### LINKED ISSUE #1199: PlausibleAnalytics: injectable client seam and typed prop-value tests
Follow-up to #1192 (flagged by the test-coverage review). `PlausibleAnalytics` constructs its `Plausible` client inside `init` with no injection point, so its behavior — including the prop-value typing added in #1192 (String/NSNumber pass through typed; anything else is stringified) — has zero test coverage. `UmamiAnalytics` shows the target shape: injectable `dataLoader`, plus `UmamiAnalyticsTests.testJSONValueCoercion`.

Work:
1. Add a seam (inject the client or its transport) so events can be captured in tests.
2. Pin the value-typing behavior: String stays String, Int/Double/Bool (as NSNumber via the @objc bridge) stay numeric, unknown types stringify.
3. Consider aligning the three backends — Firebase and Umami receive the raw `Any?` while Plausible receives the coerced value, so the same event can differ per backend.

Borisserz:
## Summary
- Adds a `PlausibleEventClient` seam (production: `AviaryInsights.Plausible`; tests: recording actor) so `PlausibleAnalytics` can post without the network.
- Extracts `sendablePropValue(from:)` and locks the #1192 typing contract: String and `NSNumber` (Int/Double/Bool via the `@objc` bridge) pass through typed; unknown types stringify.
- Adds `PlausibleAnalyticsTests` covering coercion, event posting, stop/search helpers, and default-property merge.

Closes #1199

### Out of scope
Issue item 3 (“align Firebase / Umami / Plausible value handling”) is left as a follow-up — the backends intentionally have different contracts today (raw `Any?` vs coerced Sendable), and unifying them is a broader change than this coverage seam.

## Test plan
- [x] `xcodebuild test-without-building -only-testing:OBAKitTests/PlausibleAnalyticsTests` — **8 tests, 0 failures**

## Summary by CodeRabbit

* **Refactor**
  * Improved analytics event handling with more reliable property conversion and event delivery.
  * Added support for configurable analytics event clients, including test environments.

* **Bug Fixes**
  * Analytics events now preserve typed values, search queries, stop-view details, and merged user properties more consistently.

### DIFF
--- a/Apps/Shared/Analytics/PlausibleAnalytics.swift
+++ b/Apps/Shared/Analytics/PlausibleAnalytics.swift
@@ -1,13 +1,56 @@
+//
+//  PlausibleAnalytics.swift
+//  App
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
 import Foundation
 import OBAKit
 import AviaryInsights
 
+/// Seam for posting Plausible events. Production uses `AviaryInsights.Plausible`;
+/// tests inject a recording double (#1199).
+///
+/// Explicitly `nonisolated`: App's default MainActor isolation would otherwise
+/// make this protocol MainActor-bound, which blocks both `Plausible` (a plain
+/// Sendable struct) and actor-based test doubles from conforming (CI Xcode 27).
+nonisolated protocol PlausibleEventClient: Sendable {
+    func postEvent(_ event: Event) async throws
+}
+
+extension Plausible: PlausibleEventClient {}
+
 class PlausibleAnalytics: NSObject {
-    private let client: Plausible
+    private let client: any PlausibleEventClient
     private var defaultProperties: [String: (any Sendable)?] = [:]
 
     init(defaultDomainURL: URL, analyticsServerURL: URL) {
         self.client = Plausible(defaultDomain: defaultDomainURL.host!, serverURL: analyticsServerURL)
+    }
+
+    /// Test / alternate-transport entry point. Prefer the URL-based init in production.
+    init(client: any PlausibleEventClient) {
+        self.client = client
+    }
+
+    /// Coerces an `@objc Analytics` `Any?` value into a Plausible-safe `Sendable` prop.
+    ///
+    /// String and `NSNumber` (Int/Double/Bool via the ObjC bridge) pass through typed so
+    /// server-side prop filters keep working; everything else is stringified.
+    static func sendablePropValue(from value: Any?) -> (any Sendable)? {
+        switch value {
+        case nil:
+            return nil
+        case let string as String:
+            return string
+        case let number as NSNumber:
+            return number
+        default:
+            return value.map { String(describing: $0) }
+        }
     }
 
     private func postEvent(pageURL: String, props: [String: (any Sendable)?]) async {
@@ -20,21 +63,7 @@ class PlausibleAnalytics: NSObject {
     }
 
     func reportEvent(pageURL: String, label: String, value: Any?) async {
-        // The Analytics protocol is @objc, so `value` arrives as Any?; Plausible's
-        // client requires Sendable props. Pass the known scalar types through
-        // typed (numbers must stay numbers for server-side prop filtering) and
-        // stringify only as a last resort.
-        let sendableValue: (any Sendable)?
-        switch value {
-        case nil:
-            sendableValue = nil
-        case let string as String:
-            sendableValue = string
-        case let number as NSNumber:
-            sendableValue = number
-        default:
-            sendableValue = value.map { String(describing: $0) }
-        }
+        let sendableValue = Self.sendablePropValue(from: value)
         await postEvent(pageURL: pageURL, props: [label: sendableValue])
     }
 
--- a/OBAKitTests/Analytics/PlausibleAnalyticsTests.swift
+++ b/OBAKitTests/Analytics/PlausibleAnalyticsTests.swift
@@ -1,0 +1,124 @@
+//
+//  PlausibleAnalyticsTests.swift
+//  OBAKitTests
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import Foundation
+import Testing
+import AviaryInsights
+@testable import App
+
+/// Records Plausible events so `PlausibleAnalytics` can be tested without a network.
+private actor RecordingPlausibleClient: PlausibleEventClient {
+    private(set) var events: [Event] = []
+
+    func postEvent(_ event: Event) async throws {
+        events.append(event)
+    }
+}
+
+@Suite(.serialized)
+final class PlausibleAnalyticsTests: OBATestCase {
+
+    // MARK: - sendablePropValue coercion
+    //
+    // Mirrors the @objc Analytics bridge: Int/Double/Bool arrive as NSNumber;
+    // unknown types stringify. Matches UmamiAnalyticsTests JSON-value coercion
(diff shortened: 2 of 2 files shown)


Borisserz:
CI build failed because App's default MainActor isolation made `PlausibleEventClient` MainActor-bound, so the actor-based `RecordingPlausibleClient` could not conform (error only surfaced under CI's Xcode 27).

Fix: mark the protocol `nonisolated` so both `Plausible` (Sendable struct) and actor test doubles can conform. Local `PlausibleAnalyticsTests`: 8/8 passing after the change.

aaronbrethorst:
### Code review

Found 2 issues:

1. The new test file is written in the XCTest + Nimble style this repo no longer uses. CLAUDE.md says OBAKitTests is "Written with **Swift Testing** (`@Suite` / `@Test` / `#expect`), not XCTest", that "Nimble was removed", and that "`OBATestCase` is a plain `@MainActor` base class, not an `XCTestCase`". Nimble is gone from `Apps/Shared/app_shared.yml` and `Package.resolved` as of 185b0b06, so `import Nimble` will not resolve; and because `OBATestCase` is not an `XCTestCase`, the `func testX()` methods carry no `@Test` attribute and would never be discovered even if the import resolved. The sibling file in the same directory, `OBAKitTests/Analytics/UmamiAnalyticsTests.swift`, shows the current shape (`@Suite(.serialized)` + `@Test` + `#expect`). The branch appears to be based on main from 2026-07-27, one day before the Nimble removal landed — a rebase plus a conversion is needed.

https://github.com/OneBusAway/onebusaway-ios/blob/8b5bc77e1747a0a0970504897648565f76c0d294/OBAKitTests/Analytics/PlausibleAnalyticsTests.swift#L9-L25

2. The last assertion in `testDefaultProperties_mergeIntoPostedEvent` does not test what its comment claims. `Event.props` is `[String: any Sendable]?` while the initializer takes `[String: (any Sendable)?]?`, so a nil-valued prop survives the conversion as a *present* key whose value is a boxed `Optional.none`. `props?["tap"]` therefore evaluates to `.some(...)`, not `nil`, so `beNil()` does not hold for the merged-in nil label; and in the other direction, a `nil` subscript result cannot distinguish "key present with a nil value" (the documented behavior) from "key absent" at all. Asserting on `props?.keys.contains("tap")` is what the comment is describing.

https://github.com/OneBusAway/onebusaway-ios/blob/8b5bc77e1747a0a0970504897648565f76c0d294/OBAKitTests/Analytics/PlausibleAnalyticsTests.swift#L106-L109

aaronbrethorst (changes requested):
The production side of this is good — pulling the `Any?` → `Sendable` coercion out into `sendablePropValue(from:)` and introducing a `PlausibleEventClient` seam is the right way to make this testable without a network, and marking the protocol `nonisolated` so an actor double can conform was the correct fix for the isolation error you hit. Shipping behavior is unchanged. That part I'd take as-is.

The blocker is that the test payload — which is the whole point of the PR — is written against a testing stack that no longer exists on `main`.

`OBAKitTests/Analytics/PlausibleAnalyticsTests.swift` at head opens with:

```swift
import XCTest
import Nimble
```

Nimble was removed from the project in `185b0b06` (2026-07-28 21:32 UTC). Your last CI run was 2026-07-28 10:50 UTC — about eleven hours *before* that landed, so the green check on this PR predates the removal and isn't telling you the truth about current `main`. Rebased onto today's `main`, `import Nimble` won't resolve. And even setting the import aside, the file uses bare `func testX()` methods with no `@Test` attributes, so nothing in it would be discovered by Swift Testing.

That also means the "8 tests, 0 failures" line in your test plan was measured on the stale base — worth re-running after the rebase rather than carrying it forward.

So: rebase onto `main` and convert the suite to Swift Testing — `@Suite(.serialized)`, `@Test`, `#expect` — with `OBATestCase` as a plain `@MainActor` base class rather than an `XCTestCase` (setup goes in `override init() async throws`). `OBAKitTests/Analytics/UmamiAnalyticsTests.swift` is the pattern to copy.

Two things to fix while you're in there:

**The nil-prop assertion doesn't test what its comment claims.** `expect(events[0].props?["tap"]).to(beNil())` — `AviaryInsights.Event` takes `[String: (any Sendable)?]?` and stores into `[String: any Sendable]?`, so a nil-valued key survives as a *present* key holding a boxed `Optional.none`. The subscript yields `.some(...)`, not `nil`. And even if it did read as nil, a dictionary subscript can't distinguish "key present holding nil" from "key absent" — which is exactly the distinction the adjacent comment says it's locking in. Assert on `props?.keys.contains("tap")` instead.

**`RecordingPlausibleClient` is declared at file scope with no access control**, so it leaks into every other test file's namespace in the target. `private` fixes it.

Minor, no action needed: `testReportStopViewed_postsIdAndDistance` passes `name: "Pine St"` and never asserts on it — production `reportStopViewed` silently discards `name`, which is pre-existing, but the test neither covers nor documents that.

Everything here is mechanical. Ping me once it's rebased and converted and I'll turn it around quickly.

Borisserz:
Addressed the review:

1. **Rebased** onto current `main` (post-Nimble / Swift Testing).
2. **Suite converted** to `@Suite(.serialized)` + `@Test` + `#expect`, following `UmamiAnalyticsTests`.
3. **Nil-prop assertion** now checks `props?.keys.contains("tap")` — that is the “label kept with a nil value” contract.
4. **`RecordingPlausibleClient` is `private`.**
5. Documented that `reportStopViewed` discards `name` and assert that key is absent.

Local: `PlausibleAnalyticsTests` 8/8 green after rebase.

Ready for re-review.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
All three blockers are genuinely fixed, and I checked each rather than taking the summary on faith.

**The suite is real Swift Testing now.** No `Nimble`, no `XCTest` — imports are `Foundation`, `Testing`, `AviaryInsights`, `@testable import App`. It's `@Suite(.serialized) final class PlausibleAnalyticsTests: OBATestCase` with eight backtick-named `@Test func`s and `#expect` throughout, matching `UmamiAnalyticsTests`. The rebase is real too: head's parent is `01774b93`, and CI here ran *after* the Nimble removal, so unlike last round the green check is actually telling you the truth.

**The nil-prop assertion is now correct** — `#expect(events[0].props?.keys.contains("tap") == true)`. That's the right assertion for the documented contract, since `AviaryInsights.Event` stores `[String: (any Sendable)?]?` into `[String: any Sendable]?` and a nil-valued key survives as a *present* key holding a boxed `Optional.none`.

**`RecordingPlausibleClient` is `private`**, so it no longer leaks into the target namespace. And you picked up the optional extra: `Report stop viewed posts id and distance` now documents that production `reportStopViewed` discards `name` and asserts the key is absent.

The production side I'd already signed off on, and it's unchanged — `sendablePropValue(from:)` plus the `nonisolated PlausibleEventClient` seam, with `AnalyticsOrchestrator` still on the URL-based init. Shipping behavior is identical.

Three small things for whenever you're next in this file, none worth another round:

- The section comment says "Int/Double/Bool arrive as NSNumber," but `Sendable prop value passes NSNumber through` passes `NSNumber(value: 42)` directly rather than a bare `42` as `Any?`. So the `case let number as NSNumber` branch is covered, but the Swift-scalar → NSNumber bridge that production actually relies on isn't. Passing the raw literals would close that for free.
- Relatedly, `NSNumber(value: true) == NSNumber(value: 1)` under `isEqual`, so the Bool assertion verifies pass-through but not that boolness survived.
- In the stop-viewed test, `#expect(events[0].props?["name"] == nil)` is subsumed by the `keys.contains("name")` line below it — and it's the weak form I complained about last time, sitting right next to the strong one. Just delete it.

Thanks for turning the conversion around cleanly. Merging.

PR #1237, opened 2026-07-28
Move DepartureTimeDisplay into OBAKitCore so every surface shows corrected departure times

### LINKED ISSUE #1225: Move DepartureTimeDisplay into OBAKitCore so every surface can show corrected departure times
Follow-up to #1214 / #1224.

## The core issue

#1224 fixed #1214 on the Stop page by adding `DepartureTimeDisplay` (decides *which* clock time to show, and whether to strike through a superseded scheduled one) and `DepartureTimeText` (renders it). Both live in **OBAKit**, the UI framework.

Three other surfaces render a departure time and none of them can use it:

| Surface | File | Situation |
|---|---|---|
| Widget | `OBAWidget/Views/WidgetRowView.swift` | **Can't reach it.** OBAWidget imports only `OBAKitCore`. |
| Live Activity | `OBAKitCore/UI/TripLiveActivityCardView.swift:69` | **Can't reach it.** Lives in OBAKitCore. |
| Bookmark cards | `OBAKit/Bookmarks/BookmarkCardView.swift:126-131` | *Could* import it — same framework — but still hand-rolls the old `Text(timeFormatter.string(from:)).font(.footnote).monospacedDigit().foregroundStyle(.secondary)`. |

To be precise about severity: **none of these display a wrong time.** They all render `arrivalDepartureDate`, which is already prediction-derived. What they lack is the correction affordance — the struck-through scheduled time and the "scheduled X, now expected Y" VoiceOver clause. So the app now has one screen that shows a rider *how far off* their bus is from the timetable and three that quietly don't.

The real cost is the guaranteed duplication. This codebase has already run this exact experiment: `CountdownView` and `RealtimeGlyph` each exist **twice**, once in `OBAKit/Stops/StopPage/Shared/` and once in `OBAKitCore/UI/` marked public "so the widget extension can use it". The internal copy shadows the public one inside OBAKit. Every tweak to those is a two-file edit, and `stop_page.countdown.now` consequently exists in *both* string bundles. `DepartureTimeDisplay` is on precisely that trajectory: the next person who wants a strikethrough in the widget will either copy the type or invent second wording for the same sentence, across 13 locales.

## Acceptance criteria

- [ ] `DepartureTimeDisplay` is defined **once**, in OBAKitCore, and is `public`. No second copy in OBAKit — verify with `grep -rn "struct DepartureTimeDisplay"` returning exactly one hit.
- [ ] Same for `DepartureTimeText`, so the widget and Live Activity get the rendering and not just the decision.
- [ ] OBAKitCore stays application-extension safe (`APPLICATION_EXTENSION_API_ONLY: true` still builds). The type is pure Foundation + `Formatters` today, so this should be free — but the build is the proof.
- [ ] The two strings move from `OBAKit/Strings/*.lproj` to `OBAKitCore/Strings/*.lproj`, all 13 locales, using `OBAKitCore`'s own `OBALoc` shim (`OBAKitCore/Strings/CoreLocalization.swift`). Translations carry over verbatim — they were adjudicated in #1224, don't re-translate.
- [ ] Keys are renamed off the `stop_page.` prefix, since they're no longer stop-page-specific.
- [ ] `LocalizationTests.testEveryLocaleHasSameKeysAsEnglish` passes for **both** bundles, and no key is left behind in OBAKit.
- [ ] `BookmarkCardView.timeText(for:)` uses the shared component; bookmark cards show the strikethrough and speak the corrected-time clause.
- [ ] `WidgetRowView` uses it.
- [ ] `TripLiveActivityCardView:69` uses it, replacing the bare `Text(arrival.departureDate, style: .time)`.
- [ ] Existing `DepartureTimeDisplayTests` move with the type and still pass; add coverage for the Live Activity's derived scheduled date (below).
- [ ] Full `OBAKitTests` green.

## The one non-obvious part

`TripAttributes.ContentState.ArrivalInfo` carries `departureTime` and `scheduleDeviation` but **no scheduled time**:

```swift
public let departureTime: Int      // epoch seconds, prediction-derived
public let scheduleDeviation: Int  // seconds late (+) or early (-)
```

So the Live Activity looks like it needs a payload change — it doesn't. The scheduled time is `departureTime - scheduleDeviation`. Derive it rather than adding a field, because the push payload is a wire contract shared with the notification server and changing it is a coordinated deploy. Worth a test pinning the derivation, including the `scheduleStatus == .unknown` case where there's no meaningful deviation.

Note the doc comment on `TripLiveActivityCardView:13` already calls that line "scheduled time" when it is in fact the predicted time. Fix the comment while you're there.

## Out of scope

- Whether strikethrough is the right treatment on a widget or a Lock Screen Live Activity at those type sizes — that's a design call, and this issue is about making it *possible*, not mandatory. If the answer is "no strikethrough on the widget," the shared type still pays for itself via `expectedTimeText` and the VoiceOver clause.
- De-duplicating `CountdownView` / `RealtimeGlyph`. Same underlying problem, worth its own issue.
- The pre-existing old-stop-page inconsistency in `Formatters.formattedScheduleDeviation` (picks wording from `deviation == 0` while color uses `scheduleStatus`'s ±1.5 min band, so a 1-min-late trip renders a green "arrives 1 min late").

diveshpatil9104:
<h2>Summary</h2>
<p>Fixes #1225.</p>
<p><code>DepartureTimeDisplay</code> and <code>DepartureTimeText</code> (from #1224) lived in OBAKit, so only the Stop page could render the corrected departure time. This moves both types into OBAKitCore — defined once, <code>public</code> — and adopts them on the three surfaces that couldn't reach them: bookmark cards, the widget, and the Live Activity card.</p>
<p>Each of those surfaces previously showed a single plain time. All three now show the scheduled time struck through ahead of the corrected one whenever a prediction moves a trip off its timetable, with the correction spoken to VoiceOver on every surface. A strikethrough is inaudible, so each consumer carries the clause in its combined accessibility label.</p>
<h2>What changed</h2>
<ul>
<li><strong><code>OBAKitCore/UI/DepartureTimeDisplay.swift</code> (moved, public)</strong> — localization keys renamed off the <code>stop_page.</code> prefix to <code>departure_time.*</code>. Also gained <code>DepartureTimeText.text(for:)</code>, a <code>Text</code> builder for call sites that splice the time into a larger wrapping line.</li>
<li><strong>Strings</strong> — both VoiceOver keys moved from OBAKit to OBAKitCore across all 13 locales, translations carried over verbatim. <code>LocalizationTests</code> passes for both bundles.</li>
<li><strong><code>BookmarkCardView</code></strong> — <code>timeText(for:)</code> renders the shared component; <code>Formatters.accessibilityValue(for:)</code> appends the corrected-time clause only when a correction actually renders, since the base value already names the expected time.</li>
<li><strong><code>WidgetRowView</code></strong> — the second line is now "time · deviation", built as one concatenated <code>Text</code> so it wraps at the fixed column width instead of truncating the correction. Replaced <code>nextDepartureLabel</code> / <code>updateNextDepartureLabel</code> with a single <code>fallbackLabel</code> (dead code removed).</li>
<li><strong><code>TripLiveActivityCardView</code></strong> — uses the shared component via a new <code>TripActivityPresenter.timeDisplay(for:)</code>. <code>ContentState</code> doesn't carry the scheduled instant, so it is derived by walking <code>scheduleDeviation</code> back off the departure time — except when <code>scheduleStatus == .unknown</code>, where the deviation is untrusted and <code>departureDate</code> <em>is</em> the timetable time. Also fixed the stale "scheduled time" doc comment called out in the issue.</li>
</ul>

### Screenshots

All three captured in the simulator against live Puget Sound data:

| Bookmark cards | Widget | Live Activity |
|---|---|---|
| Bookmark cards | Widget | Live Activity |

On the Live Activity, the derived schedule checks out against real data: 12:50 AM expected − 3 min late = the struck-through 12:47 AM.

<h2>Notes</h2>
<ul>
<li>The spoken clause's sentence punctuation is handled in code rather than as a second localized key. The string is spoken-only, and a sentence-form key would force all 13 locales to re-adjudicate punctuation for translations that carried over verbatim. Happy to switch to a distinct key if you'd prefer.</li>
<li>The widget's second line now gains the clock time (previously deviation-only). That read as the natural meaning of "WidgetRowView uses it" — easy to adjust if you had a different arrangement in mind.</li>
</ul></body></html>

---
<a href="https://app.blacksmith.sh/OneBusAway/codesmith/onebusaway-ios/pr/1237"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787858985&installation_model_id=429412&pr_number=1237&repository=OneBusAway%2Fonebusaway-ios&return_to=https%3A%2F%2Fgithub.com%2FOneBusAway%2Fonebusaway-ios%2Fpull%2F1237&signature=0b862e6400dbc9cceb5592c1ab84cab7e5e1bf994474adcc924f326974cecd92"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

## Summary by CodeRabbit

* **New Features**
  * Improved departure-time rendering across bookmarks, live activity cards, and widgets, showing corrected predicted times alongside scheduled times (scheduled time is struck through when updated).
  * VoiceOver now speaks clearer scheduled vs expected (“corrected”) departure timing with consistent phrasing across supported languages.

* **Bug Fixes**
  * Updated stop-page accessibility strings to ensure departure-time voiceover reflects the corrected schedule logic.

* **Tests**
  * Added/expanded tests validating corrected-time display behavior and VoiceOver accessibility output ordering.

### DIFF
--- a/OBAKit/Bookmarks/BookmarkCardView.swift
+++ b/OBAKit/Bookmarks/BookmarkCardView.swift
@@ -124,9 +124,13 @@ struct BookmarkCardView: View {
     }
 
     private func timeText(for departure: ArrivalDeparture) -> some View {
-        Text(formatters.timeFormatter.string(from: departure.arrivalDepartureDate))
+        // Shared corrected-time component (#1225): when a prediction moves the
+        // trip off its timetable, the scheduled time renders struck through
+        // ahead of the time the rider should act on. The spoken equivalent is
+        // appended in `Formatters.accessibilityValue(for:)` — this card's label
+        // uses `children: .ignore`, so the visual view can't carry it.
+        DepartureTimeText(display: DepartureTimeDisplay(arrivalDeparture: departure, formatters: formatters))
             .font(.footnote)
-            .monospacedDigit()
             .foregroundStyle(.secondary)
     }
 
--- a/OBAKit/Bookmarks/Formatters+BookmarkArrival.swift
+++ b/OBAKit/Bookmarks/Formatters+BookmarkArrival.swift
@@ -39,6 +39,22 @@ extension Formatters {
 
         var value = self.accessibilityValue(for: firstArrivalDeparture)
 
+        // The visual card renders the corrected time with the schedule struck
+        // through (`DepartureTimeText`); a strikethrough is inaudible, so the
+        // correction is carried here in words. Only the corrected case is
+        // spoken: the base value above already names the expected clock time,
+        // so the plain "at %@" clause would merely repeat it. Appended as its
+        // own sentence — the base value ends with terminal punctuation, so a
+        // comma join would read "…at 9:57 PM., scheduled…". The joiner and
+        // period are code-level rather than a localized sentence-form key on
+        // purpose: this string is only ever spoken, punctuation surfaces as a
+        // pause, and a second key would force all 13 locales to re-adjudicate
+        // punctuation for translations that carried over verbatim (#1225).
+        let timeDisplay = DepartureTimeDisplay(arrivalDeparture: firstArrivalDeparture, formatters: self)
+        if timeDisplay.scheduledTimeText != nil {
+            value += " " + timeDisplay.accessibilityTimeDescription + "."
+        }
+
         if arrivalDepartures.count >= 2 {
             let textToAppend: String
             let secondArrivalDeparture = arrivalDepartures[1]
--- a/OBAKit/Strings/ar.lproj/Localizable.strings
+++ b/OBAKit/Strings/ar.lproj/Localizable.strings
@@ -1236,12 +1236,6 @@
 /* Adherence label for a departure with no real-time signal; deliberately avoids claiming the bus is on time. */
 "stop_page.status.schedule_data" = "بيانات الجدول";
 
-/* VoiceOver clause naming the clock time of a departure. %@ is the time. */
-"stop_page.time.a11y_at_fmt" = "في %@";
-
-/* VoiceOver clause for a departure whose real-time prediction differs from its scheduled time. %1$@ is the scheduled time, %2$@ the predicted time. */
-"stop_page.time.a11y_rescheduled_fmt" = "حسب الجدول %1$@، والمتوقع الآن %2$@";
-
 /* Gap marker in the approach timeline. %d is the number of elided stops between the vehicle and the stops shown. Plural forms live in Localizable.stringsdict; the value above is only the not-found fallback. */
 "stop_page.timeline.skipped_stops_fmt" = "%d مواقف";
 
--- a/OBAKit/Strings/en.lproj/Localizable.strings
+++ b/OBAKit/Strings/en.lproj/Localizable.strings
@@ -1238,12 +1238,6 @@
 /* Adherence label for a departure with no real-time signal; deliberately avoids claiming the bus is on time. */
 "stop_page.status.schedule_data" = "schedule data";
 
-/* VoiceOver clause naming the clock time of a departure. %@ is the time. */
-"stop_page.time.a11y_at_fmt" = "at %@";
-
-/* VoiceOver clause for a departure whose real-time prediction differs from its scheduled time. %1$@ is the scheduled time, %2$@ the predicted time. */
-"stop_page.time.a11y_rescheduled_fmt" = "scheduled %1$@, now expected %2$@";
-
 /* Gap marker in the approach timeline. %d is the number of elided stops between the vehicle and the stops shown. Plural forms live in Localizable.stringsdict; the value above is only the not-found fallback. */
 "stop_page.timeline.skipped_stops_fmt" = "%d stops";
 
--- a/OBAKit/Strings/es.lproj/Localizable.strings
+++ b/OBAKit/Strings/es.lproj/Localizable.strings
@@ -1236,12 +1236,6 @@
 /* Adherence label for a departure with no real-time signal; deliberately avoids claiming the bus is on time. */
 "stop_page.status.schedule_data" = "datos de horario";
 
-/* VoiceOver clause naming the clock time of a departure. %@ is the time. */
-"stop_page.time.a11y_at_fmt" = "a las %@";
-
-/* VoiceOver clause for a departure whose real-time prediction differs from its scheduled time. %1$@ is the scheduled time, %2$@ the predicted time. */
-"stop_page.time.a11y_rescheduled_fmt" = "según el horario a las %1$@, ahora se espera a las %2$@";
-
 /* Gap marker in the approach timeline. %d is the number of elided stops between the vehicle and the stops shown. Plural forms live in Localizable.stringsdict; the value above is only the not-found fallback. */
 "stop_page.timeline.skipped_stops_fmt" = "%d paradas";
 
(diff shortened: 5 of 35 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is a really clean execution of #1225. Moving the component rather than duplicating it is the right call, and the three surfaces that previously couldn't reach it — bookmark cards, widget row, Live Activity card — now get corrected times for free.

I checked the things a cross-framework move can quietly break, and they all hold:

- **Extension safety.** `OBAKitCore/UI/DepartureTimeDisplay.swift` imports only SwiftUI, so `APPLICATION_EXTENSION_API_ONLY: true` on OBAKitCore is unaffected.
- **Isolation.** OBAKitCore is the one target pinned back to `nonisolated` default isolation, so a type moving in can silently change meaning. This one holds no mutable or global state and every call site is a MainActor SwiftUI view — and `TripLiveActivityCardView` is already precedent for a `View` in that target.
- **XcodeGen.** `OBAKitCore/project.yml` uses a `path: "."` synced folder and `OBAKitCore/UI/` already exists, so the file is picked up with no yml change. Correctly untouched.
- **Localization.** 26 keys out of OBAKit and 26 into OBAKitCore — 13 locales times two keys, symmetric, no `stop_page.time.a11y_*` leftovers.
- **The derivation.** `scheduleDeviation` is late-positive, so `departureDate - deviation` recovers the timetable instant correctly, and skipping that walk-back when `scheduleStatus == .unknown` is exactly right. The comment explaining why an untrusted deviation must not shift the displayed schedule is the kind I'd want to find in six months.

Merging. Three things worth a follow-up, none blocking:

1. **Widget row width.** The second line grows from deviation-only (~110 pt) to "9:57 AM · departs 3 min late" (~178 pt) against a 180 pt column, and wider when struck through. `lineLimit(2)` absorbs it, but on `.systemLarge` with seven rows the cumulative growth may clip the last one. You screenshotted medium — worth a look at large before we hear about it.

2. **"9:57 AM · departs on time" for schedule-only departures.** `formattedScheduleDeviation(for:)` derives its wording from `deviationFromScheduleInMinutes`, which is 0 when `predicted == false`, so a non-real-time trip claims to be on time. `TripActivityPresenter.deviationLabel` gets this right via `Strings.scheduledNotRealTime`; the widget doesn't. The text is pre-existing, but pairing it with a concrete clock time makes the false claim louder.

3. **The ASCII `"."`** appended to the spoken clause in `Formatters+BookmarkArrival.swift` applies to all 13 locales, including CJK and Arabic, which use `，`. You flagged the trade-off yourself and it's spoken-only, so I'm fine shipping it — just noting it's real.

Thanks for the care on the locale files and the presenter comment.

PR #1241, opened 2026-07-29
refactor: replace DispatchQueue debounce/throttle with MainActor utilities

### LINKED ISSUE #1196: Replace DispatchQueue.debounce/throttle with a @MainActor Debouncer utility
Follow-up to #1192 (flagged by the altitude review). `DispatchQueue.debounce`/`throttle` (`OBAKit/Extensions/DispatchExtensions.swift`) are now main-actor-only APIs bolted onto DispatchQueue: the type system can't stop `DispatchQueue.global().debounce { }`, which currently fails via a debug `assert` and would trap in `MainActor.assumeIsolated` in release. The bookkeeping lives in `@MainActor` file-globals.

A small `@MainActor` Debouncer/Throttler utility (or `Task.sleep`-based logic) would make misuse a compile error and delete the globals and both `assumeIsolated` bridges. There are exactly two call sites: `TripFloatingPanelController.swift` and `StopViewController.swift`, both `DispatchQueue.main`. `DispatchExtensionsTests` (including the debounce-suppression test added in #1192) should carry over.

Borisserz:
## Summary
- Adds `@MainActor` `Debouncer` / `Throttler` in `OBAKit/Utilities/Debouncer.swift`, owning per-instance bookkeeping (no file-level globals, no `MainActor.assumeIsolated`).
- Updates the two call sites: stop refresh cooldown (`StopViewController`) and post-scroll blink delay (`TripFloatingPanelController`).
- Removes `DispatchExtensions.swift` and carries the old suite over as `DebouncerTests`, plus a throttle-replace coverage test.

Closes #1196

### Why
`DispatchQueue.debounce` / `throttle` were main-actor-only APIs bolted onto `DispatchQueue`, so `DispatchQueue.global().debounce { }` still type-checked and failed only via assert / `assumeIsolated` at runtime. Dedicated `@MainActor` types make that a compile error.

No localization changes — no user-facing strings.

## Test plan
- [x] `xcodebuild test-without-building -only-testing:OBAKitTests/DebouncerTests` — **4 tests, 0 failures**
  - executes action (debounce + throttle)
  - suppresses second debounce within interval
  - throttle replaces pending action
- [ ] Manual smoke (optional): mash refresh on a stop; open a trip panel and confirm blink still fires after scroll

## Summary by CodeRabbit

* **New Features**
  * Added reusable controls for managing delayed and repeated actions.
  * Improved pull-to-refresh behavior by preventing repeated refresh triggers.
  * Improved trip stop highlighting during programmatic scrolling.

* **Bug Fixes**
  * Refresh and scrolling actions now execute more consistently while avoiding unnecessary duplicate work.

* **Tests**
  * Added coverage for delayed actions, timing behavior, suppression, and replacement of overlapping actions.


aaronbrethorst:
### Code review

Found 1 issue:

1. `Throttle replaces pending action` cannot fail if replacement is broken. Both calls use the same deadline and assign to the same variable, so the final value is `2` whether or not the first worker is cancelled — if `workItems[key]?.cancel()` were removed entirely, `value = 1` would run first and `value = 2` second, and `#expect(value == 2)` would still pass. It only re-tests what `Throttle executes action` already covers. Incrementing a counter and asserting it equals `1` (or giving the two calls distinguishable effects) would actually exercise the cancellation path this test is named for. (bug due to `OBAKitTests/Extensions/DebouncerTests.swift`)

https://github.com/OneBusAway/onebusaway-ios/blob/7b4759863df1177fc2e38bf56e4e9e6164f4e010/OBAKitTests/Extensions/DebouncerTests.swift#L63-L71

aaronbrethorst (changes requested):
This is the right shape for #1196. Making the utilities `@MainActor` classes turns off-actor misuse into a compile error instead of a runtime `assumeIsolated` trap, and moving the bookkeeping onto instances quietly fixes a real bug: the old file-level `lastDebounceCallTimes` keyed on a shared `nilContext`, so two `StopViewController`s on the nav stack shared one debounce window and could suppress each other's refresh. Good change.

Two things in the new bookkeeping I'd like fixed first.

### 1. Deferred cleanup can wipe a fresh window

In `Debouncer`, the cleanup work item doesn't remove state directly — it enqueues a `Task`:

```swift
let cleanup = DispatchWorkItem { [weak self] in
    Task { @MainActor in
        self?.lastCallTimes.removeValue(forKey: key)
        self?.cleanupWorkItems.removeValue(forKey: key)
    }
}
```

If a new `debounce` lands after the work item fires but before that `Task` runs, the stale `Task` deletes the *just-written* `lastCallTimes[key]` and the *just-stored* `cleanupWorkItems[key]`. The next call isn't suppressed, and the new cleanup is no longer cancellable. The old code removed the entry synchronously inside the work item, which is what made the ordering safe — so this is a small regression rather than something pre-existing.

The window is sub-millisecond and the worst case is one extra API refresh, so I'm not worried about users hitting it. I care because it's the kind of pattern that gets copied. Dropping `DispatchWorkItem` entirely keeps the removal synchronous-on-main:

```swift
cleanupTasks[key]?.cancel()
cleanupTasks[key] = Task { @MainActor in
    try? await Task.sleep(for: .seconds(interval))
    guard !Task.isCancelled else { return }
    lastCallTimes.removeValue(forKey: key)
    cleanupTasks.removeValue(forKey: key)
}
```

`Throttler` has the same deferred-removal pattern and wants the same treatment.

### 2. `Throttle replaces pending action` can't fail

```swift
throttler.throttle(deadline: .now() + .milliseconds(150)) { value = 1 }
throttler.throttle(deadline: .now() + .milliseconds(150)) { value = 2 }
try await Task.sleep(for: .milliseconds(300))
#expect(value == 2)
```

Both calls use the same deadline and assign to the same variable. With cancellation completely broken, both run, FIFO gives `value = 1` then `value = 2`, and the assertion still passes — so the test adds nothing over `Throttle executes action`.

This matters more than a normal weak test would, because cancellation is exactly the logic issue 1 lives in: the one test that should have caught that race can't. Counting instead of overwriting makes it real:

```swift
var runs = 0
throttler.throttle(deadline: .now() + .milliseconds(150)) { runs += 1 }
throttler.throttle(deadline: .now() + .milliseconds(150)) { runs += 1 }
try await Task.sleep(for: .milliseconds(300))
#expect(runs == 1)
```

### Minor

The suite's header comment still says "The first two use `poll` … The suppression test deliberately keeps a fixed wait" — there are four tests now, and the new throttle test also uses a fixed wait.

One more worth knowing, though I don't want it changed here: the names are inverted relative to how most codebases use them — `Debouncer` implements a leading-edge throttle and `Throttler` a trailing-edge debounce. They're carried over from the old API and the doc comments describe the real behavior, so renaming now would just churn call sites. Flagging it only because these are `public` in a white-label framework, and someone will eventually reach for them expecting the usual semantics.

Ping me when the two fixes are up and I'll re-review right away.

Borisserz:
Addressed both review points:

1. **Deferred cleanup** — dropped the `DispatchWorkItem` → nested `Task` hop. `Debouncer` and `Throttler` now schedule cancellable `Task`s that clear bookkeeping (and run the throttled action) synchronously on the main actor after `Task.sleep`, with an explicit `Task.isCancelled` guard.
2. **`Throttle replaces pending action`** — asserts `runs == 1` instead of overwriting a value, so a broken cancel path would fail the test.

Also refreshed the suite header comment. Ready for another look.

aaronbrethorst (approved):
Approved.

I read this one line by line, since a debounce/throttle rewrite is the kind of change where the semantics quietly drift. They did not. Specifically:

- Leading-edge cooldown is preserved in `Debouncer` (`last + interval > .now()` returns early), and trailing-delay-with-replacement is preserved in `Throttler`.
- `pendingTasks[key]?.cancel()` runs before the replacement is stored, and the `guard !Task.isCancelled` after the sleep closes the stale-resumption window. Every mutation happens in a synchronous main-actor segment with no interleaved await, so there is no gap to exploit.
- `Throttler` now clears its bookkeeping before calling `action()` instead of in a `defer`. That is a real fix: the old ordering let a reentrant `throttle()` from inside the action get clobbered.

The part I appreciated most is one you may not have set out to fix. The old `debounce` scheduled its cleanup through `throttle(deadline:)` with no context, so it shared the `nilContext` slot with `TripFloatingPanelController`'s blink throttle — meaning a stop refresh could cancel a pending blink and vice versa. Per-instance keys make that impossible. That was a live bug hiding in a helper nobody had reason to re-read.

The tests are honest, too. The suppression test uses a fixed wait rather than polling for `count == 1` (which would pass the instant the first call landed), and the replacement test counts runs instead of overwriting a value, so a broken cancel yields 2 rather than silently passing.

Two notes for later, neither worth another round:

- `Throttler`'s delay is computed synchronously but `Task.sleep` only starts once the task is scheduled, so under main-actor congestion the 750ms blink drifts by however long the task waits to start. `asyncAfter` started its timer at scheduling time. In practice this is a runloop turn.
- The naming inversion we discussed last time is still there (`Debouncer` is a leading-edge throttle, `Throttler` a trailing-edge debounce). I said I would not block on it and I am not, but if you ever touch this file again, that is the rename to make.

Merging.

PR #1242, opened 2026-07-29
fix: stabilize survey checkbox JSON and lock fresh-path hero filter

### LINKED ISSUE #1169: Survey/Routes VM polish follow-ups (deferred from #1144)
Minor, non-blocking follow-ups deferred from the merge of #1144 (Tier 2 ViewModel migration for surveys & routes). None affect correctness of the shipped flows.

### 1. Deterministic checkbox answer serialization
`SurveyViewModel.toggleCheckbox` (`OBAKit/ViewModels/SurveyViewModel.swift:115`) converts a `Set<String>` straight to `Array`, so the JSON-encoded checkbox answer ordering is non-deterministic. Sorting before encoding (`checkboxSelections[question.id, default: []].sorted()`) gives stable payloads and avoids noisy diffs/retries for logically identical selections.

### 2. Fresh-path wire-level test parity
`test_submit_retryPath_filtersHeroFromAdditionalResponses` decodes the actual PUT body to prove the hero answer is excluded on the retry path. The fresh two-leg path (`SurveyViewModel.swift:179`, `remainingResponses` filter) is covered by POST/PUT request counting but not by a wire-level hero-exclusion assertion. Add the parallel body-inspection assertion for symmetry.

### 3. `updateSubmitRow` behavior is untested
The user-facing payoff of `@Published isSubmitting` — disabling the submit row + swapping the title to "Submitting…" while leaving Cancel enabled (`OBAKit/Surveys/SurveyViewController.swift:88-99`) — has no coverage. UI-level, acceptable to defer.

### 4. Future-facing type-design parity (for the SwiftUI consumer)
- `RoutePickerViewModel` spreads one state machine across `filteredRoutes` + `didFinishLoading` + `loadError`, where the illegal combo `didFinishLoading == false && loadError != nil` is representable. A single `@Published loadState` enum would collapse it (mirroring `HeroSubmissionOutcome`).
- `StopViewModel.surveySubmissionError` emits a raw `Error` while `SurveyViewModel` emits the typed `SubmissionError`; typing the stream would let both submission paths speak the same vocabulary.

Borisserz:
## Summary
- Sort checkbox selections before `formatCheckboxAnswer` so logically identical answers encode to stable JSON (no Set-iteration order noise).
- Add a wire-level assertion that the fresh hero+follow-up submit path excludes the hero `question_id` from the additional PUT body — parity with the existing retry-path test.
- Expose a small `storedAnswer(for:)` seam so the checkbox ordering test can inspect encoding without a submit round-trip.
- Cover Eureka `updateSubmitRow`: in-flight disables submit + "Submitting…" title; idle restores; Cancel (nav close) stays enabled.

Closes #1169

### Out of scope
Issue item 4 deferred:
- `RoutePickerViewModel.loadState` / typed `surveySubmissionError` redesign — future SwiftUI consumer work.

## Test plan
- [x] `xcodebuild test-without-building -only-testing:OBAKitTests/SurveyViewModelTests` — **29 tests, 0 failures**
- [x] `xcodebuild test-without-building -only-testing:OBAKitTests/SurveyViewControllerTests` — **3 tests, 0 failures**

### DIFF
--- a/OBAKit/Surveys/SurveyViewController.swift
+++ b/OBAKit/Surveys/SurveyViewController.swift
@@ -88,7 +88,10 @@ class SurveyViewController: FormViewController {
     /// submit is in flight, so a second tap can't silently be dropped by the VM's in-flight
     /// guard. Cancel button is intentionally left enabled — letting the user back out of a
     /// hung submit matches the rest of the app's modal sheets.
-    private func updateSubmitRow(isSubmitting: Bool) {
+    ///
+    /// Internal (not `private`) so `SurveyViewControllerTests` can drive the Eureka
+    /// affordance directly without standing up a live submit (#1169).
+    func updateSubmitRow(isSubmitting: Bool) {
         guard let row = form.rowBy(tag: "submit") as? ButtonRow else { return }
         row.disabled = Condition(booleanLiteral: isSubmitting)
         row.evaluateDisabled()
--- a/OBAKit/ViewModels/SurveyViewModel.swift
+++ b/OBAKit/ViewModels/SurveyViewModel.swift
@@ -112,10 +112,19 @@ final class SurveyViewModel: ObservableObject {
             checkboxSelections[question.id, default: []].remove(option)
         }
 
-        let selections = Array(checkboxSelections[question.id, default: []])
+        // Sort so identical selections always encode the same JSON (#1169).
+        let selections = checkboxSelections[question.id, default: []].sorted()
         // swiftlint:disable:next force_try
         let jsonAnswer = try! SurveyService.formatCheckboxAnswer(selections)
         updateAnswer(for: question, answer: jsonAnswer)
+    }
+
+    /// The currently stored answer string for `question`, if any.
+    ///
+    /// Exposed for unit tests that need to assert encoding (e.g. checkbox
+    /// JSON order) without going through a submit round-trip.
+    func storedAnswer(for question: SurveyQuestion) -> String? {
+        responses.first { $0.questionId == question.id }?.answer
     }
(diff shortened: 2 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
The `.sorted()` fix is the right one and I checked the blast radius before merging: checkbox answers are in-memory VM state that's only ever encoded outward to the server, never decoded back or persisted, so there's no wire-compat or migration exposure. Stable JSON for identical selections, full stop.

Also nice that the new fresh-path test is a genuine structural parallel to the retry-path one — asserting on the actual PUT body rather than on VM internals is what makes it worth having.

Two things for a follow-up, neither blocking:

1. The three `SurveyViewControllerTests` call `updateSubmitRow(...)` directly, which is why it had to be relaxed from `private`. That means the `$isSubmitting` sink in `bindViewModel` could be deleted and all three would still pass — the binding itself is uncovered. Driving them with `await vm.submit()` would cover the real path and let the method go back to `private`. Relatedly, the docstring on `Submit row starts idle with Cancel enabled` says "the sink fires with isSubmitting == false"; with `.receive(on: DispatchQueue.main)` it hasn't fired yet at that point — the row reads "Submit Survey" because `setupForm` set it.

2. `SurveyCell.handleCheckboxSelection` still builds its titles in declared-option order, so the same question encodes differently depending on whether it was answered on the launcher card or in the full form. Pre-existing, not a regression here — but sorting by the option's index in `content.options` would make both paths agree and preserve the ordering an analyst would expect.

Thanks!

PR #1243, opened 2026-07-29
fix: switch Dynamic Island to the newest tracked trip

### LINKED ISSUE #1189: Live Activities: tracking creates duplicate activities.

Borisserz:
## Summary
- Assign a monotonic `relevanceScore` when starting a Live Activity from bookmarks or the stop page, so ActivityKit promotes the trip the rider just tracked into the Dynamic Island.
- Demote other live peers after start.
- **Re-Track of an already-live trip** promotes it in place (bump score + demote peers) instead of only showing the success toast.
- Preserve `relevanceScore` on local content refreshes via `contentPreservingRelevance` (rebuilding with the default score of `0` previously handed the Island back to the first-started activity).

Addresses Problem 2 of #1189 (app side). Does **not** close #1189 — OBACloud push `relevance-score` still needs a follow-up so production pushes don't re-tie scores.

### Follow-up
https://github.com/OneBusAway/onebusaway-ios/issues/1252

## Test plan
- [x] `xcodebuild test-without-building -only-testing:OBAKitTests/TripLiveActivityRelevanceTests`
- [x] `xcodebuild test-without-building -only-testing:OBAKitTests/TripAttributesIdentityTests`
- [ ] On device: Track A → Track B (Island shows B) → Track A again (Island returns to A)

## Summary by CodeRabbit

* **New Features**
  * Selecting a trip now promotes its Live Activity to the Dynamic Island.
  * Newly created trip activities receive relevance ranking so the most important trip is highlighted.
  * Other active trip activities are automatically de-emphasized.

* **Bug Fixes**
  * Refreshing trip information now preserves the activity’s existing prominence instead of resetting it.
  * Matching existing activities are promoted rather than producing duplicate warnings.

### DIFF
--- a/OBAKit/Bookmarks/BookmarksViewController.swift
+++ b/OBAKit/Bookmarks/BookmarksViewController.swift
@@ -238,13 +238,28 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
         }
 
         let attributes = TripAttributes(staticData: staticData)
+        // Prominence so the Dynamic Island switches to this Track when another
+        // trip is already live (#1189 Problem 2). Default score is 0 and equal
+        // scores keep the first-started activity.
+        let prominence = TripLiveActivityRelevance.prominenceScore()
         do {
             let activity = try Activity.request(
                 attributes: attributes,
-                content: .init(state: contentState, staleDate: nil),
+                content: TripLiveActivityRelevance.content(
+                    state: contentState,
+                    staleDate: nil,
+                    relevanceScore: prominence
+                ),
                 pushType: .token
             )
             trackLiveActivity(activity, arrivalDepartures: arrivalDepartures)
+            let activityID = activity.id
+            Task {
+                await Activity<TripAttributes>.demoteLivePeers(
+                    exceptActivityID: activityID,
+                    relativeTo: prominence
+                )
+            }
             Logger.info("Started Live Activity with ID: \(activity.id)")
             showLiveActivityStartedToast()
         } catch {
@@ -298,6 +313,10 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
                 // `update`. Re-fetch by ID inside a detached task instead — that copy
                 // never crosses an isolation boundary.
                 let activityID = activity.id
+                // Preserve prominence — rebuilding with the default score of 0
+                // would hand the Dynamic Island back to the first-started trip
+                // after every arrivals refresh (#1189 Problem 2).
+                let relevanceScore = activity.content.relevanceScore
                 Task.detached {
                     guard let activity = Activity<TripAttributes>.activities.first(where: { $0.id == activityID }) else {
                         // The activity ended between the loop and this task; dropping
@@ -307,7 +326,11 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
                         return
                     }
                     await activity.update(
-                        .init(state: contentState, staleDate: nil)
+                        TripLiveActivityRelevance.content(
+                            state: contentState,
+                            staleDate: nil,
+                            relevanceScore: relevanceScore
+                        )
                     )
                     Logger.info("Updated Live Activity for stop: \(staticData.stopID) route: \(staticData.routeShortName)")
                 }
--- a/OBAKit/Bookmarks/BookmarksViewController.swift
+++ b/OBAKit/Bookmarks/BookmarksViewController.swift
@@ -221,9 +221,15 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
         // Tapping Track again on a bookmark that is already tracked — or tracking
         // the same trip from the stop page — would otherwise mint a second
         // activity for one stop, leaving the user with duplicate Lock Screen
-        // cards and duplicate OBACloud push registrations.
-        if Activity<TripAttributes>.running(matching: staticData) != nil {
-            Logger.info("Live Activity already running for stop \(staticData.stopID) route \(staticData.routeShortName); not starting a duplicate.")
+        // cards and duplicate OBACloud push registrations. Re-Track still needs
+        // to promote the existing activity: after A→B the Island is on B with A
+        // demoted to 0, so tapping Track on A again must bump A (not just toast).
+        if let existing = Activity<TripAttributes>.running(matching: staticData) {
+            Logger.info("Live Activity already running for stop \(staticData.stopID) route \(staticData.routeShortName); promoting instead of duplicating.")
+            let existingID = existing.id
+            Task {
+                await Activity<TripAttributes>.promoteToDynamicIsland(activityID: existingID)
+            }
             showLiveActivityStartedToast()
             return
         }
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. Re-tracking a trip that is already live never promotes it, because #1215's duplicate guard returns before the new `prominenceScore()` assignment. Track A, then Track B (Island switches to B), then tap Track on A again: the guard matches, re-shows the "Tracking on Lock Screen" confirmation, and returns — `prominenceScore()`/`demoteLivePeers` are never reached, so the Island stays on B while the UI reports success. The PR's "Out of scope" section only mentions the OBACloud push side, so this case looks unintended. Both start paths need the same treatment (bump the existing activity's score and demote peers) in the already-running branch.

https://github.com/OneBusAway/onebusaway-ios/blob/448098e4bea16b0049084b0a50ff2bebaa64099a/OBAKit/Bookmarks/BookmarksViewController.swift#L224-L244

https://github.com/OneBusAway/onebusaway-ios/blob/448098e4bea16b0049084b0a50ff2bebaa64099a/OBAKit/Stops/StopPage/StopPageViewController.swift#L461-L478

aaronbrethorst (changes requested):
The mechanism here is sound, and the part I most wanted to see is present: preserving `relevanceScore` in `updateRunningLiveActivities` so an arrivals refresh doesn't rebuild content at the default `0` and hand the Island back to whichever activity started first. That was the subtle half of #1189 Problem 2 and you got it. `demoteLivePeers` reads cleanly too.

One gap I verified in the code, and I think it should be closed before this lands.

### Re-tracking an already-live trip shows a success toast but never promotes

Both start paths hit the duplicate guard from #1215 and `return` before `prominenceScore()` is ever reached — `BookmarksViewController.startLiveActivity` and the equivalent in `StopPageViewController`:

```swift
    Logger.info("Live Activity already running for stop …; not starting a duplicate.")
    showLiveActivityStartedToast()
    return
}
```

So: Track A, then Track B (A gets demoted to 0), then Track A again. The guard finds A already running, shows the "started" toast, and returns. A is still sitting at `0` and the Dynamic Island stays on B — while the toast tells the rider it worked.

That's the exact user intent this PR exists to serve, and the toast actively misreports it. The fix is small — promote the existing activity in place instead of bailing:

```swift
if let existing = Activity<TripAttributes>.running(matching: staticData) {
    let prominence = TripLiveActivityRelevance.prominenceScore()
    Task {
        await existing.update(
            TripLiveActivityRelevance.content(
                state: existing.content.state,
                staleDate: nil,
                relevanceScore: prominence
            )
        )
        await Activity<TripAttributes>.demoteLivePeers(
            exceptActivityID: existing.id,
            relativeTo: prominence
        )
    }
    showLiveActivityStartedToast()
    return
}
```

Worth applying to both call sites.

### Two notes, not blocking

**The fourth test doesn't test its own name.** `Preserved score helper keeps existing prominence on refresh` builds an `ActivityContent` with a literal score and asserts the score comes back — substantively the same as `Content carries the supplied relevance score` above it. The actual fix (`let relevanceScore = activity.content.relevanceScore` in `updateRunningLiveActivities`) could be deleted and this test would still pass. Since that line is the subtle part of the change, it's the one most worth pinning down.

**The push caveat you flagged is bigger than "out of scope".** Activities use `pushType: .token`, and `docs/LiveActivityPushNotifications.md` shows OBACloud sending a flat `"relevance-score": 100` on every update. Once both the promoted and demoted activities take a push they both land on `100`, tie, and ActivityKit falls back to first-started — which is the original bug. I'm fine deferring the server work, but it means the on-device checkbox in your test plan can pass while production still misbehaves, so let's not close #1189 on this PR alone. Worth filing the OBACloud side now so it doesn't get lost.

Happy to re-review as soon as the promotion path is in.

Borisserz:
Addressed the re-Track promotion gap on both start paths (bookmarks + stop page):

- If `Activity.running(matching:)` finds an existing live activity, we call `promoteToDynamicIsland(activityID:)` (fresh prominence + `demoteLivePeers`) instead of returning after the toast alone.
- Strengthened the refresh test around `contentPreservingRelevance` so it actually pins the `updateRunningLiveActivities` score-preservation path (and contrasts with the default `ActivityContent` score of `0`).

Per your note I also removed `Closes #1189` from this PR — app-side Problem 2 only — and filed the OBACloud push follow-up: '"${CLOUD_URL:-see onebusaway-cloud-services}"'.

Ready for re-review.

Borisserz:
Follow-up for the OBACloud push side (so production pushes don't flatten scores back to a tie):

Borisserz:
Correction on the OBACloud follow-up: `onebusaway-cloud-services` is archived for new issues, so I filed the tracking issue on this repo instead: https://github.com/OneBusAway/onebusaway-ios/issues/1252

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Approved.

The blocking issue from my last review is genuinely fixed, and fixed on both paths — `BookmarksViewController` and `StopPageViewController` now both bind the existing activity, capture its id, and call `promoteToDynamicIsland(activityID:)` rather than only showing a toast.

Worth calling out: you did not copy the snippet I suggested, and you were right not to. My version captured the non-`Sendable` `Activity` into a `Task {}` from a MainActor context, which would not have compiled in the Swift 6 language mode. Re-looking-up by id is the correct shape, and the comment explaining why is the kind of thing that saves the next person an hour.

The other two points are handled as well: `Closes #1189` is off the body, and #1252 tracks the server-side `"relevance-score": 100` work.

1. `updateRunningLiveActivities` snapshots `let existingContent = activity.content` on the MainActor, then re-fetches the activity inside `Task.detached` but still reads the score from the pre-hop snapshot. A refresh whose snapshot predates a concurrent promotion will write the stale score back and quietly undo it. The window is milliseconds and the monotonic wall-clock score makes it self-correcting on the next tick, but reading `relevanceScore` off the re-fetched copy closes it for free.

2. `promoteToDynamicIsland` rebuilds content with `staleDate: nil` while `demoteLivePeers` deliberately preserves `activity.content.staleDate`. Per our own push docs the server always sets `stale-date`, so a re-Track clears the stale marker until the next push arrives. Bounded and consistent with the existing local refresh path, so I am not holding on it.

On the fourth test: it is stronger than it was, and it does now prove the score comes from `existing` rather than from the new content. My original concern still technically stands — the call site in `updateRunningLiveActivities` could be deleted and the suite would stay green — but ActivityKit is not injectable, so there is no cheap way to close that. I would rather take the coverage you have than ask you to build a seam for it.

Merging.

PR #1245, opened 2026-07-30
Rental range filter and fuel labels on the map

aaronbrethorst:
Stacked on #1244. Base is `bikeshare`, so this diff is just the filter work.

## Summary

- **Minimum-range filter** for the Bikes and Scooters layers — a menu row in the Map sheet lets a rider hide micromobility with too little range left to be useful. One shared threshold, persisted beside the existing per-layer enablement so the sheet's Reset button covers it.
- **Fuel labels** beneath each rental pin: battery percent when the feed provides it, remaining range otherwise. On the launch feed `percent` is null fleet-wide, so range is the live path.
- **`RentalVisibility`** caches every delivered entity rather than only visible ones. `VehicleRentalSource` emits only diffs, so a vehicle dropped for being under threshold would never be re-sent — without the cache, lowering the threshold wouldn't bring it back until a refetch.

Design notes: `docs/superpowers/specs/2026-07-29-rental-range-filter-design.md`.

## Test plan

- [x] Full target: **1614 tests in 180 suites**, 0 failures. Seven new suites, 76 tests.
- [x] **Ran on device** (iPhone 17 Pro / iOS 26.3, live Seattle OTP data). Labels render as abbreviated ranges in rental purple beneath the markers — confirming the range fallback is the live path, that `.abbreviated` gives "10 mi" not "10 miles", and that the label sits at the balloon tip.
- [x] Filter verified deterministically in `RentalLayerCoordinatorTests`: raising the threshold removes the short-range vehicle, **lowering it restores from cache with no refetch**, and a vehicle with no range data survives an active filter (fail-open).
- [x] Fuel-label zoom gate covered three ways (inside window, outside, gate closed).
- [x] Map sheet row confirmed visually from a rendered capture: bolt glyph, trailing value, correct placement under Bikes/Scooters. Reset correctly absent when nothing differs from defaults.
- [x] `LocalizationTests` green — no `.strings` entries added; strings are `OBALoc` defaults, matching the convention the key-parity test enforces.
- [ ] **VoiceOver on a rental pin — not yet verified, see below.**

## Review notes

**One thing I could not verify.** Assigning `accessibilityLabel` does not make a `UIView` an accessibility element, and Apple documents neither whether `MKAnnotationView` is one by default nor what it derives from the annotation's `title`/`subtitle`. If it isn't, the whole VoiceOver story here is a no-op — *including the pre-existing behavior this change had to preserve* — and all 14 annotation-view tests still pass. Expect one announcement carrying vehicle name, station occupancy, and fuel figure. Silence means `isAccessibilityElement = true` is needed; a doubled name means MapKit composes its own label and the composition should be overridden rather than appended to.

Related and fixed here: the composed label originally dropped `RentalAnnotation.subtitle`, which silently cost stations their availability count for VoiceOver users, since assigning `accessibilityLabel` overrides MapKit's default.

**Deferred deliberately.** The fuel label's text shadow has no `shadowPath` (impossible for text), so Core Animation derives it from the glyph alpha mask — an offscreen pass per label when content invalidates. That's zoom-driven, not pan-driven. Worth watching during a dense-fleet pan; the fixes if it janks are an `NSAttributedString` halo or `shouldRasterize`.

**Considered and declined.** `RentalVisibility.Changes` mirrors OTPKit's `VehicleRentalSnapshot` field-for-field. Reusing it properly needs an upstream change making `fetchedAt` optional; without that you'd fabricate a timestamp for a diff that involved no fetch. Filed rather than forced.

**Follow-up worth having.** `MapLayerZoomWindow` could grow a second "draw detail at Y" threshold once a live-vehicle layer wants the same two-tier disclosure. One consumer today, so generalizing now would be guessing at the shape.

---
<a href="https://app.blacksmith.sh/OneBusAway/codesmith/onebusaway-ios/pr/1245"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787973635&installation_model_id=429412&pr_number=1245&repository=OneBusAway%2Fonebusaway-ios&return_to=https%3A%2F%2Fgithub.com%2FOneBusAway%2Fonebusaway-ios%2Fpull%2F1245&signature=f0aaf28d1451c19aab1c6bd978bc4fce89cade2612a22bd24c78d01f466c662b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

## Summary by CodeRabbit

* **New Features**
  * Added a minimum-range filter for rental vehicles in the map, with localized presets and persistent settings.
  * Added battery percentage or remaining-range labels to rental map pins, shown based on map zoom.
  * Added accessibility support for rental fuel information and improved label behavior when pins are reused.
* **Bug Fixes**
  * Rental filter changes now update visible map results immediately without requiring a data refresh.
  * Resetting map layers also clears the rental range filter.

### DIFF
--- a/OBAKit/Analytics/Analytics.swift
+++ b/OBAKit/Analytics/Analytics.swift
@@ -65,6 +65,10 @@ public class AnalyticsLabels: NSObject {
     /// layer's conversion into trip planning.
     @objc public static let rentalPlanTripTapped = "Rental Plan Trip Tapped"
 
+    /// Label used when the rental minimum-range filter changes. Value: the
+    /// threshold in meters, or "0" for no minimum.
+    @objc public static let rentalRangeFilterChanged = "Rental Range Filter Changed"
+
     /// Label used when 'Learn More About Donations' screen is displayed
     @objc public static let donationLearnMoreShown = "Donation Learn More Shown"
 
--- a/OBAKit/Mapping/Layers/MapLayer.swift
+++ b/OBAKit/Mapping/Layers/MapLayer.swift
@@ -133,4 +133,9 @@ extension Notification.Name {
     /// Posted (object: the layer's `id`) when a layer is toggled on or off —
     /// drives the basemap button's active-layer count badge.
     public static let mapLayerEnabledStateDidChange = Notification.Name("OBAMapLayerEnabledStateDidChange")
+
+    /// Posted when the shared rental minimum-range filter changes. Lets
+    /// `MapViewController` push the new value into the rental coordinator without
+    /// `MapRegionManager` needing to know the coordinator exists.
+    public static let rentalRangeFilterDidChange = Notification.Name("OBARentalRangeFilterDidChange")
 }
--- a/OBAKit/Mapping/Layers/MapSheetView.swift
+++ b/OBAKit/Mapping/Layers/MapSheetView.swift
@@ -39,6 +39,16 @@ import UIKit
                 self?.objectWillChange.send()
             }
             .store(in: &cancellables)
+
+        // Settings may mirror this same UserDefaults key while the sheet stays its
+        // owner, so an external writer must not leave an open sheet showing a stale
+        // rung.
+        NotificationCenter.default.publisher(for: .rentalRangeFilterDidChange)
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in
+                self?.objectWillChange.send()
+            }
+            .store(in: &cancellables)
     }
 
     // MARK: - Basemap
--- a/OBAKit/Mapping/Layers/MapSheetView.swift
+++ b/OBAKit/Mapping/Layers/MapSheetView.swift
@@ -73,6 +73,27 @@ import UIKit
         objectWillChange.send()
         mapRegionManager.resetMapLayersToDefaults()
     }
+
+    // MARK: - Rental Range Filter
+
+    /// Computed once per sheet presentation: the ladder involves a
+    /// MeasurementFormatter and a localized string, neither worth redoing per render.
+    let rangeFilterPresets = RentalRangePreset.presets()
+
+    /// The rung to highlight. A stored value that isn't on the current ladder — the
+    /// rider's locale changed since they chose it — highlights the closest rung
+    /// without the stored preference being rewritten.
+    var selectedRangePresetID: Int {
+        RentalRangePreset.nearest(
+            toMeters: mapRegionManager.rentalRangeFilter.minimumRangeMeters,
+            in: rangeFilterPresets
+        )?.id ?? 0
+    }
+
+    func selectRangePreset(id: Int) {
+        objectWillChange.send()
+        mapRegionManager.rentalRangeFilter = RentalRangeFilter(minimumRangeMeters: id)
+    }
 }
 
 struct MapSheetView: View {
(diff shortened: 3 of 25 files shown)


aaronbrethorst:
### Code review

Five lenses over the diff: CLAUDE.md compliance, bug scan, git history, prior PR feedback, and code-comment accuracy. Three came back clean — including a history pass that confirmed no deliberate fix from `c1b05cc4` or `ab225fe9` was reversed, notably the `view.annotation` reassignment that keeps glyph and tint tracking live data.

Two findings were fixed in `e0c397fb`:

1. The coordinator's class comment claimed "all debounce, cancellation, and diffing live in the framework." This branch falsified it by adding `RentalVisibility`, a second client-side diffing pass in OBAKit. Reworded.

https://github.com/OneBusAway/onebusaway-ios/blob/e0c397fbb06d734de90adf2ff0afcdda834da214/OBAKit/Mapping/Layers/RentalLayerCoordinator.swift#L13-L22

2. `MapSheetModel` had no subscription for `.rentalRangeFilterDidChange`, so an external writer would leave an open sheet showing a stale rung. The design notes contemplate Settings mirroring these keys, so this is an anticipated case. Added, mirroring the existing availability subscription.

https://github.com/OneBusAway/onebusaway-ios/blob/e0c397fbb06d734de90adf2ff0afcdda834da214/OBAKit/Mapping/Layers/MapSheetView.swift#L44-L52

### Two pre-existing issues this PR builds on, not fixed here

Both were already raised on #1244 and remain open. Flagging them because this PR adds code to the affected functions, so they now interact with the filter.

3. `setLayer` and `viewportDidChange` each fire a bare unstored `Task { }` with no awaiting or cancellation, so two calls in quick succession can leave the source on a stale form-factor/viewport pair. This PR edits both functions without touching the race.

https://github.com/OneBusAway/onebusaway-ios/blob/e0c397fbb06d734de90adf2ff0afcdda834da214/OBAKit/Mapping/Layers/RentalLayerCoordinator.swift#L100-L120

4. `configureRentalLayers()` tears down and rebuilds the service, coordinator and layers on every call, with no guard for an unchanged `graphQLURL`. This PR adds the filter seeding inside it, so a routine regions-list refresh now also discards the rental cache and re-primes the threshold — eroding the no-refetch cache restore this feature exists for.

https://github.com/OneBusAway/onebusaway-ios/blob/e0c397fbb06d734de90adf2ff0afcdda834da214/OBAKit/Mapping/MapViewController+MapLayers.swift#L31-L55

Left alone deliberately: fixing either means changing async or region-lifecycle behavior that neither this PR's tests nor its on-device checks cover. Worth their own change, with #1244's review.

Also not fixed, same reasoning: `MapSheetModel` still doesn't observe `.mapLayerEnabledStateDidChange` — the original instance of the pattern in finding 2, on a line this branch doesn't touch.

PR #1253, opened 2026-07-31
feat: hide MapKit Points of Interest from the Map sheet

### LINKED ISSUE #1246: Add an option to the Map config screen to hide POIs
We’ve gotten some complaints about POIs appearing in the map

Image

Let’s add an option to the new map options sheet (which is available in the PR that adds bike share to the map) to hide or show these POIs.

Borisserz:
## Summary

People have been complaining that Apple’s Points of Interest (restaurants, shops, the usual pin soup) clutter the browse map. This adds a show/hide control on the Map sheet that landed with bikeshare.

- New **Map display** section on the Map sheet with a Points of Interest toggle (works even in regions that have no rental layers).
- Off means `pointOfInterestFilter = .excludingAll`, and `.pointsOfInterest` is removed from `selectableMapFeatures` so you can’t tap ghosts that aren’t drawn.
- Preference lives in UserDefaults (`mapRegionManager.mapViewShowsPointsOfInterest`). Settings mirrors the same key; the Map sheet stays the owner.
- Map sheet **Reset** turns POIs back on with the other layer defaults. Default remains on — same as stock MapKit.

Closes #1246

## Test plan

- [x] `xcodebuild test-without-building -only-testing:OBAKitTests/MapRegionManagerPointsOfInterestTests` (7/7)
- [ ] On device / simulator: open Map sheet → turn off Points of Interest → pins for restaurants/shops disappear; turn back on → they return
- [ ] Settings → Map → flip the same toggle while the Map sheet is open → sheet updates
- [ ] Hide POIs → Reset on the Map sheet → POIs show again

### DIFF
--- a/OBAKit/Mapping/Layers/MapLayer.swift
+++ b/OBAKit/Mapping/Layers/MapLayer.swift
@@ -138,4 +138,10 @@ extension Notification.Name {
     /// `MapViewController` push the new value into the rental coordinator without
     /// `MapRegionManager` needing to know the coordinator exists.
     public static let rentalRangeFilterDidChange = Notification.Name("OBARentalRangeFilterDidChange")
+
+    /// Posted when Points of Interest visibility changes. Lets an open Map sheet
+    /// refresh if Settings (or another writer) flips the same preference.
+    public static let mapPointsOfInterestVisibilityDidChange = Notification.Name(
+        "OBAMapPointsOfInterestVisibilityDidChange"
+    )
 }
--- a/OBAKit/Mapping/Layers/MapSheetView.swift
+++ b/OBAKit/Mapping/Layers/MapSheetView.swift
@@ -49,6 +49,13 @@ import UIKit
                 self?.objectWillChange.send()
             }
             .store(in: &cancellables)
+
+        NotificationCenter.default.publisher(for: .mapPointsOfInterestVisibilityDidChange)
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in
+                self?.objectWillChange.send()
+            }
+            .store(in: &cancellables)
     }
 
     // MARK: - Basemap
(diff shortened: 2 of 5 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Approved.

This is a clean, well-scoped change. A few things I checked that hold up:

- `mapView` is a stored (non-lazy) `let` and `applyPointsOfInterestVisibility()` runs in `init` after `registerDefaults`, so dropping the inline `selectableMapFeatures` assignment loses nothing.
- The `mapLayersDifferFromDefaults` change keeps Reset honest — POIs are always-visible chrome and never region-gated, so Reset always changes something on screen. That was the constraint we landed on in #1245 and you respected it.
- The tests follow the conventions properly: `.serialized`, `OBATestCase` with `override init() async throws`, and a per-test UserDefaults suite rather than leaning on `.standard`.

One thing I want to flag as a follow-up rather than a blocker, because it is a product call and not a defect: `.excludingAll` is broader than the toggle's label suggests. The row says "Points of Interest (restaurants, shops, etc.)", but `.excludingAll` also suppresses `.publicTransport`, airports, hospitals, and parks. We already have a curated `MKPointOfInterestFilter(including:)` list in `MapSnapshotter.swift` built on the reasoning that "POIs to show are public services and first/last-mile transportation" — so the codebase has an opinion here that this path does not share. Hiding transit stations in a transit app is at least worth a second thought. Reusing that including-list, inverted, would match the label more closely.

Also noting for later, not for this PR: the Settings row is inert when `OBAUseMapPanelExperience` is on, since `MapPanelRootView` uses `.mapStyle(.standard(emphasis: .muted))` with no `pointsOfInterest:` argument. Its three sibling rows (scale, traffic, heading) are equally inert there, so this is a pre-existing gap in the experimental panel rather than something you introduced.

Merging.

PR #1256, opened 2026-08-02
Honor colorScheme for stop page on-time green

### LINKED ISSUE #1255: Stop page header card ignores colorScheme for on-time green
## Context

Follow-up from #1233 / #599.

The light-mode on-time green (`ThemeColors.departureOnTime` → `#007300`) now clears WCAG AA against white. One pre-existing gap remains outside that PR's scope:

The stop page **header card** (and related SwiftUI surfaces) can still render with a hardcoded light appearance via `.colorScheme(.light)` (or equivalent), so the dynamic `UIColor` that switches to `.systemGreen` in dark mode never gets a dark trait collection on those views.

## Expected

Header / sheet surfaces that show on-time badges should honor the user's interface style so dark mode gets the intended green.

## Notes

Sibling of the contrast fix in #1233 — called out in review as a separate issue rather than expanding that PR.

Borisserz:
## Summary
- Stop page header, sheet walk pill, and departure rows resolve `ThemeColors.departureOnTime` against the SwiftUI `colorScheme`, so dark mode gets `.systemGreen` instead of the light hex green.
- Adds `ThemeColors.departureOnTime(for:)` / `DepartureStatus.color(for:)` plus a small `ThemeSwiftUI` bridge.
- Closes #1255.

There wasn't a hardcoded `.colorScheme(.light)` on `main` — `Color(uiColor:)` was just resolving the dynamic color against light traits. Explicit resolution fixes that.

## Test plan
- [x] `DepartureStatusTests` + `UIColorWCAGTests` (19 tests) on iPhone Air simulator
- [ ] Dark mode: open a stop sheet / push header and confirm on-time green matches system green
- [ ] Light mode: on-time green still uses the high-contrast hex

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved departure, status, alarm, walking-direction, and divider colors across stop pages for light and dark mode.
  * Ensured on-time departure colors consistently adapt to the current appearance.

* **Tests**
  * Added coverage validating correct color resolution in both light and dark interface styles.

### DIFF
--- a/OBAKit/Stops/StopPage/Departures/DepartureRowView.swift
+++ b/OBAKit/Stops/StopPage/Departures/DepartureRowView.swift
@@ -35,6 +35,7 @@ struct DepartureRowView: View {
 
     @Environment(\.dynamicTypeSize) private var dynamicTypeSize
     @Environment(\.obaFormatters) private var formatters
+    @Environment(\.colorScheme) private var colorScheme
     @ScaledMetric(relativeTo: .body) private var alarmCircleSize: CGFloat = 34
     @AppStorage(UserDefaultsStore.stopUIReducedColorsKey) private var reducedColors = false
 
@@ -125,7 +126,7 @@ struct DepartureRowView: View {
     private var statusText: some View {
         Text(status.label)
             .font(.footnote.weight(.semibold))
-            .foregroundStyle(Color(uiColor: status.color))
+            .foregroundStyle(ThemeSwiftUI.departureStatus(status, colorScheme))
     }
 
     @ViewBuilder
@@ -135,7 +136,7 @@ struct DepartureRowView: View {
                 .font(.subheadline.weight(.semibold))
                 .foregroundStyle(hasAlarm ? Color.white : Color.secondary)
                 .frame(width: alarmCircleSize, height: alarmCircleSize)
-                .background(hasAlarm ? Color(uiColor: ThemeColors.shared.departureOnTime) : Color.clear, in: Circle())
+                .background(hasAlarm ? ThemeSwiftUI.departureOnTime(colorScheme) : Color.clear, in: Circle())
                 .overlay(Circle().strokeBorder(Color(uiColor: .separator), lineWidth: hasAlarm ? 0 : 1.5))
                 // Use onTapGesture, not Button: inner gestures beat the outer
                 // .onTapGesture(perform: onTap) on the row VStack, so tapping the
@@ -157,7 +158,7 @@ struct DepartureRowView: View {
         CountdownView(
             minutes: departure.arrivalDepartureMinutes,
             isRealTime: status.isRealTime,
-            color: dimmed ? Color(uiColor: .tertiaryLabel) : Color(uiColor: status.color)
+            color: dimmed ? Color(uiColor: .tertiaryLabel) : ThemeSwiftUI.departureStatus(status, colorScheme)
         )
     }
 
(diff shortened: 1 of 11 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The new resolution path doesn't change any rendered color, so it can't fix #1255. Apple documents `Color(uiColor:)` as preserving the UIColor's adaptability ("The new color preserves the adaptability of the original… automatically changes its appearance when the user turns on Dark Mode"), and SwiftUI's `\.colorScheme` environment is derived from the same trait collection UIKit uses to resolve that dynamic color — so `ThemeSwiftUI.departureOnTime(colorScheme)` returns exactly what `Color(uiColor: ThemeColors.shared.departureOnTime)` already returned. There is also no `.colorScheme(_:)`, `.preferredColorScheme(_:)`, or `overrideUserInterfaceStyle` anywhere in app code (the PR description concedes this), so no stop-page surface was pinned to light traits. And if the environment's `colorScheme` *were* wrongly `.light`, `uiStyle(_:)` would force `.light` too, so the fix would still not help. Worth confirming the reported symptom against a real dark-mode build before landing a parallel color-resolution API (`ThemeColors.departureOnTime(for:)`, `DepartureStatus.color(for:)`, `ThemeSwiftUI`) plus ~50 call-site changes — and note the same `Color(uiColor: status.color)` pattern is left in place in `BookmarkCardView`, `TripCardView`, and `TripActionBar`, which would be equally "broken" if the premise held.

https://github.com/OneBusAway/onebusaway-ios/blob/f385ef48bd5198d71948d7351f624fb77b5463cc/OBAKitCore/Theme/Theme.swift#L174-L185

aaronbrethorst (changes requested):
This is careful, tidy work — the `ThemeSwiftUI` bridge is a clean shape, and threading a `ColorScheme` through explicitly instead of relying on ambient traits is a defensible pattern in general.

The trouble is that I don't think it changes what's rendered, which means it can't fix #1255.

`ThemeColors.departureOnTime` is already built as a dynamic `UIColor { traitCollection in ... }` in `Theme.swift`, and `Color(uiColor:)` preserves that adaptivity — Apple's documentation is explicit that the resulting `Color` keeps the original's Dark Mode behavior. SwiftUI's `\.colorScheme` and the trait collection UIKit resolves that dynamic color against come from the same place, so `ThemeSwiftUI.departureOnTime(colorScheme)` should hand back exactly the color the old `Color(uiColor:)` call already produced.

I also went looking for something that might be pinning these views to light traits, since that would make the premise real. There is no `.colorScheme(_:)`, no `.preferredColorScheme(_:)`, and no `overrideUserInterfaceStyle` anywhere in `OBAKit/`, `OBAKitCore/`, or `Apps/`. The issue describes a forced light scheme that doesn't exist in the codebase.

So before we take ~20 call-site rewrites and a parallel color-resolution API, I'd like a reproduction: a screenshot from a dark-mode build showing the wrong green, and ideally the view it came from. If the symptom is real then something else is causing it, and this change wouldn't address that either — I'd rather find the actual cause than add an API around a guess.

Two notes for whenever this comes back:

- The two new tests in `UIColorWCAGTests.swift` don't compute a contrast ratio. They compare a color against a literal copied out of `Theme.swift`, which is the identity check the existing comment in that file specifically warns is not what the suite is for. Dark mode gets no contrast assertion at all.
- While you were in here: `StopPageHeaderView`'s walk capsule and the alarm badges put white text on a solid `departureOnTime` fill, which in dark mode resolves to `.systemGreen` — roughly 2:1 against white, well under AA. That's pre-existing and not yours to fix in this PR, but it's a real accessibility gap and worth its own issue if you're interested.

PR #1257, opened 2026-08-02
Use ellipsis for overflowing stop route labels

### LINKED ISSUE #514: Extra Bus Route Consistency (UX)
Issue: 
There’s inconsistency for extra bus route expression below the bus stops, some are using “+n more”, and some are using “+...”

Recommendation: 
Use “...” if there are more than 3 bus routes for a single stop.

Desktop - 12

Borisserz:
## Summary
- When a stop serves more routes than the display limit (map pins use 3), `Formatters.formattedRoutes` now appends `...` instead of `+ N more`.
- Closes #514.

## Test plan
- [x] `FormattersTests` — 10/10 on iPhone Air simulator (includes overflow / no-overflow cases)
- [ ] On the map, find a stop with 4+ routes and confirm the pin subtitle ends with `...` and never shows `+ N more`

## Summary by CodeRabbit

- **Improvements**
  - Route lists exceeding the display limit now end with an ellipsis (`...`) for a cleaner, more concise presentation.
  - Route lists at or below the limit continue to display all available routes.

- **Bug Fixes**
  - Removed the previous overflow-count format (`+ N more`) to provide more consistent route labels.

### DIFF
--- a/OBAKitCore/Utilities/Formatters.swift
+++ b/OBAKitCore/Utilities/Formatters.swift
@@ -597,14 +597,15 @@ public class Formatters: NSObject {
             return nil
         }
 
+        // When the list overflows the limit (map pins use 3), append "..." rather
+        // than "+ N more". Callers were inconsistent — some labels truncated to
+        // "+..." while others showed a count — and #514 asks for a single "...".
+        let fmt = OBALoc("formatters.routes_label_fmt", value: "Routes: %@", comment: "A format string used to denote the list of routes served by this stop. e.g. 'Routes: 10, 12, 49'")
         if routeNames.count > limit {
-            let fmt = OBALoc("formatters.routes_label_plus_more_fmt", value: "Routes: %@, + %d more", comment: "A format string used to denote the overflowing list of routes served by this stop. e.g. 'Routes: 10, 12, 49, + 3 more")
             let shortList = routeNames.prefix(limit).joined(separator: ", ")
-            let overflowCount = routeNames.count - limit
-            return String(format: fmt, shortList, overflowCount)
+            return String(format: fmt, shortList) + "..."
         }
         else {
-            let fmt = OBALoc("formatters.routes_label_fmt", value: "Routes: %@", comment: "A format string used to denote the list of routes served by this stop. e.g. 'Routes: 10, 12, 49'")
             return String(format: fmt, routeNames.joined(separator: ", "))
         }
     }
--- a/OBAKitTests/Application/FormattersTests.swift
+++ b/OBAKitTests/Application/FormattersTests.swift
@@ -102,4 +102,32 @@ final class FormattersTests: OBATestCase {
 
         #expect(label == Formatters.formattedAccessibilityLabel(stop: stop))
     }
+
+    // MARK: - Route overflow labels (#514)
+
+    @Test func `Formatted routes within the limit lists every route`() throws {
+        let routes = try makeRoutes(shortNames: ["10", "20", "30"])
+        let label = try #require(Formatters.formattedRoutes(routes, limit: 3))
+        #expect(label == "Routes: 10, 20, 30")
+        #expect(!label.contains("..."))
+        #expect(!label.contains("more"))
+    }
+
+    @Test func `Formatted routes over the limit appends ellipsis`() throws {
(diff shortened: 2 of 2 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. This PR and your still-open #1267 are two competing fixes for the same symptom, and they can't both land as written. `Stop.mapTitle` is the **only** caller anywhere in the repo that passes a `limit` — every other `formattedRoutes` call site (`formattedAccessibilityLabel`, `Stop.subtitle`, `BookmarkRowViewModel`, `StopPageSheetHeaderView`, `StopHeaderController`) uses the default `.max`, so `routeNames.count > limit` is unreachable for them. #1267 repoints `mapTitle` at a new `formattedMapRoutes`, which would leave the branch this PR edits with no production caller at all, and would put the pin label back under a different formatter. The two also collide mechanically: both add a byte-identical `private func makeRoutes(shortNames:)` to `FormattersTests`, so merging both yields a duplicate declaration. Worth deciding which of the two approaches you want and closing the other.

https://github.com/OneBusAway/onebusaway-ios/blob/f3395395c310afaded20feca13c62c1131651fab/OBAKitCore/Utilities/Formatters.swift#L599-L607

aaronbrethorst (changes requested):
The change itself is small and correct, and the over-limit test is a genuine regression guard — I checked, and reverting the source produces `"Routes: 10, 20, 30, + 2 more"` and fails the assertion. Good.

The problem isn't this diff, it's that **you have two open PRs editing the same overflow branch**. #1267 adds `Formatters.formattedMapRoutes` with its own `"..."` marker and repoints `Stop.mapTitle` at it. This PR changes the overflow branch of `formattedRoutes`.

That collides in two ways. Mechanically, both append a byte-identical `private func makeRoutes(shortNames:)` to `FormattersTests` — duplicate declaration if both land. Semantically, and this is the part no tool catches: `Stop.mapTitle` is the **only** caller in the repo that passes a `limit`. The other five call sites take the default `.max`, so `count > limit` is unreachable for them. If #1267 lands, the branch this PR modifies has zero production callers and the test you just wrote guards dead code.

So one of these two should close. My read is that #1267's approach is the better one — a dedicated map formatter, prefix-free, without overloading `formattedRoutes`' meaning for its five other callers. If you agree, close this and fold the ellipsis decision into #1267 (which needs another round anyway — I left notes there about the bookmark loop placement). If you'd rather keep this one, then #1267 should drop `formattedMapRoutes` and reuse `formattedRoutes`.

Whichever survives, three things to carry across:

- **Use `…` (U+2026), not three periods.** This matters more than it sounds: `StopAnnotationView.titleLabel` sets `lineBreakMode = .byTruncatingTail`, and UIKit's own truncation renders `…`. So the same pin label ends in `...` when the text fits and `…` when it doesn't — which is a new flavor of exactly the inconsistency #514 is about. (#514's text does literally ask for `"..."`, so I don't blame you for the reading; I just think the glyph is right.)
- **Keep the marker inside `OBALoc`.** `String(format: fmt, shortList) + "..."` puts a user-visible marker beyond translator reach, where it used to live inside `formatters.routes_label_plus_more_fmt`. zh-Hans/zh-Hant conventionally use `……`. Related: that key is now orphaned in all 13 catalogs and should be removed with it.
- **The marker is itself truncatable.** `titleLabel` is two lines with `.byTruncatingTail`, so with longer names (route-type fallbacks like "Light Rail", or long short-names) the trailing `"..."` gets truncated away and the rider sees no overflow hint at all. #1267 mitigates this by dropping the `"Routes: "` prefix; this PR doesn't.

Also: the doc comment above `formattedRoutes` still says `/// For example: "Routes: 10, 12, 49, + 3 more".`, which is now false.

On your implied question about `lineBreakMode`/`.truncationMode` — string manipulation is the right tool here, because this truncates a *list of items* ("3 of N routes"), which UIKit truncation can't express. The issue is only that the explicit marker isn't protected from UIKit truncating on top of it.

This is also currently conflicted against `main`. Let me know which of the two you want to keep and I'll prioritize that one.

Borisserz:
Closing in favor of #1267, as discussed in review: a dedicated `formattedMapRoutes` is cleaner than overloading `formattedRoutes` for its other callers. The ellipsis glyph (`…` / U+2026) and `OBALoc` notes from this review will be folded into #1267.

PR #1258, opened 2026-08-02
Filter implausible epoch arrival times from stop lists

### LINKED ISSUE #638: Incorrect bus arrival time
Since around January 25, the app has intermittently shown a bus listed as arriving ~2.8 million minutes early. The error was seen again today Feb 2. Converting it to an exact time showed it was Jan 1 1970, so I assume that the bus arrival time is 0 and it thinks the bus is arriving at the Unix epoch time. Unlike other previously arrived buses, the time traveling bus is stuck at the top of the Arrivals and Departures list. It sometimes appears in the Bookmarks view as well. So far I have seen the error on King County Metro Bus 271 (both routes) in Puget Sound. 

Device: iPhone 13, iOS 16.2
App Version: 23.1.0

Screenshots:
IMG_3882
IMG_3955
IMG_3861

Borisserz:
## Summary
- Arrival/departures whose best time is before the existing decode threshold (`timeIntervalSinceReferenceDate < 1`) are treated as implausible.
- Filtered at `StopArrivals` decode and again in stop list UI paths, so epoch sentinels no longer pin to the top as millions of minutes early.
- Closes #638.

## Test plan
- [x] `ArrivalDepartureTests` — 13/13 on iPhone Air simulator
- [ ] If a feed emits epoch times again, confirm the row is absent from stop / bookmarks lists

### DIFF
--- a/OBAKit/Stops/StopPage/StopPageView.swift
+++ b/OBAKit/Stops/StopPage/StopPageView.swift
@@ -179,7 +179,9 @@ struct StopPageView: View {
     private var filteredDepartures: [ArrivalDeparture] {
         let all = viewModel.stopArrivals?.arrivalsAndDepartures ?? []
         let visible = viewModel.isListFiltered ? all.filter(preferences: viewModel.stopPreferences) : all
-        return visible.filteringTerminalDuplicates()
+        return visible
+            .filteringImplausibleDates()
+            .filteringTerminalDuplicates()
     }
 
     private var attributionText: String {
--- a/OBAKit/Stops/StopViewController.swift
+++ b/OBAKit/Stops/StopViewController.swift
@@ -698,9 +698,11 @@ public class StopViewController: UIViewController,
             if isListFiltered {
                 arrDeps = stopArrivals.arrivalsAndDepartures
                     .filter(preferences: stopPreferences)
+                    .filteringImplausibleDates()
                     .filteringTerminalDuplicates()
             } else {
                 arrDeps = stopArrivals.arrivalsAndDepartures
+                    .filteringImplausibleDates()
                     .filteringTerminalDuplicates()
             }
 
@@ -720,7 +722,9 @@ public class StopViewController: UIViewController,
 
             sections = groups.flatMap { group -> [OBAListViewSection] in
                 var groupSections: [OBAListViewSection] = []
-                let filtered = group.arrivalDepartures.filteringTerminalDuplicates()
+                let filtered = group.arrivalDepartures
+                    .filteringImplausibleDates()
+                    .filteringTerminalDuplicates()
                 let pastDeps = filtered.filter { $0.arrivalDepartureMinutes < 0 }
                 let upcomingDeps = filtered.filter { $0.arrivalDepartureMinutes >= 0 }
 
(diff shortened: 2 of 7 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is the right fix for #638, and the thing that convinces me is that the threshold isn't invented. `Date(timeIntervalSinceReferenceDate: 1.0)` is the same 2001 boundary already used at `ArrivalDeparture.swift:150` and `:156` to nilify predicted times at decode — you're extending an established cutoff from predicted times to the resolved `arrivalDepartureDate`, which is exactly the residual case #638 describes (predicted already nilified, scheduled still an epoch sentinel).

The part I checked hardest is whether this can silently eat real data, because that's the usual failure of a plausibility heuristic. It can't: the threshold is **absolute**, not relative to `now`, so a bus that departed five minutes ago and a departure scheduled a week out both pass untouched. A `now - 1 month` style rule would have quietly eaten the past-departures section; you avoided that.

Filtering at the `StopArrivals` decode is the right layer, and I want to be explicit that I checked the blast radius rather than assuming: it reaches bookmarks (`BookmarkDataLoader.swift:174`), nearby-trip matching (`NearbyTripMatcher.swift:121`), the map layer, and report-a-problem. That's correct — #638 reports the bad row in Bookmarks too, so a view-layer-only fix would have left it.

The tests are load-bearing. `Epoch scheduled times are implausible` decodes `scheduledArrivalTime: 0` at `stopSequence: 5` and asserts the predicate is false; flip the comparison in production and it goes red. Good to see after several PRs today where that wasn't true.

Four follow-ups, none blocking:

- **The threshold literal is duplicated.** You added `earliestPlausibleDate`, but the two decode sites at `ArrivalDeparture.swift:150`/`:156` still hardcode `Date(timeIntervalSinceReferenceDate: 1.0)`. The new doc comment says the constant "matches" them — point those two at `Self.earliestPlausibleDate` and the comment becomes true by construction instead of by vigilance.
- **The filter runs after `filteringDuplicateVehicleReports()`.** Dedupe keeps whichever colliding entry has the later `lastUpdated`; if an AVL ghost carrying the epoch time is the fresher report, it evicts the good entry and then gets dropped itself, erasing the visit entirely. Unlikely — duplicate reports for one trip share a scheduled time from the same GTFS stop time — but swapping the two lines is free and strictly safer.
- **Nothing tests the decode integration.** Every new test exercises the predicate; the actual claim (the bad row disappears from a decoded payload) has no coverage. A short test decoding a `StopArrivals` fixture with one epoch entry and asserting the count would close it.
- **`Nilified predicted time with valid scheduled remains plausible` duplicates the existing `Decode arrival departure with invalid predicted times`**, which this PR already augmented with the same three assertions. Only `stopSequence` differs and both resolve to `.arriving`. One can go.

Also: the doc comment at `StopViewModel.swift:336-341` explains why `predicted == true` isn't proof of a usable time — written to justify *not* checking date usability there. You now add exactly that check on the next line, so it reads oddly. Worth a trim.

The five UI-layer filter sites are all downstream of the decode filter and can't change a result. Your PR body calls that deliberate defense-in-depth and I'll take it, but a future reader will assume those lines are load-bearing — a one-line comment saying "belt-and-braces; decode already filters" would save them the trace.

Merging.

PR #1259, opened 2026-08-02
Hide redundant schedule callouts on the trip map

### LINKED ISSUE #713: Bus "flag" obscures bus location on the bus position route.
Annoying that the link flag for the bus's schedule obscures the bus's actual location much of the time. The flag is redundant. The back button in the app does the same thing.

Borisserz:
## Summary
- Trip map stop pins no longer show schedule-time callout flags that covered the vehicle annotation.
- Opening a trip from a departure arms `skipNextStopTimeHighlight` before the load-time `selectAnnotation`, so the rider is not bounced straight back to the stop page.
- A rider tap with callouts off still highlights the matching list row and opens the stop (the same gesture `StopAnnotationView` uses when `canShowCallout` is false).
- Unreachable `mapView(_:annotationView:calloutAccessoryControlTapped:)` on this screen is deleted.
- Refs #713.

## Test plan
- [x] `TripMapAnnotationPolicyTests` — `viewFor` is built from an `arrivalDeparture` fixture so the policy actually runs (not MKAnnotationView's default `canShowCallout`).
- [x] Programmatic `didSelect` with the skip flag does not push a stop page.
- [x] A rider tap with callouts off does push the stop page.
- [ ] Open a trip from a stop's departure: stay on the trip screen; the origin stop is selected on the map and highlighted in the list, but the stop page is not pushed.
- [ ] Tap a different stop pin: list row highlights and the stop page opens.

## Summary by CodeRabbit

* **Improvements**
  * Simplified trip map stop annotations by removing callouts and accessory controls.
  * Selecting a stop highlights the matching itinerary row.
  * Tapping a stop opens its details directly.
  * Programmatic stop selection no longer navigates unexpectedly.
  * Initial arrival/departure selection no longer highlights the next stop.
  * Refreshes preserve rider-selected or deselected stops and avoid stale highlights.

* **Tests**
  * Added coverage for stop annotation behavior, selection, navigation, and refresh handling.

* **Documentation**
  * Documented trip map stop selection and refresh behavior.

### DIFF
--- a/OBAKit/Trip/TripViewController.swift
+++ b/OBAKit/Trip/TripViewController.swift
@@ -427,6 +427,13 @@ class TripViewController: UIViewController,
         didSet {
             guard !isBeingPreviewed else { return }
 
+            // Value-equal write (30s refresh of the same stop): nothing to
+            // select. Drop a leaked skip so the next pin tap still opens.
+            if oldValue == selectedStopTime {
+                skipNextStopTimeHighlight = false
+                return
+            }
+
             var animated = true
             if isFirstStopTimeLoad {
                 animated = false
@@ -434,8 +441,7 @@ class TripViewController: UIViewController,
             }
             self.mapView.deselectAnnotation(oldValue, animated: animated)
 
-            guard oldValue != self.selectedStopTime,
-                let selectedStopTime = self.selectedStopTime else { return }
+            guard let selectedStopTime = self.selectedStopTime else { return }
 
             // Fixes #220: Find matching trip stop using stop ID instead of using pointers.
             if let annotation = self.mapView.annotations
@@ -447,18 +453,23 @@ class TripViewController: UIViewController,
     }
     private var isFirstStopTimeLoad = true
 
-    /// Auto-selects the rider's origin stop when trip details load. Arms the
-    /// skip flag only on the first load: `tripDetails` republishes every 30s
-    /// with a value-equal `TripStopTime`, so arming on every emission left the
-    /// flag stuck and swallowed the next pin tap.
+    /// Auto-selects the rider's origin stop when trip details first load.
+    /// Later `$tripDetails` emissions (30s refresh) must not write origin
+    /// back if the rider deselected or picked another stop — that
+    /// `selectAnnotation` would fire `didSelect` → `openStop`.
     func applyOriginStopSelection(from details: TripDetails) {
         guard let arrivalDeparture = tripConvertible.arrivalDeparture else { return }
-        if isFirstStopTimeLoad {
-            skipNextStopTimeHighlight = true
-        } else {
-            skipNextStopTimeHighlight = false
+        guard let origin = details.stopTimes.first(where: { $0.stopID == arrivalDeparture.stopID }) else { return }
+
+        if let current = selectedStopTime, current.stopID != origin.stopID {
+            return
         }
-        selectedStopTime = details.stopTimes.filter { $0.stopID == arrivalDeparture.stopID }.first
+        if selectedStopTime == nil && !isFirstStopTimeLoad {
+            return
+        }
+
+        skipNextStopTimeHighlight = true
+        selectedStopTime = origin
     }
 }
 
--- a/OBAKit/Trip/TripViewController.swift
+++ b/OBAKit/Trip/TripViewController.swift
@@ -427,17 +427,7 @@ class TripViewController: UIViewController,
         }
         else if let view = annotationView as? MinimalStopAnnotationView, let arrivalDeparture = tripConvertible.arrivalDeparture {
             view.selectedArrivalDeparture = arrivalDeparture
-
-            if let stopTime = annotation as? TripStopTime {
-                view.rightCalloutAccessoryView = UIButton.chevronButton
-
-                let calloutLabel = UILabel.autolayoutNew()
-                calloutLabel.textColor = ThemeColors.shared.secondaryLabel
-                calloutLabel.text = application.formatters.timeFormatter.string(from: stopTime.arrivalDate)
-                view.detailCalloutAccessoryView = calloutLabel
-            }
-
-            view.canShowCallout = true
+            TripMapAnnotationPolicy.apply(to: view)
         }
 
         return annotationView
--- a/OBAKit/Trip/TripViewController.swift
+++ b/OBAKit/Trip/TripViewController.swift
@@ -339,28 +339,8 @@ class TripViewController: UIViewController,
         defer { skipNextStopTimeHighlight = false }
         guard !skipNextStopTimeHighlight else { return }
 
-        func highlightStopInList() {
-            self.tripDetailsController.highlightStopInList(stopTime.stop)
-        }
-
-        // Callouts are off on this map. A rider tap opens the stop — the same
-        // gesture StopAnnotationView uses when `canShowCallout` is false —
-        // and still highlights the matching list row so the two surfaces agree.
-        guard view.canShowCallout else {
-            highlightStopInList()
-            openStop(stopTime)
-            return
-        }
-
-        if self.mapView.hasBeenTouched {
-            highlightStopInList()
-        } else {
-            if traitCollection.horizontalSizeClass == .regular {
-                floatingPanel.move(to: .full, animated: true, completion: highlightStopInList)
-            } else {
-                floatingPanel.move(to: .half, animated: true, completion: highlightStopInList)
-            }
-        }
+        tripDetailsController.highlightStopInList(stopTime.stop)
+        openStop(stopTime)
     }
 
     public func mapView(_ mapView: MKMapView, didDeselect view: MKAnnotationView) {
@@ -466,6 +446,20 @@ class TripViewController: UIViewController,
         }
     }
     private var isFirstStopTimeLoad = true
+
+    /// Auto-selects the rider's origin stop when trip details load. Arms the
+    /// skip flag only on the first load: `tripDetails` republishes every 30s
+    /// with a value-equal `TripStopTime`, so arming on every emission left the
+    /// flag stuck and swallowed the next pin tap.
+    func applyOriginStopSelection(from details: TripDetails) {
+        guard let arrivalDeparture = tripConvertible.arrivalDeparture else { return }
+        if isFirstStopTimeLoad {
+            skipNextStopTimeHighlight = true
+        } else {
+            skipNextStopTimeHighlight = false
+        }
+        selectedStopTime = details.stopTimes.filter { $0.stopID == arrivalDeparture.stopID }.first
+    }
 }
 
 // MARK: - ViewModel Binding
@@ -495,10 +489,7 @@ private extension TripViewController {
                     mapView.showAnnotations(annotationsToShow, animated: true)
                 }
 
-                if let arrivalDeparture = tripConvertible.arrivalDeparture {
-                    skipNextStopTimeHighlight = true
-                    selectedStopTime = details.stopTimes.filter { $0.stopID == arrivalDeparture.stopID }.first
-                }
+                applyOriginStopSelection(from: details)
             }
             .store(in: &cancellables)
     }
--- a/OBAKit/Trip/TripViewController.swift
+++ b/OBAKit/Trip/TripViewController.swift
@@ -337,10 +337,16 @@ class TripViewController: UIViewController,
     public func mapView(_ mapView: MKMapView, didSelect view: MKAnnotationView) {
         guard let stopTime = view.annotation as? TripStopTime else { return }
         defer { skipNextStopTimeHighlight = false }
-        guard !skipNextStopTimeHighlight else { return }
+        if skipNextStopTimeHighlight {
+            // Programmatic select leaves the pin selected. MapKit will not
+            // re-fire `didSelect` for an already-selected annotation, so the
+            // first tap on the origin stop would do nothing.
+            mapView.deselectAnnotation(view.annotation, animated: false)
+            return
+        }
 
         tripDetailsController.highlightStopInList(stopTime.stop)
-        openStop(stopTime)
+        openStop(stopTime, on: mapView)
     }
 
     public func mapView(_ mapView: MKMapView, didDeselect view: MKAnnotationView) {
@@ -350,12 +356,16 @@ class TripViewController: UIViewController,
         self.selectedStopTime = nil
     }
 
-    private func openStop(_ stopTime: TripStopTime) {
+    private func openStop(_ stopTime: TripStopTime, on mapView: MKMapView) {
         var transferContext: TransferContext?
         if let arrivalDeparture = tripConvertible.arrivalDeparture,
            stopTime.stopID != arrivalDeparture.stopID {
             transferContext = .from(arrivalDeparture: arrivalDeparture, arrivalDate: stopTime.arrivalDate)
         }
+
+        // Same reason as `MapViewController`: leaving the tapped pin selected
+        // only strands a highlight. MapKit will not fire `didSelect` again.
+        mapView.deselectAnnotation(stopTime, animated: false)
 
         application.viewRouter.navigateTo(stop: stopTime.stop, from: self, transferContext: transferContext)
     }
--- a/OBAKit/Trip/TripViewController.swift
+++ b/OBAKit/Trip/TripViewController.swift
@@ -421,17 +421,7 @@ class TripViewController: UIViewController,
         }
         else if let view = annotationView as? MinimalStopAnnotationView, let arrivalDeparture = tripConvertible.arrivalDeparture {
             view.selectedArrivalDeparture = arrivalDeparture
-
-            if let stopTime = annotation as? TripStopTime {
-                view.rightCalloutAccessoryView = UIButton.chevronButton
-
-                let calloutLabel = UILabel.autolayoutNew()
-                calloutLabel.textColor = ThemeColors.shared.secondaryLabel
-                calloutLabel.text = application.formatters.timeFormatter.string(from: stopTime.arrivalDate)
-                view.detailCalloutAccessoryView = calloutLabel
-            }
-
-            view.canShowCallout = true
+            TripMapAnnotationPolicy.apply(to: view)
         }
 
         return annotationView
--- a/OBAKit/Trip/TripViewController.swift
+++ b/OBAKit/Trip/TripViewController.swift
@@ -339,6 +339,11 @@ class TripViewController: UIViewController,
         defer { skipNextStopTimeHighlight = false }
         guard !skipNextStopTimeHighlight else { return }
 
+        guard view.canShowCallout else {
+            openStop(stopTime)
+            return
+        }
+
         func mapViewAnnotationSelectionComplete() {
             self.tripDetailsController.highlightStopInList(stopTime.stop)
         }
@@ -363,7 +368,10 @@ class TripViewController: UIViewController,
 
     public func mapView(_ mapView: MKMapView, annotationView view: MKAnnotationView, calloutAccessoryControlTapped control: UIControl) {
         guard let stopTime = view.annotation as? TripStopTime else { return }
+        openStop(stopTime)
+    }
 
+    private func openStop(_ stopTime: TripStopTime) {
         var transferContext: TransferContext?
         if let arrivalDeparture = tripConvertible.arrivalDeparture,
            stopTime.stopID != arrivalDeparture.stopID {
--- a/OBAKit/Trip/TripViewController.swift
+++ b/OBAKit/Trip/TripViewController.swift
@@ -339,22 +339,26 @@ class TripViewController: UIViewController,
         defer { skipNextStopTimeHighlight = false }
         guard !skipNextStopTimeHighlight else { return }
 
+        func highlightStopInList() {
+            self.tripDetailsController.highlightStopInList(stopTime.stop)
+        }
+
+        // Callouts are off on this map. A rider tap opens the stop — the same
+        // gesture StopAnnotationView uses when `canShowCallout` is false —
+        // and still highlights the matching list row so the two surfaces agree.
         guard view.canShowCallout else {
+            highlightStopInList()
             openStop(stopTime)
             return
         }
 
-        func mapViewAnnotationSelectionComplete() {
-            self.tripDetailsController.highlightStopInList(stopTime.stop)
-        }
-
         if self.mapView.hasBeenTouched {
-            mapViewAnnotationSelectionComplete()
+            highlightStopInList()
         } else {
             if traitCollection.horizontalSizeClass == .regular {
-                floatingPanel.move(to: .full, animated: true, completion: mapViewAnnotationSelectionComplete)
+                floatingPanel.move(to: .full, animated: true, completion: highlightStopInList)
             } else {
-                floatingPanel.move(to: .half, animated: true, completion: mapViewAnnotationSelectionComplete)
+                floatingPanel.move(to: .half, animated: true, completion: highlightStopInList)
             }
         }
     }
@@ -364,11 +368,6 @@ class TripViewController: UIViewController,
             stopTime == self.selectedStopTime else { return }
 
         self.selectedStopTime = nil
-    }
-
-    public func mapView(_ mapView: MKMapView, annotationView view: MKAnnotationView, calloutAccessoryControlTapped control: UIControl) {
-        guard let stopTime = view.annotation as? TripStopTime else { return }
-        openStop(stopTime)
     }
 
     private func openStop(_ stopTime: TripStopTime) {
@@ -497,6 +496,7 @@ private extension TripViewController {
                 }
 
                 if let arrivalDeparture = tripConvertible.arrivalDeparture {
+                    skipNextStopTimeHighlight = true
                     selectedStopTime = details.stopTimes.filter { $0.stopID == arrivalDeparture.stopID }.first
                 }
             }
--- a/OBAKit/Trip/TripMapAnnotationPolicy.swift
+++ b/OBAKit/Trip/TripMapAnnotationPolicy.swift
@@ -14,9 +14,10 @@ import MapKit
 /// Schedule-time callouts duplicated the trip list and obscured the vehicle.
 /// Pins therefore have `canShowCallout = false`. A rider tap opens the stop
 /// (the same gesture `StopAnnotationView` uses without a callout) and
-/// highlights the matching list row. Programmatic selection on load is skipped
-/// via `TripViewController.skipNextStopTimeHighlight` so opening a trip from a
-/// departure does not immediately push that stop back on top of the trip.
+/// highlights the matching list row. Programmatic selection (first load,
+/// list → map) arms `skipNextStopTimeHighlight` so that `didSelect` does
+/// not push the stop. A 30s refresh must not re-select the origin after
+/// the rider has moved the pin.
 enum TripMapAnnotationPolicy {
     static func apply(to view: MinimalStopAnnotationView) {
         view.canShowCallout = false
--- a/OBAKit/Trip/TripMapAnnotationPolicy.swift
+++ b/OBAKit/Trip/TripMapAnnotationPolicy.swift
@@ -18,13 +18,8 @@ import MapKit
 /// via `TripViewController.skipNextStopTimeHighlight` so opening a trip from a
 /// departure does not immediately push that stop back on top of the trip.
 enum TripMapAnnotationPolicy {
-    static let showsStopCallout = false
-
     static func apply(to view: MinimalStopAnnotationView) {
-        view.canShowCallout = showsStopCallout
-
-        guard !showsStopCallout else { return }
-
+        view.canShowCallout = false
         view.rightCalloutAccessoryView = nil
         view.detailCalloutAccessoryView = nil
     }
--- a/OBAKit/Trip/TripMapAnnotationPolicy.swift
+++ b/OBAKit/Trip/TripMapAnnotationPolicy.swift
@@ -11,9 +11,12 @@ import MapKit
 
 /// Callout behavior for stop annotations on the trip map.
 ///
-/// Trip stop pins are selected to highlight the corresponding row in the trip
-/// details list. Schedule-time callouts ("flags") obscure the vehicle annotation
-/// and duplicate information already shown in the list and navigation bar.
+/// Schedule-time callouts duplicated the trip list and obscured the vehicle.
+/// Pins therefore have `canShowCallout = false`. A rider tap opens the stop
+/// (the same gesture `StopAnnotationView` uses without a callout) and
+/// highlights the matching list row. Programmatic selection on load is skipped
+/// via `TripViewController.skipNextStopTimeHighlight` so opening a trip from a
+/// departure does not immediately push that stop back on top of the trip.
 enum TripMapAnnotationPolicy {
     static let showsStopCallout = false
 
--- a/OBAKit/Trip/TripMapAnnotationPolicy.swift
+++ b/OBAKit/Trip/TripMapAnnotationPolicy.swift
@@ -1,0 +1,28 @@
+//
+//  TripMapAnnotationPolicy.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import MapKit
+
+/// Callout behavior for stop annotations on the trip map.
+///
+/// Trip stop pins are selected to highlight the corresponding row in the trip
+/// details list. Schedule-time callouts ("flags") obscure the vehicle annotation
+/// and duplicate information already shown in the list and navigation bar.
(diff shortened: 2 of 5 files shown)


aaronbrethorst:
### Code review

Found 2 issues:

1. Neither new test would fail if the production change were reverted. `Trip map suppresses stop callouts` asserts that a `false` literal is false, and `Applying policy clears callout accessories` asserts that `apply(to:)` does what its own body says. Nothing binds `TripViewController.mapView(_:viewFor:)` to the policy, so restoring `view.canShowCallout = true` plus the accessory setup at the call site — the actual regression these tests exist to catch — leaves both green (tests that pass regardless of behavior).

https://github.com/OneBusAway/onebusaway-ios/blob/70fa6bd475001c98253b8c3c42f1b554131726a3/OBAKitTests/Trip/TripMapAnnotationPolicyTests.swift#L16-L33

2. `mapView(_:annotationView:calloutAccessoryControlTapped:)` is now unreachable, and with it the trip map's only "open this stop" affordance. The chevron in the stop callout was its sole trigger; the map's other annotations are the vehicle (`PulsingVehicleAnnotationView`, which the method's `as? TripStopTime` guard rejects and which carries no accessory view) and the user location. Either delete the stranded method, or adopt the pattern this codebase already documents in `StopAnnotationView` — "`MapViewController` treats selection itself as the open gesture whenever `canShowCallout` is `false`" — so a pin tap can still reach the stop. (bug due to `OBAKit/Trip/TripViewController.swift` lines 368-378 being left with no caller)

https://github.com/OneBusAway/onebusaway-ios/blob/70fa6bd475001c98253b8c3c42f1b554131726a3/OBAKit/Trip/TripViewController.swift#L367-L379

aaronbrethorst (changes requested):
The behavior change is sound, and I checked the thing that would have made it not sound: whether hiding the callout loses information the rider can't get elsewhere on that screen. It doesn't. The callout's time came from `stopTime.arrivalDate` through `formatters.timeFormatter`, and the trip list row renders the identical value from the identical formatter (`TripStopListItem.swift:202`) — and selecting the pin still scrolls to and blinks that row. Schedule-only trips are no different, since the callout showed scheduled time either way. So the flag really was redundant with the list, which is what #713 complains about.

Three things to sort out before I take it.

**The tests don't test the change.** `TripMapAnnotationPolicyTests` asserts that a `false` literal is false, and that `apply(to:)` does what its own body says. Nothing connects `TripViewController.mapView(_:viewFor:)` to the policy. Revert the behavioral hunk in `TripViewController.swift` — restore `view.canShowCallout = true` and the accessory setup — leave `TripMapAnnotationPolicy.swift` in place, and both tests stay green. This is the fourth PR in today's batch with that shape, so I'm being firm about it: a test that passes against unmodified `main` isn't coverage, it's decoration.

**`calloutAccessoryControlTapped` is now stranded.** `TripViewController.swift:368-378` navigates to the stop detail page with transfer context, and the removed chevron was its only trigger — the vehicle view sets no accessory and is rejected by the method's `as? TripStopTime` guard. So the trip map loses its only affordance for opening a stop, and the dead method stays behind. This codebase already documents the alternative: `StopAnnotationView.swift:204-206` notes that when `canShowCallout` is false, *selection* becomes the open gesture. Either adopt that here or delete the stranded method — but don't leave the map with no way to reach a stop and unreachable code explaining how it used to.

**"Closes #713" probably over-claims.** `FeatureFlags.isNewStopPageEnabled` defaults to `true`, so the default path now opens `TripPageViewController`, which is deliberately map-free and sits over the route-projection `TripFocus` layer — and *that* layer's `TripStopAnnotationView` sets `canShowCallout = true` (`TripFocusAnnotations.swift:77`). `TripViewController` is still genuinely reachable (alarms, deep links, recent stops, vehicle search, map-panel mode, flag off), so this isn't dead code — but merging it likely won't fix #713 for default users, and the keyword would auto-close the issue. Worth checking which surface the reporter actually saw, especially since the screenshot never arrived.

One design note, take it or leave it: `showsStopCallout` is a compile-time `false`, so `view.canShowCallout = showsStopCallout` is just `= false` and the `guard !showsStopCallout else { return }` branch is unreachable. That's a new file, type, and test suite around a constant. If the policy type is meant to grow (per-annotation rules, a flag), say so in the doc comment; otherwise a one-line edit at the call site does the same job and there's nothing tautological left to test.

Sort the affordance question and the test, and I'll take another look.

Borisserz:
Updated the trip-map behavior so selecting a stop opens it when its callout is suppressed. The regression test now exercises TripViewController.mapView(_:viewFor:) and verifies it keeps canShowCallout disabled. The commit now references #713 rather than closing it.

aaronbrethorst:
### Code review

Found 3 issues:

1. Opening a trip from a departure now bounces straight back to the stop page. `bindTripContent()` auto-selects the rider's own stop on every trip-details load — `selectedStopTime = details.stopTimes.filter { $0.stopID == arrivalDeparture.stopID }.first` — without arming `skipNextStopTimeHighlight`, and that assignment's `didSet` calls `mapView.selectAnnotation`, which fires `didSelect`. Since the policy forces `canShowCallout = false`, the new guard falls through to `openStop(stopTime)` and pushes a `StopViewController` for the stop the user just came from as soon as the trip loads. That same auto-selection is what made the schedule flag pop up unbidden in #713, so this path runs on every trip open, not just on a tap. (bug due to `OBAKit/Trip/TripViewController.swift` lines 505-507 selecting programmatically while only `showStopOnMap(_:)` sets the skip flag)

https://github.com/OneBusAway/onebusaway-ios/blob/1d26ba285397034770d6f5123311fd4381cb10db/OBAKit/Trip/TripViewController.swift#L344-L351

2. Both new tests still pass with the production change reverted. `TripConvertible(tripDetails:)` leaves `arrivalDeparture` nil, so `viewFor`'s `else if let view = annotationView as? MinimalStopAnnotationView, let arrivalDeparture = tripConvertible.arrivalDeparture` binding fails and `TripMapAnnotationPolicy.apply(to:)` is never called — the assertion only observes `MKAnnotationView`'s default `canShowCallout == false`. Restore `view.canShowCallout = true` plus the chevron and time accessories at the call site and this test stays green. Building the controller with `TripConvertible(arrivalDeparture:)` would actually enter the branch. (This is the same gap as the first point of the 2026-08-05 review.)

https://github.com/OneBusAway/onebusaway-ios/blob/1d26ba285397034770d6f5123311fd4381cb10db/OBAKitTests/Trip/TripMapAnnotationPolicyTests.swift#L27-L37

3. Selecting a stop pin no longer highlights the matching row in the trip list, and `TripMapAnnotationPolicy`'s own doc comment still says it does. `showsStopCallout` is a compile-time `false`, so `view.canShowCallout` can never be true for a `TripStopTime` view, which makes `mapViewAnnotationSelectionComplete()` and the `floatingPanel.move(to: .half/.full)` calls below the new guard unreachable — the highlight-and-move-the-panel behavior added in db1aeb41 and 4e9f54cb is gone. Either preserve it for user taps or update the comment, which currently documents the opposite of what the code now does.

https://github.com/OneBusAway/onebusaway-ios/blob/1d26ba285397034770d6f5123311fd4381cb10db/OBAKit/Trip/TripMapAnnotationPolicy.swift#L13-L18

aaronbrethorst (changes requested):
Removing the callout is a reasonable answer to #713 — the chevron-and-time bubble genuinely was redundant against the list below, and routing pin selection straight to the stop follows the `StopAnnotationView`/`MapViewController` pattern I pointed you at last time.

The problem is that this map, unlike `MapViewController`'s, selects an annotation programmatically on load, and the new selection-as-open path doesn't account for it.

In `TripViewController`, `bindTripContent()` sets `selectedStopTime` from `tripConvertible.arrivalDeparture` without arming `skipNextStopTimeHighlight` — `showStopOnMap(_:)` is the only place that sets that flag. The `didSet` on `selectedStopTime` calls `mapView.selectAnnotation`, MapKit fires `didSelect`, the guard at the top passes because the flag is false, and the new `guard view.canShowCallout else { openStop(stopTime); return }` fires. Since `showsStopCallout` is a compile-time `false`, `canShowCallout` is always false. So opening a trip from a stop's departure — the ordinary path into this screen — immediately pushes the stop page back on top of the trip the rider just opened.

The tests don't catch it because they build `TripConvertible(tripDetails:)` with a nil `arrivalDeparture`, which is precisely the case that skips the auto-selection. That's also why neither of them can fail: with `arrivalDeparture` nil, `viewFor`'s `else if ... let arrivalDeparture` binding never succeeds, `apply(to:)` never runs, and the assertion just observes `MKAnnotationView`'s default `canShowCallout == false`. Revert the production diff and both stay green.

The fix is probably to arm the skip flag on the programmatic selection in `bindTripContent`, and to build the test fixture with an `arrivalDeparture` so it exercises the branch under test.

One smaller thing: `mapView(_:annotationView:calloutAccessoryControlTapped:)` is now unreachable — no annotation on this map has an accessory view. Worth deleting rather than keeping as a wrapper.

Glad to look again once selection-on-load is handled.

aaronbrethorst:
### Code review

Thanks — the three items from the last round are addressed: `bindTripContent()` now arms `skipNextStopTimeHighlight` before the load-time selection, the fixture is built from a real `ArrivalDeparture` so `viewFor` actually reaches `TripMapAnnotationPolicy.apply(to:)` (reverting the production diff would flip `canShowCallout` back to `true` and fail that test), and the unreachable `calloutAccessoryControlTapped` is gone.

Found 2 issues:

1. The skip flag is armed on *every* trip-details emission, not just the first load, and it leaks. `TripViewModel` republishes `tripDetails` every 30 seconds (`refreshInterval = 30.0`, timer started from `viewWillAppear`), so this sink re-runs while the rider sits on the screen. By then `selectedStopTime` already holds the origin stop, and `TripStopTime.isEqual` compares only `arrival`/`departure`/`regionIdentifier`/`stopID` — all unchanged across a refresh — so the re-assignment is value-equal and the `didSet` bails at `guard oldValue != self.selectedStopTime` without ever calling `mapView.selectAnnotation`. No `didSelect` fires, and `didSelect`'s `defer { skipNextStopTimeHighlight = false }` is the only thing that clears the flag, so it stays armed. The rider's next pin tap then hits `guard !skipNextStopTimeHighlight else { return }` and does nothing at all — no list highlight, no stop page — which is exactly the gesture this PR makes the map's only "open this stop" affordance. Repro: open a trip from a departure, wait 30s without touching the map, tap a stop pin. Scoping the arm to the first load (there is already an `isFirstStopTimeLoad`) or clearing the flag when the assignment doesn't change the selection would fix it. (bug due to `OBAKit/Trip/TripViewController.swift` lines 505-506 arming a one-shot flag inside a repeating Combine sink)

https://github.com/OneBusAway/onebusaway-ios/blob/bff958a158cae0df4f492092f1d03fbafe3d9573/OBAKit/Trip/TripViewController.swift#L503-L508

2. Everything below the new guard in `didSelect` is now unreachable. `TripMapAnnotationPolicy.showsStopCallout` is a compile-time `false` and nothing else in OBAKit sets `canShowCallout` on a `MinimalStopAnnotationView` — `viewFor` is the only place these views are configured, and `MKAnnotationView` defaults to `false` — so `view.canShowCallout` can never be true for a `TripStopTime`. The `hasBeenTouched` branch and both `floatingPanel.move(to:)` calls are dead. Same call as the `calloutAccessoryControlTapped` deletion last round: better removed than left as a branch that reads like live behavior.

https://github.com/OneBusAway/onebusaway-ios/blob/bff958a158cae0df4f492092f1d03fbafe3d9573/OBAKit/Trip/TripViewController.swift#L357-L368

OBAKit/Trip/TripViewController.swift
```diff
@@ -505,6 +502,7 @@ private extension TripViewController {
                 }
 
                 if let arrivalDeparture = tripConvertible.arrivalDeparture {
+                    skipNextStopTimeHighlight = true
```
aaronbrethorst: This arms a one-shot flag inside a sink that re-fires every 30 seconds.

`TripViewModel` republishes `tripDetails` on a 30s timer, so this line runs repeatedly. The `selectedStopTime` reassignment below is value-equal across a refresh (`TripStopTime.isEqual` compares arrival/departure/regionIdentifier/stopID, none of which change), so `didSet` returns at `guard oldValue != self.selectedStopTime` without calling `selectAnnotation` — which means no `didSelect`, which means the `defer` that clears this flag never runs.

The flag stays `true` and swallows the rider's next pin tap. Scope the arm to the first load, or clear it when the assignment does not actually change the selection.

OBAKit/Trip/TripViewController.swift
```diff
@@ -343,17 +343,26 @@ class TripViewController: UIViewController,
...
 
-        func mapViewAnnotationSelectionComplete() {
+        func highlightStopInList() {
             self.tripDetailsController.highlightStopInList(stopTime.stop)
         }
 
+        // Callouts are off on this map. A rider tap opens the stop — the same
+        // gesture StopAnnotationView uses when `canShowCallout` is false —
+        // and still highlights the matching list row so the two surfaces agree.
+        guard view.canShowCallout else {
```
aaronbrethorst: Since `TripMapAnnotationPolicy.showsStopCallout` is a `static let false` and `apply(to:)` assigns it directly to `canShowCallout` — and nothing else in the codebase sets `canShowCallout = true` on a `MinimalStopAnnotationView` — this guard always takes the else branch.

That makes the `hasBeenTouched` check and both `floatingPanel.move(to:)` calls below unreachable. Please delete them rather than leaving a branch that reads as live behavior.

aaronbrethorst (changes requested):
All three items from last round are genuinely fixed, and I checked each in the code rather than from the commit messages: the skip flag is now armed before the load-time assignment in `bindTripContent()`, the test fixtures use `TripViewController(application:arrivalDeparture:)` so `viewFor`'s `MinimalStopAnnotationView` binding actually succeeds, and `calloutAccessoryControlTapped` is gone with its body living on as `private func openStop(_:)`. The tests would now genuinely fail if the production diff were reverted — I traced that specifically, since it was the complaint last time.

Unfortunately the fix for the first item introduces a new problem, and it lands on the exact interaction this PR makes the map's only affordance.

**The skip flag leaks across the 30-second auto-refresh and eats the rider's next pin tap.**

The chain, all verified at the head SHA:

1. `TripViewModel.refreshInterval` is 30s, and `loadData()` assigns `tripDetails = newDetails` on a `@Published` property with no equality check — so the `bindTripContent()` sink re-runs every 30 seconds, not just on load.
2. Each re-run hits the new `skipNextStopTimeHighlight = true` and reassigns `selectedStopTime`.
3. `TripStopTime.isEqual` compares `arrival`, `departure`, `regionIdentifier`, and `stopID` — all scheduled values that don't change across a refresh. So the reassignment is value-equal.
4. `selectedStopTime`'s `didSet` therefore returns at `guard oldValue != self.selectedStopTime` and never calls `selectAnnotation`.
5. No `selectAnnotation` means MapKit fires no `didSelect` — and `defer { skipNextStopTimeHighlight = false }` inside `didSelect` is the only place the flag is ever cleared.

So the flag sticks at `true`, and the rider's next tap on a stop pin returns at the guard: no list highlight, no stop page, nothing. Worse, `didSet` calls `deselectAnnotation(oldValue)` *before* that early return, so each refresh visually clears the selection — which makes a re-tap the most likely next gesture, and it's precisely the one that gets swallowed.

Scoping the arm to the first load would do it, or clearing the flag when the assignment doesn't actually change the selection. A one-shot flag armed inside a repeating sink is the shape to get away from.

**Second, smaller: the branch below the new guard is now dead.** `TripMapAnnotationPolicy.showsStopCallout` is a `static let false` and `apply(to:)` assigns it straight to `view.canShowCallout`; nothing anywhere sets `canShowCallout = true` on a `MinimalStopAnnotationView`. So `guard view.canShowCallout else { … return }` always takes the else branch, and the `hasBeenTouched` check plus both `floatingPanel.move(to:)` calls underneath are unreachable. That's the same category of leftover I asked you to delete last round — please take it out rather than leaving it as a branch that reads like live behavior.

One coverage note while you're in there: `Programmatic selection does not open the stop` sets `skipNextStopTimeHighlight` by hand instead of driving `bindTripContent`, so it exercises the pre-existing guard rather than the line you added to fix blocker 1. Driving it through the real path would also catch the leak above.

Also worth reconsidering: `guard !showsStopCallout else { return }` inside `apply(to:)` is a permanently-true branch guarding a constant. If callouts are genuinely gone, a constant with one position is just future confusion.

Send it back once the flag is scoped to first load and the dead branch is out.

Borisserz:
`applyOriginStopSelection` arms `skipNextStopTimeHighlight` only on the first load. Later `$tripDetails` emissions (every 30s) clear the flag, so a value-equal `TripStopTime` cannot leave skip stuck and swallow the next pin tap.

`didSelect` no longer has the dead `canShowCallout` true branch (callouts are compile-time off). `TripMapAnnotationPolicy.apply` always sets `canShowCallout = false` and nils accessories.

New coverage drives `applyOriginStopSelection` twice without loading `controller.view`: first load arms skip, a refresh clears it, then a tap still opens the stop. `TripMapAnnotationPolicyTests` 5/5 passed locally.

aaronbrethorst:
### Code review

The four items from the last round are addressed: the skip flag is now scoped by `isFirstStopTimeLoad`, the dead `canShowCallout == true` branch under the guard is gone, `showsStopCallout` and its tautological guard are gone, and a new test drives `applyOriginStopSelection` through two loads. But the fix for the leak inverts the failure rather than removing it.

Found 1 issue:

1. Clearing the skip flag on every non-first `tripDetails` emission makes the 30-second refresh navigate to the origin stop on its own. `applyOriginStopSelection` reasons that later emissions carry a value-equal `TripStopTime`, so the reassignment is inert — but that only holds while `selectedStopTime` is still the origin stop. Any ordinary gesture moves it off:

   - Tapping a stop row in the trip list runs `showStopOnMap(_:)`, which sets `selectedStopTime = tripStop.stopTime` for stop B.
   - Tapping the map background (or another pin) deselects the origin annotation, and `didDeselect` sets `selectedStopTime = nil`.
   - Returning from a pushed stop page re-runs `viewWillAppear` → `viewModel.start()` → `loadData()`, and the selection was cleared by the `didDeselect` that preceded the push.

   In each case the next emission takes the `else` branch, sets `skipNextStopTimeHighlight = false`, and assigns the origin stop back into `selectedStopTime`. `oldValue != self.selectedStopTime` now passes, `didSet` calls `mapView.selectAnnotation`, MapKit fires `didSelect`, the guard falls through because the flag was just cleared, and `openStop(stopTime)` pushes a `StopViewController` the rider never asked for. Repro: open a trip from a departure, tap any stop row in the list, wait 30 seconds without touching anything — the stop page for your origin stop appears. On `main` that same re-selection only highlighted the list row, so this is new; it is the round-2 bounce, moved from load time to refresh time. The new test locks the behavior in: `#expect(!controller.skipNextStopTimeHighlight)` after the second `applyOriginStopSelection` asserts exactly the state that lets the auto-push through, and it passes only because the test never lets `selectedStopTime` diverge from the origin between calls.

   The shape to aim for is the one suggested last round: arm the flag for *every* programmatic selection, and clear it when the assignment did not actually change the selection (i.e. when `didSet` returns before `selectAnnotation`), rather than deciding from load ordinality. (bug due to `OBAKit/Trip/TripViewController.swift` clearing a one-shot guard in the `else` branch of a repeating Combine sink)

https://github.com/OneBusAway/onebusaway-ios/blob/49c0f4fe42d16c92dfd528a2489f881d0af3f6bb/OBAKit/Trip/TripViewController.swift#L459-L468

One non-blocking note, no action needed unless you disagree: arming the flag on first load also skips `tripDetailsController.highlightStopInList(_:)` and the `floatingPanel.move(to: .half/.full)` that used to run on load, since both sat below the same guard. The panel's `initialState` is `.tip`, so opening a trip from a departure now leaves the list collapsed where it previously animated up to half height and blinked the origin row. That is an unavoidable consequence of the guard placement you were asked for, but the PR description still says the origin stop "is highlighted in the list", which is no longer true.

aaronbrethorst (changes requested):
Thanks for sticking with this one through three rounds — and to be clear, all four items from the last round are genuinely fixed. I checked them in the code rather than taking the commit messages for it. Pulling the callout policy out into `TripMapAnnotationPolicy` is a real improvement over the inline `canShowCallout` juggling, and the tests around it are the right shape.

There is one blocker left, and it's a new one that came in with the fix rather than something you missed.

`applyOriginStopSelection` clears `skipNextStopTimeHighlight` on every non-first emission. The reasoning in the doc comment — later emissions reassign a value-equal `TripStopTime`, so the `didSet` bails — only holds while `selectedStopTime` is still the origin stop. It stops holding as soon as the rider does anything:

- Tap the map background, or another pin: `didDeselect` fires and sets `selectedStopTime = nil`.
- Tap a row in the trip list: `showStopOnMap` sets it to a different stop.

Now the next 30-second `tripDetails` emission runs `applyOriginStopSelection`, takes the `else` branch, and reassigns the origin stop. This time `oldValue != selectedStopTime` passes, so the `didSet` calls `selectAnnotation`, `didSelect` runs with the flag cleared, and — because this PR makes `didSelect` call `openStop` — the app pushes a `StopViewController` the rider never asked for. On main the same re-selection only highlighted a list row, which is why this is new.

I confirmed the refresh really does fire: `TripViewModel.refreshInterval` is 30s, `loadData()` assigns `tripDetails` unconditionally, and `shouldSkipProgrammaticRefresh` is never assigned anywhere in the codebase, so the `?? false` always takes the refresh. So this is every rider, every 30 seconds, after any map interaction.

The shape I'd suggest is the one from the last round: arm the flag on every programmatic selection, and clear it when the assignment didn't actually change the selection — rather than deciding from load ordinality. That keeps the guard tied to "did we select this, or did the rider" instead of "is this the first load".

Two smaller things while you're in here, neither blocking:

Arming on first load also skips `highlightStopInList` and the `floatingPanel.move(to:)` below the same guard, so opening a trip from a departure now leaves the list collapsed at a tip where it used to animate to half height with the origin row highlighted. The PR description still says the origin stop "is highlighted in the list", which isn't true anymore. That's a consequence of the guard placement I asked for, so it's on me as much as you — but worth deciding deliberately rather than by accident.

`Programmatic selection does not open the stop` still sets the flag by hand instead of driving `bindTripContent`. The new test gets closer by calling `applyOriginStopSelection` directly, but neither exercises the sink, which is where this bug lives — a test that drove the sink twice with an interaction in between would have caught it.

Happy to re-review as soon as the selection guard is reworked.

Borisserz:
`$tripDetails` still republishes ~every 30s. `applyOriginStopSelection` now writes the origin only on first load, or while the origin is still selected. After `didDeselect` (`selectedStopTime = nil`) or a list tap onto another stop, a later emission leaves that choice alone — it does not `selectAnnotation` the origin, so `didSelect` does not `openStop`.

`selectedStopTime`’s `didSet` clears `skipNextStopTimeHighlight` on a value-equal assign, so a leaked arm cannot swallow the next pin tap. Skip is still armed on every programmatic assign (first load, `showStopOnMap`).

`didSelect` still skips list highlight and the panel move when the flag is set: those share that entry, and tests must not load `.view`. The origin pin is selected on the map. Documented in `docs/trip-map-callouts.md`.

New test: apply → `selectedStopTime = nil` → apply again → still `nil`. `TripMapAnnotationPolicyTests` (6) passed on iPhone 17 Pro. Re-assigning origin on every emission fails that test.

aaronbrethorst:
### Code review

The round-4 blocker is genuinely fixed, and I traced it rather than taking the commit message for it: `applyOriginStopSelection` now writes the origin only on first load or while the origin is still the selected stop, so a 30s emission after a deselect or a list tap onto another stop returns early and never calls `selectAnnotation` — no self-navigating stop page. The value-equal branch in `selectedStopTime`'s `didSet` clears a leaked arm, and `TripStopTime.isEqual` really is stable across a refresh (`arrival`/`departure` are scheduled offsets from `serviceDate`), so that branch is reachable. `TripMapAnnotationPolicy.apply(to:)` is exercised for real by `Trip controller applies the no-callout policy to stop views`.

Found 1 issue:

1. `openStop` never deselects the annotation it opened, so a stop pin cannot be opened twice in a row — and the origin pin cannot be opened at all on first tap. MapKit does not re-fire `didSelect` for an annotation that is already selected, and nothing here clears the selection: `didSelect` doesn't touch `selectedStopTime`, `openStop` doesn't call `deselectAnnotation`, and the only `deselectAnnotation` in the file operates on `oldValue` inside `selectedStopTime`'s `didSet`. The 30s refresh doesn't clear it either — `mapView.updateAnnotations(with:)` diffs by `Hashable`, and `TripStopTime.hash` combines `arrival`/`departure`/`regionIdentifier`/`stopID`, all unchanged, so the annotation objects are never removed and re-added. Concretely: open a trip from a departure, `applyOriginStopSelection` selects the origin pin programmatically, and the rider's tap on that pin does nothing — it's already selected. Same after any tap: tap pin B, read the stop page, come back, tap pin B again, nothing. The rider has to tap a different pin or the background first. Since this PR makes pin selection the map's only "open this stop" affordance, that's the whole affordance failing on the most likely gesture. The codebase already handles this: `MapViewController`'s selection-as-open path calls `show(stop:deselecting: view.annotation)` → `prepareMapForPushedStopPage` → `deselectAnnotation`, and the comment right above it warns about leaving "annotations that select but never open." A `mapView.deselectAnnotation(stopTime, animated: false)` in `openStop` (or passing the annotation through) matches that. (bug due to `OBAKit/Trip/TripViewController.swift` `didSelect` calling `openStop` without deselecting, vs. `OBAKit/Mapping/MapViewController.swift:771-778` and `:819-824`)

https://github.com/OneBusAway/onebusaway-ios/blob/a39cbe3c1bee31ff0e95586d5d5f2fd5d1261c68/OBAKit/Trip/TripViewController.swift#L340-L349

Two non-blocking notes, no action required unless you agree:

`docs/trip-map-callouts.md` says "`didSelect` still skips list highlight and the floating-panel move when the flag is set." The list highlight part is right, but the `floatingPanel.move(to: .half/.full)` calls were deleted from `didSelect` in this PR — there's no panel move left under that guard for the flag to skip.

The PR description's test plan still says the origin stop "is highlighted in the list" on load. It isn't, for the reason from last round: `highlightStopInList` sits under the skip guard. The doc file gets this right; the description is stale.

aaronbrethorst (changes requested):
Round four's blocker is genuinely fixed, and I want to say so plainly because this one has taken a lot of passes.

`applyOriginStopSelection` no longer decides by load ordinality alone. It bails when the rider has picked a different stop, and bails when they've deselected, so the only write left on a later emission is a value-equal assign that `didSet` short-circuits before it can reach `selectAnnotation`. I walked the four gestures I listed last time — row tap, background tap, another pin, return from the stop page — and none of them lets the 30-second refresh drag the map back to the origin stop. `Refresh does not reselect origin after the rider deselects` covers it. The `oldValue == selectedStopTime` early return from round three holds up too; `TripStopTime`'s equality is over scheduled offsets that are stable across a refresh, so that branch is genuinely reachable rather than decorative.

The tests are also real now, which was my round-one complaint. Building the controller with an `arrivalDeparture` means the `MinimalStopAnnotationView` binding in `viewFor` actually succeeds, so `TripMapAnnotationPolicy.apply(to:)` runs and reverting the production hunk fails the suite.

One thing still needs fixing, and it's a consequence of what this PR set out to do.

Removing the callout makes selection the trip map's only way to open a stop. But nothing ever deselects. `didSelect` doesn't write `selectedStopTime` — it calls `highlightStopInList` and `openStop` and stops — and `openStop` navigates away without touching map selection. The only `deselectAnnotation` in the file acts on `oldValue` inside `selectedStopTime.didSet`, which never runs for a rider tap.

So the pin stays selected after the stop page opens, and MapKit doesn't re-fire `didSelect` for an annotation that's already selected. Tap a pin, look at the stop, come back, tap the same pin: nothing. Worse on load — `applyOriginStopSelection` selects the origin pin programmatically, so the very first tap on the rider's own stop, which is the most likely thing they'll touch on this screen, does nothing at all. They have to tap the map background first to break the selection, with no way to know that.

The map side of the app already solved this. `MapViewController.show(stop:deselecting:)` routes through `prepareMapForPushedStopPage`, which calls `deselectAnnotation` with a comment saying that leaving the tapped pin selected "only strands a highlight for the rider to find on the way back." `openStop` wants the same thing — deselect the annotation before or as you navigate. Worth a test that taps a pin twice.

Two small stale bits while you're in there. `docs/trip-map-callouts.md` says `didSelect` "still skips list highlight and the floating-panel move when the flag is set," but this PR deleted the `floatingPanel.move(to:)` calls from `didSelect` entirely, so there's no panel move left under that guard. And the PR description's test plan still says the origin stop is highlighted in the list on load; it isn't, because `highlightStopInList` sits below the skip guard. I mentioned that one last round.

Separately from the code: the branch now conflicts with `main` and needs a rebase before it can land. That's merge state, not a defect — no re-review needed for the resolution unless it changes behavior.

Fix the deselect and this is done.

Borisserz:
Rebased onto current `main`.

`didSelect` now deselects on the `MKMapView` MapKit passed in — both on the skip path (load-time origin select) and in `openStop` after a rider tap. Same reason as `MapViewController`: leaving the pin selected means MapKit will not fire `didSelect` again, so the first tap on the origin stop (and a second tap after coming back) did nothing.

Tests use that same passed-in map, not `controller.view`. `Opening a stop deselects the pin so a second tap still opens` fails if `deselectAnnotation` is removed (`selectedAnnotations` stays non-empty; nav count stays 2).

`docs/trip-map-callouts.md` no longer claims a panel move or a list highlight on load.

`TripMapAnnotationPolicyTests` (8) passed on iPhone 17 Pro.

aaronbrethorst:
### Code review

Round five's blocker is fixed in the production code: `didSelect` now deselects on the skip path and `openStop` deselects before navigating, so the origin pin is tappable after load and a pin can be re-tapped after coming back from the stop page. Both paths use the `MKMapView` MapKit passes in, matching `MapViewController.prepareMapForPushedStopPage`.

Found 1 issue:

1. Neither test added for that fix can fail if the fix is reverted. Both build an `MKMapView()` — zero frame, never in a window, no delegate, no annotation views — then call `mapView.selectAnnotation(stopTime, animated: false)`. Apple documents that method as: "If the specified annotation isn't onscreen, and, therefore, doesn't have an associated annotation view, this method has no effect." So `selectedAnnotations` is empty before `didSelect` is ever called, and `#expect(mapView.selectedAnnotations.isEmpty)` — the only assertion in `Programmatic selection deselects so the origin pin can be tapped`, and the only deselect-sensitive assertion in `Opening a stop deselects the pin so a second tap still opens` — passes with or without the `deselectAnnotation` lines in `TripViewController`. The suite's own first test works around exactly this limitation by calling `registerAnnotationView` and invoking `mapView(_:viewFor:)` by hand, which is the evidence that MapKit isn't building views for this map. The nav-count assertions don't cover the gap either: the second `didSelect` is invoked directly rather than being re-fired by MapKit, so `nav.viewControllers.count == 3` holds whether or not the pin was deselected — the re-fire is the entire mechanism the fix exists to restore.

https://github.com/OneBusAway/onebusaway-ios/blob/8ba2c935f90cff4eaaa1337adb7594fc9b8657c0/OBAKitTests/Trip/TripMapAnnotationPolicyTests.swift#L187-L205

aaronbrethorst (approved):
Approving. Six rounds is a lot to ask of anyone and you worked through every item — thank you for sticking with it.

I went back through all five earlier reviews and checked each item against the current head rather than taking the replies at face value. All of them are genuinely closed: the tests no longer pass against unmodified `main` for the policy hunk, `calloutAccessoryControlTapped` is gone with the body living on as `openStop`, "Closes #713" became "Refs", the tautological `showsStopCallout` guard is deleted, the dead `canShowCallout == true` branch in `didSelect` is gone, the skip-flag leak across the 30s refresh is closed by the `oldValue == selectedStopTime` early return, the self-navigating refresh is fixed, and the dead-second-tap problem is fixed by deselecting on both paths.

I then walked the behavior myself — load, list-row tap, pin tap, return from the stop page, deselect, and the 30-second refresh. None of them self-navigate, and none swallow a tap. That's what I needed to see.

**One correction to the automated review comment above:** it claims the two new deselect tests are inert because `selectAnnotation` is a no-op on a zero-frame, unhosted `MKMapView`. I tested that directly — on a zero-frame map that was never added to a window, with the annotation added first, `selectAnnotation` **did** take effect (`selectedAnnotations.count == 1`, and `view(for:)` returned non-nil). So `#expect(selectedAnnotations.isEmpty)` is a real assertion, not a vacuous one. Please disregard that finding; I don't want you chasing a non-issue after five rounds of real ones.

Filed #1349 for the leftovers, all tidiness rather than behavior: now that `selectedStopTime` is nil except transiently, `applyOriginStopSelection`'s first guard and the `didSet`'s `oldValue ==` branch are unreachable, and `docs/trip-map-callouts.md`'s "Refresh" section describes a state machine that no longer exists. The PR description's test plan is also still stale — it claims the origin stop is selected and highlighted on load, and neither is true now.

PR #1260, opened 2026-08-02
Clarify Report Problem trip section copy

### LINKED ISSUE #451: Clarify what the reporting options for trip issues mean
Feedback from a beta user:
- Find a way to make what "report a problem for this trip" does clearer
- 'Providing a way to send feedback to the service provider on the bus service itself would also be valuable'
- "When I see buses leaving 3-5 minutes early I want to report it (as a former driver I know there's no excuse) but don't know if it's your problem or theirs. Having a provider feedback link might be an easy way to add that value for me and perhaps improve transit service too."

Borisserz:
## Summary
- Renames the Report Problem hub section from “Problem with a Vehicle at the Stop” to “Problem with a Trip”.
- Doc comments spell out that these reports are agency/service feedback, not app bugs.
- Closes #451.

## Test plan
- [x] `ReportProblemCopyTests` — 2/2 on iPhone Air simulator
- [ ] Open Report a Problem from a stop and confirm the trip section header reads “Problem with a Trip”

## Summary by CodeRabbit

* **Improvements**
  * Clarified report categories by distinguishing stop-related issues from trip-related issues.
  * Updated the reporting interface to use consistent section labels.
  * Refined translations for trip-problem reporting across supported languages.

### DIFF
--- a/OBAKit/Reporting/ReportProblemCopy.swift
+++ b/OBAKit/Reporting/ReportProblemCopy.swift
@@ -1,0 +1,29 @@
+//
+//  ReportProblemCopy.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import OBAKitCore
+
+/// User-facing copy for the Report Problem hub screen.
+enum ReportProblemCopy {
+    static var stopProblemHeader: String {
+        OBALoc(
+            "report_problem_controller.stop_problem.header",
+            value: "Problem with the Stop",
+            comment: "Section header for reporting incorrect stop names, numbers, locations, or missing routes or trips, not app bugs."
+        )
+    }
+
+    static var vehicleProblemHeader: String {
+        OBALoc(
+            "report_problem_controller.trip_problem.header",
+            value: "Problem with a Trip",
+            comment: "Section header for reporting a problem with a specific trip or vehicle service. Feedback helps improve transit data."
+        )
+    }
+}
--- a/OBAKit/Reporting/ReportProblemViewController.swift
+++ b/OBAKit/Reporting/ReportProblemViewController.swift
@@ -12,7 +12,9 @@ import OBAKitCore
 
 /// The 'hub' view controller for reporting problems about stops and trips.
 ///
-/// From here, a user can report a problem either about a `Stop` or about a trip at that stop.
+/// From here, a user can report incorrect stop details or a problem with a specific
+/// trip's service. These reports are sent to the transit agency to improve data and
+/// service — they are not app bug reports.
 ///
 /// - Note: This view controller expects to be presented modally.
 class ReportProblemViewController: TaskController<StopArrivals>,
@@ -98,7 +100,7 @@ class ReportProblemViewController: TaskController<StopArrivals>,
             self.navigationController?.pushViewController(stopProblemController, animated: true)
         }
 
-        return OBAListViewSection(id: "stop_problem_section", title: OBALoc("report_problem_controller.stop_problem.header", value: "Problem with the Stop", comment: "A table header in the 'Report Problem' view controller."), contents: [row])
+        return OBAListViewSection(id: "stop_problem_section", title: ReportProblemCopy.stopProblemHeader, contents: [row])
     }
 
     private var vehicleProblemSection: OBAListViewSection? {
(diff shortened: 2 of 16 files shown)


aaronbrethorst:
### Code review

Found 2 issues:

1. The copy change reuses the existing key `report_problem_controller.vehicle_problem.header` rather than introducing a new one, so only English actually changes. The other twelve catalogs still hold that key with translations of the *old* wording — `fr` "Problème concernant un véhicule à cet arrêt", `es` "Problema con un vehículo en la parada", `pl` "Problem z pojazdem na przystanku", and so on. `NSLocalizedString` falls back to the `value:` argument only when the key is *missing*; here it is present in every locale, so the stale translation wins and non-English riders keep seeing exactly the copy #451 asks to fix. The convention in this repo for a changed string is a new key, translated into all thirteen catalogs, with the old key removed — commit `5a95d4fe` did precisely that when `stop_page.mode.chronological` became `stop_page.mode.time`.

https://github.com/OneBusAway/onebusaway-ios/blob/4217560ce8985f4e757c52676ca44506c118fab2/OBAKit/Strings/en.lproj/Localizable.strings#L833-L836

2. The new translator comment describes the stop section as "stop amenities or signage", but `StopProblemCode` (OBAKitCore/Models/REST/ProblemReporting.swift) offers only name-wrong, number-wrong, location-wrong, route-or-trip-missing, and other — there is no amenity or signage reporting path anywhere in the app. This string ships into `en.lproj` as the `/* … */` comment that the thirteen translators read, and it replaces a comment that was accurate. The same wording is repeated in the `ReportProblemViewController` doc comment. (The vehicle-section rewording itself checks out — `TripProblemCode` is never-came / came-early / came-late / wrong-headsign / does-not-stop-here, which "Problem with a Trip" describes well.)

https://github.com/OneBusAway/onebusaway-ios/blob/4217560ce8985f4e757c52676ca44506c118fab2/OBAKit/Reporting/ReportProblemCopy.swift#L17-L19

aaronbrethorst (changes requested):
The copy change is the right call — "Problem with a Vehicle at the Stop" really does read like it's about a bus shelter rather than a trip, and #451 is a fair complaint. Pulling the two headers into `ReportProblemCopy` so they're not buried in view-controller code is a good move, and I checked the rewording against `TripProblemCode`: it's accurate about what that section actually reports.

The blocker is that for 12 of 13 locales, this PR changes nothing.

**`report_problem_controller.vehicle_problem.header` keeps its key while its English value changes.** `NSLocalizedString` falls back to `value:` only when a key is *missing* — and this key is present in every non-English catalog, translated to the old wording. I checked:

```
fr: "Problème concernant un véhicule à cet arrêt"
es: "Problema con un vehículo en la parada"
pl: "Problem z pojazdem na przystanku"
ru: "Проблема с транспортом на остановке"
```

...and the same holds for ko, zh-Hans, zh-Hant, ar, it, vi, fil, pt-BR. So the stale translation wins and non-English riders keep reading exactly the copy #451 asks to fix. Only English users see the improvement.

The fix is the pattern already established here — mint a new key (`report_problem_controller.trip_problem.header`), translate it into all thirteen catalogs, and drop the old one. `5a95d4fe` did precisely this for `stop_page.mode.chronological` → `stop_page.mode.time`, with the commit message "New keys, translated into all thirteen catalogs, with the old ones removed."

One smaller thing worth fixing in the same pass:

**The new translator comment on `stopProblemHeader` says "stop amenities or signage,"** but `StopProblemCode` only offers name-wrong / number-wrong / location-wrong / route-or-trip-missing / other. There's no amenity or signage path in the app. That comment ships into `en.lproj` as the `/* */` guidance 13 translators work from, and it replaced a comment that was accurate — so it'd actively mislead the translations you're about to request. Same wording is repeated in the `ReportProblemViewController` doc comment.

On the tests: `ReportProblemCopyTests` asserts a localized string equals its own hardcoded English value. That's not wrong, but it's worth knowing it would have stayed green through the exact bug above — it passes while 12 locales show the old string. If you want the suite to earn its place, asserting that the *key* resolves differently from its fallback, or simply dropping the tests and relying on the catalogs, would both be more honest. Your call; not a blocker.

Get the key renamed and the translations in and I'll merge this happily.

Borisserz:
Addressed the review items: introduced report_problem_controller.trip_problem.header in all 13 catalogs with trip-specific translations, removed the old vehicle key, and updated the translator and controller documentation to match StopProblemCode. Removed the self-equality copy tests rather than retain coverage that missed localization regressions. Merged current upstream/main and force-pushed a rewritten commit authored by boris without co-author trailers.\n\nVerification: plutil validated all 13 catalogs and the new key is present exactly once in each. The full OBAKitTests build did not reach tests because current upstream/main fails compiling OBAKit TripPage sources under this Xcode with the unknown RegionIsolation warning group (exit 65).

aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is exactly the fix I was hoping for. Minting a new key instead of reusing the old one was the right call, and you carried it all the way through: I verified 14 additions of `report_problem_controller.trip_problem.header` against 14 removals of the vehicle key, across all 13 catalogs plus the source — no orphan, no duplicate, correct alphabetical placement. The translations read as trip/service sense rather than the old vehicle-at-stop phrasing.

The reworded comment on the stop key now matches what `StopProblemCode` actually offers, too.

Merging.

PR #1261, opened 2026-08-02
Add More tab tutorial, phone, and text config keys

### LINKED ISSUE #614: Customize "more" page

Borisserz:
## Summary
- Adds optional `TutorialURL`, `PhoneURL`, and `TextURL` under `OBAKitConfig.MoreTab`.
- When set, the More tab shows a Contact & Help section (Tutorials / Call Agency / Text Agency).
- Documents the keys in `README.markdown`. `CustomLinks` still covers arbitrary extra URLs.
- Closes #614.

## Test plan
- [x] `MoreTabConfigurationTests` — 11/11 on iPhone Air simulator
- [ ] Set `PhoneURL: tel:+1…` in a white-label `project.yml`, regenerate, confirm Call Agency opens Phone
- [ ] Leave keys unset on OneBusAway — no Contact & Help section appears

### DIFF
--- a/OBAKit/Settings/MoreViewController.swift
+++ b/OBAKit/Settings/MoreViewController.swift
@@ -71,6 +71,7 @@ public class MoreViewController: UIViewController,
             donateSection,
             updatesAndAlertsSection,
             myLocationSection,
+            agencyContactSection,
             helpOutSection,
             customLinksSection,
             aboutSection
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is the right shape for a white-label feature, and I checked the part that matters most for agencies that don't opt in: neither `Apps/OneBusAway/project.yml` nor `Apps/KiedyBus/project.yml` sets any of the three keys, `Bundle.moreTabConfiguration` falls back to `.default` with all three `nil`, `guard !contents.isEmpty else { return nil }` drops the section, and `items(for:)` ends in `.compactMap { $0 }`. So an unconfigured agency sees no section at all — no empty row, no stray header, no crash. The new memberwise-init parameters are defaulted too, so nothing existing breaks.

Following the `helpOutSection`/`customLinksSection` pattern was the right call, and treating an empty string as `nil` in `url(from:key:)` is a nice touch — it means an agency can blank a key out in a build config without getting a dead row. (`TranslateURL`/`DevelopURL` don't do that; yours is the better behavior, and worth propagating eventually.)

- The four new `OBALoc` keys — `more_controller.tutorials`, `more_controller.call_agency`, `more_controller.text_agency`, `more_controller.agency_contact.header` — aren't in any catalog under `OBAKit/Strings/` yet. English is unaffected because `OBALoc` passes `value:` through to `NSLocalizedString`, so this is a "next `scripts/extract_strings` run" item rather than a bug. Mentioning it because recent work here has tended to land the keys in-PR.
- A scheme-less `PhoneURL` like `"2065551212"` passes `URL(string:)` as a relative URL, so the row renders and `application.open` then quietly fails — same on hardware that can't place calls. That's identical to how the existing `translateURL`/`developURL`/`CustomLinks` rows behave, so it's not something you introduced, and your README documents the correct `tel:`/`sms:` form. A `canOpenURL` gate across all of those rows would be a decent standalone cleanup.

Merging.

PR #1262, opened 2026-08-02
Add setup script and local signing overrides

### LINKED ISSUE #326: Improve documentation for initial setup, maybe a setup script, and probably xcodegen tweaks
Key takeaways:

- [ ] Try running through the setup process without ever opening Xcode on the system and seeing what breaks (probably easiest with a VM)
- [ ] Update documentation accordingly
- [ ] Consider writing a small setup script that will check for dependencies and offer a clearer list of what's needed to the user
- [ ] Improve documentation about development teams and bundle identifiers
- [ ] Look into creating some sort of 'local settings' file that can be used to customize the bundle identifiers for development without being part of the OBA org. The settings file should be `.gitignore`d.

------

I received a report about the documentation for initial setup needing some clarification. Here are the notes:

## xcodebuild missing

> I ran into my first issue. I got xcrun: error: unable to find utility "xcodebuild", not a developer tool or in PATH when I tried to run
scripts/carthage_build
> 
> I think it was because I didn't open Xcode yet on my system. I opened it and it now says "Installing components" which sounds good.
> 
> I was able to fix it by running `sudo xcode-select -s /Applications/Xcode.app/Contents/Developer`

## running on device

> Hm, I can't get it to deploy to my device
> Failed to register bundle identifier: The app identifier "org.onebusaway.iphone.TodayView" cannot be registered to your development team because it is not available. Change your bundle identifier to a unique string to try again.
> I set all the projects to my personal team, following Apple's instructions.

Alan replied:

> after setting projects to your personal team, change the app identifiers to something different than the default. you can add your name or something unique, “uniquename.onebusaway.iphone” and “uniquename.onebusaway.iphone.TodayView” for example.

The user replied:

> Got it, and I also had to update the references to that in the plist files
> Well, one plist, on the NSUserActivityTypes iirc

Alan said:

> it may be helpful to keep a Git Patch handy to automatically update these values because everytime you generate_project, it will overwrite what you just did with the default

And finally I said:

> I think we could set something up with xcodegen and a local settings file to make this easier too.

Borisserz:
## Summary
- Adds `scripts/setup` to check Xcode / XcodeGen and generate the project without opening Xcode first.
- Makes a present `Apps/*/local.yml` authoritative for `DEVELOPMENT_TEAM` in the generated project while retaining the shared include list.
- Documents why a standalone `PRODUCT_BUNDLE_IDENTIFIER` override cannot produce a working personal device build.
- Closes #326.

## Test plan
- [x] `ruby -c scripts/generate_project`
- [x] Generate OneBusAway with a temporary `local.yml`; verify the temporary team ID appears four times in `project.pbxproj`
- [x] Substitute a failing `xcodegen`; verify `scripts/generate_project` exits nonzero without success output
- [ ] On a clean machine: install XcodeGen, run setup, open `OBAKit.xcodeproj`

### DIFF
--- a/scripts/generate_project
+++ b/scripts/generate_project
@@ -1,5 +1,7 @@
 #!/usr/bin/env ruby
 
+require "yaml"
+
 app = ARGV[0]
 
 if app.nil?
@@ -38,23 +40,54 @@ BEGIN {
     `rm -rf OBAKit.xcodeproj`
     `cp #{project_yml_path(app)} .`
     append_local_overrides(app)
-    puts `xcodegen`
+    system("xcodegen")
+    abort "XcodeGen failed; OBAKit.xcodeproj was not generated." unless $?.success?
     `mkdir -p OBAKit.xcodeproj/project.xcworkspace/xcshareddata/swiftpm`
     `cp #{package_resolved_path(app)} OBAKit.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/`
   end
 
   # XcodeGen's `optional: true` include still errors when the file is absent, so
   # we only wire Apps/<App>/local.yml when the contributor created it (#326).
+  #
+  # The including project.yml takes precedence over includes, so DEVELOPMENT_TEAM
+  # must also be copied into its target settings for local.yml to be authoritative.
   def append_local_overrides(app)
     local = "Apps/#{app}/local.yml"
     return unless File.exist?(local)
 
-    File.open("project.yml", "a") do |f|
-      f.puts ""
-      f.puts "include:"
-      f.puts "  - path: #{local}"
+    local_settings = YAML.load_file(local)
+    project = File.read("project.yml")
+    project = splice_include(project, local)
+
+    local_settings.fetch("targets", {}).each do |target, target_settings|
+      team = target_settings.dig("settings", "base", "DEVELOPMENT_TEAM")
+      next unless team
+
+      project = replace_development_team(project, target, team)
     end
+
+    File.write("project.yml", project)
     puts "Applying local overrides from #{local}"
+  end
+
+  def splice_include(project, local)
+    include = project.index(/^include:\s*$/)
+    abort "Could not find include list in project.yml." unless include
+
+    line_end = project.index("\n", include)
+    project.insert(line_end + 1, "  - path: #{local}\n")
+  end
+
+  def replace_development_team(project, target, team)
+    target_start = project.index(/^  #{Regexp.escape(target)}:\s*$/)
+    abort "Could not find target #{target} in project.yml." unless target_start
+
+    next_target = project.index(/^  \S/, target_start + 1) || project.length
+    target_spec = project[target_start...next_target]
+    replacement_count = target_spec.sub!(/^(\s{8}DEVELOPMENT_TEAM:\s*).+$/, "\\1#{team}")
+    abort "Could not find DEVELOPMENT_TEAM for target #{target} in project.yml." unless replacement_count
+
+    project[0...target_start] + target_spec + project[next_target..]
   end
 
   def project_yml_path(app)
--- a/scripts/generate_project
+++ b/scripts/generate_project
@@ -37,9 +37,24 @@ BEGIN {
   def generate_project(app)
     `rm -rf OBAKit.xcodeproj`
     `cp #{project_yml_path(app)} .`
+    append_local_overrides(app)
     puts `xcodegen`
     `mkdir -p OBAKit.xcodeproj/project.xcworkspace/xcshareddata/swiftpm`
     `cp #{package_resolved_path(app)} OBAKit.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/`
+  end
+
+  # XcodeGen's `optional: true` include still errors when the file is absent, so
+  # we only wire Apps/<App>/local.yml when the contributor created it (#326).
+  def append_local_overrides(app)
+    local = "Apps/#{app}/local.yml"
+    return unless File.exist?(local)
+
+    File.open("project.yml", "a") do |f|
+      f.puts ""
+      f.puts "include:"
+      f.puts "  - path: #{local}"
+    end
+    puts "Applying local overrides from #{local}"
   end
 
   def project_yml_path(app)
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,7 @@ OBAKitTests.xcresult
 
 # Review artifacts (e.g. /aaron-pr-review)
 pr-*.md
+
+# Personal XcodeGen overrides (team ID / bundle ID). See Apps/*/local.yml.example.
+Apps/*/local.yml
+
--- a/Apps/KiedyBus/local.yml.example
+++ b/Apps/KiedyBus/local.yml.example
@@ -1,0 +1,12 @@
+# Copy to Apps/KiedyBus/local.yml (gitignored) and adjust for your Apple team.
+# See Apps/OneBusAway/local.yml.example for notes.
+
+targets:
+  App:
+    settings:
+      base:
+        DEVELOPMENT_TEAM: YOUR_TEAM_ID
+  OBAWidget:
+    settings:
+      base:
+        DEVELOPMENT_TEAM: YOUR_TEAM_ID
--- a/Apps/KiedyBus/local.yml.example
+++ b/Apps/KiedyBus/local.yml.example
@@ -1,5 +1,8 @@
 # Copy to Apps/KiedyBus/local.yml (gitignored) and adjust for your Apple team.
-# See Apps/OneBusAway/local.yml.example for notes.
+# `scripts/generate_project` copies DEVELOPMENT_TEAM into the generated
+# project, so this file is authoritative for local signing. See
+# Apps/OneBusAway/local.yml.example before changing a bundle identifier:
+# associated identifiers must be changed together in a white-label config.
 
 targets:
   App:
--- a/Apps/KiedyBus/project.yml
+++ b/Apps/KiedyBus/project.yml
@@ -92,3 +92,4 @@ include:
   - path: OBAKit/project.yml
   - path: OBAKitTests/project.yml
   - path: OBAWidget/project.yml
+
--- a/Apps/OneBusAway/local.yml.example
+++ b/Apps/OneBusAway/local.yml.example
@@ -1,0 +1,17 @@
+# Copy to Apps/OneBusAway/local.yml (gitignored) and adjust for your Apple team.
+# Regenerated into the Xcode project via `scripts/setup` / `scripts/generate_project`.
+#
+# XcodeGen merges this file over the shared project.yml, so you can develop
+# without belonging to the OneBusAway organization team.
+
+targets:
+  App:
+    settings:
+      base:
+        DEVELOPMENT_TEAM: YOUR_TEAM_ID
+        # PRODUCT_BUNDLE_IDENTIFIER: com.example.onebusaway
+  OBAWidget:
+    settings:
+      base:
+        DEVELOPMENT_TEAM: YOUR_TEAM_ID
+        # PRODUCT_BUNDLE_IDENTIFIER: com.example.onebusaway.widget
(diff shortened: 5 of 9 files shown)


aaronbrethorst:
### Code review

Found 2 issues:

1. `append_local_overrides` appends a **second top-level `include:` key** to the copied `project.yml`, which already has one (`Apps/OneBusAway/project.yml:104`). YAML is last-key-wins, so the appended block *replaces* the original nine includes rather than extending them — `Apps/Shared/app_shared.yml` (which supplies `name: OBAKit`), `OBAKitCore/`, `OBAKit/`, `OBAKitTests/`, `OBAWidget/` and the rest are all dropped. Verified against XcodeGen 2.46.0 with the real spec plus the exact text this method emits: `Parsing project spec failed: Decoding failed at "name": Nothing found`. Because `generate_project` runs `rm -rf OBAKit.xcodeproj` first and never checks xcodegen's exit status, a contributor who follows the new README steps 1–3 is left with no working project and a "success" message. This is the unchecked third box in the test plan. The fix is to splice the path into the existing `include:` list instead of appending a new key.

https://github.com/OneBusAway/onebusaway-ios/blob/0a27cf3e22738ad6aec1811183a017f104c90735/scripts/generate_project#L51-L57

2. Even with issue 1 fixed, an XcodeGen `include` **cannot** override a setting the including file sets — the including file takes precedence over its includes, not the other way round. `Apps/OneBusAway/project.yml:82` and `:102` set `DEVELOPMENT_TEAM: 4ZQCMA634J` (KiedyBus sets `85APG5FQFE` at `:66`/`:86`), so `local.yml` is silently ignored. Confirmed by splicing `Apps/OneBusAway/local.yml` into the existing include list and dumping the merged spec: `DEVELOPMENT_TEAM: 4ZQCMA634J` for both `App` and `OBAWidget`. The comment on lines 4–5 below states the inverse of XcodeGen's actual merge order, and the README's signing instructions inherit the same wrong assumption. Overriding these will need a different mechanism (e.g. removing `DEVELOPMENT_TEAM`/`PRODUCT_BUNDLE_IDENTIFIER` from the shared spec so the include is the only source, or post-processing the spec in `generate_project`).

https://github.com/OneBusAway/onebusaway-ios/blob/0a27cf3e22738ad6aec1811183a017f104c90735/Apps/OneBusAway/local.yml.example#L3-L11

Two notes that did check out, for what it's worth: the `optional: true` comment in `generate_project` is accurate (XcodeGen 2.46 still hard-errors on a missing optional include), and CI/Fastlane signing is unaffected, since a fresh checkout never has a gitignored `local.yml` and the new code returns early without one.

scripts/generate_project
```diff
@@ -37,11 +37,26 @@ BEGIN {
...
+  # XcodeGen's `optional: true` include still errors when the file is absent, so
+  # we only wire Apps/<App>/local.yml when the contributor created it (#326).
+  def append_local_overrides(app)
+    local = "Apps/#{app}/local.yml"
+    return unless File.exist?(local)
+
+    File.open("project.yml", "a") do |f|
+      f.puts ""
+      f.puts "include:"
+      f.puts "  - path: #{local}"
```
aaronbrethorst: This appends a **second** top-level `include:` to a file that already has one (`Apps/OneBusAway/project.yml:104`, nine entries). YAML is last-key-wins, so this replaces the whole list rather than extending it:

```
include -> [{'path': 'local.yml'}]
```

That drops `Apps/Shared/app_shared.yml`, which defines `name: OBAKit`, so XcodeGen can't parse the spec at all. Combined with `rm -rf OBAKit.xcodeproj` earlier and no exit-status check on `xcodegen`, a contributor following the README ends up with no project and a success message.

You need to splice into the existing list instead of appending a new key — e.g. parse the YAML, push `{'path' => local}` onto `spec['include']`, and rewrite the file.

Note that fixing this alone isn't enough: `project.yml` sets `DEVELOPMENT_TEAM` directly at `:82`/`:102`, and in XcodeGen the including file beats its includes, so the override will still be ignored — just silently.

aaronbrethorst (changes requested):
A `scripts/setup` that checks dependencies before generating is a genuinely good idea, and #326 is a real papercut — a contributor cloning this repo shouldn't have to discover the signing dance by hitting errors. The script itself is also written carefully: `set -euo pipefail`, a properly quoted `ROOT="$(cd "$(dirname "$0")/.." && pwd)"`, no `rm`, no `curl | sh`, no unquoted expansions, and `$APP` guarded by an `[[ -f "Apps/${APP}/project.yml" ]]` check before use. The `.gitignore` pattern `Apps/*/local.yml` correctly catches real overrides while leaving `local.yml.example` tracked, so a contributor can't accidentally commit their team ID. I checked release signing too: CI and Fastlane both run `generate_project OneBusAway` on fresh checkouts where the gitignored `local.yml` never exists, so `append_local_overrides` returns early and TestFlight behavior is byte-identical. That part I have no concerns about.

The problem is that the headline feature doesn't work, in two separate ways.

**1. Appending a second `include:` deletes the first one.** `Apps/OneBusAway/project.yml:104` already has an `include:` key with nine entries. `append_local_overrides` appends another top-level `include:` to the copied `project.yml`, and YAML is last-key-wins — so the appended one-entry block *replaces* all nine. I confirmed the semantics directly:

```
include -> [{'path': 'local.yml'}]
```

That drops `Apps/Shared/app_shared.yml`, which is where `name: OBAKit` lives, so XcodeGen fails to parse the spec at all. And because `generate_project` does `rm -rf OBAKit.xcodeproj` first and never checks xcodegen's exit status, a contributor who follows the new README steps 1–3 ends up with no project *and* a cheerful "Next steps: Open OBAKit.xcodeproj". That's a worse first-run experience than the one this PR is fixing. It's also the unchecked third box in your own test plan.

**2. Even with that fixed, `include` can't override what the including file sets.** In XcodeGen the including file wins over its includes. `Apps/OneBusAway/project.yml` sets `DEVELOPMENT_TEAM: 4ZQCMA634J` at both `:82` and `:102` (KiedyBus: `85APG5FQFE` at `:66`/`:86`), so a `local.yml` pulled in via `include` is silently ignored. The comment in `local.yml.example:4-5` — "XcodeGen merges this file over the shared project.yml" — asserts the opposite of the real merge order, and the README signing instructions inherit that assumption.

Worth being explicit about the failure ordering: fixing only #1 makes things *quieter but not better* — generation succeeds and the override silently does nothing, which is harder to debug than a parse error.

For the mechanism, I'd suggest making the local file authoritative rather than included: have `append_local_overrides` merge `local.yml`'s settings into the copied `project.yml` (splice into the existing `include:` list *and* strip or overwrite the hardcoded `DEVELOPMENT_TEAM` lines), or read the team ID out of `local.yml` and substitute it directly. Either way it needs a test-run proving a personal team ID actually reaches the generated project.

Two related things while you're in there:

- `generate_project` captures `xcodegen` in backticks and never checks `$?`, so a failed generation still exits 0. `scripts/setup` has `set -e` but can't see it, then creates the swiftpm directory and prints success over a broken tree. Pre-existing, but the new script is what turns it into a confident false success — worth a `abort unless $?.success?`.
- `local.yml.example` advertises `PRODUCT_BUNDLE_IDENTIFIER` as an override, but the app group (`group.org.onebusaway.iphone`, in both the entitlements and `OBAKitConfig.AppGroup`), the merchant ID, the `applinks:` domains, and the `org.onebusaway.iphone.user_activity.*` identifiers are all hardcoded. Changing just the bundle ID won't get a personal team to a working device build, which is the actual complaint in #326.

Minor: stray trailing blank lines added to both `project.yml` files and `.gitignore`, and the README rewrite drops the note that dependencies come in via Xcode's SPM integration.

I'd like this to land — please get a working end-to-end run on a personal team and check that third box, and I'll take another look.

Borisserz:
@aaronbrethorst Addressed both override failures. The generator now splices local.yml into the existing include list and copies each local DEVELOPMENT_TEAM into the copied target settings. An end-to-end run with a temporary team ID found it in the generated pbxproj four times. I also made XcodeGen failure abort generation and documented why bundle-ID-only overrides are unsupported.

aaronbrethorst:
### Code review

Both blockers from the previous round are genuinely fixed, and I verified the mechanism rather than taking the description's word for it. I ran the two string transforms (`splice_include` + `replace_development_team`) against real copies of `Apps/OneBusAway/project.yml` and `Apps/KiedyBus/project.yml`: the include list is spliced, not clobbered (10 entries for OneBusAway, 7 for KiedyBus, all originals intact), the merged spec still parses, and `DEVELOPMENT_TEAM` is rewritten in place for both `App` and `OBAWidget` in the top-level spec — which is the copy that actually wins over includes. `abort ... unless $?.success?` closes the silent-failure hole too. Signing for CI and Fastlane is still unaffected: both run `generate_project OneBusAway` on fresh checkouts where the gitignored `local.yml` never exists, so `append_local_overrides` returns early. `.gitignore`'s `Apps/*/local.yml` correctly ignores real overrides while leaving `local.yml.example` tracked, so a team ID can't be committed, and the rewritten root `project.yml` is covered by the existing `/project.yml` rule.

Found 1 issue:

1. `scripts/setup` tells contributors that `local.yml.example` can override "DEVELOPMENT_TEAM / bundle ID", but `local.yml.example:7-10` in this same PR says "PRODUCT_BUNDLE_IDENTIFIER is intentionally not supported here", and the README repeats that. `append_local_overrides` only reads `settings.base.DEVELOPMENT_TEAM`; a `PRODUCT_BUNDLE_IDENTIFIER` in `local.yml` reaches the spec only through the spliced `include`, where it loses to the hardcoded `PRODUCT_BUNDLE_IDENTIFIER` at `Apps/OneBusAway/project.yml:83`/`:103`. So a contributor who follows this message gets a silent no-op — the exact "quieter but not better" failure mode the last round was about. Looks like a leftover from the previous revision; dropping "/ bundle ID" fixes it.

https://github.com/OneBusAway/onebusaway-ios/blob/24c280b949714c28d466f422a59b9ef3457912fd/scripts/setup#L51-L53

Below the bar for a separate item, but worth a pass while you're in there:

- The stray trailing blank lines are still there. `Apps/OneBusAway/project.yml` and `Apps/KiedyBus/project.yml` are each a one-line diff adding nothing but an empty line, plus one on `.gitignore`.
- `OBAKit/Trip/TripPage/TripPageViewController.swift` gains `import CoreLocation` in a build-tooling PR. The import is correct — line 140 uses `CLLocationCoordinate2D` and reaches it transitively today — but it's unrelated to #326 and belongs in its own change.
- `ok "Xcode / xcodebuild: $(xcodebuild -version ...)"` runs inside the branch where `command -v xcodebuild` already succeeded, so a broken developer dir prints a green check with an empty version string.
- An empty `Apps/<App>/local.yml` makes `YAML.load_file` return `false`, and `false.fetch` raises — after `rm -rf OBAKit.xcodeproj` has already run. Same for a `local.yml` naming a target defined in an included spec rather than the app spec: `replace_development_team` aborts, and the project is already gone.

aaronbrethorst (approved):
Both mechanical problems from my last review are fixed, and I checked them rather than taking the description's word for it: the include list is spliced into the existing `include:` block rather than appended as a second one, and `DEVELOPMENT_TEAM` is written directly into the top-level spec so it actually wins over the includes. Adding the `$?` check after `xcodegen` is a real improvement too — a failed generation used to pass silently.

The part I most wanted to be sure about is release signing, and it's safe: both `.github/workflows/tests.yml` and `fastlane/Fastfile` call `generate_project OneBusAway` on fresh checkouts where `Apps/*/local.yml` cannot exist, so `append_local_overrides` returns early and CI signing is byte-identical. `.gitignore` already covers the rewritten root `project.yml`, so nobody can commit their team ID by accident either.

Lowering the barrier for contributors outside the org is worth having. Merging.

One thing I've filed rather than held this up for: `scripts/setup` still advertises overriding "DEVELOPMENT_TEAM / bundle ID", while `local.yml.example` and the README both say bundle ID is deliberately unsupported — and the code agrees with the docs, so that message sends people down a dead end. Tracked in #1282, along with a suggestion to validate `local.yml` before `rm -rf OBAKit.xcodeproj` runs. Both small if you want them.

PR #1263, opened 2026-08-02
Tick Live Activity countdowns with Text.timer

### LINKED ISSUE #1187: Live Activity countdown should tick on its own, not only when a push arrives
The Live Activity countdown is a static string baked in at push time. `TripLiveActivityCardView.body` does `let now = Date()` and renders `"\(minutes)m"`, so the card only changes when a push arrives — there's no process running behind a Live Activity, and nothing re-renders the archived snapshot on its own.

Everything currently keeping that number honest is a workaround for this:

- obacloud pushes a keepalive every 55s purely to advance the clock (`LiveActivity::KEEPALIVE_INTERVAL`), even when the prediction hasn't moved
- those pushes go out at APNs priority 10, paired with `NSSupportsLiveActivitiesFrequentUpdates` (#1179), because priority 5 is power-deferred and an idle phone simply never receives it

It also leaves one hole we can't close: a user can turn frequent updates off in Settings (`ActivityAuthorizationInfo().frequentPushesEnabled`). For that user we'd hammer priority 10 into their normal budget, get throttled, and their countdown would freeze — with no fallback, because pushes *are* the clock.

If the card could tick on its own, pushes would become *corrections* (the prediction moved) rather than the clock itself, and all of the above relaxes.

## Mechanism

iOS 18 (our deployment target) has a `Text` initializer that's generic over **any** `DiscreteFormatStyle`, not just Apple's:

```swift
init<V, F>(_ source: TimeDataSource<V>, format: F)
    where V == F.FormatInput, F: DiscreteFormatStyle, F.FormatOutput == String
```

So a custom `TripCountdownFormatStyle` can render our exact `8m` / `NOW` typography and still self-update:

- `format(_ now: Date) -> String` — minutes until `departureDate`, `NOW` under a minute
- `discreteInput(after:)` / `discreteInput(before:)` — the instants at which the output flips (minute boundaries anchored on `departureDate`, *not* on wall-clock minutes). This is the actual trick: it's how SwiftUI knows when to schedule the next render.

```swift
Text(.currentDate, format: TripCountdownFormatStyle(departure: arrival.departureDate))
```

## The catch: always-on display redacts it

From [the initializer's docs](https://developer.apple.com/documentation/swiftui/text/init(_:format:)-8sfgg), under *Redaction for Reduced Luminance*:

> If the `format` is known to SwiftUI and allows removing units or fields, SwiftUI removes parts that change more frequently than the frame rate allows. **Otherwise, all digits in the formatted output are redacted using dashes.**

A custom format style is by definition not "known to SwiftUI". So on an always-on display (iPhone 14 Pro and later, Lock Screen dimmed — *the* dominant state for a Live Activity), `8m` would render as `—m`.

**This is a documentation claim, not something we've observed.** Confirming it on a real device is the first task here and it gates the whole design.

## Two approaches

**A. Custom style + luminance fallback.** Branch on `@Environment(\.isLuminanceReduced)`: bright screen → self-ticking custom style; dimmed → a plain static `Text` of the last pushed value (static text isn't redacted). Keeps the exact typography in both states; the number is always correct the moment you pick the phone up, even with zero pushes. Never worse than today, materially better when someone is actually looking. The weakness: the fallback applies exactly when the phone is idle, which is when priority-10 delivery is least reliable.

**B. Adopt a system format style.** `SystemFormatStyle`'s styles are known to SwiftUI, so they degrade gracefully in AOD and tick everywhere with no push at all. But none of them render `8m` / `NOW` — we'd get "in 8 min" or "8:03", a visible redesign of the card's most important element. Worth a spike to see how close any of them actually gets before ruling it out.

## The trap

**Making the big number tick does not make the card tick.** The chips, the deviation label, and the `upcomingArrivals(now:)` filter that drops departed buses are all snapshots taken from the same `let now = Date()`. Fix only the countdown and the card goes internally inconsistent — a ticking headline above frozen chips and a bus that already left. (We hit a transient version of this during #1179; by design it would be permanent.)

## Checklist

- [ ] Put a custom `DiscreteFormatStyle` on a device and check what an always-on display does to it. Everything below depends on the answer.
- [ ] `TripCountdownFormatStyle` in OBAKitCore, with unit tests on the `discreteInput` boundaries — that's the part that fails silently (wrong boundaries → it never ticks, or it re-renders constantly and burns battery).
- [ ] `CountdownView` takes a `Date` rather than `minutes: Int`. It's shared with the Stop page and bookmark rows, so either a parallel init, or those get self-ticking too (a small win in both).
- [ ] Chips, deviation label, and the departed-bus filter, or the card is internally inconsistent.
- [ ] Then the obacloud relaxations: raise `KEEPALIVE_INTERVAL` back up, let `content_changed?` become load-bearing again (its comment already anticipates this: *"the right predicate to lean on again if the widget ever gains the ability to self-tick"*), consider dropping non-urgent corrections back to priority 5, and give `frequentPushesEnabled: off` a real fallback.

Context: #1179, and obacloud PRs #963 / #966 / #968.

Borisserz:
## Summary
- Replaces unbounded Live Activity timer labels with `Text(timerInterval:pauseTime:countsDown:showsHours:)`, clamped at departure.
- Removes the unreachable NOW branch, keeps Dynamic Island labels compact, and gives them localized VoiceOver countdown labels.
- Scope: the countdown is deliberately live while deviation/status text remains content-snapshot data; making the whole card internally time-consistent needs a separate redesign.
- Closes #1187.

## Test plan
- [x] `LiveActivityCountdownTests` — 2/2 on iPhone 17 Pro simulator
- [ ] Start a Live Activity on hardware and confirm all Island layouts fit the bounded timer
- [ ] Verify Lock Screen and always-on-display redaction behavior on hardware

### DIFF
--- a/OBAKit/Trip/TripPage/TripPageViewController.swift
+++ b/OBAKit/Trip/TripPage/TripPageViewController.swift
@@ -9,6 +9,7 @@
 
 import ActivityKit
 import Combine
+import CoreLocation
 import SwiftUI
 import OBAKitCore
 
--- a/OBAKitCore/UI/LiveActivityCountdownView.swift
+++ b/OBAKitCore/UI/LiveActivityCountdownView.swift
@@ -1,0 +1,62 @@
+//
+//  LiveActivityCountdownView.swift
+//  OBAKitCore
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import SwiftUI
+
+/// Live Activity countdown that ticks on its own via `Text(_:style: .timer)`.
+///
+/// Unlike `CountdownView`'s static `"\(minutes)m"` string (which only changes when
+/// a push or local `Activity.update` arrives), the system redraws `.timer` text
+/// as the departure approaches — so the Lock Screen / Dynamic Island stay honest
+/// even when keepalive pushes are sparse (#1187).
+public struct LiveActivityCountdownView: View {
+    public let departureDate: Date
+    public let isRealTime: Bool
+    public let color: Color
+    /// `true` = card-header size, `false` = compact chips / island.
+    public var emphasized: Bool = true
+
+    public init(departureDate: Date, isRealTime: Bool, color: Color, emphasized: Bool = true) {
+        self.departureDate = departureDate
+        self.isRealTime = isRealTime
+        self.color = color
+        self.emphasized = emphasized
+    }
+
+    public var body: some View {
+        HStack(alignment: .top, spacing: 2) {
+            countdownLabel
+                .font(emphasized ? .system(.title2, design: .rounded, weight: .heavy) : .system(.callout, design: .rounded, weight: .heavy))
+                .monospacedDigit()
+                .foregroundStyle(color)
+                .multilineTextAlignment(.trailing)
+            RealtimeGlyph(isRealTime: isRealTime, color: color, size: emphasized ? 11 : 9)
+                .padding(.top, 1)
+        }
+        .accessibilityHidden(true)
+    }
+
+    @ViewBuilder
+    private var countdownLabel: some View {
+        if LiveActivityCountdown.shouldShowNow(departureDate: departureDate) {
+            Text(OBALoc("stop_page.countdown.now", value: "NOW", comment: "Shown in place of the minutes countdown when the vehicle is departing now"))
+        } else {
+            Text(departureDate, style: .timer)
+        }
+    }
+}
+
+/// Pure helpers for Live Activity countdown presentation (#1187).
+public enum LiveActivityCountdown {
+    /// When the departure is at or before `now`, show "NOW" instead of a timer
+    /// that would start counting up.
+    public static func shouldShowNow(departureDate: Date, now: Date = Date()) -> Bool {
+        departureDate.timeIntervalSince(now) <= 0
+    }
+}
--- a/OBAKitCore/UI/LiveActivityCountdownView.swift
+++ b/OBAKitCore/UI/LiveActivityCountdownView.swift
@@ -9,7 +9,7 @@
 
 import SwiftUI
 
-/// Live Activity countdown that ticks on its own via `Text(_:style: .timer)`.
+/// Live Activity countdown that ticks on its own via a bounded timer interval.
 ///
 /// Unlike `CountdownView`'s static `"\(minutes)m"` string (which only changes when
 /// a push or local `Activity.update` arrives), the system redraws `.timer` text
@@ -42,21 +42,24 @@ public struct LiveActivityCountdownView: View {
         .accessibilityHidden(true)
     }
 
-    @ViewBuilder
     private var countdownLabel: some View {
-        if LiveActivityCountdown.shouldShowNow(departureDate: departureDate) {
-            Text(OBALoc("stop_page.countdown.now", value: "NOW", comment: "Shown in place of the minutes countdown when the vehicle is departing now"))
-        } else {
-            Text(departureDate, style: .timer)
-        }
+        Text(
+            timerInterval: LiveActivityCountdown.boundedTimerInterval(departureDate: departureDate),
+            pauseTime: departureDate,
+            countsDown: true,
+            showsHours: false
+        )
     }
 }
 
 /// Pure helpers for Live Activity countdown presentation (#1187).
 public enum LiveActivityCountdown {
-    /// When the departure is at or before `now`, show "NOW" instead of a timer
-    /// that would start counting up.
-    public static func shouldShowNow(departureDate: Date, now: Date = Date()) -> Bool {
-        departureDate.timeIntervalSince(now) <= 0
+    /// Returns a countdown interval that reaches zero at `departureDate` and
+    /// never becomes an elapsed-time counter after that instant.
+    public static func boundedTimerInterval(
+        departureDate: Date,
+        now: Date = Date()
+    ) -> ClosedRange<Date> {
+        min(now, departureDate)...departureDate
     }
 }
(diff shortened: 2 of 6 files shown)


aaronbrethorst:
### Code review

Found 2 issues:

1. Once the departure passes, the countdown keeps running — upward — until the next push arrives, still in the departure's own color. `Text(_:style: .timer)` renders the difference from *now* with no bound, and the only thing that can swap it for "NOW" is a re-evaluation of the view body, which (per this PR's own premise, and #1187: "nothing re-renders the archived snapshot on its own") happens only on a push. That is the same failure #1187 filed, pointed the other way: instead of freezing at `2m`, a bus that already left reads `9:12` and climbing, indistinguishable from a countdown. `Text(timerInterval: now...departureDate, pauseTime: departureDate, countsDown: true, showsHours: false)` clamps at `0:00` and is the bounded form of this API; a `staleDate` on the `ActivityContent` would also cap it.

https://github.com/OneBusAway/onebusaway-ios/blob/1e44e6944832a220a5d502a0b4c3c8d8ef275cb3/OBAKitCore/UI/LiveActivityCountdownView.swift#L45-L52

2. The `shouldShowNow` guard — and all three new tests — cover a state that cannot reach these views. Every call site takes its arrival from `upcomingArrivals(now:)`, which has already kept only `departureDate >= now`. In `departurePill` the guard is handed the *same* captured `now` used for that filter, so it can only be true if `departureDate == now` to the nanosecond against a whole-second departure time — provably dead. The other three sites re-read `Date()`, leaving a window of microseconds. So the branch is false in production, the doc comment's promise ("show `NOW` instead of a timer that would start counting up") does not hold, and the tests would go on passing through any regression in issue 1. This is the trap #1187 names: the departed-bus filter is a snapshot taken from the same `let now = Date()`.

https://github.com/OneBusAway/onebusaway-ios/blob/1e44e6944832a220a5d502a0b4c3c8d8ef275cb3/OBAKitCore/UI/TripLiveActivityCardView.swift#L121-L131

aaronbrethorst (changes requested):
The instinct here is right: a countdown frozen at whatever the last push said is a real problem, and #1187 is worth solving. But I don't think this trades up, and the safeguard that's supposed to prevent the obvious failure mode doesn't run.

**`Text(date, style: .timer)` doesn't stop at zero — it counts up, without bound.** Apple's bounded form is `Text(timerInterval:pauseTime:countsDown:showsHours:)`; the `style: .timer` variant just renders the delta from now. So a bus that has already left reads `9:12` and climbing, and it's visually identical to a countdown — the rider can't tell the difference. In the healthy case the 55s keepalive bounds this, but the unhealthy case (`frequentPushesEnabled` off, throttled) is the entire justification for this PR, and there's no bound there at all.

**The `shouldShowNow` guard that's meant to catch this is unreachable.** Every call site sources its arrival from `upcomingArrivals(now:)`, which is `arrivals.filter { $0.departureDate >= now }`. So by construction `departureDate >= now`, and `shouldShowNow` (`departureDate.timeIntervalSince(now) <= 0`) can only fire on exact nanosecond equality against a whole-second departure time. In `departurePill` it's provably dead — the guard gets the *same* captured `now` the filter used. The other three re-read `Date()`, which leaves a microsecond window. So the doc comment's promise — "show NOW instead of a timer that would start counting up" — doesn't hold anywhere.

Those two compound: the three new tests only exercise `shouldShowNow`, the branch that never runs, so they'd stay green through any regression in the counting-up behavior. Nothing pins the thing the PR actually changes.

There's also a consistency problem that #1187 specifically names as the trap to avoid. After this change the headline and chips tick every second while `deviationLabel` / `statusText` ("3:26 PM - arrives on time") and the departed-bus filter stay frozen at the last push. A card where one number moves and the text next to it contradicts it is arguably worse than one that's uniformly stale.

Two more I'd want addressed before this ships:

- **Dynamic Island width.** `5m`/`62m` becomes `5:00`/`1:02:33`. `minimal` is a tiny circle, `compactTrailing` has `.frame(minWidth: 20)` with no max, and expanded trailing is 32pt bold. That needs `showsHours: false` and a `maxWidth`, verified on a device.
- **VoiceOver in the Island.** `Text(date, style: .timer)` carries no accessibility label there, and `5:00` will most likely be read as a clock time. `Formatters` documents this hazard and offers `formattedTimeUntilArrivalDeparture` for labels. (The lock-screen card is fine — `.accessibilityHidden(true)` matches the old `CountdownView` and `timeStatusLine` supplies the label.)

Smaller: second-resolution countdowns on schedule-only departures (`scheduleStatus == .unknown`) imply precision the feed doesn't have — the old `15m` was honestly coarse. And `TripActivityPresenter.minuteText` becomes production-dead, referenced only by its own tests.

The good news is the scheduled-vs-real-time distinction survives intact — `RealtimeGlyph` on the card and `presenter.color(for:)` everywhere — so nothing is lost there.

Concretely, what I'd want to see: the bounded `Text(timerInterval:pauseTime:countsDown:)` form so it stops at the departure rather than counting up; the stale-text problem addressed or explicitly scoped out; both device-verification boxes in your description actually checked, including the always-on-display redaction check #1187 lists. I flagged this one in the issue as "a visible redesign of the card's most important element — worth a spike," and I still think that's the right framing. Happy to talk through the approach before you sink more time into it.

Borisserz:
@aaronbrethorst Replaced every timer with the bounded API, removed shouldShowNow and its dead-branch tests, and added tests for future and past bounded intervals. Island labels use Formatters through the presenter, use showsHours: false, and have maximum widths. The deviation/status consistency concern is now explicitly scoped in both code and the PR body; I have not checked the hardware/AOD boxes, so they remain unchecked.

aaronbrethorst:
### Code review

aaronbrethorst:
Thanks for opening this PR and working through the fixes. I tried this out on my phone and the live activity 0:00 time makes me uncomfortable. It's a significant departure from the way we've presented this information in the past, and I think changing it in general is not where we want to go—doubly so for only changing it in one place.

PR #1264, opened 2026-08-02
Coalesce Live Activity updates and re-fetch relevance

### LINKED ISSUE #1197: Live Activity updates: ordering and observability of the detached re-fetch path
Follow-up to #1192 (flagged by the silent-failure review). `BookmarksViewController.updateRunningLiveActivities()` now updates each Live Activity from a `Task.detached` that re-fetches the activity by ID (a non-Sendable `Activity` can't cross the main-actor boundary to ActivityKit's @concurrent `update`). Two consequences to improve on:

1. **Ordering**: detached tasks have no ordering guarantee, so two rapid refresh cycles can in principle apply content states out of order (older arrival times winning). Consider coalescing per activity ID (keep only the latest pending state) or serializing updates through a single consumer.
2. **Observability**: if the ID lookup misses (activity ended in between), the update is skipped — this is logged since #1192, but if stale-Lock-Screen reports appear, this path is the first suspect.

Also noted in review: unit-testing this requires wrapping ActivityKit behind a protocol (its static `activities` isn't injectable) — judged disproportionate for the migration PR, but worth doing if this code grows. A manual device smoke test of Live Activity tracking is the current verification story.

Borisserz:
## Summary
- Rebases the coalescer onto upstream/main and preserves the content-refresh `staleDate: nil` rationale.
- Keeps per-ID serial draining, injects the apply operation, and logs only after a successful ActivityKit update.
- Adds a drain-ordering test that holds the first apply, queues two more states, and verifies serial first/latest application.
- Closes #1197.

## Test plan
- [ ] `xcodebuild build-for-testing` is blocked on rebased upstream/main: `TripPageViewController.swift:140` cannot find `CLLocationCoordinate2D` in scope.
- [ ] On hardware: refresh a tracked bookmark rapidly and verify the final state and Dynamic Island ranking.

### DIFF
--- a/OBAKit/Bookmarks/BookmarksViewController.swift
+++ b/OBAKit/Bookmarks/BookmarksViewController.swift
@@ -314,38 +314,15 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
                 if !application.liveActivityTracker.isForwardingPushToken(activityID: activity.id) {
                     trackLiveActivity(activity, arrivalDepartures: arrivalDepartures)
                 }
-                // `Activity` is not Sendable and this loop's instance lives in the
-                // main-actor region, so it can't be sent to ActivityKit's @concurrent
-                // `update`. Re-fetch by ID inside a detached task instead — that copy
-                // never crosses an isolation boundary.
+                // Coalesce per activity ID. The worker re-fetches the activity
+                // before applying content, so its current relevance score is
+                // preserved across the asynchronous hop (#1197).
                 let activityID = activity.id
-                Task.detached {
-                    guard let activity = Activity<TripAttributes>.activities.first(where: { $0.id == activityID }) else {
-                        // The activity ended between the loop and this task; dropping
-                        // the update is correct, but log it so a stale Lock Screen is
-                        // diagnosable.
-                        Logger.info("Live Activity \(activityID) is no longer running; skipping update.")
-                        return
-                    }
-                    // Preserve prominence via the shared helper, reading the score
-                    // off this freshly re-fetched activity — not a snapshot taken
-                    // before the hop, which can predate a concurrent
-                    // promote/demote and would silently write the stale score
-                    // back, undoing it (#1243 review follow-up).
-                    //
-                    // `staleDate` stays nil here on purpose: unlike the
-                    // score-only promote/demote touches (which must carry the
-                    // push-set marker through), this update installs fresh local
-                    // content, and carrying a possibly-past stale-date forward
-                    // could mark it stale on arrival. The next push re-arms it.
-                    await activity.update(
-                        TripLiveActivityRelevance.contentPreservingRelevance(
-                            state: contentState,
-                            staleDate: nil,
-                            existing: activity.content
-                        )
+                Task {
+                    await LiveActivityUpdateCoalescer.shared.schedule(
+                        activityID: activityID,
+                        state: contentState
                     )
-                    Logger.info("Updated Live Activity for stop: \(staticData.stopID) route: \(staticData.routeShortName)")
                 }
             } else if !application.liveActivityTracker.isTracking(activityID: activity.id) {
                 // No bookmark/arrival data to register push updates with, but the activity is
--- a/OBAKit/Trip/TripPage/TripPageViewController.swift
+++ b/OBAKit/Trip/TripPage/TripPageViewController.swift
@@ -8,6 +8,7 @@
 //
 
 import ActivityKit
+import CoreLocation
 import Combine
 import SwiftUI
 import OBAKitCore
--- a/OBAKitCore/LiveActivities/LiveActivityUpdateCoalescer.swift
+++ b/OBAKitCore/LiveActivities/LiveActivityUpdateCoalescer.swift
@@ -1,0 +1,93 @@
+//
+//  LiveActivityUpdateCoalescer.swift
+//  OBAKitCore
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import ActivityKit
+import Foundation
+
+/// Serializes Live Activity content refreshes per activity ID (#1197).
+///
+/// `BookmarksViewController.updateRunningLiveActivities` used to spawn an
+/// unbounded `Task.detached` per activity on every poll. Two rapid cycles could
+/// apply an older `ContentState` after a newer one. This coalescer keeps only
+/// the latest pending state per ID and drains them one at a time.
+///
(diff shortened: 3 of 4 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. Neither new test can fail if the fix is reverted. Both tests exercise only `LiveActivityUpdateMailbox`, which is a one-line wrapper over `Dictionary` — `enqueueKeepsOnlyLatestStatePerActivity` asserts that subscript-assign overwrites, and `differentActivityIDsStayIndependent` asserts that distinct keys are independent. Neither constructs a `LiveActivityUpdateCoalescer`, calls `schedule`, or touches `drain`/`draining`. The actual #1197 fix is the serialization: the `draining` guard plus the `while let take()` drain loop. Delete the `draining` set, collapse `drain` to a single `take` + apply, or revert `BookmarksViewController` to `Task.detached` — all three leave both tests green. The mailbox's own doc comment bills it as existing "so coalescing rules can be unit-tested without ActivityKit (#1197)," and `pendingCount(for:)` is annotated "Test seam" but is not called by any test or by production code. This is the same note left on #1243 and #1271; ActivityKit not being injectable is a fine reason the real test can't be written against this seam, but then the test plan and doc comments shouldn't bill these as covering the ordering fix.

https://github.com/OneBusAway/onebusaway-ios/blob/52ae138b92e58869072a14274cc665bdbdce90a9/OBAKitTests/Models/LiveActivityUpdateMailboxTests.swift#L23-L36

aaronbrethorst (changes requested):
I want to start with the part that's genuinely good, because it's the hard part: **the coalescer is correct.** I traced the reentrancy timeline specifically looking for a dropped final update, and it isn't there. `drain` awaits `Self.apply`, which releases the actor, so a concurrent `schedule` can enqueue mid-flight; its spawned drain task early-returns on the `draining` guard, and the in-flight `while let take()` picks the new state up on resume. The part that makes it airtight is that there's no suspension point between the final `take()` returning nil and `defer { draining.remove(activityID) }` — the executor runs that stretch atomically, so there's no lost-wakeup window, and any `schedule` after removal spawns a fresh drain. That's a subtle thing to get right.

Two problems, one of them my doing.

**This is now conflicted, and that's partly my fault.** I merged #1271 a few minutes ago, which touched the same lines in `updateRunningLiveActivities`. Your branch was cut from `01774b93`, before that landed. So this needs a rebase onto `main` — sorry for the churn.

Worth knowing before you do: **#1271 already shipped the second half of this PR.** Reading relevance off the re-fetched `Activity.content` at apply time is on `main` now. So roughly half of this becomes a no-op on rebase, and what's left — the coalescing — is the genuinely new contribution. Two things to preserve while you rebase:

- Main's call site now carries a detailed comment explaining why `staleDate` stays `nil` on the content-refresh path, as distinct from the score-only promote/demote touches. That point took two review rounds to settle; please don't let the rebase drop it.
- The old code logged *after* a successful `update`; yours logs before the attempt. The new line is richer (activity ID, relevance score), but you lose the "this actually landed" signal — on a PR whose linked issue is half about observability, I'd keep both, or move the richer line after the await.

**And the tests, again.** Both new tests exercise only `LiveActivityUpdateMailbox` — that subscript-assign overwrites, and that distinct keys are independent. That's a test of `Dictionary`. Neither constructs the coalescer or touches `drain`/`draining`, which is the entire mechanism this PR exists for. Delete the `draining` set, collapse the `while let take()` loop to a single take, or revert the call site to `Task.detached`, and both tests stay green. `pendingCount(for:)` is even annotated "Test seam" and is called by nothing.

This is the third PR in this sweep with the same shape, so let me be concrete about the way out rather than just flagging it. The reason the tests degenerate into testing a dictionary is that `apply` is `nonisolated static` and hardcodes `Activity<TripAttributes>.activities`, so there's no seam. #1197 anticipated exactly this. Give the coalescer an injectable apply step — a stored `(String, ContentState) async -> Void` closure defaulting to the ActivityKit call — and the drain-ordering behavior becomes directly testable: schedule three states for one ID while a slow apply is in flight, assert the applies land serially and the last state wins. That test would fail against every one of the reverts above, and it's the test that would actually protect this work.

Two smaller notes:

- The doc comment says the apply step is `static` "so the actor isn't held across the ActivityKit `update` await." Swift actors are reentrant and release at every await regardless — an actor-isolated `apply` would behave identically. The `nonisolated static` form is fine; the stated reason isn't.
- The class doc says "rapid polls can't apply out of order," but the call site still wraps `schedule` in an unstructured `Task {}`, so *enqueue* order isn't language-guaranteed — only the drain is serialized. Near-theoretical on the main actor, and not a regression, but the guarantee as written isn't quite there.

Rebase, keep the coalescer, and give it a test that bites — then I'll take it.

Borisserz:
@aaronbrethorst Rebased onto upstream/main, retained the staleDate:nil explanation in the coalescer apply path, and moved the rich log after ActivityKit update. The coalescer now takes an injectable apply closure; its new test blocks the first apply, schedules two newer states, and asserts serial [first, latest] application. I could not execute it because the rebased target build stops in upstream TripPageViewController.swift:140 ( missing), unrelated to this PR; the PR test plan records that blocker.

Borisserz:
Correction to the build blocker: it is `TripPageViewController.swift:140`, where `CLLocationCoordinate2D` is not in scope after the rebase. This is outside the coalescer diff; I have left the test-plan checkbox unchecked.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is the right shape for #1197. Replacing the unbounded per-poll `Task.detached` with an actor that keeps only the latest state per activity ID, and re-fetching the activity inside the worker so the current relevance score survives the hop, addresses the ordering problem properly rather than papering over it.

I worked through the drain logic looking for a lost update and didn't find one: `schedule` enqueues and spawns the drain as a single uninterrupted actor job, and there's no suspension point between `mailbox.take()` returning nil and the `defer` clearing `draining` — so there's no window where a pending state is left with no drain in flight. An enqueue arriving while `applyState` is suspended gets picked up when the loop re-enters. The new test also fails deterministically if the `draining` set is removed, which is what was missing last time.

1. The test has no `.timeLimit` trait. Collapsing `drain` to a single `take` + apply would leave `waitForApplyCount(2)` waiting forever, so that regression burns a CI job to timeout instead of reporting a failure. Worth adding.
2. `LiveActivityUpdateCoalescer.shared` is a global singleton, while its sibling `liveActivityTracker` hangs off `Application`. The DI pattern would put it there too.
3. The file is `LiveActivityUpdateMailboxTests.swift` but the suite inside is `LiveActivityUpdateCoalescerTests`.

Merging — happy to take those as a follow-up.

PR #1265, opened 2026-08-02
Polish EditBookmarkViewModel follow-ups from #1138

### LINKED ISSUE #1145: Follow-ups from PR #1138 (Bookmarks ViewModel Layer)
Minor, non-blocking polish items surfaced while reviewing #1138. None affect correctness.

### Cleanup
- [ ] **Remove the dead `isFavorite` parameter from `EditBookmarkViewModel.prepareToSave(name:isFavorite:)`** (`OBAKit/ViewModels/EditBookmarkViewModel.swift:118`). It is never read in that method — the favorite flag is only applied in `persist(...)`, as the method's own doc comment notes. Drop it from the signature (callers already pass `isFavorite` to `persist` separately).

### Type design refinement
- [ ] **Eliminate the `isNewBookmark` two-sources-of-truth.** `prepareToSave` decides newness and carries it in `SaveOutcome.readyToSave(_, isNew:)`, then `EditBookmarkViewController.save()` threads that boolean back into `persist(..., isNewBookmark:)` and independently hard-codes `true` on the duplicate-confirmation path. Nothing structurally guarantees these agree. Consider deriving newness inside `persist` from the outcome/mode, or splitting `.readyToSave` into `.readyToSaveNew`/`.readyToSaveExisting` so an inconsistent (bookmark, isNew) pair is unrepresentable.

### Test coverage
- [ ] **Add a test for the duplicate → cancel → no-analytics path.** This is #1138's headline behavior change (analytics now fires *after* duplicate confirmation), but the guarantee lives in `EditBookmarkViewController.save()`, which is untested. The VM tests cover that `persist` emits and that `prepareToSave` returns `.duplicateRequiresConfirmation` without persisting, but nothing asserts cancelling the duplicate alert suppresses the `addBookmark` event.
- [ ] **AgencyAlertsStore concurrency test only bites under Thread Sanitizer.** `AgencyAlertsStoreConcurrencyTests` can pass even with the lock removed unless run under TSan, and it doesn't concurrently exercise `deleteAgencyAlerts()`'s `removeAll()` (the mutation most likely to corrupt state during a concurrent read). Confirm CI runs OBAKitTests with `-enableThreadSanitizer YES`, and consider adding a delete-vs-read fanout.

### Docs / consistency
- [ ] **Reword the `preloadedArrivals` empty-array comment** (`OBAKit/Bookmarks/AddBookmarkViewController.swift:110`). It says an empty array "surfaces the 'no upcoming departures' view," but there is no dedicated empty-state view — an empty array renders the inline 'route bookmarks unavailable' explanation in the trip section while still showing the whole-stop bookmark row.
- [ ] **Log parity in `ManageBookmarksViewController.restoreEmptyBookmarkNames()`** (`ManageBookmarksViewController.swift:188`). `saveNameChange` now `Logger.warn`s when a bookmark ID from the form isn't found in the store; the parallel not-found case in `restoreEmptyBookmarkNames` silently `continue`s. Make them consistent.

Borisserz:
## Summary
- Removes the unused `isFavorite` argument from `prepareToSave`.
- Splits `SaveOutcome` into `readyToSaveNew` / `readyToSaveExisting` so newness cannot disagree with the bookmark.
- Drops the vacuous duplicate-cancellation ViewModel test and strengthens duplicate-confirmation coverage with a two-bookmark assertion.
- Corrects the `prepareToSave` documentation.
- Refs #1145.

## Test plan
- [ ] `EditBookmarkViewModelTests` — blocked before tests by the current upstream/main OBAKit TripPage build failure under this Xcode.
- [ ] Controller-level duplicate-cancellation coverage remains follow-up work for #1145.

### DIFF
--- a/OBAKitTests/ViewModels/EditBookmarkViewModelTests.swift
+++ b/OBAKitTests/ViewModels/EditBookmarkViewModelTests.swift
@@ -240,7 +240,7 @@ final class EditBookmarkViewModelTests: OBATestCase {
         let app = createApplicationWithoutRegion(dataLoader: dataLoader)
 
         let vm = EditBookmarkViewModel(application: app, source: .stop(stop), bookmark: nil)
-        let outcome = vm.prepareToSave(name: "My Stop", isFavorite: true)
+        let outcome = vm.prepareToSave(name: "My Stop")
 
         guard case .regionUnavailable = outcome else {
             Issue.record("Expected .regionUnavailable, got \(outcome)")
@@ -255,14 +255,13 @@ final class EditBookmarkViewModelTests: OBATestCase {
         let app = createApplication(dataLoader: dataLoader)
 
         let vm = EditBookmarkViewModel(application: app, source: .stop(stop), bookmark: nil)
-        let outcome = vm.prepareToSave(name: "My Stop", isFavorite: true)
+        let outcome = vm.prepareToSave(name: "My Stop")
 
-        guard case .readyToSave(let bookmark, let isNew) = outcome else {
-            Issue.record("Expected .readyToSave, got \(outcome)")
+        guard case .readyToSaveNew(let bookmark) = outcome else {
+            Issue.record("Expected .readyToSaveNew, got \(outcome)")
             return
         }
         #expect(bookmark.name == "My Stop")
-        #expect(isNew)
     }
 
     @Test @MainActor
@@ -272,10 +271,10 @@ final class EditBookmarkViewModelTests: OBATestCase {
         let app = createApplication(dataLoader: dataLoader)
 
         let vm = EditBookmarkViewModel(application: app, source: .stop(stop), bookmark: nil)
-        let outcome = vm.prepareToSave(name: "   ", isFavorite: true)
+        let outcome = vm.prepareToSave(name: "   ")
 
-        guard case .readyToSave(let bookmark, _) = outcome else {
-            Issue.record("Expected .readyToSave"); return
+        guard case .readyToSaveNew(let bookmark) = outcome else {
+            Issue.record("Expected .readyToSaveNew"); return
         }
         #expect(bookmark.name == Formatters.formattedTitle(stop: stop))
     }
@@ -290,7 +289,7 @@ final class EditBookmarkViewModelTests: OBATestCase {
         app.userDataStore.add(existing, to: nil)
 
         let vm = EditBookmarkViewModel(application: app, source: .stop(stop), bookmark: nil)
-        let outcome = vm.prepareToSave(name: "Stop", isFavorite: true)
+        let outcome = vm.prepareToSave(name: "Stop")
 
         guard case .duplicateRequiresConfirmation(let dup) = outcome else {
             Issue.record("Expected .duplicateRequiresConfirmation, got \(outcome)")
@@ -312,17 +311,16 @@ final class EditBookmarkViewModelTests: OBATestCase {
         app.userDataStore.add(bookmark, to: nil)
 
         let vm = EditBookmarkViewModel(application: app, source: .stop(stop), bookmark: bookmark)
-        let outcome = vm.prepareToSave(name: "Updated Name", isFavorite: false)
+        let outcome = vm.prepareToSave(name: "Updated Name")
 
-        guard case .readyToSave(let saved, let isNew) = outcome else {
-            Issue.record("Expected .readyToSave, got \(outcome)")
+        guard case .readyToSaveExisting(let saved) = outcome else {
+            Issue.record("Expected .readyToSaveExisting, got \(outcome)")
             return
         }
         #expect(saved.name == "Original")
         #expect(saved.isFavorite)
-        #expect(!isNew)
 
-        vm.persist(saved, name: "Updated Name", isFavorite: false, to: nil, isNewBookmark: isNew)
+        vm.persistExisting(saved, name: "Updated Name", isFavorite: false, to: nil)
         #expect(saved.name == "Updated Name")
         #expect(!saved.isFavorite)
     }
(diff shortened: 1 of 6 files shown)


aaronbrethorst:
### Code review

Found 2 issues:

1. The new "Cancelling duplicate confirmation" test asserts a no-op and cannot fail. `resolveDuplicate(.cancelled, …)` is literally `case .cancelled: return`, so "no bookmark added, no `addBookmark` event" holds by construction — the test passes regardless of what the app does. #1145 asked for coverage of the guarantee that lives in `EditBookmarkViewController.save()` ("nothing asserts cancelling the duplicate alert suppresses the `addBookmark` event"), and that wiring is still untested: change the Cancel action's handler to `.createDuplicate(bookmark)` and every test here still passes. Cancel behavior is also unchanged from `main` (`handler: nil` → a call whose body is `return`), so nothing new is actually guarded.

https://github.com/OneBusAway/onebusaway-ios/blob/f033664bfccd2860e7b9fb14f2fe1a175e8f3746/OBAKitTests/ViewModels/EditBookmarkViewModelTests.swift#L520-L526

2. "Closes #1145" will auto-close the issue while one of its checkboxes is untouched. The AgencyAlertsStore concurrency item — "confirm CI runs OBAKitTests with `-enableThreadSanitizer YES`, and consider adding a delete-vs-read fanout" — is not addressed by this PR and is still absent on `main`: no `enableThreadSanitizer` anywhere under `.github/` or `Apps/`, and `AgencyAlertsStoreConcurrencyTests` still has a single test that never calls `deleteAgencyAlerts()`. Combined with item 1, "Refs #1145" would be more accurate than "Closes".

https://github.com/OneBusAway/onebusaway-ios/blob/f033664bfccd2860e7b9fb14f2fe1a175e8f3746/OBAKitTests/Networking/AgencyAlertsStoreConcurrencyTests.swift#L99-L102

OBAKitTests/ViewModels/EditBookmarkViewModelTests.swift
```diff
@@ -489,15 +486,65 @@ final class EditBookmarkViewModelTests: OBATestCase {
...
+        vm.persistNew(bookmark, name: "Stop", isFavorite: true, to: nil)
 
         let addBookmarkEvents = analyticsMock.reportedEvents.filter { $0.label == AnalyticsLabels.addBookmark }
         #expect(addBookmarkEvents.isEmpty)
     }
+
+    // MARK: - Duplicate confirmation (#1145 / #1138)
+
+    @Test @MainActor
+    func `Cancelling duplicate confirmation does not persist or report analytics`() throws {
```
aaronbrethorst: This test cannot fail. `resolveDuplicate(.cancelled, …)` hits a branch whose entire body is `return`, so there's no behavior to assert on — and Cancel is byte-identical to `main`, where the alert's `handler: nil` was already a no-op.

The guarantee #1145 asked for lives one layer up, in `EditBookmarkViewController.save()`: that tapping Cancel on the duplicate alert suppresses the `addBookmark` analytics event. Nothing here or on `main` asserts that — rewire the Cancel handler to `.createDuplicate(bookmark)` and the suite stays green.

Either drive it at the controller level with an analytics spy, or drop this test and leave the checklist item open.

aaronbrethorst (changes requested):
The three code-level follow-ups here are done properly. Dropping the dead `isFavorite` parameter, splitting `SaveOutcome` into `.readyToSaveNew`/`.readyToSaveExisting` so an inconsistent `(bookmark, isNew)` pair can't be built, the `Logger.warn` parity in `restoreEmptyBookmarkNames` — all correct. I checked the `AddBookmarkViewController` comment reword against the actual `else` branch and it now describes what really renders (the inline "route bookmarks unavailable" row alongside the whole-stop row), which is what #1145 asked for. The new log line matches `saveNameChange`'s wording.

Two things to fix, both cheap.

**The new cancel test can't fail.** `Cancelling duplicate confirmation does not persist or report analytics` calls `vm.resolveDuplicate(.cancelled, …)`, and that branch's entire body is `return`. #1145 was specific that the guarantee lives in `EditBookmarkViewController.save()` — nothing asserts that tapping Cancel on the alert suppresses the `addBookmark` event. Rewire the alert's Cancel handler to `.createDuplicate(bookmark)` and the whole suite stays green. Cancel behavior is also byte-identical to `main` (the old `handler: nil` was already a no-op), so no new guarantee has been established.

I'd rather you either write the controller-level test — a spy on the analytics client, drive `save()`, tap Cancel, assert zero `addBookmark` events — or drop this test and leave the checklist item open. What I don't want is a green test standing in for a guarantee that isn't there. This is the same note I left on #1243 and #1271, so I'd like to get ahead of it: a test that passes against unmodified `main` isn't coverage.

**"Closes #1145" is premature.** The Thread Sanitizer item is untouched — I grepped, and there's no `enableThreadSanitizer` anywhere under `.github/` or `Apps/`, so CI still doesn't run OBAKitTests under TSan, and the delete-vs-read fanout isn't there. Merging this would auto-close the issue with that outstanding. Change it to "Refs #1145" (or tick only the boxes you actually did), and the issue stays open for the concurrency work.

Non-blocking observations:

- `persistNew`/`persistExisting` narrow the hazard but don't remove it — the controller still picks between two free functions, and calling `persistExisting` on a `.readyToSaveNew` bookmark compiles and silently drops analytics. #1145's other suggestion (derive newness inside `persist` from the outcome) would have made it unrepresentable. The doc comment "Prefer the typed helpers so callers can't pass a mismatched new/existing flag" claims more than the API enforces.
- The Cancel call site passes `name`, `isFavorite`, and `groupID` into a branch that ignores all three. A `.createDuplicate(Bookmark, …)` payload carrying them only where they're used would read more honestly.
- `Confirming duplicate persists…` asserts the bookmark is findable but never that a *second* bookmark now exists (`bookmarks.count == 2`) — which is the actual create-duplicate semantic.
- `prepareToSave`'s doc still says form values "(`name`, `isFavorite`) are applied inside `persist`", vestigial now that `isFavorite` left the signature.

Fix the closing keyword and the test story and this is a quick merge.

Borisserz:
Addressed the review items: dropped the ViewModel cancellation no-op test, strengthened duplicate confirmation to assert two bookmarks, corrected the prepareToSave documentation, and changed the issue reference to Refs #1145. Merged current upstream/main and force-pushed a rewritten commit authored by boris without co-author trailers.\n\nVerification: git diff --check passed. The requested EditBookmarkViewModelTests build did not reach tests because current upstream/main fails compiling OBAKit TripPage sources under this Xcode with the unknown RegionIsolation warning group (exit 65).

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Nice cleanup. Splitting `readyToSave(Bookmark, isNew:)` into `.readyToSaveNew` / `.readyToSaveExisting` and making `persist` private behind the two typed helpers does what #1145 was after — the pair can't disagree anymore.

I traced all four save branches against main to be sure this is behavior-preserving: persist-then-delegate ordering is unchanged, the `isNew` values match, Cancel is still a true no-op, and dropping `isFavorite` from `prepareToSave` is safe because only `name` ever fed bookmark construction and duplicate detection. The new duplicate test asserts `bookmarks.count == 2` and exactly one analytics event, so it can actually fail.

Both points from my last review are handled — the vacuous cancel test is gone and the issue reference is correctly downgraded to "Refs".

Merging.

PR #1266, opened 2026-08-02
Fix bookmark callout dismissed on map pan

### LINKED ISSUE #782: Bus stop information disappears on bookmarked bus stop
In the map view, when a bookmarked bus stop is close to the top of the screen to where the bubble showing route information would go offscreen, tapping the bus stop will move the map and then open the bubble as expected, but then immediately close the bubble. Tapping a second time opens the bubble as expected (it stays open).

I have reproduced the issue in several cases, it must be a bookmarked bus stop close to the top of the screen. Non-bookmarked bus stops function correctly - the map moves and the bubble opens and stays open.

1. Open map view with favorited stop near top of screen.
IMG_1262

2. Tap on favorited stop. Map shifts, bubble opens for a second then closes.
IMG_1264

3. Tap on favorited stop again. Bubble opens and stays open.
IMG_1263

Borisserz:
## Summary
- Preserve selected bookmark annotations when `displayUniqueStopAnnotations` removes a sibling `Stop` during stop reloads
- Skip `prepareForReuse`/rebind for selected annotations in `refreshAnnotationViews` so open callouts stay open
- Fold `Bookmark.stopID` into `selectedStopIDs` so `removeStopAnnotationsPreservingSelection` matches its existing "(bookmarks get the same exemption)" comment
- Bookmark label refresh now runs **before** the zoom and stops-layer guards (same arrangement as #1267). Bookmarks survive both of those returns; leaving the loop below them froze expanded labels when zoomed out or with the stops layer off.

Closes #782

## Test plan
- [x] `Display unique stop annotations preserves selected bookmark` — hosts the map in a window, selects the bookmark, then asserts `mapView.view(for:)` is the map-vended spy subclass (`prepareForReuseCount == 0`, same instance after reload). Calling `viewFor` directly is not used; that handed the test a view the map never owned.
- [x] `Remove stop annotations preserving selection keeps selected bookmark` — adds both the bookmarked `Stop` and a sibling `Stop`, then disables the stops layer. This fails on `main` because `selectedStopIDs` did not include bookmark stop IDs.
- [x] Manual: bookmark a stop, pan so the pin sits near the top of the screen, tap once — map shifts and the callout stays open. Reproduced on iPhone 17 Pro simulator (this branch). Non-bookmarked stops were already fine.

### DIFF
--- a/OBAKitTests/Application/Mapping/MapRegionManagerTests.swift
+++ b/OBAKitTests/Application/Mapping/MapRegionManagerTests.swift
@@ -527,13 +527,25 @@ final class MapRegionManagerTests: OBATestCase {
             PrepareForReuseCountingStopAnnotationView.self,
             forAnnotationViewWithReuseIdentifier: MKMapView.reuseIdentifier(for: StopAnnotationView.self)
         )
+
+        // Host the map so MapKit vends annotation views. Calling `viewFor`
+        // directly hands the dequeued view to the test, not the map, so
+        // `mapView.view(for:)` would observe a different instance (or nil).
+        let window = UIWindow(frame: CGRect(x: 0, y: 0, width: 390, height: 844))
+        mgr.mapView.frame = window.bounds
+        window.addSubview(mgr.mapView)
+        window.makeKeyAndVisible()
+
         let stop = try #require(Fixtures.loadSomeStops().first)
         let bookmark = Bookmark(name: "Home", regionIdentifier: pugetSoundRegionIdentifier, stop: stop)
 
         mgr.mapView.addAnnotations([stop, bookmark])
+        mgr.mapView.setCenter(stop.coordinate, animated: false)
+        mgr.mapView.layoutIfNeeded()
         mgr.mapView.selectAnnotation(bookmark, animated: false)
+
         let bookmarkView = try #require(
-            mgr.mapView(mgr.mapView, viewFor: bookmark) as? PrepareForReuseCountingStopAnnotationView
+            mgr.mapView.view(for: bookmark) as? PrepareForReuseCountingStopAnnotationView
         )
 
         // Simulate a stale sibling Stop lingering after a stops reload.
@@ -543,6 +555,7 @@ final class MapRegionManagerTests: OBATestCase {
         #expect(mgr.mapView.annotations.contains { $0 === bookmark })
         #expect(!mgr.mapView.annotations.contains { ($0 as? Stop)?.id == stop.id })
         #expect(bookmarkView.prepareForReuseCount == 0)
+        #expect(mgr.mapView.view(for: bookmark) === bookmarkView)
     }
 
     /// Disabling the stops layer removes browse stops but must not deselect a
--- a/OBAKitTests/Application/Mapping/MapRegionManagerTests.swift
+++ b/OBAKitTests/Application/Mapping/MapRegionManagerTests.swift
@@ -497,4 +497,58 @@ final class MapRegionManagerTests: OBATestCase {
         // Zoomed out past it → hide.
         #expect(MapRegionManager.shouldShowExtraStopData(forVisibleMapRectHeight: 7_001) == false)
     }
+
+    // MARK: - Bookmark selection preservation
+
+    private func makeBookmarkManager() -> MapRegionManager {
+        let queue = OperationQueue()
+        let dataLoader = MockDataLoader(testName: name)
+        let application = buildApplication(queue: queue, dataLoader: dataLoader)
+        return MapRegionManager(application: application)
+    }
+
+    /// When stop annotations reload, a sibling `Stop` for a bookmarked stopID is
+    /// removed but the selected `Bookmark` must stay selected — rebinding its view
+    /// dismisses the callout (#782).
+    @Test @MainActor
+    func `Display unique stop annotations preserves selected bookmark`() throws {
+        let mgr = makeBookmarkManager()
+        let stop = try #require(Fixtures.loadSomeStops().first)
+        let bookmark = Bookmark(name: "Home", regionIdentifier: pugetSoundRegionIdentifier, stop: stop)
+
+        mgr.mapView.addAnnotations([stop, bookmark])
+        mgr.bookmarks = [bookmark]
+        mgr.mapView.selectAnnotation(bookmark, animated: false)
+        _ = mgr.mapView(mgr.mapView, viewFor: bookmark)
+
+        // Simulate a stale sibling Stop lingering after a stops reload.
+        mgr.mapView.addAnnotation(stop)
+        mgr.bookmarks = [bookmark]
+
+        #expect(mgr.mapView.selectedAnnotations.contains { $0 === bookmark })
+        #expect(mgr.mapView.annotations.contains { $0 === bookmark })
+        #expect(!mgr.mapView.annotations.contains { ($0 as? Stop)?.id == stop.id })
+    }
+
+    /// Disabling the stops layer removes browse stops but must not deselect a
+    /// bookmark the rider is actively previewing.
+    @Test @MainActor
+    func `Remove stop annotations preserving selection keeps selected bookmark`() throws {
+        let mgr = makeBookmarkManager()
+        mgr.registerMapLayer(StopsMapLayer(manager: mgr))
+
+        let stops = try Fixtures.loadSomeStops()
+        let stop = try #require(stops.first)
+        let otherStop = try #require(stops.dropFirst().first)
+        let bookmark = Bookmark(name: "Work", regionIdentifier: pugetSoundRegionIdentifier, stop: stop)
+
+        mgr.bookmarks = [bookmark]
+        mgr.mapView.addAnnotation(otherStop)
+        mgr.mapView.selectAnnotation(bookmark, animated: false)
+
+        mgr.setMapLayerEnabled(false, id: StopsMapLayer.layerID)
+
+        #expect(mgr.mapView.selectedAnnotations.contains { $0 === bookmark })
+        #expect(mgr.mapView.annotations.contains { $0 === bookmark })
+    }
 }
--- a/OBAKitTests/Application/Mapping/MapRegionManagerTests.swift
+++ b/OBAKitTests/Application/Mapping/MapRegionManagerTests.swift
@@ -20,6 +20,16 @@ import Testing
 final class MapRegionManagerTests: OBATestCase {
     var queue: OperationQueue!
 
+    @MainActor
+    private final class PrepareForReuseCountingStopAnnotationView: StopAnnotationView {
+        var prepareForReuseCount = 0
+
+        override func prepareForReuse() {
+            prepareForReuseCount += 1
+            super.prepareForReuse()
+        }
+    }
+
     override init() async throws {
         try await super.init()
 
@@ -513,21 +523,26 @@ final class MapRegionManagerTests: OBATestCase {
     @Test @MainActor
     func `Display unique stop annotations preserves selected bookmark`() throws {
         let mgr = makeBookmarkManager()
+        mgr.mapView.register(
+            PrepareForReuseCountingStopAnnotationView.self,
+            forAnnotationViewWithReuseIdentifier: MKMapView.reuseIdentifier(for: StopAnnotationView.self)
+        )
         let stop = try #require(Fixtures.loadSomeStops().first)
         let bookmark = Bookmark(name: "Home", regionIdentifier: pugetSoundRegionIdentifier, stop: stop)
 
         mgr.mapView.addAnnotations([stop, bookmark])
-        mgr.bookmarks = [bookmark]
         mgr.mapView.selectAnnotation(bookmark, animated: false)
-        _ = mgr.mapView(mgr.mapView, viewFor: bookmark)
+        let bookmarkView = try #require(
+            mgr.mapView(mgr.mapView, viewFor: bookmark) as? PrepareForReuseCountingStopAnnotationView
+        )
 
         // Simulate a stale sibling Stop lingering after a stops reload.
-        mgr.mapView.addAnnotation(stop)
         mgr.bookmarks = [bookmark]
 
         #expect(mgr.mapView.selectedAnnotations.contains { $0 === bookmark })
         #expect(mgr.mapView.annotations.contains { $0 === bookmark })
         #expect(!mgr.mapView.annotations.contains { ($0 as? Stop)?.id == stop.id })
+        #expect(bookmarkView.prepareForReuseCount == 0)
     }
 
     /// Disabling the stops layer removes browse stops but must not deselect a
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. Both new tests pass unchanged against `main` — they can't detect the regression they're named for. Every assertion is about `mapView.selectedAnnotations` / `mapView.annotations` membership, but none of the three changed code paths can affect either collection: `affectedStopIDs.subtract(selectedBookmarkStopIDs)` only narrows the argument to `refreshAnnotationViews`, and the new `guard !mapView.selectedAnnotations.contains(...)` only skips `prepareForReuse()`/rebind — neither adds, removes, selects, nor deselects an annotation. In `Display unique stop annotations preserves selected bookmark`, the sibling `Stop` is removed by the unchanged `stopAnnotationsToRemove` path on `main` too, and the `Bookmark` stays selected on `main` because nothing in `refreshAnnotationViews` deselects. In `Remove stop annotations preserving selection keeps selected bookmark`, `Bookmark` is an `NSObject`, not a `Stop`, so `mapView.annotations.compactMap { $0 as? Stop }` never yielded it even before this change; and since `stop` itself is never added as an annotation (only `otherStop` is), the one line this test nominally covers — `selectedStopIDs` now including bookmark stop IDs — is never exercised either. The symptom in #782 is the callout closing, which happens via `prepareForReuse()` + `view.delegate = self` on a live `StopAnnotationView`; `mapView.view(for:)` returns nil in a headless test, so that loop body no-ops in both versions. A test that fails before the fix has to observe the annotation view (or callout visibility), not selection membership. With the manual repro checkbox still unchecked in the description, there's currently no evidence the change fixes the reported behavior.

https://github.com/OneBusAway/onebusaway-ios/blob/1ebf8bc0631ba9d47421a1f4bfce1736736c459c/OBAKitTests/Application/Mapping/MapRegionManagerTests.swift#L418-L462

OBAKitTests/Application/Mapping/MapRegionManagerTests.swift
```diff
@@ -406,4 +406,58 @@ final class MapRegionManagerTests: OBATestCase {
...
+        _ = mgr.mapView(mgr.mapView, viewFor: bookmark)
+
+        // Simulate a stale sibling Stop lingering after a stops reload.
+        mgr.mapView.addAnnotation(stop)
+        mgr.bookmarks = [bookmark]
+
+        #expect(mgr.mapView.selectedAnnotations.contains { $0 === bookmark })
+        #expect(mgr.mapView.annotations.contains { $0 === bookmark })
+        #expect(!mgr.mapView.annotations.contains { ($0 as? Stop)?.id == stop.id })
+    }
```
aaronbrethorst, line 559: These three assertions all hold on `main` too, so this test can't fail if the fix is reverted.

The sibling `Stop` is removed by the pre-existing `stopAnnotationsToRemove` path regardless of this PR, and nothing on `main` deselects the bookmark. Headless, `mapView.view(for:)` returns nil, so `refreshAnnotationViews`' loop body no-ops in both versions.

To bite, assert on something the loop actually mutates — e.g. that the selected bookmark's view did *not* get `prepareForReuse()` called on it (a spy subclass counting the call), or that its `isHidingExtraStopAnnotationData` survived the reload.

OBAKitTests/Application/Mapping/MapRegionManagerTests.swift
```diff
@@ -406,4 +406,58 @@ final class MapRegionManagerTests: OBATestCase {
...
+        let bookmark = Bookmark(name: "Work", regionIdentifier: pugetSoundRegionIdentifier, stop: stop)
+
+        mgr.bookmarks = [bookmark]
+        mgr.mapView.addAnnotation(otherStop)
+        mgr.mapView.selectAnnotation(bookmark, animated: false)
+
+        mgr.setMapLayerEnabled(false, id: StopsMapLayer.layerID)
+
+        #expect(mgr.mapView.selectedAnnotations.contains { $0 === bookmark })
+        #expect(mgr.mapView.annotations.contains { $0 === bookmark })
```
aaronbrethorst, line 581: This one is vacuous for a more basic reason: `removeStopAnnotationsPreservingSelection` filters `mapView.annotations.compactMap { $0 as? Stop }`, and `Bookmark` is `@objc(OBABookmark) public final class Bookmark: NSObject` — not a `Stop`. The cast never matched a `Bookmark`, before or after this PR, so a bookmark could never have been removed here.

Also, `stop` is never added as an annotation (only `otherStop` is), so the line this test is meant to cover — `selectedStopIDs` now collecting `Bookmark.stopID` — is never exercised. Adding `stop` as an annotation and asserting it's preserved when a bookmark for it is selected would at least reach the changed line.

aaronbrethorst (changes requested):
The production change is written the way I'd want it written. I went looking specifically for a stuck-callout or leaked-selection bug and didn't find one: there's no persistent flag anywhere — you read `mapView.selectedAnnotations` live at each call site, so there's no state to reset on screen exit. Legitimate deselection still works, and the ordering is right: `displayUniqueStopAnnotations` deselects annotations it's about to remove *before* `selectedBookmarkStopIDs` is computed, so a stale bookmark gets deselected first and its stop ID is still correctly refreshed. Tapping elsewhere and selecting another pin go through MapKit untouched. That's careful work.

What I can't do is merge it, because nothing here shows it fixes anything.

**Both new tests pass unchanged against `main`.** That's not a suspicion — it's mechanical:

- `Remove stop annotations preserving selection keeps selected bookmark` asserts a `Bookmark` survives `removeStopAnnotationsPreservingSelection`. But that method filters `mapView.annotations.compactMap { $0 as? Stop }`, and `Bookmark` is `@objc(OBABookmark) public final class Bookmark: NSObject` — it is not a `Stop` and never was. The cast could never have matched it, on this branch or on `main`. On top of that, `stop` is never added as an annotation (only `otherStop` is), so the one line this test nominally covers — `selectedStopIDs` now including bookmark stop IDs — is never reached.
- `Display unique stop annotations preserves selected bookmark` asserts the sibling `Stop` is gone and the `Bookmark` is still selected. The `Stop` is removed by the pre-existing `stopAnnotationsToRemove` path on `main` too, and nothing on `main` deselects the bookmark. Headless, `mapView.view(for:)` returns nil, so `refreshAnnotationViews`' loop body no-ops in both versions anyway.

I traced what the three changed paths can actually affect: they change the *argument* to `refreshAnnotationViews` and whether its loop body runs. Nothing inside that loop — `prepareForReuse`, `view.annotation =`, `view.delegate =`, the delegate customize hook — mutates `mapView.annotations` or `mapView.selectedAnnotations`, which is everything these tests assert on. So there is no version of these assertions that could distinguish the fix from its absence.

The repro checkbox in the description is also unchecked, so right now there's no evidence of any kind — automated or manual — that #782 is fixed.

**And I have a specific reason to doubt it is.** On a pan, `setStops` → `displayUniqueStopAnnotations` only puts anything in `affectedStopIDs` when a `Stop` annotation exists for a bookmarked stop ID, or a `Bookmark` annotation is stale. But `stopsToAdd` already filters out `!bookmarksHash.keys.contains($0.id)`, so a sibling `Stop` for a bookmarked ID is never added in the first place — which suggests `affectedStopIDs` is empty on a routine pan and `refreshAnnotationViews` is already a no-op there. If that's right, the callout dismissal on pan is coming from somewhere else entirely and this change won't move it. I can't run the app to rule out a path I've missed, which is exactly why I want the manual repro.

So: please run the repro on a device or simulator and confirm the callout survives a pan, and check the box. If it does fix it, I'd like at least one test that fails without the change — asserting on `isHidingExtraStopAnnotationData` or on a spy counting `prepareForReuse` calls would bite where these don't. If it turns out not to fix it, that's genuinely useful to know and worth saying in the thread rather than quietly reworking.

Smaller things while you're in here, none blocking:

- The bookmark label loop was added *after* the zoom and stops-layer early returns, so bookmark pins never get `isHidingExtraStopAnnotationData` refreshed with the stops layer off — where bookmark pins are the only pins on screen. (#1267 has the same issue; whichever lands second can pick it up.)
- `@Test @MainActor` is redundant — `OBATestCase` is already `@MainActor` per CLAUDE.md, and no other test in this file annotates.
- `makeBookmarkManager()` creates a local `OperationQueue()` that shadows the suite's `queue` property.
- `affectedStopIDs.subtract(...)` and the `refreshAnnotationViews` guard are two mechanisms for the same outcome in the bookmark case. Not wrong, just worth knowing you only need one for bookmarks (the guard additionally covers selected `Stop`s).

Heads up that this branches from `01774b93` and `main` is well past it now — the test file has moved ~91 lines, so expect a rebase.

Borisserz:
Addressed the review with regression coverage for selected bookmark views not being rebound and for preserving the selected stop annotation when the stops layer is disabled.

aaronbrethorst:
### Code review

Found 2 issues:

1. `Display unique stop annotations preserves selected bookmark` still passes unchanged against `main`, so neither of the two production changes it exists to cover is actually exercised. The `prepareForReuseCount` spy watches a view the production code can never reach: `bookmarkView` is produced by invoking the delegate method directly (`mgr.mapView(mgr.mapView, viewFor: bookmark)`), and `dequeueReusableAnnotationView(withIdentifier:for:)` hands that instance to the *caller* — the map never receives it, so it is not registered as the annotation's view. `refreshAnnotationViews` looks the view up via `mapView.view(for: annotation)`, which returns either nil (headless map, never laid out, so no annotation views were built) or a *different* instance the map dequeued for itself. Either way `bookmarkView.prepareForReuse()` is never called, on this branch or on `main`, and `#expect(bookmarkView.prepareForReuseCount == 0)` is vacuous. The other three assertions hold on `main` too: the sibling `Stop` is removed by the pre-existing `stopAnnotationsToRemove` path, and nothing on `main` deselects or removes the `Bookmark`. So `affectedStopIDs.subtract(selectedBookmarkStopIDs)` and the new `guard !mapView.selectedAnnotations.contains(...)` in `refreshAnnotationViews` remain untested. (The second test, `Remove stop annotations preserving selection keeps selected bookmark`, *does* now discriminate — `stop` is added as an annotation and only survives the layer-off removal because bookmark stop IDs were added to `selectedStopIDs` — so that half of the previous review is addressed.) A spy that bites would have to be the view the map itself vends, e.g. by asserting through `mapView.view(for:)` after forcing the map to build views, or by testing `refreshAnnotationViews`' input set directly.

https://github.com/OneBusAway/onebusaway-ios/blob/2eca75e2a0dc6632910b8650332b98a643d08458/OBAKitTests/Application/Mapping/MapRegionManagerTests.swift#L533-L546

2. The new bookmark under-pin label loop is still placed *after* the `guard isStopsLayerEnabled` early return, so it never runs in the one situation where it matters most: with the stops layer toggled off, bookmark pins are the only pins on the map, and they keep whatever `isHidingExtraStopAnnotationData` value they were built with across every subsequent zoom. Same for the zoom guard directly above it — past `requiredHeightToShowStops` the `Stop` annotations are removed but the bookmarks stay on screen with stale expanded labels. This is the point raised in the 2026-08-05 review and by CodeRabbit; the PR description claims the review was addressed, but this block is unchanged. Moving the two annotation-view loops above both guards (keeping `shouldHideExtraStopAnnotationData` as the source of truth) is what makes "extend zoom label updates to bookmark pins" actually true.

https://github.com/OneBusAway/onebusaway-ios/blob/2eca75e2a0dc6632910b8650332b98a643d08458/OBAKit/Mapping/MapRegionManager.swift#L999-L1019

aaronbrethorst (changes requested):
The production logic here is sound. I traced the deselect ordering, the replaced-bookmark path, and the selected-`Stop` path looking for a leaked selection or a stuck callout and didn't find one, and folding `Bookmark.stopID` into `selectedStopIDs` finally makes the code match the "(bookmarks get the same exemption)" comment that was already sitting above it. Test 2 is also properly fixed — adding `stop` as an annotation alongside `otherStop` means it would genuinely fail on `main`.

Two things still block, both carried over from the last round.

**The bookmark label loop is still below the early returns.** In `reloadStopAnnotations()`, the new loop sits after both the zoom guard and the `isStopsLayerEnabled` guard. The zoom guard removes `type: Stop.self` only and returns, so bookmarks survive with stale expanded labels when zoomed out; the layer guard returns while bookmarks are the *only* pins on screen. The net effect is that the refresh runs only in the case where it matters least. It needs to move above both guards — #1267 does exactly this, so it's worth looking at how that one is arranged.

**Test 1 still can't fail on `main`.** `bookmarkView` comes from calling the delegate method directly, so `dequeueReusableAnnotationView(withIdentifier:for:)` hands that view to the test, not to the map. `refreshAnnotationViews` looks the view up via `mapView.view(for:)`, which returns nil or a different instance, so `prepareForReuseCount` observes an object the production path never touches. The other three assertions already hold on `main`. That leaves `affectedStopIDs.subtract(...)` and the new `refreshAnnotationViews` guard — the actual substance of the change — untested.

Also, the manual repro checkbox is still unchecked. I'd like to know #782 is actually fixed on a device before this lands, because the trace I re-ran suggests a sibling `Stop` for a bookmarked ID is only ever added by the search paths, not by a routine pan.

Send it back my way once the loop moves and the test observes a map-vended view.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Both blockers are genuinely fixed, and I checked each in the code rather than taking the PR body's word for it.

**The label loop moved.** `refreshBookmarkAnnotationLabels()` now sits immediately after `updateZoomWarningOverlay()` and above all three early returns — the search override, the 40,000-height zoom guard that does `removeAnnotations(type: Stop.self)`, and the `isStopsLayerEnabled` guard. So bookmarks get their label gate re-applied in exactly the zoomed-out and layer-off cases where the old placement skipped it. The comment you added ("Bookmark pins are user content, so their label gate must refresh even when stop loading is suppressed") says the right thing and now matches the code.

**Test 1 can now fail.** It registers a `PrepareForReuseCountingStopAnnotationView` under the identifier `reuseIdentifier(for:)` actually returns for a `Bookmark`, hosts the map in a key window, and pulls the view via `mgr.mapView.view(for: bookmark)` — the same lookup `refreshAnnotationViews` uses. That was the whole problem last round: the test was observing an object the production path never touched. Now it observes the map-vended instance, so `prepareForReuseCount == 0` genuinely fails on `main`.

Two things worth carrying forward, neither blocking:

- **The manual repro box is ticked, but against the simulator.** I'd asked about a device, and the reason still stands: the `affectedStopIDs.subtract` fix only bites when a `Stop` annotation coexists with a `Bookmark` for the same stopID, and that's clearly reachable via the search path but not obviously via a routine pan — `stopsToAdd` already filters bookmarked stops. The change is a strict improvement either way, so I'm not holding the merge for it, but I wouldn't close #782 until someone confirms on hardware.
- **Test 1 leaks a key `UIWindow`** — `makeKeyAndVisible()` with no teardown. The two other window-hosting suites in the repo (`StopSheetPresenterTests`, `StopPageSheetHeaderLayoutTests`) both set `window.isHidden = true` in an `isolated deinit` with a comment explaining why. Serialization makes the practical risk low, but it does keep the `MKMapView` and its manager alive for the rest of the run.

Also: `affectedStopIDs.subtract(selectedBookmarkStopIDs)` and the new `guard !mapView.selectedAnnotations.contains(...)` inside `refreshAnnotationViews` cover the same case — either alone would do. Two mechanisms for one invariant is fine today, just something for whoever edits this next to be aware of.

Approving.

PR #1267, opened 2026-08-02
Fix map stop route labels and zoom hide gate

### LINKED ISSUE #132: small UI problems with route labels on stop icons on map
The route labels on the stop icons on the map are great, but there are a couple of UI bugs as of  0c8e4795c79fb9c65fdf836bbd6b1a6abc71800f.

If there are too many routes to display under the stop icon, the "..." is not always shown.  See screenshot with the 3rd Ave & Pine St (NW) stop.  I got confused by this trying to find that stop last night, since it had routes under it but not the 62 that I was looking for.
IMG_3087

The route labels are omitted if zoomed out, but they should be omitted a little more aggressively -- right now they can overlap other routes.  See 2nd screenshot.
IMG_3090

Borisserz:
## Summary
- Add `Formatters.formattedMapRoutes` for compact under-pin labels (short names + explicit overflow marker, no `Routes:` prefix). `formattedRoutes` and the map formatter share `sortedRouteDisplayNames` so they cannot drift.
- Lower `requiredHeightToShowExtraStopData` from 7000 to 5000. At 7000, route lists were still painted at a city-block zoom where pins overlap and the text is unreadable. 5000 keeps labels at street-level zoom and hides them sooner when zoomed out.
- Bookmark pin labels refresh on every camera settle, **before** the search, zoom, and stops-layer early returns. Bookmarks stay on the map in all three states; leaving the loop below those guards froze their labels.

The previous `bookmarkLabelRefreshHandler` seam is gone. MKMapView does not vend annotation views headlessly in CI, so there is no `view(for:)` assertion for this path. The wiring is the call-site ordering: `refreshBookmarkAnnotationLabels()` runs immediately after `updateZoomWarningOverlay()` and before every early return.

Closes #132

## Test plan
- [x] `FormattersTests` — map route formatter within/over limit cases
- [x] `MapRegionManagerTests` — 5000-point threshold (`5_000 == true`, `5_001 == false`)
- [ ] Manual: zoom out on map with many-route stops — labels should show an overflow marker and hide earlier than before
- [ ] Manual: bookmark pins should lose/gain labels when zooming, including with the stops layer off and during a route search

## Summary by CodeRabbit

- **New Features**
  - Map stop labels now display concise route names with comma separation and an ellipsis when additional routes are available.
  - Additional stop information becomes visible at a closer zoom level.

- **Bug Fixes**
  - Selected stop and bookmark callouts now remain visible during map updates.
  - Bookmark labels refresh more reliably after the map camera settles.
  - Map updates are less likely to interrupt selected bookmark details.

### DIFF
--- a/OBAKit/Mapping/MapRegionManager.swift
+++ b/OBAKit/Mapping/MapRegionManager.swift
@@ -944,6 +944,10 @@ public class MapRegionManager: NSObject,
 
     /// Above this visible-map-rect height (map points), stop pins are too
     /// zoomed-out for their under-pin label. Shared with `MapPanelRootView`.
+    ///
+    /// 7,000 still showed route lists at a city-block zoom where pins overlap
+    /// and the text is unreadable. 5,000 keeps labels for street-level zoom
+    /// where they help, and hides them sooner when zoomed out (#132).
     public static let requiredHeightToShowExtraStopData = 5000.0
 
     /// Height half of the under-pin label gate. Callers combine it with the
--- a/OBAKit/Mapping/MapRegionManager.swift
+++ b/OBAKit/Mapping/MapRegionManager.swift
@@ -944,7 +944,7 @@ public class MapRegionManager: NSObject,
 
     /// Above this visible-map-rect height (map points), stop pins are too
     /// zoomed-out for their under-pin label. Shared with `MapPanelRootView`.
-    public static let requiredHeightToShowExtraStopData = 7000.0
+    public static let requiredHeightToShowExtraStopData = 5000.0
 
     /// Height half of the under-pin label gate. Callers combine it with the
     /// standard-map-type and "show labels" default checks.
--- a/OBAKit/Mapping/MapRegionManager.swift
+++ b/OBAKit/Mapping/MapRegionManager.swift
@@ -838,7 +838,7 @@ public class MapRegionManager: NSObject,
 
     /// Above this visible-map-rect height (map points), stop pins are too
     /// zoomed-out for their under-pin label. Shared with `MapPanelRootView`.
-    public static let requiredHeightToShowExtraStopData = 7000.0
+    public static let requiredHeightToShowExtraStopData = 5000.0
 
     /// Height half of the under-pin label gate. Callers combine it with the
     /// standard-map-type and "show labels" default checks.
@@ -893,9 +893,18 @@ public class MapRegionManager: NSObject,
             return
         }
 
-        let visibleStops = mapView.annotations(in: mapView.visibleMapRect).filter(type: Stop.self)
-        for s in visibleStops {
-            if let stopView = mapView.view(for: s) as? StopAnnotationView {
+        // Stop and Bookmark pins share `StopAnnotationView`; both need the zoom gate
+        // refreshed on every region change (bookmarks were previously skipped).
+        let visibleRect = mapView.visibleMapRect
+        let visibleStopAnnotations = mapView.annotations(in: visibleRect).filter(type: Stop.self)
+        let visibleBookmarkAnnotations = mapView.annotations(in: visibleRect).filter(type: Bookmark.self)
+        for stop in visibleStopAnnotations {
+            if let stopView = mapView.view(for: stop) as? StopAnnotationView {
+                stopView.isHidingExtraStopAnnotationData = shouldHideExtraStopAnnotationData
+            }
+        }
+        for bookmark in visibleBookmarkAnnotations {
+            if let stopView = mapView.view(for: bookmark) as? StopAnnotationView {
                 stopView.isHidingExtraStopAnnotationData = shouldHideExtraStopAnnotationData
             }
         }
--- a/OBAKit/Extensions/ModelExtensions.swift
+++ b/OBAKit/Extensions/ModelExtensions.swift
@@ -57,7 +57,7 @@ nonisolated extension Stop: @retroactive MKAnnotation {
     }
 
     public var mapTitle: String? {
-        return Formatters.formattedRoutes(routes, limit: 3)
+        Formatters.formattedMapRoutes(routes)
     }
 }
 
--- a/OBAKitCore/Utilities/Formatters.swift
+++ b/OBAKitCore/Utilities/Formatters.swift
@@ -589,15 +589,9 @@ public class Formatters: NSObject {
 
     // MARK: - Routes
 
-    /// Generates a formatted, human readable list of routes.
-    ///
-    /// For example: "Routes: 10, 12, 49, + 3 more".
-    ///
-    /// - Parameter routes: An array of `Route`s from which the string will be generated.
-    /// - Parameter limit: The number of `Route`s that be displayed in the returned string. By default, this is `Int.max`.
-    /// - Returns: A human-readable list of the passed-in `Route`s.
-    public class func formattedRoutes(_ routes: [Route], limit: Int = .max) -> String? {
-        guard routes.count > 0 else { return nil }
+    /// Sorted short names (or route-type labels when short names are absent) for display.
+    private class func sortedRouteDisplayNames(from routes: [Route]) -> [String] {
+        guard routes.count > 0 else { return [] }
 
         var routeNames = routes
             .map { $0.shortName }
@@ -609,6 +603,40 @@ public class Formatters: NSObject {
             routeNames = routes.map { $0.routeType }.uniqued.compactMap { routeTypeToString($0) }.sorted { $0.localizedStandardCompare($1) == .orderedAscending }
         }
 
+        return routeNames
+    }
+
+    /// Compact route list for map pin labels under stop icons.
+    ///
+    /// For example: "10, 12, 49..." — no "Routes:" prefix, and an explicit "..." when the
+    /// list overflows so UIKit truncation doesn't silently drop the overflow hint (#132).
+    ///
+    /// - Parameter routes: An array of `Route`s from which the string will be generated.
+    /// - Parameter limit: The number of route names shown before appending "...".
+    /// - Returns: A comma-separated list of route short names, or `nil` when none are available.
+    public class func formattedMapRoutes(_ routes: [Route], limit: Int = 3) -> String? {
+        let routeNames = sortedRouteDisplayNames(from: routes)
+        guard !routeNames.isEmpty else {
+            return nil
+        }
+
+        if routeNames.count > limit {
+            return routeNames.prefix(limit).joined(separator: ", ") + "..."
+        }
+        else {
+            return routeNames.joined(separator: ", ")
+        }
+    }
+
+    /// Generates a formatted, human readable list of routes.
+    ///
+    /// For example: "Routes: 10, 12, 49, + 3 more".
+    ///
+    /// - Parameter routes: An array of `Route`s from which the string will be generated.
+    /// - Parameter limit: The number of `Route`s that be displayed in the returned string. By default, this is `Int.max`.
+    /// - Returns: A human-readable list of the passed-in `Route`s.
+    public class func formattedRoutes(_ routes: [Route], limit: Int = .max) -> String? {
+        let routeNames = sortedRouteDisplayNames(from: routes)
         guard !routeNames.isEmpty else {
             return nil
         }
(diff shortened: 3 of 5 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The new bookmark-pin refresh sits after both early returns in `reloadStopAnnotations`, so it doesn't run in the two cases where bookmark pins are the only pins left on the map. With the "Bus stops" layer toggled off, `guard isStopsLayerEnabled else { ... return }` (line 887) returns before the loop on *every* region change — bookmark labels then never update at any zoom, which is the exact behavior the PR's test plan claims to fix ("bookmark pins should lose/gain labels when zooming, same as stop pins"). The same applies to a single pinch-out that lands past `requiredHeightToShowStops`: the 40,000-height guard returns first, stop pins are removed, and the bookmark pins keep their labels painted over a region-scale map. Bookmarks stay on the map in both states by design (`displayUniqueStopAnnotations`: "bookmarks are user content and stay visible regardless"), their views are `StopAnnotationView`s (`reuseIdentifier(for:)` — `case is Bookmark`), and nothing else re-runs `prepareForDisplay` on a zoom, so `labelStack.isHidden` stays frozen. The SwiftUI path already treats this as a requirement — `MapPanelRootView.onMapCameraChange` sets `showStopLabels = false` *before* its own zoomed-out return, commented "so bookmark pins don't keep theirs painted over the map". Moving the bookmark loop above both guards (leaving the stop loop where it is) would make the added comment's "on every region change" accurate.

https://github.com/OneBusAway/onebusaway-ios/blob/c350cb5efe6f85595fed0f77e55c19c6b7e6c181/OBAKit/Mapping/MapRegionManager.swift#L881-L911

OBAKit/Mapping/MapRegionManager.swift
```diff
@@ -893,9 +893,18 @@ public class MapRegionManager: NSObject,
...
+        for stop in visibleStopAnnotations {
+            if let stopView = mapView.view(for: stop) as? StopAnnotationView {
+                stopView.isHidingExtraStopAnnotationData = shouldHideExtraStopAnnotationData
+            }
+        }
+        for bookmark in visibleBookmarkAnnotations {
+            if let stopView = mapView.view(for: bookmark) as? StopAnnotationView {
                 stopView.isHidingExtraStopAnnotationData = shouldHideExtraStopAnnotationData
             }
         }
```
aaronbrethorst: This loop needs to move above both early returns.

As placed, it sits below `guard visibleMapRect.height <= requiredHeightToShowStops` and below `guard isStopsLayerEnabled`. Bookmark pins stay on the map in both of those states — `removeStopAnnotationsPreservingSelection` only removes `Stop` annotations, and line 693 says "bookmarks are user content and stay visible regardless" — so with the stops layer off, bookmark labels freeze at whatever the gate last set and never refresh on zoom.

Suggest hoisting just the bookmark work to the top of the method, right after `updateZoomWarningOverlay`:

```swift
// Bookmark pins stay visible regardless of the stops layer and the stop
// zoom gate, so their labels must be refreshed before either early return.
let visibleRect = mapView.visibleMapRect
for bookmark in mapView.annotations(in: visibleRect).filter(type: Bookmark.self) {
    if let stopView = mapView.view(for: bookmark) as? StopAnnotationView {
        stopView.isHidingExtraStopAnnotationData = shouldHideExtraStopAnnotationData
    }
}
```

and leaving the `Stop` loop down here where it is. The comment on line 896 ("both need the zoom gate refreshed on every region change") will then be accurate.

aaronbrethorst (changes requested):
The formatter work here is good. Extracting `sortedRouteDisplayNames` so `formattedRoutes` and the new `formattedMapRoutes` can't drift apart is the right refactor, and a prefix-free compact list with an explicit overflow marker is a real improvement over what the map pin showed before.

I checked the threshold change carefully because that's where this kind of fix usually goes wrong, and it's correct — `shouldShowExtraStopData` is `height <= requiredHeightToShowExtraStopData`, so lowering 7000 to 5000 hides labels *sooner* when zoomed out, which is what you describe. The boundary tests (`5_000 == true`, `5_001 == false`) match the `<=` semantics, and no stale `7000` references survive anywhere in the tree. I also confirmed there's no annotation-reuse hazard: `StopAnnotationView.prepareForReuse` nils `titleLabel.text` and hides `labelStack`, and `prepareForDisplay` repopulates both the `Stop` and `Bookmark` branches, so a recycled view can't carry a stale label. And it composes fine with the route-projection work that just landed — while a stop sheet is up those annotations become `BackgroundDotAnnotationView`, the `as? StopAnnotationView` casts fail, and both loops correctly no-op.

One blocker, and it's a one-hunk move.

**The bookmark half of the fix is placed below both early returns**, so it doesn't run in the two states where bookmark pins are the only pins on screen. In `reloadStopAnnotations`, the new `for bookmark in visibleBookmarkAnnotations` loop sits under `guard mapView.visibleMapRect.height <= requiredHeightToShowStops` and under `guard isStopsLayerEnabled`. I verified bookmark pins survive both: `removeStopAnnotationsPreservingSelection` only removes `Stop` annotations, and the comment at `MapRegionManager.swift:693` says it outright — "bookmarks are user content and stay visible regardless." So with the Bus stops layer off, a bookmark pin keeps whatever `isHidingExtraStopAnnotationData` it was last given and never refreshes at any zoom, because nothing else re-runs `prepareForDisplay` on a region change.

That's precisely the case your test plan says this fixes, and the new comment — "both need the zoom gate refreshed on every region change" — doesn't hold as written. Worth noting the SwiftUI surface already treats this as a requirement: `MapPanelRootView.onMapCameraChange` sets `showStopLabels = false` *before* its zoomed-out return, commented "so bookmark pins don't keep theirs painted over the map."

Move the bookmark loop above both guards and leave the stop loop where it is.

- `formattedMapRoutes` hardcodes `", "` and `"..."` while every sibling in `Formatters` routes user-visible text through `OBALoc`. It's punctuation only and CLAUDE.md doesn't require it, but a localizer may want a different list separator or the `…` glyph.
- There's no test for the bookmark refresh itself — `MapRegionManagerTests` only re-asserts the 5,000 constant. Once you move the loop, a test that the gate reaches bookmark views would keep this from regressing.

Fix the placement and I'll get this in.

aaronbrethorst:
Heads up, and this one's on me: I merged #1270 a few minutes ago, which touched `Formatters.swift`, so this branch is now conflicted on top of the placement fix I asked for. Merge `main` in while you're addressing that — no re-review needed for the conflict resolution itself unless it changes behavior.

Also relevant to the ordering question I raised on #1257: that PR is still open and edits the same overflow branch of `formattedRoutes` that this one routes around. My preference is to keep this one and close #1257, since a dedicated map formatter is cleaner than overloading `formattedRoutes` for its five other callers. Let me know if you'd rather go the other way.

Borisserz:
@AaronBrethorst Addressed the review: bookmark label visibility is now refreshed before every early return, map-route overflow uses , and the regression test covers the search-suppressed path. Also, #1257 was closed in favor of this PR.

Borisserz:
@AaronBrethorst Correction: map-route overflow now uses the Unicode ellipsis character (U+2026). Bookmark label visibility is refreshed before every early return, and the regression test covers the search-suppressed path. #1257 was closed in favor of this PR.

Borisserz:
CI failure was the new bookmark-label regression test under a search result: calling the `viewFor` delegate alone does not attach the view, so `mapView.view(for:)` no-oped in headless CI.

Fixed by dequeuing with `for: annotation` (so the view is attached) and refreshing every bookmark annotation before the search early-return, not only `annotations(in:)`.

Borisserz:
Same test failed again: even with `dequeue(for:)`, headless CI does not keep a stable `mapView.view(for:)` for the assertion.

Replaced that MapKit view assertion with a `bookmarkLabelRefreshHandler` seam that fires immediately before bookmark label refresh — so the test fails if the refresh is moved below the search early-return, without depending on annotation view attachment.

aaronbrethorst:
### Code review

aaronbrethorst (changes requested):
The blocker from last time is properly fixed: `refreshBookmarkAnnotationLabels()` now runs above the search guard, the zoom guard, and the layer guard, and broadening it from `annotations(in:)` to all `mapView.annotations` means offscreen-but-attached views are covered too. The comment's claim that it runs on every region change is now accurate.

I also checked the rest against the head rather than taking it on trust. `formattedMapRoutes` gets the overflow boundary right (exactly `limit` routes gets no ellipsis, `limit + 1` elides), the `sortedRouteDisplayNames` extraction is behavior-preserving for `formattedRoutes`, and the 5,000 threshold matches the `<=` semantics in the tests. Good, careful work.

One thing I want changed before this lands, and it's small:

`bookmarkLabelRefreshHandler` is production API that exists only so a test can observe a call. Nothing outside the test suite ever assigns it. And the test it enables — `A camera settle updates a bookmark label while a search result is displayed` — asserts only that the closure fired, so if someone deleted the `refreshBookmarkAnnotationLabels()` call and left the handler in place, the test would still pass. It can't catch the regression it's named for.

You hit a real obstacle here and you documented it honestly, which I appreciate: MKMapView won't vend annotation views headlessly, so `view(for:)` assertions are genuinely unreliable in CI. But the answer isn't a seam on the manager. Either pull the gate decision into something pure that can be asserted directly, or drop the test and say in the PR that the wiring is covered by the call-site ordering. Both are fine by me; shipping the closure isn't.

Separately, and not a blocker: lowering `requiredHeightToShowExtraStopData` from 7000 to 5000 is a product judgment with no rationale in the diff. I'm inclined to take it, but say a word about what you saw that prompted it.

Ping me when the handler's gone and I'll take another look.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
The blocker is resolved, and resolved the right way: you removed the production seam rather than papering over it with a better-looking assertion. I confirmed `bookmarkLabelRefreshHandler` appears nowhere at the head SHA — zero occurrences in `MapRegionManager.swift`, and it's absent from the diff entirely.

Dropping the test and covering the wiring by call-site ordering was one of the two options I said I'd take, and it's the honest one given that MKMapView genuinely won't vend annotation views headlessly. `refreshBookmarkAnnotationLabels()` is called once, immediately after `updateZoomWarningOverlay()` and above all three early returns, which is the ordering that matters.

The 7,000 → 5,000 rationale is now in the doc comment where the next reader will actually find it — "7,000 still showed route lists at a city-block zoom where pins overlap and the text is unreadable" is precisely the kind of thing I wanted recorded, because otherwise it reads as an arbitrary constant change three years from now. Thank you for putting it in the code rather than only in the PR description.

The two new `FormattersTests` cases assert exact output (`"10, 20, 30"` and `"10, 20, 30..."`), so they'd fail on a regression in either the at-limit or over-limit branch. That's real coverage.

One small thing, non-blocking: back on 2026-08-05 you noted that map-route overflow "now uses the Unicode ellipsis character (U+2026)", but at head `Formatters.swift` is three ASCII periods and the test asserts the ASCII form. Looks like the rebase dropped that change. I'd already called hardcoded punctuation non-blocking, so this only matters as a claim that no longer matches the code — worth a one-character fix whenever you're next in there.

Approving.

aaronbrethorst:
Approved on the merits — this needs a rebase before it can land, and the conflict is worth a word because it isn't mechanical.

#1266 merged first, and it introduces a `refreshBookmarkAnnotationLabels()` on `MapRegionManager` that is **byte-for-byte identical** to yours, with the same call-site placement above the early returns. So `main` already has the method. Git flagged the conflict rather than merging both, which is lucky — two definitions in the same type would have been an invalid redeclaration and a broken build.

Resolution: merge `main` in and drop your copy of the function and its call, keeping the one already there. Everything else in this PR — the `formattedMapRoutes` overflow work, `sortedRouteDisplayNames`, and the 5,000 threshold with its rationale — is yours alone and unaffected.

No re-review needed once it's mergeable, unless the resolution changes behavior.

Borisserz:
Rebased onto current `main`.

`refreshBookmarkAnnotationLabels()` is the copy #1266 already landed — this branch no longer redefines it or re-calls it. The remaining `MapRegionManager` delta is the 5,000-point extra-stop label threshold and its comment.

`formattedMapRoutes` / `sortedRouteDisplayNames` and the Formatters tests are unchanged in behavior.

Local: `FormattersTests` + `MapRegionManagerTests` 29/29 on iPhone 17 Pro.

aaronbrethorst:
### Code review

PR #1268, opened 2026-08-02
Silence Bookmarks-tab 404 bulletins for missing stops

### LINKED ISSUE #1181: San Diego iOS 404 Bookmark Error
Hello, I receive a 404 not found error (screenshot attached) when opening the bookmarks tab on iOS. The bookmarks themselves are fine, I hit dismiss, select my bookmark, and all the bus route ETAs and mapping tools work as expected. But when I exit the bookmark back into the bookmark tab, the error pops up before I can navigate to the other tabs (eg Map).

I haven't seen this on Samsung (my previous phone), but this has been a consistent error recently (a few months at minimum - I forget to report bugs timely). Don't remember if the iOS 18 had issues, but iOS 26 definitely does. Currently running iOS 26.5.2, if that helps.

First image was taken yesterday upon opening the app, and second image was taken right now when opening the app. Please let me know if you need more information!

Borisserz:
## Summary
- Literal **HTTP 404** on a trip bookmark is silent: no bulletin, no failure haptic, the card settles on "No upcoming departures", and any previous departures for that stop are cleared.
- **Empty HTTP 200** (also thrown as `APIError.requestNotFound` by `APIService+GetData`) still bulletins. A live San Diego stop is not that case.
- Docs on `fetchedStopIDs` / `hasFetchedData` now match the 404 path.

Closes #1181

## San Diego trace (2026-08-14)

Hit `realtime.sdmts.com` the way the app builds URLs (`base https://realtime.sdmts.com/api/` + path `/api/where/...` → `/api/api/where/...`):

- Live stop `MTS_11589` arrivals: **HTTP 200**, full JSON body (`code: 200`, real `arrivalsAndDepartures`). Not empty. Same for `stop/MTS_11589` and `current-time`.
- The short path `/api/where/...` (without the doubled `/api`) is HTTP 404 — that is not the URL the app uses.
- Missing IDs (`MTS_NOPE`, `MTS_99999999`, unprefixed `11589`) came back as HTTP 500, which still takes the generic `displayError` path.

So a working San Diego bookmark is a normal 200. The 404 bulletin in #1181 is a *literal* missing resource (or a sibling bookmark that 404s while the one you tap still resolves). Empty-body 200 is not what this region returns for a live stop.

## Test plan
- [x] Literal HTTP 404 — no `displayError`, `hasLoadedArrivalData == true`
- [x] Empty HTTP 200 — still `displayError`, `lastRefreshHadError`
- [x] HTTP 500 — still `displayError`
- [x] HTTP 404 after a successful fetch — stale departures cleared
- [ ] Manual: Bookmarks tab in San Diego — no 404 bulletin on a live stop; a genuinely gone stop shows "No upcoming departures"

## Summary by CodeRabbit

* **Bug Fixes**
  * Bookmark refreshes now complete gracefully when a stop returns no available service, without showing an error.
  * Previously displayed departure information is cleared when a stop is confirmed unavailable.
  * Genuine refresh failures continue to display an error and correctly indicate that the refresh encountered a problem.
  * Outdated refresh results are ignored to prevent stale information from replacing current data.

### DIFF
--- a/OBAKitCore/Bookmarks/BookmarkDataLoader.swift
+++ b/OBAKitCore/Bookmarks/BookmarkDataLoader.swift
@@ -52,13 +52,14 @@ public class BookmarkDataLoader: NSObject {
     /// Resumed when that batch drains (`taskFinished`) or is retired (`cancelUpdates`).
     @MainActor private var batchContinuations: [UInt64: [CheckedContinuation<Void, Never>]] = [:]
 
-    /// Stops whose arrival fetch has completed successfully at least once this
-    /// session. Lets consumers distinguish "still loading" from "loaded, but no
-    /// upcoming departures".
+    /// Stops whose arrival fetch has finished this session — a successful
+    /// payload **or** a literal HTTP 404. Lets consumers distinguish "still
+    /// loading" from "loaded, but no upcoming departures". Empty HTTP 200
+    /// (also thrown as `APIError.requestNotFound`) is not recorded here.
     @MainActor private var fetchedStopIDs = Set<StopID>()
 
-    /// `true` once at least one arrival fetch for `stopID` has completed
-    /// successfully this session.
+    /// `true` once an arrival fetch for `stopID` has finished this session
+    /// (success or HTTP 404).
     @MainActor public func hasFetchedData(forStopID stopID: StopID) -> Bool {
         fetchedStopIDs.contains(stopID)
     }
@@ -178,14 +179,17 @@ public class BookmarkDataLoader: NSObject {
 
                     self.delegate?.dataLoaderDidUpdate(self)
                 }
-            } catch APIError.requestNotFound {
-                // A bookmarked stop that no longer resolves in the current region is
-                // not a failure the rider watched happen — mirror StopViewModel's 404
-                // handling: no bulletin, no batch-error flag. Still mark it fetched so
-                // the card settles on "No upcoming departures" rather than "Loading...".
+            } catch APIError.requestNotFound(let response) where response.statusCode == 404 {
+                // Literal HTTP 404: the stop no longer exists in this region.
+                // San Diego trace 2026-08-14 against realtime.sdmts.com: a live
+                // stop (`MTS_11589`) returns HTTP 200 with a full JSON body on
+                // the app URL (`/api/api/where/...`). Empty HTTP 200 — also
+                // thrown as `requestNotFound` by `APIService+GetData` — is a
+                // transient blip and falls through to `displayError` below.
                 await MainActor.run {
                     guard batchID == self.currentBatchID else { return }
                     self.fetchedStopIDs.insert(bookmark.stopID)
+                    self.tripBookmarkKeys = self.tripBookmarkKeys.filter { $0.key.stopID != bookmark.stopID }
                     self.delegate?.dataLoaderDidUpdate(self)
                 }
             } catch {
--- a/OBAKitCore/Bookmarks/BookmarkDataLoader.swift
+++ b/OBAKitCore/Bookmarks/BookmarkDataLoader.swift
@@ -181,7 +181,13 @@ public class BookmarkDataLoader: NSObject {
             } catch APIError.requestNotFound {
                 // A bookmarked stop that no longer resolves in the current region is
                 // not a failure the rider watched happen — mirror StopViewModel's 404
-                // handling: no bulletin, no batch-error flag.
+                // handling: no bulletin, no batch-error flag. Still mark it fetched so
+                // the card settles on "No upcoming departures" rather than "Loading...".
+                await MainActor.run {
+                    guard batchID == self.currentBatchID else { return }
+                    self.fetchedStopIDs.insert(bookmark.stopID)
+                    self.delegate?.dataLoaderDidUpdate(self)
+                }
             } catch {
                 // Same staleness gate as the success path: if cancelUpdates() retired
                 // the batch (or a newer batch started) while this fetch was in flight,
--- a/OBAKitCore/Bookmarks/BookmarkDataLoader.swift
+++ b/OBAKitCore/Bookmarks/BookmarkDataLoader.swift
@@ -178,6 +178,10 @@ public class BookmarkDataLoader: NSObject {
 
                     self.delegate?.dataLoaderDidUpdate(self)
                 }
+            } catch APIError.requestNotFound {
+                // A bookmarked stop that no longer resolves in the current region is
+                // not a failure the rider watched happen — mirror StopViewModel's 404
+                // handling: no bulletin, no batch-error flag.
             } catch {
                 // Same staleness gate as the success path: if cancelUpdates() retired
                 // the batch (or a newer batch started) while this fetch was in flight,
--- a/OBAKitTests/ViewModels/BookmarksViewModelTests.swift
+++ b/OBAKitTests/ViewModels/BookmarksViewModelTests.swift
@@ -64,6 +64,59 @@ final class BookmarksViewModelTests: OBATestCase {
         return Application(config: config)
     }
 
+    @MainActor
+    private final class DisplayErrorSpyApplication: Application {
+        private(set) var displayErrorCallCount = 0
+
+        override func displayError(_ error: Error) async {
+            displayErrorCallCount += 1
+        }
+    }
+
+    private func createSpyApplication(dataLoader: MockDataLoader) -> DisplayErrorSpyApplication {
+        stubRegions(dataLoader: dataLoader)
+        stubAgenciesWithCoverage(dataLoader: dataLoader, baseURL: Fixtures.pugetSoundRegion.OBABaseURL)
+        Fixtures.stubAllAgencyAlerts(dataLoader: dataLoader)
+
+        let locManager = MockAuthorizedLocationManager(
+            updateLocation: TestData.mockSeattleLocation,
+            updateHeading: TestData.mockHeading
+        )
+        let locationService = LocationService(userDefaults: userDefaults, locationManager: locManager)
+        locationService.startUpdates()
+
+        let config = AppConfig(
+            regionsBaseURL: regionsURL,
+            apiKey: apiKey,
+            appVersion: appVersion,
+            userDefaults: userDefaults,
+            analytics: AnalyticsMock(),
+            queue: queue,
+            locationService: locationService,
+            bundledRegionsFilePath: bundledRegionsPath,
+            regionsAPIPath: regionsAPIPath,
+            dataLoader: dataLoader,
+            fixedRegionName: Fixtures.pugetSoundRegion.name
+        )
+
+        return DisplayErrorSpyApplication(config: config)
+    }
+
+    private func addTripBookmark(to app: Application) throws -> Bookmark {
+        let stopArrivals = try Fixtures.loadRESTAPIPayload(
+            type: StopArrivals.self,
+            fileName: "arrivals-and-departures-for-stop-1_10914.json"
+        )
+        let arrivalDep = try #require(stopArrivals.arrivalsAndDepartures.first)
+        let bookmark = Bookmark(
+            name: "Route 49",
+            regionIdentifier: pugetSoundRegionIdentifier,
+            arrivalDeparture: arrivalDep
+        )
+        app.userDataStore.add(bookmark, to: nil)
+        return bookmark
+    }
+
     // MARK: - Tests
 
     /// `init` defaults to `true` (set via `register(defaults:)`) on a clean UserDefaults.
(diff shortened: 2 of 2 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The 404 branch silences the bulletin but leaves the bookmark card stuck on "Loading..." forever. Only the success path does `self.fetchedStopIDs.insert(bookmark.stopID)`, so after a 404 `hasFetchedData(forStopID:)` stays `false`, `BookmarkRowViewModel.hasLoadedArrivalData` stays `false`, and `BookmarkCardView.loadingText` renders "Loading..." rather than "No upcoming departures" — for the life of the session, on every refresh. The new branch also never fires `dataLoaderDidUpdate`. Previously the perpetual-loading row was at least accompanied by an error bulletin and a failure haptic (`lastBatchHadError`); this PR removes both, so the rider is left with a row that claims to be loading and nothing else. `BookmarkCardView` carries the guidance directly above the code this breaks — "a fetched-but-empty result must not look like it's loading forever" — and `fetchedStopIDs`/`hasFetchedData` exists precisely because PR #1191's review flagged "Do not render an empty result as perpetual loading." Worth noting too that `StopViewModel`, which the new comment says it mirrors, does not just go quiet: it sets `isBrokenBookmark`, which gives the rider an explanation and a way out. This change adopts the silent half without the explanatory half. Suggest the 404 branch insert into `fetchedStopIDs` and notify the delegate (terminal "No upcoming departures" state), or carry a broken-bookmark signal through to the row.

https://github.com/OneBusAway/onebusaway-ios/blob/6263cf6a002949b8fdaa2522926ccb5512a672c1/OBAKitCore/Bookmarks/BookmarkDataLoader.swift#L179-L186

OBAKitCore/Bookmarks/BookmarkDataLoader.swift
```diff
@@ -178,6 +178,10 @@ public class BookmarkDataLoader: NSObject {
 
                     self.delegate?.dataLoaderDidUpdate(self)
                 }
+            } catch APIError.requestNotFound {
+                // A bookmarked stop that no longer resolves in the current region is
+                // not a failure the rider watched happen — mirror StopViewModel's 404
+                // handling: no bulletin, no batch-error flag.
```
aaronbrethorst: This branch needs to mark the stop as fetched, or the card never leaves "Loading...".

`fetchedStopIDs.insert(bookmark.stopID)` happens only on the success path above (line 172). With this branch empty, `hasFetchedData(forStopID:)` stays `false` for a 404'd stop, so `BookmarkCardView.loadingText` renders "Loading..." indefinitely instead of "No upcoming departures" — the exact thing that label's doc comment says must not happen.

Minimum fix:

```swift
} catch APIError.requestNotFound {
    // A bookmarked stop that no longer resolves in the current region is
    // not a failure the rider watched happen — mirror StopViewModel's 404
    // handling: no bulletin, no batch-error flag. Still mark it fetched so
    // the card settles on "No upcoming departures" rather than "Loading...".
    await MainActor.run {
        guard batchID == self.currentBatchID else { return }
        self.fetchedStopIDs.insert(bookmark.stopID)
        self.delegate?.dataLoaderDidUpdate(self)
    }
}
```

Note the staleness guard — the success path gates on `batchID == self.currentBatchID` and this should too.

aaronbrethorst (changes requested):
The error-classification half of this is right, and I checked the blast radius rather than assuming it: network failures, captive-portal, auth, other 4xx/5xx, decoding errors and `noResponseBody` all still fall through to the generic catch and `displayError`. That's a genuinely tight narrowing, and a 404 on a stop that no longer resolves in the region really isn't a failure the rider watched happen. The `DisplayErrorSpyApplication` seam is a clean way to test it, too.

The blocker is what the new branch *doesn't* do.

**A 404 now leaves the bookmark card stuck on "Loading..." forever.** `fetchedStopIDs.insert(bookmark.stopID)` happens only on the success path (line 172). The new `catch APIError.requestNotFound` branch is empty, so the stop is never marked fetched — `hasFetchedData(forStopID:)` stays `false`, `BookmarksViewModel.buildRow` passes `hasLoadedArrivalData: false`, and `BookmarkCardView.loadingText` renders "Loading..." rather than "No upcoming departures".

Two things make me confident this is a real defect and not a nitpick. First, the doc comment directly above `loadingText` states the rule this breaks: *"a fetched-but-empty result must not look like it's loading forever."* Second, `fetchedStopIDs`/`hasFetchedData` exists **only** because a review on #1191 demanded it — the note was "Do not render an empty result as perpetual loading," and commit `6220c8a1` added the mechanism in response. This change walks that back for the 404 case.

It's worth being precise about why this lands on you rather than being pre-existing: the perpetual-"Loading..." state did exist before, but it came with a bulletin and a failure haptic, so the rider knew something had gone wrong. This PR's whole purpose is to declare that state benign — and having removed the last signal, the row needs a terminal state of its own.

The comment also says it mirrors `StopViewModel`'s 404 handling, but `StopViewModel` sets `isBrokenBookmark`, which drives an explanatory page with a way out. This adopts the silent half and not the explanatory half.

Smallest fix that would make this landable: insert into `fetchedStopIDs` and notify the delegate in the 404 branch, the same as the success path, so the card settles on "No upcoming departures." If you'd rather say something truer, plumbing a broken-bookmark state through to the card would be better still — but the one-liner is enough for me.

One thing I'd like you to confirm before this lands, separate from the above. Issue #1181's reporter says *"The bookmarks themselves are fine, I hit dismiss, select my bookmark, and all the bus route ETAs and mapping tools work as expected"* — which contradicts the premise that the stop no longer resolves. `APIService+GetData.swift:54` throws `requestNotFound` for a GET returning **HTTP 200 with `expectedContentLength == 0`**, not just a literal 404. If San Diego is actually returning empty bodies, this suppresses the symptom without establishing the cause, and a transient empty-body response across all bookmarks would now report as a *successful* refresh — success haptic, `lastRefreshHadError == false`. Can you check a real trace from that region?

Fix the loading state and I'll turn this around quickly.

Borisserz:
Addressed the loading-state blocker:

- `catch APIError.requestNotFound` now inserts into `fetchedStopIDs` and calls `dataLoaderDidUpdate` (same staleness gate as success), still without `displayError` / `lastBatchHadError`.
- Test asserts `hasLoadedArrivalData` after a 404 so the row leaves perpetual "Loading...".

On the San Diego empty-body vs 404 question: `APIService+GetData` maps both HTTP 404 and HTTP 200 with `expectedContentLength == 0` to `requestNotFound`, so this path covers both. I don't have a live San Diego trace in this environment; if empty bodies are transient across *all* bookmarks, they'd now look like a successful empty refresh. Happy to follow up with a tighter distinction (literal 404 vs empty body) if you want that as a separate change.

`BookmarksViewModelTests` — 18/18 on iPhone Air.

aaronbrethorst:
### Code review

The loading-state blocker from the previous review is genuinely fixed: the 404 branch now inserts into `fetchedStopIDs` and fires `dataLoaderDidUpdate` under the same staleness gate as the success path, and the new test asserts `hasLoadedArrivalData`. The narrowing is still tight — network, auth, other 4xx/5xx, decoding and `noResponseBody` all fall through to `displayError`.

Found 1 issue:

1. The 404 branch now reports a *successful* refresh for a case that isn't necessarily a missing stop, and the "settles on No upcoming departures" claim only holds for a never-fetched stop. `APIError.requestNotFound` is thrown by `APIService+GetData.swift:54` for a GET returning **HTTP 200 with `expectedContentLength == 0`**, not just a literal 404 — the codebase's own comment there explains this is a REST API quirk. So a server-side empty-body blip on a stop that resolves fine now takes this branch: no bulletin, `lastBatchHadError` untouched (success haptic, `lastRefreshHadError == false`), and the stop positively marked fetched. That is strictly less signal than before this PR, and it renders a definitive "No upcoming departures" for a stop that may have departures. Issue #1181's reporter says "The bookmarks themselves are fine, I hit dismiss, select my bookmark, and all the bus route ETAs and mapping tools work as expected" — which contradicts the premise written into the new comment ("A bookmarked stop that no longer resolves in the current region"), and suggests the empty response is transient rather than a genuinely gone stop. This is the same confirmation that was requested on 2026-08-05 and it hasn't been resolved; a literal-404-vs-empty-body distinction, or plumbing a broken/stale signal to the card the way `StopViewModel` does with `isBrokenBookmark`, would close it. Corollary worth noting on the same branch: `tripBookmarkKeys` is only ever written on the success path and never cleared (`BookmarkDataLoader.swift:176`, `:230`), so a 404 arriving *after* a prior successful fetch leaves the previous departures on the card — the rider sees a stale, silently-frozen countdown with a success haptic, not "No upcoming departures".

https://github.com/OneBusAway/onebusaway-ios/blob/9626e02f2a41a945a3e1b2601a9f34b81c3a0f77/OBAKitCore/Bookmarks/BookmarkDataLoader.swift#L180-L192

Lower-confidence, non-blocking: the doc comments on `fetchedStopIDs` and `hasFetchedData` still say the set holds stops whose fetch "completed **successfully**" — the 404 branch now inserts on a failure, so those two comments are stale.

aaronbrethorst (changes requested):
The loading-state half of this is genuinely fixed, and fixed the way I asked: `fetchedStopIDs.insert` and `dataLoaderDidUpdate` now happen inside the same `batchID == currentBatchID` gate as the success path, and the new test asserts `hasLoadedArrivalData` after a 404 so it would catch a regression. Narrowing the catch to a specific error case rather than swallowing everything was the right instinct too.

What still blocks it is the other half, which is the part I asked you to confirm before this landed. `APIError.requestNotFound` is not only thrown on a literal 404 — `OBAKitCore/Network/APIService/APIService+GetData.swift` throws it for any GET that returns **HTTP 200 with an empty body**, and the comment right above it explains that this is how the REST API signals a bogus stop ID. So a valid stop that momentarily returns an empty body now produces: no bulletin, `lastBatchHadError` untouched (success haptic), and the stop positively marked as fetched — which renders a confident "No upcoming departures" that is simply wrong. That is worse than the noisy failure it replaces, because the rider has no signal anything went wrong.

You acknowledged this conflation in your reply, which I appreciate. I'd still like the San Diego trace before we commit to treating the state as benign.

Two smaller things to fold in:

1. The doc comments on `fetchedStopIDs` and `hasFetchedData` in `BookmarkDataLoader.swift` both say "completed **successfully**". The 404 branch now inserts on failure, so they're false as written.
2. `tripBookmarkKeys` is only written on the success path and never cleared, so a 404 arriving after a prior successful fetch leaves stale departures on the card with no error — a frozen countdown rather than the "No upcoming departures" the new comment describes. Partly pre-existing, but before this change it at least came with a bulletin.

Happy to re-review as soon as you have the trace.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is exactly what I was asking for, and you fixed the hard part rather than the easy part.

The blocker was that `APIError.requestNotFound` is thrown from two places, and only one of them is a real 404 — `APIService+GetData` also throws it for a GET returning HTTP 200 with an empty body. I verified the new discrimination end to end rather than trusting the commit message:

- `case requestNotFound(HTTPURLResponse)` carries the real response, so `statusCode` is 200 for the empty-body case and 404 for the literal one.
- The new `catch APIError.requestNotFound(let response) where response.statusCode == 404` therefore genuinely excludes the empty-200 case, which falls through to the generic catch, sets `lastBatchHadError`, and still shows the bulletin.
- No intermediate wrapping in between — `RESTAPIService+Get` calls `getData` directly, and the only other `requestNotFound` throw site is in `SendData`, which this path never touches.

The empty-200 test isn't vacuous either: `MockDataLoader` passes a real `contentLength`, so the mock genuinely produces `requestNotFound(200)` and the test would fail if the `where` clause were dropped.

Both smaller items are done too — the `fetchedStopIDs`/`hasFetchedData` doc comments now say "success **or** a literal HTTP 404" and explicitly note that empty-200 isn't recorded, and the 404 branch now clears `tripBookmarkKeys` for that stop under the same batch gate, so a card drops to "No upcoming departures" instead of freezing a stale countdown.

The narrowing is appropriately tight: network, auth, other 4xx/5xx, decoding, `noResponseBody`, and empty-200 all still bulletin.

One thing to carry forward rather than fix here. Your own trace in the PR body reports that `realtime.sdmts.com` returns **500** for a missing stop ID, not 404, and that the only 404 you saw came from a URL the app doesn't build. If San Diego is the reporter's region, then the bulletin in #1181 wasn't a literal 404 and this branch would never fire for them. The code is right regardless — I just wouldn't auto-close #1181 on merge without confirming.

Related and accepted: a region with a wrong base URL or API path now 404s every bookmark and shows a confident "No upcoming departures" with a success haptic and no bulletin. That's the cost of the behavior we asked for, but worth knowing it's strictly less signal than before in that one scenario.

Approving.

PR #1269, opened 2026-08-02
Narrow PanelDetent docs to UIKit FloatingPanel

### LINKED ISSUE #1141: ViewModel layer fit-and-finish follow-ups

Borisserz:
## Summary
- Narrows `PanelDetent` / `requestedPanelDetent` docs to the current UIKit `FloatingPanel` mapping (`tip` / `half` / `full`).
- Removes the SwiftUI `presentationDetent` TODO that the closed enum can't deliver.
- The other #1141 follow-ups already landed on main: `bindListData` split, dropped `type_body_length` disable, and surveys fetched once via `performInitialStopSetupIfNeeded` (locked by `StopViewModelTests`).
- Closes #1141.

## Test plan
- [x] Confirmed on `upstream/main`: `StopViewModelTests` includes `Surveys refreshed exactly once across refreshes`
- [x] Doc-only change — no behavior delta
- [ ] Skim `MapPanelViewModel` comments in Xcode for accuracy

### DIFF
--- a/OBAKit/ViewModels/MapPanelViewModel.swift
+++ b/OBAKit/ViewModels/MapPanelViewModel.swift
@@ -11,9 +11,8 @@ import Foundation
 import Combine
 import OBAKitCore
 
-/// Describes the desired expansion state of the bottom panel (EC11).
+/// Desired expansion level for the map bottom panel's UIKit `FloatingPanel`.
 /// `MapFloatingPanelController` maps each case to a `FloatingPanelState`.
-/// - TODO: add a SwiftUI `PresentationDetent` mapping when the SwiftUI sheet lands.
 enum PanelDetent: Equatable {
     case tip
     case half
@@ -29,9 +28,8 @@ class MapPanelViewModel: ObservableObject {
 
     // MARK: - Published State
 
-    /// Desired panel expansion level. `MapFloatingPanelController` observes this and calls
-    /// `floatingPanel.move(to:)` (EC11).
-    /// - TODO: bind to SwiftUI `presentationDetent` when the SwiftUI sheet lands.
+    /// Desired panel expansion level. Observed by `MapFloatingPanelController`, which
+    /// calls `floatingPanel.move(to:)`.
     @Published var requestedPanelDetent: PanelDetent = .tip
 
     /// Nearby stops visible in the current map region.


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Good cleanup. I checked the claim rather than taking it on faith, and it holds: `MapFloatingPanelController.bindPanelDetent()` does switch every `PanelDetent` case to a `FloatingPanelState`, and the SwiftUI sheet that the deleted TODOs were waiting on has already landed — `HomeSheetView` is backed by its own `HomeSheetViewModel`, with detents supplied by `FloatingSheetContainer`'s `.presentationDetents(...)`, not by `PanelDetent`. So those TODOs described work that will never happen on this enum, and deleting them is right.

Dropping the `(EC11)` markers is also an improvement — they were opaque to anyone who wasn't around for that ticket.

Two stale comments in the same neighborhood, if you want to finish the sweep in a follow-up — both on lines this PR didn't touch, so not a blocker:

- `MapPanelViewModel`'s class doc still says it's consumed "by future `HomeSheetView` (SwiftUI, via `@StateObject`)".
- `MapFloatingPanelController.swift:380` still says "so UIKit and future SwiftUI share the same source of truth".

Both are stale for exactly the reason this PR cites.

Doc-only, no behavior change. Merging.

PR #1270, opened 2026-08-02
Widget: say Scheduled/not real-time for schedule-only trips, and stop large-family rows clipping

diveshpatil9104:
## Summary

Implements the two widget follow-ups from the review of #1237.

1. **Schedule-only trips claimed "arrives on time".** The widget's deviation text comes from `deviationFromScheduleInMinutes`, which is 0 when `predicted == false` — and pairing that with a clock time made the false claim louder. The row (and its VoiceOver label) now says "Scheduled/not real-time" via a new `Formatters.deviationLabel(for:)`, the same gate `TripActivityPresenter.deviationLabel` and `DepartureTimeDisplay` already apply. `StopArrivalView`'s hand-rolled copy of the gate now delegates to it too.
2. **`.systemLarge` could clip whole rows.** The label's `fixedSize` forced full wrapped height even when seven rows didn't fit. Without it, labels still wrap where there's room and truncate under height pressure instead of pushing rows off the canvas.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved schedule deviation labels for early, late, scheduled, and stale trip data.
  * Clearly identifies trips that are scheduled only and not updated in real time.
  * Improved departure label sizing in widgets to support wrapping or truncation when space is limited.

### DIFF
--- a/OBAKit/Stops/StopArrivalView.swift
+++ b/OBAKit/Stops/StopArrivalView.swift
@@ -287,12 +287,7 @@ class StopArrivalView: UIView {
 
         accessibilityTimeLabel.text = formatters.timeFormatter.string(from: arrivalDeparture.arrivalDepartureDate)
 
-        if arrivalDeparture.scheduleStatus == .unknown {
-            accessibilityScheduleDeviationLabel.text = Strings.scheduledNotRealTime
-        }
-        else {
-            accessibilityScheduleDeviationLabel.text = formatters.formattedScheduleDeviation(for: arrivalDeparture)
-        }
+        accessibilityScheduleDeviationLabel.text = formatters.deviationLabel(for: arrivalDeparture)
 
         accessibilityScheduleDeviationLabel.textColor = formatters.colorForScheduleStatus(arrivalDeparture.scheduleStatus)
         accessibilityRelativeTimeBadge.configure(with: arrivalDeparture, formatters: formatters)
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is exactly the follow-up I was hoping for from #1237, and the central claim checks out. `ArrivalDeparture.scheduleStatus` opens with `guard predicted else { return .unknown }`, so `!predicted` and `scheduleStatus == .unknown` are the same condition — the `StopArrivalView` refactor is behavior-preserving, and routing both call sites through one helper means the gate can't drift between them. It also lines up with what `DepartureTimeDisplay` already keys off (`isRealTime: arrivalDeparture.predicted`).

The reasoning in the helper's doc comment is the part I most appreciate: a schedule-only trip has zero deviation *by definition*, so the raw phrase would assert "on time" when no vehicle has reported at all. Pairing that with a concrete clock time in the widget is exactly where a false real-time claim does the most damage.

No new localization work needed, which I checked — `common.scheduled_not_real_time` already lives in `OBAKitCore/Strings/Strings.swift:93` and is present across all 13 locales, and OBAKitCore's `OBALoc` resolves against the OBAKitCore bundle, so the widget extension picks it up for free.

The four tests are all load-bearing — each would fail if the gate were removed — and the fourth (predicted timestamps present but `predicted: false`) is the one I'd have asked for if it weren't there. Good call documenting the 2054 reference-date quirk inline rather than leaving a mysterious literal.

Two notes, neither blocking:

- The second paragraph of the `departureLabel` doc comment still says a single text run "wraps to the second line the way this label always has, where an `HStack` would truncate the deviation." With `fixedSize` gone that's now less absolute — under `.systemLarge` pressure this label can truncate too. The comparison against `HStack` still holds; the sentence just overstates the guarantee a little. Worth a trim next time you're in the file.
- The consolidation is 2-of-3: `StopArrivalItem.swift:202` still hand-rolls `scheduleStatus == .unknown`. It works on a view model rather than an `ArrivalDeparture`, so it can't call the new helper without a second overload — noting it only so the remaining copy isn't a surprise later.

Merging.

PR #1271, opened 2026-08-02
Live Activity: close the stale-score refresh race and preserve staleDate on promotion

diveshpatil9104:
## Summary

Implements the two follow-ups from the review of #1243.

1. **A refresh could silently undo a promotion.** `updateRunningLiveActivities` snapshotted `activity.content` on the main actor, then applied it later inside `Task.detached` — a promote/demote landing in between got overwritten with the stale score. The score is now read off the freshly re-fetched activity inside the task. This shrinks the window to the unavoidable read-before-commit gap inside one update (ActivityKit has no compare-and-swap), rather than the unbounded snapshot-to-task gap.
2. **Promotion cleared the push-set stale marker.** `promoteToDynamicIsland` rebuilt content with `staleDate: nil` while `demoteLivePeers` preserved it. Promotion is a score-only touch of unchanged content, so it now carries `staleDate` through too. The refresh path deliberately keeps `staleDate: nil` — it installs fresh local content, and carrying a possibly-past marker forward could flag it stale on arrival; that distinction is now documented at the call site.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Live Activity updates to preserve current relevance and prominence information.
  * Preserved expiration timing when promoting trips to the Dynamic Island.
  * Prevented stale or missing activity data during updates.

* **Tests**
  * Added regression coverage to verify that activity expiration dates and relevance scores are retained correctly.

### DIFF
--- a/OBAKit/Bookmarks/BookmarksViewController.swift
+++ b/OBAKit/Bookmarks/BookmarksViewController.swift
@@ -319,10 +319,6 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
                 // `update`. Re-fetch by ID inside a detached task instead — that copy
                 // never crosses an isolation boundary.
                 let activityID = activity.id
-                // Preserve prominence via the shared helper — rebuilding with the
-                // default score of 0 would hand the Island back to the first-
-                // started trip after every arrivals refresh (#1189 Problem 2).
-                let existingContent = activity.content
                 Task.detached {
                     guard let activity = Activity<TripAttributes>.activities.first(where: { $0.id == activityID }) else {
                         // The activity ended between the loop and this task; dropping
@@ -331,11 +327,22 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
                         Logger.info("Live Activity \(activityID) is no longer running; skipping update.")
                         return
                     }
+                    // Preserve prominence via the shared helper, reading the score
+                    // off this freshly re-fetched activity — not a snapshot taken
+                    // before the hop, which can predate a concurrent
+                    // promote/demote and would silently write the stale score
+                    // back, undoing it (#1243 review follow-up).
+                    //
+                    // `staleDate` stays nil here on purpose: unlike the
+                    // score-only promote/demote touches (which must carry the
+                    // push-set marker through), this update installs fresh local
+                    // content, and carrying a possibly-past stale-date forward
+                    // could mark it stale on arrival. The next push re-arms it.
                     await activity.update(
                         TripLiveActivityRelevance.contentPreservingRelevance(
                             state: contentState,
                             staleDate: nil,
-                            existing: existingContent
+                            existing: activity.content
                         )
                     )
                     Logger.info("Updated Live Activity for stop: \(staticData.stopID) route: \(staticData.routeShortName)")
--- a/OBAKitCore/LiveActivities/TripLiveActivityRelevance.swift
+++ b/OBAKitCore/LiveActivities/TripLiveActivityRelevance.swift
@@ -67,10 +67,14 @@ extension Activity where Attributes == TripAttributes {
             return
         }
         let prominence = TripLiveActivityRelevance.prominenceScore()
+        // A promotion only touches the score; the content is unchanged, so the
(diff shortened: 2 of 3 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The new regression test can't fail if the fix is reverted. `Content carries the supplied stale date through a score change` calls `TripLiveActivityRelevance.content(state:staleDate:relevanceScore:)` directly with a literal `staleDate` — but this PR doesn't change `content(...)`; it's an unmodified one-line pass-through to `ActivityContent.init`. Flip `promoteToDynamicIsland` back to `staleDate: nil` and this test still passes, so it doesn't pin the change it documents ("Pins the score-only touch pattern `promoteToDynamicIsland` and `demoteLivePeers` share"). It also only adds the non-nil arm of an assertion the test above it already makes (`#expect(content.staleDate == nil)` at L50). This is the same note from the #1243 review — "The actual fix could be deleted and this test would still pass" — landing on the same helper a second time. ActivityKit isn't injectable, so a real test likely needs a seam over the promote/demote content-rebuild step; short of that, the honest fix is to stop billing it as regression coverage in the doc comment and PR description.

https://github.com/OneBusAway/onebusaway-ios/blob/f5d9191e1381d7609534a309454283fe61bb35f6/OBAKitTests/Models/TripLiveActivityRelevanceTests.swift#L52-L71

aaronbrethorst (approved):
Both production changes are correct, and they're the right shape.

Moving the `activity.content` read inside the detached task genuinely narrows the window. Reading the score off the freshly re-fetched activity instead of a snapshot taken before the hop means a promote/demote that lands during the hop is no longer silently written back over. The residual read-to-update gap is real and unavoidable without a compare-and-swap that ActivityKit doesn't offer — and I'm glad you documented it at the call site rather than implying it was airtight.

Carrying `staleDate` through `promoteToDynamicIsland` is right for the reason you give: a promotion only touches the score, the content is unchanged, so the push-set marker should survive it. Matching `demoteLivePeers` closes a genuine asymmetry.

One thing to be straight about, since it's the second time it's come up. **The new test doesn't guard the fix.** `Content carries the supplied stale date through a score change` calls `TripLiveActivityRelevance.content(state:staleDate:relevanceScore:)` with a literal date — and that helper is untouched by this PR. Revert `promoteToDynamicIsland` back to `staleDate: nil` and the test stays green. Its doc comment says it "pins the score-only touch pattern `promoteToDynamicIsland` and `demoteLivePeers` share," and it doesn't pin either. That's the same note I left on #1243.

I'm merging anyway, and I want to explain why rather than have it look inconsistent. `promoteToDynamicIsland` is an extension on `Activity<TripAttributes>` — it needs a live ActivityKit activity, which a unit test process can't create. So the test that would actually guard this fix can't be written against the current seam. That's a fine answer. What I'd rather not have is a test whose comment claims coverage it doesn't provide: either drop it, or reword the comment to say it pins `content(state:staleDate:relevanceScore:)`'s pass-through and nothing more. If you want the real thing, extracting the score-only touch into a pure function over `(existingContent, newScore) -> ActivityContent` would give you something testable — worth considering, not required here.

Two follow-ups I noticed while reading, both pre-existing and out of scope. Please open issues rather than folding them in:

- `TripPageViewController.swift:317` starts a Live Activity with a bare `.init(state:staleDate:)`, so `relevanceScore` defaults to 0 and it never calls `demoteLivePeers` — a third start path that bypasses the prominence scheme entirely. That's the #1189 Problem 2 failure mode this type exists to prevent, and `TripLiveActivityRelevance`'s own doc says "Both start paths…" as if there were two.
- On the refresh path, `staleDate: nil` leaves a locally-refreshed card un-stale-able until the next push. Your comment correctly argues against carrying a possibly-past date forward, but `docs/LiveActivityPushNotifications.md:689` implies the alternative is a fresh `now + 2min`, not `nil`.

For the record on the CodeRabbit note asking you to serialize through "the existing shared per-activity non-reentrant update queue" — there is no such queue in this repo. Its underlying point about the residual window is real, but it's a redesign, not a defect you introduced here. Right to leave it.

PR #1272, opened 2026-08-04
Stop Details Sheet 

mosliem:
Replaces the stop details sheet's UIKit host with a native SwiftUI view, and extracts the
logic it shares with the pushed Stop page into testable pieces used by both surfaces.

## What changed

- **The sheet.** `StopDetailsSheetView` renders the stop details route directly in SwiftUI
instead of hosting `StopPageViewController`. It presents modally rather than pushing, so
the map stays put behind it.

- **Chrome.** Nav-bar buttons are replaced by a pinned top bar (Refresh, Close) and a row of
circular actions (Schedule, Filter, Bookmark, More) that stays beneath the map header as
the list scrolls. Both wear the same Liquid Glass treatment, with a `.regularMaterial`
fallback below iOS 26.

- **Header.** The sheet header carries a map thumbnail of the stop, rendered through a
snapshot loader, and collapses as the list scrolls — the pinned title fades in as the card
scrolls away.

- **Scroll to top.** A floating glass button appears once the rider has scrolled more than
one viewport and returns the list to the top. There is no separate cell-count gate: a short
list cannot scroll a full viewport, so distance is its own content gate — and it adapts to
Dynamic Type, expanded trip panels, and survey/donation cards for free.

- **Modal flows.** Everything the sheet presents on top of itself — the schedule, bookmark
editor, route filter, nearby stops, service alerts, report-a-problem, walking directions,
surveys, donations, the alarm picker, permission-denied alerts, live activities, and error
alerts — routes through one presenter rather than being scattered across the view. It
handles push-versus-present itself, so the same calls work from the sheet and from the
pushed page.

- **Idle timer.** The "keep the screen awake while a self-refreshing screen is visible"
behaviour becomes a `keepsScreenAwake()` view modifier — the SwiftUI counterpart of
`Idleable`. `CurrentTripView` drops its hand-rolled copy and adopts it. The per-screen
`wasDisabledByUs` latch is replaced by `ScreenAwakeCoordinator`, which counts leases keyed by
owner: a latch cannot survive two live screens overlapping (SwiftUI runs the incoming
`onAppear` before the outgoing `onDisappear`), and `Idleable` now routes through the same
coordinator so the UIKit and SwiftUI halves cannot fight over one global.

- **UIKit shrinks.** `StopPageViewController` goes from 1,024 lines to 641 as its
responsibilities move into the shared types, and `StopDetailSheetHost` — the UIKit hosting
shim the sheet used to need — is deleted outright.

- **Shared extractions.** Logic the sheet and the pushed page both need now lives in one
place, and is unit-testable for the first time:

| Type | Responsibility |
| --- | --- |
| `StopPageContent` | Derives the displayed list from the view model |
| `StopPageActionPresenter` | The modal flows (bookmarks, alarms, share, errors) |
| `StopDeparturesSections` | The departure sections both surfaces render |
| `StopPageLifecycleModifier` | Lifecycle wiring, applied once |
| `StopSheetHeaderCollapse` | Header collapse arithmetic |
| `ScrollToTopVisibility` | When the scroll-to-top button shows |

**Bug fixes along the way.** 
A main-thread hang on scroll (see below), an alarm failure
alert that left an undismissable layer behind, an alarm success message clobbered by its own
teardown, a refresh spinner too brief to see, an action row whose disabled state reset while
scrolling, and a `fillsPage` regression on the pushed page.

## Testing

- **5 new test suites**, all Swift Testing: `StopPageContentTests`,
  `StopPageActionPresenterTests`, `StopSheetHeaderCollapseTests`,
  `StopPageActionRowStateTests`, `ScrollToTopVisibilityTests`
- Full suite: **2009 tests, 2008 passing**. The one failure,
  `RentalFormatTests.rangeFallbackUsesAbbreviatedUnits()`, is pre-existing on `main` and
  unrelated

## Two carried commits, not mine

The final commit fixes two build breaks that are pre-existing on `main` and unrelated to this
work — Xcode 26 will not build the branch without them:

- `TripPageViewController.swift` was missing an `import CoreLocation`
- `StopRouteFocusModelTests.swift`'s `StubMapDeparture` needed `nonisolated` to conform under
  main-actor default isolation

Both are one-liners; happy to split them into their own PR if you'd rather they land
separately.
  
  ## Video 
  
https://github.com/user-attachments/assets/282fddeb-f541-4462-8f3f-d1fe3d15c403

## Summary by CodeRabbit

* **New Features**
  * Introduced a native stop-details sheet with collapsible headers, pinned controls, refresh, filtering, sorting, alerts, bookmarks, schedules, and departure details.
  * Added a floating scroll-to-top button after scrolling beyond one viewport.
  * Keeps the screen awake while viewing stop details and refreshes content when returning to the foreground.
* **Bug Fixes**
  * Improved alarm creation feedback and stop-action presentation reliability.
  * Prevented background interaction while the stop-details sheet is fully expanded.
* **Documentation**
  * Added design and implementation documentation for the stop-details experience.


OBAKit/Stops/StopPage/StopPageViewController.swift
```diff
@@ -48,19 +50,6 @@ class StopPageViewController: UIHostingController<StopPageRootView>,
...
-    /// first `locationMenu()` build, by which point `viewModel.stop` is set.
-    private lazy var googleMapsAvailable: Bool = {
-        guard let coordinate = viewModel.stop?.coordinate,
-              let url = AppInterop.googleMapsWalkingDirectionsURL(coordinate: coordinate)
-        else { return false }
-        return application.canOpenURL(url)
-    }()
-    #endif
-
     var bookmarkContext: Bookmark? {
```
aaronbrethorst, line 71: **Device builds are broken.** This hunk deletes `private lazy var googleMapsAvailable` (moved to `StopPageActionPresenter.googleMapsAvailable(coordinate:)`, which is `private` and on a different type), but `locationMenu()` still references it at line 472:

```swift
if let googleMapsURL = AppInterop.googleMapsWalkingDirectionsURL(coordinate: stop.coordinate),
   self.googleMapsAvailable {
```

That reference sits inside `#if !targetEnvironment(simulator)`, so the simulator build — and therefore CI — compiles fine. `xcodebuild archive`, TestFlight, and anyone running on hardware will fail with *"value of type 'StopPageViewController' has no member 'googleMapsAvailable'"*.

OBAKit/Sheet/DI/AppSheetViewFactory.swift
```diff
@@ -77,11 +95,24 @@ final class AppSheetViewFactory {
...
+    ///
+    /// It renders the same departures as the pushed screen — through the shared
+    /// `StopDeparturesSections` — but replaces the navigation bar with a pinned
+    /// Refresh/Close strip and a row of circular actions, and collapses the map
+    /// header away as the list scrolls so the actions stay reachable.
+    func stopDetailView(stopID: Stop.ID) -> StopDetailsSheetView {
+        StopDetailsSheetView(
+            stopID: stopID,
+            viewModel: StopViewModel(environment: self.application, stopID: stopID),
+            presenter: StopPageActionPresenter(
```
aaronbrethorst: **The presenter is rebuilt on every re-render, and every modal it presents holds it weakly.**

`viewModel` survives re-evaluation via `@StateObject` + the `@autoclosure`, but `presenter` is a plain `let` on the view — so each time `MapPanelRootView.body` re-evaluates (a map camera update, an annotation change, the detent binding writing back to `SheetCoordinator`), `sheetContent(route)` → `factory.view(for:)` → `stopDetailView(stopID:)` builds a **new** `StopPageActionPresenter` and drops the old one. Nothing else retains it.

`StopPreferencesWrappedView.delegate`, `AddBookmarkViewController.delegate`, `EditBookmarkViewController.delegate` and `AlarmBuilder.delegate` are all `weak`, so any modal that outlives one re-render loses its callbacks:

- Filter → route picker opens → map updates behind it → **Save applies nothing**, routes stay unfiltered.
- Bookmark → `AddBookmarkViewController.cancel()` is only `delegate?.bookmarkEditorCancelled(self)`, so **Cancel is a dead button** and the modal is stuck on screen.
- Alarm bulletin → the created alarm is never indexed and no "Alarm created" confirmation appears.

The presenter needs the same lifetime treatment as the view model — hold it in `@StateObject`/`@State` on `StopDetailsSheetView`, or have the factory vend a single instance per stop.

OBAKit/Stops/StopPage/StopPageActionPresenter.swift
```diff
@@ -0,0 +1,697 @@
...
+        // reporting success always has one — unless the view model has since
+        // been released, in which case there is nothing left to record on.
+        if let flow = alarmFlow, let viewModel = flow.viewModel {
+            // `replaceAlarm` indexes the new alarm synchronously and no-ops the
+            // delete when the departure had no prior alarm, so it serves both
+            // the create and change flows.
+            Task { await viewModel.replaceAlarm(with: alarm, for: flow.departure) }
+
+            if alarmBuilder.trackOnLockScreen {
+                startLiveActivity(for: flow.departure, viewModel: viewModel)
```
aaronbrethorst: **The success path has the bulletin-dismissal hazard you guarded against on the failure path.**

`AlarmBuilder.createAlarm` calls `delegate?.alarmBuilder(self, alarmCreated:)` *before* its `defer` dismisses the bulletin. With "Track on Lock Screen" on, this immediately calls `startLiveActivity(for:viewModel:)`; if `Activity.request` throws (activities disabled mid-flow, quota, push-token failure), `presentationHost` resolves to the still-presented bulletin and the error alert is presented on it. The `defer` then tears that bulletin down — the alert either flashes and vanishes, or the two transitions collide and leave the bulletin's view on screen and unresponsive.

`waitForBulletinDismissal` below already solves exactly this for `alarmBuilder(_:error:)`; the success path needs the same treatment before it can present anything.

OBAKit/Sheet/Content/CurrentTrip/CurrentTripView.swift
```diff
@@ -94,20 +89,7 @@ struct CurrentTripView: View {
...
-        guard !UIApplication.shared.isIdleTimerDisabled else { return }
-        UIApplication.shared.isIdleTimerDisabled = true
-        wasIdleTimerDisabledByUs = true
-    }
-
-    private func reEnableIdleTimer() {
-        guard wasIdleTimerDisabledByUs else { return }
-        UIApplication.shared.isIdleTimerDisabled = false
-        wasIdleTimerDisabledByUs = false
+        .keepsScreenAwake()
```
aaronbrethorst, line 78: **Adopting `.keepsScreenAwake()` silently drops the scene-phase re-arm.**

The deleted `disableIdleTimer()`/`reEnableIdleTimer()` pair was driven from `.onChange(of: scenePhase)` as well as appear/disappear: `.background` released the hold and the `.background → .active` edge re-took it — which also restarted the 10-minute failsafe. `KeepsScreenAwakeModifier` only acquires in `onAppear` and releases in `onDisappear`, neither of which fires when the app backgrounds with the sheet mounted, and `ScreenAwakeCoordinator.armFailsafe()` is called only from `acquire()`.

So: rider on the Current Trip sheet pockets the phone for >10 minutes and comes back. The failsafe `Timer` whose fire date has passed fires on the first run-loop pass after resume, `expireFailsafe()` restores `stateBeforeFirstLease` (false), and since the lease count never hit zero no `acquire()` follows — the live trip screen dims and auto-locks mid-trip, and stays sleep-able for the rest of that sheet's life.

The modifier should take a scene-phase parameter (or observe it itself) and re-acquire on the `.background → .active` edge.

OBAKit/Stops/StopPage/StopPageActionPresenter.swift
```diff
@@ -0,0 +1,697 @@
...
+        guard
+            application.donationsManager.donationsEnabled,
+            let donationModel = application.donationsManager.buildObservableDonationModel()
+        else { return }
+
+        let view = DonationLearnMoreView()
+            .environmentObject(donationModel)
+            .environmentObject(AnalyticsModel(application.analytics))
+
+        presentationHost(for: "donation")?.present(UIHostingController(rootView: view), animated: true)
```
aaronbrethorst: **This bypasses `PromptCoordinator`.** The code being replaced (`StopPageViewController.showDonationUI`) called `presentDonationModal(learnMoreView, coordinator: application.promptCoordinator)`, whose whole reason for existing is that all three donation call sites charge `coordinator.noteShown(.donationModal)` *in the presentation completion* — its doc comment spells out the ordering hazard. With the bare `present`, a rider shown the donation modal from the Stop page can still be hit with the App Store review prompt in the same session: two interruptions where the budget allows one. `MoreViewController` and `Application` still route through the helper, so the Stop page is now the odd one out.

```suggestion
        presentationHost(for: "donation")?.presentDonationModal(view, coordinator: application.promptCoordinator)
```

OBAKit/Sheet/Content/Stop/Details/StopDetailsSheetView.swift
```diff
@@ -0,0 +1,428 @@
...
+    @State private var userActivity: NSUserActivity?
+    /// Gates the one-shot success haptic to the first arrivals load, matching
+    /// `StopViewController.bindArrivalsSink()`; later refreshes are silent.
+    @State private var firstLoad = true
+    /// Drives the top bar's spinner. Distinct from `viewModel.isLoading`, which
+    /// also flips for the ~15s background refresh — the rider did not ask for
+    /// that one, so it should not flash chrome at them.
+    @State private var isManuallyRefreshing = false
+
+    @AppStorage("StopViewController.pastDeparturesCollapsed") private var pastCollapsed = true
```
aaronbrethorst, line 129: **This `@AppStorage` binds to `UserDefaults.standard`, not the app-group suite.**

`.defaultAppStorage(_:)` only affects `AppStorage` instances in views *below* the modified view; a view's own property wrappers resolve against the environment it was handed. This view declares `pastCollapsed` here and then applies `.defaultAppStorage(userDefaults)` at line 152 — inside its own `sheetBody`, i.e. to its descendants only.

`StopPageRootView` exists purely to work around this ("Its only job is to apply `.defaultAppStorage` so the page's `@AppStorage` reads and writes the app-group suite"), and the sheet needs the same wrapper. As written, expanding past departures on the map sheet doesn't carry over to the pushed Stop page (and vice versa), and the value lands in `UserDefaults.standard`.

OBAKit/Controls/SwiftUI/KeepsScreenAwake.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    private init() {}
+
+    /// Test seam: the lease count, so the pairing can be asserted without
+    /// driving a real view hierarchy.
+    var activeLeases: Int { leaseCount }
+
+    func acquire() {
+        if leaseCount == 0 {
+            stateBeforeFirstLease = UIApplication.shared.isIdleTimerDisabled
```
aaronbrethorst, line 125: **This is a second idle-timer arbiter that doesn't coordinate with the first.**

`Idleable.enableIdleTimer()` (`OBAKit/Extensions/ViewController.swift`) writes `application.isIdleTimerDisabled = false` unconditionally — the same global the coordinator drives. The new sheet creates the collision itself: the stop sheet holds a lease (`stateBeforeFirstLease = false`), the rider taps "View Full Trip", `StopPageActionPresenter.showTrip` presents an `Idleable` `TripViewController`, and dismissing it runs `enableIdleTimer()` → `false`. The coordinator never re-asserts, so the display sleeps while the rider is still watching live departures — the exact failure the lease count was introduced to prevent, just across the UIKit/SwiftUI boundary.

The reverse ordering is worse: pushed `StopPageViewController` (Idleable, timer disabled) → open the SwiftUI sheet (lease records `stateBeforeFirstLease = true`) → pop the pushed page (`enableIdleTimer()` → false) → close the sheet (last release restores `true`). The idle timer is now disabled with nothing holding it and no failsafe armed — the display never sleeps for the rest of the session.

The fix is for `Idleable.disableIdleTimer()`/`enableIdleTimer()` to acquire/release the same lease instead of writing the global directly.

OBAKit/Sheet/Content/Stop/Details/StopDetailsSheetView.swift
```diff
@@ -0,0 +1,428 @@
...
+    ) {
+        self.stopID = stopID
+        _viewModel = StateObject(wrappedValue: viewModel())
+        self.presenter = presenter
+        self.feedback = feedback
+        self.formatters = formatters
+        self.userDefaults = userDefaults
+    }
+
+    private var navigation: StopPageNavigationHandler {
```
aaronbrethorst: `navigation` is computed, so every access rebuilds the whole `StopPageNavigationHandler` — `TripClosures`, `AlarmClosures`, `StopClosures` and a 17-field struct. `actionRowOverlay` touches it seven times, `headerRows` once, `departuresBuilder` once: ~10 full rebuilds per body evaluation, and `StopViewModel` republishes on the ~15s refresh *and* the per-second status timer. Every closure identity changes on every pass, so nothing downstream can be diffed as equal.

The pushed presentation already gets this right — `StopPageView.navigation` is a stored `let` built once by the hosting controller. Build it once in `init` alongside `presenter`.

OBAKit/Sheet/Content/Stop/Details/StopDetailsSheetView.swift
```diff
@@ -0,0 +1,428 @@
...
+    private func headerRows(showsLoadingState: Bool) -> some View {
+        Section {
+            Group {
+                if let stop = viewModel.stop {
+                    StopPageHeaderView(
+                        stop: stop,
+                        walkTime: viewModel.walkTime,
+                        statusText: viewModel.statusText,
+                        snapshotLoader: { size in
+                            await presenter.loadSnapshot(stop: stop, size: size, traitCollection: UITraitCollection.current)
```
aaronbrethorst: `UITraitCollection.current` is only guaranteed to be the view's traits inside `traitCollectionDidChange`/`layoutSubviews`/`draw(_:)`. `StopPageHeaderView` invokes `snapshotLoader` from `.task(id: cardWidth)`, which runs outside any UIKit update pass, so this can read a default trait collection — handing `MapSnapshotter` a `displayScale` of 1.0 on a 3x device and rendering a visibly soft map card. The pushed path passes the hosting controller's own `traitCollection` and stays sharp.

Derive the traits from `@Environment(\.displayScale)` (plus the size class you need), or thread the host's trait collection through the presenter as the pushed path does.

OBAKit/Sheet/Content/Stop/Details/StopDetailsSheetView.swift
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    private func showAlarmPermissionDenied() {
+        presenter.showAlarmPermissionDeniedAlert {
+            // Reset so a later already-denied attempt re-fires the binding.
+            viewModel.clearAlarmPermissionDenied()
+        }
+    }
+
+    private func handle(scenePhase previous: ScenePhase, to phase: ScenePhase) {
```
aaronbrethorst: This switch — comment included — is a verbatim copy of the `.onChange(of: scenePhase)` block in `CurrentTripView`, which this same PR is editing. The PR factors the *idle-timer* half of that block out into a shared `keepsScreenAwake()` modifier and then copy-pastes the other half here. The next change to scene-phase policy (say, also deactivating on `.inactive` so a refresh doesn't fire behind the app switcher) has to be found in two places, and the two live screens will drift.

A `refreshesOnForeground(viewModel:)` modifier next to `stopPageLifecycle` would carry it once — and it's where the missing idle-timer re-arm from the `CurrentTripView` comment belongs too.

aaronbrethorst (changes requested):
Reviewed at high effort against `main`. Requesting changes — the branch has merge conflicts to resolve regardless, and there are a few real defects worth folding into the same pass.

The architecture here is good: extracting `StopPageContent`, `StopDeparturesSections`, `StopPageActionPresenter` and the lifecycle modifier out of the 993-line view controller is exactly the right move, and the feedback-loop analysis behind the overlay-not-inset chrome is well reasoned and well documented.

**Blocking:**

1. **Device builds don't compile** — `locationMenu()` still references the deleted `googleMapsAvailable`; it's inside `#if !targetEnvironment(simulator)`, so CI stays green and the archive fails.
2. **`StopPageActionPresenter` is rebuilt on every sheet re-render** while every modal it presents holds it weakly — Save on the route filter, Cancel on the bookmark editor, and the alarm confirmation all become dead once the map updates behind the modal.
3. **`@AppStorage` on the sheet binds to `UserDefaults.standard`** rather than the app-group suite; `.defaultAppStorage` applied inside the same view's body doesn't reach the view's own wrappers. That's what `StopPageRootView` exists for.
4. **The donation modal bypasses `PromptCoordinator`**, so the Stop page can now show a donation modal *and* a review prompt in one session.
5. **`keepsScreenAwake()` drops `CurrentTripView`'s scene-phase re-arm**, and the coordinator doesn't coordinate with `Idleable` — both leave the idle timer in the wrong state in ordinary flows.

Details inline. The remaining comments are non-blocking cleanups, but the `navigation` rebuild is worth doing before merge since it's a one-line change on a per-second-refreshing screen.

Also worth deciding separately (not blocking): with `StopDetailsSheetView` landing, there are now two independent sheet chromes — this one and `StopPageView`'s `showToolbarOnBottom` mode, still live via `MapViewController` — sharing the departures sections but not the header or toolbar. Fine if the FloatingPanel path is on its way out; a maintenance tax if it isn't.

OBAKit/Stops/StopPage/StopPageViewController.swift
mosliem: Fixed in a2cfc24f. `locationMenu()` now calls `actionPresenter.googleMapsAvailable(coordinate:)` — the memoized lookup that moved with the rest of the flow — and that method drops its `private`.

Verified with an actual device compile rather than by inspection, since the simulator build never takes this path:

```
xcodebuild build -scheme App -sdk iphoneos -destination 'generic/platform=iOS' \
  CODE_SIGNING_ALLOWED=NO
** BUILD SUCCEEDED **
```

OBAKit/Sheet/DI/AppSheetViewFactory.swift
mosliem: Fixed in aec5c830. `presenter` is now `@State` behind an `@autoclosure`, the same treatment `viewModel` gets, so it survives `MapPanelRootView.body` re-evaluating:

```swift
@State private var presenter: StopPageActionPresenter

init(..., presenter: @autoclosure () -> StopPageActionPresenter, ...) {
    _presenter = State(wrappedValue: presenter())
}
```

The factory correspondingly vends closures rather than built instances, so neither the view model nor the presenter is constructed on a re-render that will discard it.

OBAKit/Stops/StopPage/StopPageActionPresenter.swift
mosliem: Fixed in fc0c06f1. The success path now goes through the same gate the failure path does:

```swift
if alarmBuilder.trackOnLockScreen {
    let departure = flow.departure
    Task { @MainActor in
        await waitForBulletinDismissal(alarmBuilder)
        startLiveActivity(for: departure, viewModel: viewModel)
    }
}
```

`replaceAlarm` and the "Alarm created" HUD still run synchronously — neither presents on the bulletin, so neither needs to wait.

OBAKit/Sheet/Content/CurrentTrip/CurrentTripView.swift
mosliem: Fixed in 55c24d1c, though not quite where you suggested — flagging that, since you pointed at `refreshesOnForeground` in the other comment.

The re-arm went into `KeepsScreenAwakeModifier` itself:

```swift
.onChange(of: scenePhase) { _, phase in
    switch phase {
    case .active:   acquire()   // no-op unless a backgrounding released it
    case .background: release()
    case .inactive: break       // Control Center / banner: the screen is still the rider's
    @unknown default: break
    }
}
```

Reasoning: putting it in `refreshesOnForeground` would only cover screens that *also* refresh on foreground. A screen that adopts `keepsScreenAwake()` alone would keep the bug, and nothing about that combination is enforced. Inside the modifier the guarantee travels with the lease. Happy to move it if you'd rather the two halves sit together.

`ScreenAwakeCoordinatorTests` covers the lease arithmetic; the phase transitions themselves aren't asserted, since driving `scenePhase` needs a hosted view.

OBAKit/Stops/StopPage/StopPageActionPresenter.swift
mosliem: Fixed in 0f598900 — took the suggestion as written, so the Stop page charges the budget in the presentation completion like the other two call sites.

OBAKit/Sheet/Content/Stop/Details/StopDetailsSheetView.swift
mosliem: Fixed in aec5c830. Added `StopDetailsSheetRootView`, the sheet's counterpart to `StopPageRootView`: it applies `.defaultAppStorage(userDefaults)` from *outside* the sheet, and the factory now vends it. The `.defaultAppStorage` call inside `sheetBody` is gone.

You were right about the mechanism — `pastCollapsed` was landing in `UserDefaults.standard` and not carrying over to the pushed page.

OBAKit/Controls/SwiftUI/KeepsScreenAwake.swift
mosliem: Fixed in 55c24d1c, as you described — `Idleable` no longer writes `isIdleTimerDisabled` at all:

```swift
func disableIdleTimer() {
    ScreenAwakeCoordinator.shared.acquire(owner: ObjectIdentifier(self))
}

nonisolated func enableIdleTimer() {
    let owner = ObjectIdentifier(self)   // captured before the hop, so a deallocated
    Task { @MainActor in                 // controller can't strand its lease
        ScreenAwakeCoordinator.shared.release(owner: owner)
    }
}
```

The owner-keyed pair is new. The raw counting API can't be idempotent, and UIKit doesn't guarantee `viewWillAppear`/`viewWillDisappear` alternate strictly, so a repeat appear would have leaked a lease and stranded the timer for the session. `acquire(owner:)` is a no-op if that owner already holds one; `release(owner:)` is a no-op if it doesn't.

Four tests cover it, including both orderings from your comment:

- `An Idleable screen leaving keeps a SwiftUI screen's hold`
- `A pushed screen popped under a sheet does not strand the timer`
- `A repeated acquire for one owner takes a single lease`
- `Releasing an owner that holds no lease is a no-op`

`idleTimerFailsafe` is left on the protocol and its five conformers. The coordinator owns the failsafe now so nothing writes it, but deleting the requirement meant touching four view controllers this PR otherwise doesn't, which felt like the wrong thing to fold in here. It carries a two-line note saying it's unused; happy to remove it in a follow-up.

OBAKit/Sheet/Content/Stop/Details/StopDetailsSheetView.swift
mosliem: Fixed in aec5c830. `navigation` became `makeNavigation()`, called once at the top of `sheetBody` and threaded into `list(content:navigation:)`, `headerRows(showsLoadingState:navigation:)`, `departuresBuilder(navigation:)` and `actionRowOverlay(navigation:)` — one build per body evaluation instead of ten.

OBAKit/Sheet/Content/Stop/Details/StopDetailsSheetView.swift
mosliem: Fixed in aec5c830. The sheet reads `@Environment(\.displayScale)` and builds the traits from it:

```swift
private var snapshotTraits: UITraitCollection {
    UITraitCollection { traits in
        traits.displayScale = displayScale
    }
}
```

Only `displayScale` is derived — `loadSnapshot` already forces `.dark` because the header is always-dark by design, so the interface style would be overwritten anyway.

OBAKit/Sheet/Content/Stop/Details/StopDetailsSheetView.swift
mosliem: Fixed in aaec1aad. The switch is now one modifier in `OBAKit/Controls/SwiftUI/RefreshesOnForeground.swift`, used by both live screens:

```swift
.refreshesOnForeground(
    onForeground: { Task { await viewModel.start() } },
    onBackground: { viewModel.deactivate() }
)
```

Closures rather than a `viewModel:` parameter, because the two view models don't share a protocol and their `start()` differs (`CurrentTripViewModel`'s is synchronous, `StopViewModel`'s is `async`). Both the `previous == .background` rule and the `.inactive` no-op live in the modifier now.

It sits next to `KeepsScreenAwake.swift` rather than `stopPageLifecycle`, since `CurrentTripView` isn't a stop page. The idle-timer re-arm went into `keepsScreenAwake()` instead of here — reasoning in the `CurrentTripView` thread.

mosliem:
Rebased onto `main` (`613497d9`) and worked through the review. Details are in the threads; this is the shape of it.

## Rebase

Four of the 38 commits needed real resolution rather than textual merging, because `main` had meanwhile removed the departure accordion, added the Departure Type filter (#1096), replaced the mode toggle with `StopPageListHeaderRow`, and introduced `TripPageViewController`. I folded that into the extracted types rather than reverting it:

- `StopPageContent` gained `routeVisibleDepartures`, `filteringImplausibleDates()`, the type filter and `isDepartureFilterEmpty`; `StopDeparturesSections` and `StopDeparturesBuilder` lost the accordion and gained `onSelectDeparture` / `onShowAllDepartureTypes`.
- `StopPageActionPresenter.showTrip` was still calling the old `viewRouter.navigateTo(arrivalDeparture:)`. It now builds a `TripPageViewController` with `originTitle`, owns `onTripPagePush`, and pushes where a nav stack exists / presents modally from the sheet.
- Trip sharing moved into the presenter too (108642f1) — the pushed VC owned the only `TripSharingCoordinator`, so the sheet's row menu had a dead Share action.

## Blocking items

| | Commit |
|---|---|
| Device build (`googleMapsAvailable`) | a2cfc24f |
| Presenter rebuilt every re-render | aec5c830 |
| `@AppStorage` on the wrong suite | aec5c830 |
| Donation bypassing `PromptCoordinator` | 0f598900 |
| Idle timer: scene-phase re-arm + `Idleable` collision | 55c24d1c, aaec1aad |

Non-blocking: `navigation` built once (aec5c830), the alarm success path waits for the bulletin (fc0c06f1), snapshot traits from `@Environment(\.displayScale)` (aec5c830), and the scene-phase switch shared through one modifier (aaec1aad).

Two deviations worth your call, both explained in-thread: the idle-timer re-arm went into `keepsScreenAwake()` rather than `refreshesOnForeground`, and `idleTimerFailsafe` is left on the protocol rather than deleted across five view controllers.

## Verification

- Simulator test build: **succeeded**
- Device build (`-sdk iphoneos -destination 'generic/platform=iOS'`): **succeeded** — the path that would have caught the `googleMapsAvailable` break
- `OBAKitTests`: 2005 tests, one failure (below)
- SwiftLint: clean on every file touched

## Two things that aren't this PR's

`main` does not build or test clean on Xcode 26 independently of this branch. I verified both in a clean worktree at `613497d9` before touching anything:

- `TripPageViewController.swift` is missing `import CoreLocation` — `cannot find type 'CLLocationCoordinate2D' in scope`.
- `StubMapDeparture` in `StopRouteFocusModelTests.swift` needs `nonisolated` — `ConformanceIsolation`, which this repo escalates to an error.

Both are in 855af8d3, isolated so it can be dropped if you'd rather they go up separately. Nothing else depends on it.

The one test failure, `RentalFormatTests.rangeFallbackUsesAbbreviatedUnits`, also reproduces on unmodified `main` — `MKDistanceFormatter`'s spelled-out style now returns `"5.5 km"` too, so the test's inequality no longer holds. Untouched by this branch, so I left it alone.

## Still open

Your note about the two sheet chromes — this one and `StopPageView`'s `showToolbarOnBottom` mode, still live via `MapViewController` — is unaddressed, since you flagged it as a separate decision. Happy to take it either way.

aaronbrethorst:
### Code review

Found 1 issue:

1. The sheet's Filter menu drops the Departure Type filter, which the surface it replaces offered. `StopPageActionRow.filterMenu` carries only All Routes / Filtered Routes; nothing in the sheet's chrome calls `updateArrivalDepartureFilter`, and `moreMenu` doesn't carry it either. Before this PR, `AppSheetRoute.stopDetails` resolved to `StopDetailSheetHost` → `StopPageViewController` with `showToolbarOnBottom == false`, so `configureBarButtons()` installed a filter bar button whose menu is `UIMenu(children: [routesSection, departureFilterMenu()])`. Both surviving Stop page presentations still expose it — the pushed page through `departureFilterMenu()`, the FloatingPanel sheet through `StopPageToolbar`'s `onSetDepartureFilter`. `StopPageContent` still applies the filter (`.filter(by: arrivalDepartureFilter)`), and `Application.effectiveArrivalDepartureFilter` gives it a global default from Settings, so a rider whose default isn't `.all` opens the map sheet onto a silently filtered list with no in-sheet way to widen it — only the empty state's "show all departure types" escape hatch, which appears only once the filter has emptied the list entirely. Related: `StopPageActionRowState.isFilterOn` is `hasHiddenRoutes && isListFiltered`, where the bar button it replaces fills its glyph on `... || arrivalDepartureFilter != .all`, so the sheet's Filter glyph also reads "off" while that filter is holding rows back.

https://github.com/OneBusAway/onebusaway-ios/blob/322b3a49f6901a63bb92d16f5d6567f24fc520b4/OBAKit/Sheet/Content/Stop/Details/StopPageActionRow.swift#L156-L178

aaronbrethorst (changes requested):
This is strong work, and every blocking point from my last review is genuinely resolved — I checked each rather than assuming. The presenter is a `@StateObject` with an autoclosure initializer so it survives re-renders; `googleMapsAvailable` moved onto the presenter and the `#if !targetEnvironment(simulator)` path calls it; `StopDetailsSheetRootView` applies `.defaultAppStorage(userDefaults)` the same way `StopPageRootView` does; donation routes through `PromptCoordinator`; and the two competing idle-timer arbiters are now one lease-counting `ScreenAwakeCoordinator` with `idleTimerFailsafe` removed from all five conformers and real millisecond-expiry coverage behind it. Shrinking `StopPageViewController` by nearly 400 lines while extracting genuinely testable pieces is the good kind of refactor.

One gap holds it up, and it's contained.

The Departure Type filter has no control on the new sheet. `StopPageActionRow.filterMenu` carries only All Routes / Filtered Routes, and `moreMenu` doesn't carry it either, so nothing in the sheet's chrome calls `updateArrivalDepartureFilter` — the only call is the `onShowAllDepartureTypes` escape hatch in the empty state. The surface this replaces did expose it: `configureBarButtons()` in non-toolbar mode, which is the mode `StopDetailSheetHost` used, installs `UIMenu(children: [routesSection, departureFilterMenu()])`. Both other presentations still have it.

That matters because `StopPageContent` still *applies* the filter, and `effectiveArrivalDepartureFilter` gives it a Settings-level default. So a rider whose default is anything other than `.all` gets a silently filtered sheet with no way to change it from the sheet.

There's a corollary worth fixing in the same pass: `StopPageActionRowState.isFilterOn` omits `arrivalDepartureFilter != .all`, which the bar-button state it replaces includes. So even the indicator glyph reads "off" while that filter is active.

While you're in the filter menu: `filterChoice` marks the active option with `.accessibilityAddTraits(.isSelected)` only. The UIKit `filterMenu()` it ports sets a checkmark image, so a sighted user currently sees two identical plain rows in what's meant to be a mode selection.

Add the departure-type section to the Filter menu and fold that condition into `isFilterOn`, and I think this is ready. Send it back when it's in.

mosliem:
All three are in, in `8b5aa4c5`.

**Departure Type on the sheet.** `filterMenu` now carries the same two sections `filterMenu()` does — the route choices, then a "Departure Type" submenu over `ArrivalDepartureFilter.allCases` — wired through a new `onSetDepartureFilter` closure to `viewModel.updateArrivalDepartureFilter`.

One deviation worth calling out: the Filter button lost its `.disabled(!state.canFilter)`. Departure Type applies to any stop, including a single-route one, so gating the whole menu on `routeCount > 1` would have left the new control unreachable exactly where the bar button offered it. `canFilter` now gates only the route section inside the menu, which is what `filterMenu()` effectively does too — it drops "Filtered Routes" and keeps the submenu.

**`isFilterOn`.** Split in two, because the one property was doing two jobs:

- `isRouteFilterOn` — `hasHiddenRoutes && isListFiltered`, which checks one of the two route choices.
- `isFilterOn` — `isRouteFilterOn || departureFilter != .all`, which drives the glyph and the VoiceOver value, matching `configureBarButtons()`.

Folding the condition into the single property would have checked "Filtered Routes" whenever the departure filter was active on a stop with no hidden routes, so the checkmark needed the narrower one. Three tests cover it: the fill for a non-default departure type, and that the route choice doesn't move with it.

**Checkmarks.** `filterChoice` renders `Label(title, systemImage: "checkmark")` when selected and plain `Text` otherwise — a `Label` with a blank image on the unselected row indents its title away from its neighbour. Same fix in `StopPageToolbar.filterChoice`, since its own Departure Type menu already had the checkmark and the two halves of that menu disagreed.

Verification: test build on iPhone 16 / iOS 26.5 succeeded, SwiftLint clean on the four touched files, and `OBAKitTests` is 2027 tests with the one pre-existing `RentalFormatTests.rangeFallbackUsesAbbreviatedUnits` failure that also reproduces on `main`.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
All three items are fixed, and one of them is fixed better than I asked for.

**Departure Type is back on the sheet.** `filterMenu` now emits the route choices in a `Section` plus a `departureFilterMenu` submenu over `ArrivalDepartureFilter.allCases`, wired through `onSetDepartureFilter` to `viewModel.updateArrivalDepartureFilter`, which persists app-wide the same way the pushed page does. Importantly you also dropped `.disabled(!state.canFilter)` from the Filter button and moved `canFilter` to gate only the route section — so a single-route or still-loading stop can still reach Departure Type. That was the actual trap: gating the whole button would have left the same hole for a different reason.

**`isFilterOn` now includes the departure filter** — and rather than just appending the condition, you split out `isRouteFilterOn` so the route-choice checkmarks read from the narrower value. That's the right decomposition. A naive fix would have made "Filtered Routes" show a checkmark whenever the departure filter was active, which would have been a new bug in the fix for the old one. I checked it against `configureBarButtons()` and the semantics match. The new tests cover both directions, including a regression test that the departure filter doesn't move `isRouteFilterOn`.

**`menuChoiceLabel` gives sighted users a real indicator.** `Label(title, systemImage: "checkmark")` when selected, plain `Text` otherwise, with `.accessibilityAddTraits(.isSelected)` kept alongside. Applying it to `StopPageToolbar.filterChoice` too — which had the identical gap and which I hadn't called out — is the kind of thing I notice and appreciate.

For the record on style: a checkmark `Label` is the correct SwiftUI analogue here. The sibling PR #1284 moved the *UIKit* menus from checkmark images to `UIAction.state` because UIKit has that API and it's what VoiceOver reads; SwiftUI has no equivalent, so label-plus-trait is right.

One small divergence I noticed and am fine with: while the stop is still loading, UIKit's `filterMenu()` appends "Filtered Routes" whereas the sheet drops the route section entirely. The sheet's behavior is arguably better — you can't meaningfully pick routes you haven't loaded — and it's a clear improvement on the pre-fix state where the entire button was disabled.

CI is green at the head SHA. This has been a long haul across several rounds and the result is a genuinely better stop page. Approving.

PR #1279, opened 2026-08-12
Get the app building on Xcode 26

aaronbrethorst:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trip page location handling by enabling access to standard geographic coordinate types.


aaronbrethorst:
Closing in favor of the same one-line fix riding along in Boris's open PRs (#1260, #1262, #1264, #1265), which are landing now.

Those branches add `import CoreLocation` after `import ActivityKit`, while this PR added it after `import OBAKitCore`. Same import block, different position — keeping both would have conflicted every one of those branches on that hunk for no benefit.

PR #1280, opened 2026-08-12
Native search for the SwiftUI map panel

mosliem:
# Native search for the SwiftUI map panel

Brings search to the map-panel experience (`OBAUseMapPanelExperience`) as native SwiftUI, replacing the roles `MapFloatingPanelController`, `SearchResultsController`, and `RouteStopsViewController` play on the UIKit map. The classic UIKit path is untouched — both surfaces work, and the panel is still behind its feature flag.

## What's included

**Search session.** `SearchSheetView` / `SearchSheetViewModel` open search in the home sheet with a Close button back to the home sections. Reuses the existing `SearchInteractor` for quick-search, recents, bookmarks, and placemark rows; adds `SearchManager.fetchResults` so results can be awaited directly instead of arriving through `MapRegionManager.searchResponse`.

**One routing path for results.** `SearchResultRouter` turns a result into a map display plus a sheet route, shared by the single-result path and the disambiguation list so the two can't drift on what "opening a result" means. A `Route` is resolved into `StopsForRoute` before anything is drawn.

**Results drawn on the map.** `MapSearchDisplayModel` holds what the SwiftUI `Map` should draw and where the camera should point; `MapSearchOverlays` renders route polylines, the route's own stops, and searched place markers. A drawn route suppresses the ambient stop layer, matching the UIKit path's `removeAllAnnotations()` + `searchResponseOverridesStopLoading()`.

**New sheets.** `searchResults` (disambiguation, with inline vehicle progress and a retryable error row), `mapItem` (reuses `MapItemView` with sheet-native actions), `routeStops` (replaces `RouteStopsViewController`, opens at medium so the route stays visible), and `nearbyStops` (bridges the existing controller).

**Viewport scoping.** `MapViewportRecorder` records the SwiftUI map's visible rect so address and route searches are scoped to what the user is looking at.

Plus the new strings across all locales, including per-locale plural forms for the result count.

## Notable decisions

**A drawn result's lifetime is keyed to the sheet stack, not to a view's lifecycle.** Clearing the map from the owning sheet's `onDisappear` looked right but wasn't: the floating-sheet system rebuilds a sheet's content view without the user leaving, and each rebuild wiped the route off the map seconds after it was drawn while the sheet stayed visible. The display now records its owning route and is dropped when that route leaves the stack — which still covers a real exit, including the drag-down the OS routes through `truncateStacked`.

**Results are resolved before the sheet stack is unwound.** Popping first left the stacked layer empty across a network call, and set the failure message on a search view model whose view — and whose alert — had already been torn down, so a route that failed to resolve dropped the user on home with no explanation. Resolve, then unwind and push in one synchronous step.

## Verification

- Full `OBAKitTests` suite passes on iPhone 16 / iOS 26, with one pre-existing unrelated failure (`RentalFormatTests.rangeFallbackUsesAbbreviatedUnits` compares `MKDistanceFormatter`'s abbreviated and spelled-out output, which are identical under a metric host locale).
- SwiftLint clean.
- New unit coverage for the router, the search view model, the display model's ownership rule, the sheet-route factory, and row mapping.
- Manually exercised on the simulator: route search opens the stops sheet and the drawn result stays put.

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive search for stops, routes, map items, and vehicles.
  * Added search results, route stops, nearby stops, and map-item sheets with navigation and map integration.
  * Added vehicle loading, retry, and inline error states.
  * Added dynamic, region-aware search placeholders and improved map camera behavior.
* **Bug Fixes**
  * Prevented duplicate search rows from sharing identifiers.
  * Improved route-stop handling for unresolved or empty results.
* **Accessibility & Localization**
  * Added translated search states, result counts, query-clearing labels, and accessibility actions across supported languages.

### DIFF
--- a/OBAKit/Sheet/Content/Search/SearchResultRouter.swift
+++ b/OBAKit/Sheet/Content/Search/SearchResultRouter.swift
@@ -1,0 +1,109 @@
+//
+//  SearchResultRouter.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import Foundation
+import MapKit
+import OBAKitCore
+
+/// Turns a single search result into a map display plus a sheet route.
+///
+/// Shared by `SearchSheetViewModel` (single-result searches) and
+/// `SearchResultsSheetView` (a row picked out of a disambiguation list), so the two
+/// can't drift on what "opening a result" means.
+///
+/// The UIKit equivalent is split across `MapRegionManager.searchResponse.didSet` and
+/// `MapViewController.mapRegionManager(_:showSearchResult:)`.
+@MainActor
+final class SearchResultRouter {
+
+    private let application: Application
+    private let coordinator: SheetCoordinator<AppSheetRoute>
+    private let displayModel: MapSearchDisplayModel
+    private let onPresentVehicleTrip: (VehicleStatus) -> Void
+
+    /// Set when resolving a result fails, so the presenting screen can render the
+    /// failure inline instead of popping a modal over the sheet.
+    private(set) var lastError: Error?
+
+    init(
+        application: Application,
+        coordinator: SheetCoordinator<AppSheetRoute>,
+        displayModel: MapSearchDisplayModel,
+        onPresentVehicleTrip: @escaping (VehicleStatus) -> Void
+    ) {
+        self.application = application
+        self.coordinator = coordinator
+        self.displayModel = displayModel
+        self.onPresentVehicleTrip = onPresentVehicleTrip
+    }
+
+    /// Routes `response` when it holds exactly one result. Returns `false` when it
+    /// doesn't, leaving the decision (disambiguate, or report no results) to the
+    /// caller.
+    func presentSingleResult(from response: SearchResponse) async -> Bool {
+        guard response.results.count == 1, let result = response.results.first else {
+            return false
+        }
+        await present(result: result)
+        return true
+    }
+
+    func present(result: Any) async {
+        lastError = nil
+
+        switch result {
+        case let stop as Stop:
+            displayModel.show(stop: stop)
+            coordinator.push(.stopDetails(stopID: stop.id))
+
+        case let mapItem as MKMapItem:
+            // Matches the UIKit rule: only animate the recenter for nearby
+            // destinations, so a cross-region jump doesn't fly the camera across
+            // the map.
+            let animated = isWithinAnimationRange(mapItem.placemark.coordinate)
+            displayModel.show(mapItem: mapItem, animated: animated)
+            coordinator.push(.mapItem(mapItem))
+
+        case let route as Route:
+            await presentRoute(route)
+
+        case let stopsForRoute as StopsForRoute:
+            displayModel.show(stopsForRoute: stopsForRoute)
+            coordinator.push(.routeStops(stopsForRoute))
+
+        case let vehicle as VehicleStatus:
+            onPresentVehicleTrip(vehicle)
+
+        default:
+            Logger.error("SearchResultRouter: unhandled result type \(type(of: result))")
+        }
+    }
+
+    /// A `Route` carries no geometry. Resolve it into `StopsForRoute` — the polyline
+    /// and stop list — before anything can be drawn or listed.
+    private func presentRoute(_ route: Route) async {
+        guard let apiService = application.apiService else { return }
+
+        do {
+            let stopsForRoute = try await apiService.getStopsForRoute(routeID: route.id).entry
+            displayModel.show(stopsForRoute: stopsForRoute)
+            coordinator.push(.routeStops(stopsForRoute))
+        } catch {
+            Logger.error("SearchResultRouter: failed to load stops for route \(route.id): \(error)")
+            lastError = error
+        }
+    }
+
+    private static let animationDistanceThreshold: CLLocationDistance = 1609 // roughly a mile
+
+    private func isWithinAnimationRange(_ coordinate: CLLocationCoordinate2D) -> Bool {
+        guard let currentLocation = application.locationService.currentLocation else { return false }
+        return coordinate.distance(from: currentLocation.coordinate) <= Self.animationDistanceThreshold
+    }
+}
--- a/OBAKit/Sheet/Content/Search/SearchResultRouter.swift
+++ b/OBAKit/Sheet/Content/Search/SearchResultRouter.swift
@@ -43,67 +43,112 @@ final class SearchResultRouter {
         self.onPresentVehicleTrip = onPresentVehicleTrip
     }
 
-    /// Routes `response` when it holds exactly one result. Returns `false` when it
-    /// doesn't, leaving the decision (disambiguate, or report no results) to the
-    /// caller.
-    func presentSingleResult(from response: SearchResponse) async -> Bool {
-        guard response.results.count == 1, let result = response.results.first else {
-            return false
-        }
-        await present(result: result)
-        return true
+    /// A search result that has been resolved into everything needed to show it, with
+    /// any network work already done.
+    ///
+    /// Separating this from `present(_:)` lets a caller finish the slow part while its
+    /// own screen is still up — so a failure can be reported where the user is looking,
+    /// and the sheet stack isn't torn down around an in-flight request.
+    enum Resolved {
+        case stop(Stop)
+        case mapItem(MKMapItem, animated: Bool)
+        case route(StopsForRoute)
+        /// Handed to the trip bridge rather than a sheet, so it has no sheet route.
+        case vehicle(VehicleStatus)
     }
 
-    func present(result: Any) async {
+    /// Resolves the single result in `response`. Returns `nil` when the response holds
+    /// anything other than exactly one result, leaving that decision (disambiguate, or
+    /// report no results) to the caller.
+    func resolveSingleResult(from response: SearchResponse) async -> Resolved? {
+        guard response.results.count == 1, let result = response.results.first else {
+            return nil
+        }
+        return await resolve(result: result)
+    }
+
+    /// Turns a raw search result into a `Resolved`, performing whatever network work
+    /// the result needs. Records `lastError` and returns `nil` when it can't be
+    /// resolved. Touches neither the map nor the sheet stack.
+    func resolve(result: Any) async -> Resolved? {
         lastError = nil
 
         switch result {
         case let stop as Stop:
-            displayModel.show(stop: stop)
-            coordinator.push(.stopDetails(stopID: stop.id))
+            return .stop(stop)
 
         case let mapItem as MKMapItem:
             // Matches the UIKit rule: only animate the recenter for nearby
             // destinations, so a cross-region jump doesn't fly the camera across
             // the map.
-            let animated = isWithinAnimationRange(mapItem.placemark.coordinate)
-            displayModel.show(mapItem: mapItem, animated: animated)
-            coordinator.push(.mapItem(mapItem))
+            return .mapItem(mapItem, animated: isWithinAnimationRange(mapItem.placemark.coordinate))
 
         case let route as Route:
-            await presentRoute(route)
+            return await resolveRoute(route)
 
         case let stopsForRoute as StopsForRoute:
-            displayModel.show(stopsForRoute: stopsForRoute)
-            coordinator.push(.routeStops(stopsForRoute))
+            return .route(stopsForRoute)
 
         case let vehicle as VehicleStatus:
-            onPresentVehicleTrip(vehicle)
+            return .vehicle(vehicle)
 
         default:
             Logger.error("SearchResultRouter: unhandled result type \(type(of: result))")
+            return nil
         }
+    }
+
+    /// Draws the result and opens its sheet in one synchronous step, so the stacked
+    /// layer never sits empty between the two.
+    ///
+    /// Each display is handed the same route value that gets pushed: the display lives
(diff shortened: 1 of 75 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. This is why the `build` check is red (run 31734948821). `replaceMappedResponses` swaps the **entire** mock table, and this call registers only the `stops-for-route` stub — dropping the `regions-v3.json` and `agencies-with-coverage` stubs that `buildApplication` installed for the live `Application` created in `makeSelection`. When that `Application`'s background regions refresh lands (after this test has already passed) `MockDataLoader` hits `fatalError("Missing response to URL: https://regions.example.com/regions-v3.json")` and takes the test host down. `xcodebuild` attributes the crash to whichever test was in flight, which is why CI reports `SearchSheetViewModelTests."Typing rebuilds the interactor sections"` as the failure even though that test is innocent and passes on the rerun. The existing convention is to re-register the base stubs inside the closure — see the comment on `BookmarksViewModelTests.swift:450` ("replaceMappedResponses replaces the *entire* table, so re-register the regions/agencies/alerts mocks that the Application's background tasks rely on") and `registerBaseStubs(on:)` in `DestinationStopPickerControllerTests.swift`. Adding `stubRegions(dataLoader: loader)` and `stubAgenciesWithCoverage(dataLoader: loader, baseURL: Fixtures.pugetSoundRegion.OBABaseURL)` inside the closure should fix it.

https://github.com/OneBusAway/onebusaway-ios/blob/d4172cc3a8a622502fc9a727fe8a1bc204d57401/OBAKitTests/Sheet/SearchResultsSheetViewTests.swift#L330-L336

aaronbrethorst (changes requested):
This is well-built and unusually well-documented — the plan docs, the `MapSearchDisplayModel` ownership story, and keeping the existing UIKit path untouched behind the feature flag all made it much easier to follow than a change this size usually is.

CI is red, and the cause is in code this PR adds, so I want it green before this replaces anything.

`SearchResultsSheetViewTests."A successful retry clears the recorded failure"` calls `dataLoader.replaceMappedResponses` registering only the `stops-for-route` stub. That method swaps the **entire** response table — its own doc comment says so, and `BookmarksViewModelTests` carries an explicit note about it — so the `regions-v3.json` and `agencies-with-coverage` stubs that `buildApplication` installed are dropped. The live `Application`'s background regions refresh then lands about 130ms later, finds no registered response, and hits `MockDataLoader`'s `fatalError`, which kills the whole test process.

The important part, and the reason this is worth spelling out: **CI blames the wrong test.** The reported failure is `SearchSheetViewModelTests."Typing rebuilds the interactor sections"`, a three-line synchronous test that merely happened to be in flight when the host died and which passes on the automatic rerun. Chasing that one will waste your afternoon. The fix is in the retry test — re-register the regions and agencies stubs inside the `replaceMappedResponses` block.

That also explains the 27-minute job: after the crash, `simctl diagnose` stalls for its full 600s timeout.

Two things I noticed while reading, neither blocking:

- `SearchSheetView.onDisappear { ProgressHUD.dismiss() }` — your own comment on `MapSearchDisplayModel.owner` establishes that the sheet system can rebuild a sheet's content view without the user leaving, which fires `onDisappear`. If that happens mid-search the HUD goes away with no path to bring it back, since `isSearching` doesn't change and the `onChange` won't re-fire.
- `SearchSheetViewModel.close()` doesn't cancel `pendingPresentation` — only `deinit` does. Hitting Close while a stop or map-item presentation is resolving lets it land afterwards, so the rider gets a detail sheet they'd already backed out of.

Get CI green and I'll do a proper pass on the feature itself.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
CI is green, and I checked it's real rather than lucky: run 31853579628 is on the current head `df58260a`, created eleven seconds after the commit, attempt 1 with no rerun, and the "OBAKit Unit Test" step succeeded on its own. I also confirmed `tests.yml` has no `-retry-tests-on-failure`, so nothing is masking a flake.

More importantly, you fixed it at the root cause rather than around it — and then wrote down why:

```
// replaceMappedResponses replaces the *entire* table, so re-register the
// regions/agencies mocks that the Application's background tasks rely on.
// Dropping them lets the regions refresh land on an unmocked URL, and
// MockDataLoader's fatalError takes the whole test host down — attributed to
// whichever test happened to be in flight, not this one.
```

That comment is worth more than the fix. The genuinely nasty part of this bug was the misattribution, and the next person to reach for `replaceMappedResponses` now gets the warning at the point of use instead of losing an afternoon chasing an innocent three-line test. I swept the diff and this is the PR's only call site, so nothing else repeats the pattern.

Both non-blocking notes are handled properly too:

- The HUD is restored from `onAppear` when `isSearching` is true, with a comment naming the sheet-rebuild mechanism and explaining why `onChange` can't cover it.
- `close()` now cancels `pendingPresentation`, with `Task.isCancelled` guards after each resolve before touching the coordinator. And the test is real — without the fix the detached task pushes `.stopDetails` after the pop, so `#expect(coordinator.stackedRoutes.isEmpty)` genuinely fails.

Now the proper pass on the feature I owed you. It holds up. `MapSearchDisplayModel`'s lifetime is keyed to the sheet route stack rather than a view lifecycle, which is the thing that makes the ownership story work at all. Cancellation is checked at each await boundary. The UIKit path stays behind the feature flag, and the one behavior change I found there — a failed vehicle-details fetch no longer publishing an empty error response — is deliberate and correct: publishing an empty response triggers `notifyDelegatesNoSearchResults`, so the old code produced a double alert. Localization parity is clean across all 13 locales, including adding the new plural key to `LocalizationTests.pluralKeys` and using `localizedStringWithFormat` so the `few`/`many` forms are actually reachable in ar/pl/ru.

Three small things, none blocking:

- `showMapItem`/`showStop` overwrite `pendingPresentation` without cancelling the previous task. `close()` cancels, but a second row tap doesn't cancel the first — two rapid taps could each pop and present. Hard to hit since the first tap tears the list down.
- `SearchResultRouter.lastError` is shared mutable state between the search sheet and the results sheet. In practice they can't resolve concurrently and it's read synchronously on the main actor, so it's fine today — just fragile if the sheets ever overlap.
- There's a sub-millisecond window in `performSearchAndWait` between the `Task.isCancelled` check and the `.disambiguate` push where `await present(outcome)` can suspend.

Sequencing note: #1288 and #1289 are stacked on this branch, so landing this first unblocks both of them to rebase onto `main`.

Approving.

aaronbrethorst:
Approved on the merits. Heads up: now that #1272 has landed on `main` and touched the same sheet plumbing — `SheetRoute.swift`, `AppSheetViewFactory.swift`, and `MapPanelRootController.swift` — this branch has merge conflicts.

Please merge `main` in and resolve. `AppSheetViewFactory.init` is the spot to look at first: both PRs add required parameters to it, and both add cases to `AppSheetRoute` plus its exhaustiveness-guard switch.

Once it's mergeable it's good to go — no re-review needed unless the resolution changes behavior.

One sequencing note: #1288 and #1289 are stacked on this branch, so getting this rebased and landed unblocks both of them.

OBAKit/Sheet/Root/MapPanelRootController.swift
```diff
@@ -87,6 +97,39 @@ public final class MapPanelRootController: UIViewController {
...
+                    value: "The vehicle you chose doesn't appear to be on a trip right now, which means we don't know how to show it to you.",
+                    comment: "This message appears when a searched-for vehicle doesn't have an assigned trip."
+                )
+                // `presenter`, not `host`, for the reason spelled out in
+                // `presentTripController` below: UIKit ignores `present` on a
+                // controller that already has a `presentedViewController`, and by the
+                // time we get here the base sheet is still up on `host`. Presenting
+                // from `host` silently drops the alert, leaving the rider on home with
+                // no explanation for why their vehicle search went nowhere.
+                Task { await AlertPresenter.show(errorMessage: message, presentingController: presenter) }
```
aaronbrethorst, line 124: This is the right fix, and I checked the reasoning rather than just the diff.

Dropping `let host` from the guard above is safe: `host` isn't referenced anywhere else in `present(vehicleStatus:)`, and `topmostController()` is `guard let host else { return nil }` followed by a loop that always returns non-nil — so `topmostController() == nil` iff `host == nil` is strictly biconditional. The guard's failure set is unchanged, which means the `"host or application is nil"` log message is still accurate, and this now matches the identical formulation in `present(_ arrival:)` right above. Two paths that should behave the same now do.

The comment's premise checks out too: the base sheet is `.sheet(isPresented: .constant(true))` in `FloatingSheetContainer`, so it's permanently presented and `host.presentedViewController` is non-nil whenever the panel is on screen. Presenting from `host` really would have been swallowed.

One residual gap, not a regression and not blocking — `presenter` is captured synchronously but consumed inside the `Task`. On the *disambiguation* path (`.searchResults` stacks as a real modal, `prefersStacking == true`), `SearchResultsSelection.select` calls `coordinator.popToRoot()` and then `router.present(resolved)` in the same synchronous turn, so `topmostController()` hands back the results sheet SwiftUI is about to tear down, and the alert goes down with it. The single-result path — the common one — is genuinely fixed here, because `.search` is a content-swap route and the base sheet never goes away. So this commit is a strict improvement over what was there; I'm just noting where the remaining hole is for whenever you're back in this file.

OBAKit/Sheet/Content/Search/SearchSheetViewModel.swift
```diff
@@ -0,0 +1,272 @@
...
+    )
+
+    /// The in-flight `router.present(...)` kicked off by a delegate callback.
+    /// `SearchDelegate`'s methods are synchronous, so presentation has to be
+    /// detached — this handle lets tests await it instead of polling.
+    private(set) var pendingPresentation: Task<Void, Never>?
+
+    func showMapItem(_ mapItem: MKMapItem) {
+        application.userDataStore.addRecentMapItem(mapItem)
+        pendingPresentation = Task { [weak self] in await self?.leaveSearchAndPresent(mapItem) }
```
aaronbrethorst, line 231: Carrying this forward from my last pass — still open, still not blocking.

Both `showMapItem` (here) and `showStop` (line 235) assign `pendingPresentation` without cancelling whatever was already in it. `close()` cancels correctly, and a second tap is hard to land because the first tap tears the list down — but if two do get through, each runs `leaveSearchAndPresent` to completion, so you get two `coordinator.pop()` + `router.present(...)` pairs. A `pendingPresentation?.cancel()` before each assignment is the whole fix if you want to close it.

OBAKit/Sheet/Content/Search/SearchResultRouter.swift
```diff
@@ -0,0 +1,161 @@
...
+final class SearchResultRouter {
+
+    private let application: Application
+    private let coordinator: SheetCoordinator<AppSheetRoute>
+    private let displayModel: MapSearchDisplayModel
+    private let onPresentVehicleTrip: (VehicleStatus) -> Void
+
+    /// Set when resolving a result fails, so the presenting screen can render the
+    /// failure inline instead of popping a modal over the sheet.
+    private(set) var lastError: Error?
```
aaronbrethorst, line 32: Also carried forward, also not blocking. This one instance is shared by both `searchView()` and `searchResultsView()` through the lazy `searchResultRouter` in `AppSheetViewFactory`, so `lastError` is shared mutable state across the two sheets. It's fine today — they can't resolve concurrently and it's read synchronously on the main actor — but it's the kind of thing that stops being fine the moment those sheets can overlap. Worth a comment at minimum saying why the single instance is safe.

aaronbrethorst (approved):
Re-approving on the two commits that landed after my last pass — and this time I've verified the merge itself, not just the branch.

**`97bc6e15` (alert presenter)** is correct, and I traced the reasoning rather than taking the comment's word for it. Details inline. The short version: dropping `host` from the guard changes nothing about when the guard fires, because `topmostController()` is nil exactly when `host` is — so the log message stays accurate and `present(vehicleStatus:)` now behaves the same as `present(_ arrival:)` directly above it, which already presented from `presenter`. That divergence was the bug.

**`16357e48`** is fine but redundant: I have the identical one-line removal open as #1283. I'll close mine and let this carry it.

**On the stale-green problem.** Your CI run (32399886125) is on the current head, attempt 1, no rerun — real green. But it was created at 17:50Z on the 20th, and `main` moved at 00:26Z on the 21st when #1299 landed. #1299 touched `SheetRoute.swift`, `AppSheetViewFactory.swift`, and `MapPanelRootController.swift` — all files this branch also modifies. Git reports no textual conflict, but "merges cleanly" and "compiles" are different claims for a Swift change where `AppSheetViewFactory.init` is gaining parameters from several directions at once.

So I merged #1280, #1288, and #1289 onto current `main` locally and built and ran the suite against the result: **2277 tests in 246 suites passed.** The stack is safe to land as-is. No action needed from you — recording it so you know the approval is backed by a real integration run rather than by CI that predates the base it's merging into.

Two carried-forward notes from my last review are still open and still not blocking; I've left them inline so they don't get lost. The third (the `Task.isCancelled` window before the `.disambiguate` push) I looked at again and convinced myself it's genuinely empty — a same-actor `async` call doesn't hop and that arm has no `await` of its own — so consider it closed rather than deferred.

**Merge order.** I'm landing this first, then #1288, then #1289, then #1293. I'd told you in my #1293 review to land *that* one first to minimize churn — that advice is now stale, because #1293 has since grown `.stringsdict` changes across all 13 locales. I test-merged it and it conflicts with this stack in 20 files. Landing the stack first means one branch needs a rebase instead of three.

Approving. Thank you for the care on this one — the root-cause fix and the comment you left about `replaceMappedResponses` are still the thing I'd point other contributors at.

PR #1284, opened 2026-08-14
Address arrival/departure filter follow-ups

### LINKED ISSUE #1274: Follow-ups from #1096 (arrival/departure filter): English copy, VoiceOver state, docs
Follow-ups from #1096. None of these blocked the merge; collecting them so they don't evaporate.

- [ ] **English missed the arrivals-and-departures rename.** #1096's last commit renamed `settings_controller.arrival_display_section.title` to cover both arrivals and departures across all 12 non-English locales — fr is "Affichage des arrivées et départs", es is "Visualización de llegadas y salidas" — but `OBAKit/Strings/en.lproj/Localizable.strings:985` and the inline `OBALoc` value in `SettingsViewController.swift` still say "Arrival Display". English is now the only locale with the old, narrower wording. `LocalizationTests` only checks key parity, so CI can't catch this class of drift.

- [ ] **`departureFilterMenu()` marks selection with an image rather than menu state.** It sets `action.image = checkmark` where `action.state = .on` is what conveys selection to VoiceOver. It matches the local convention in `sortMenu()`/`filterMenu()` in the same file, so fixing it probably means fixing all three; `BookmarksViewController.swift:154` is the example that does it right.

- [ ] **`OBAKitConfig.DefaultArrivalDepartureFilter` is undocumented.** The white-label knob that motivates the whole feature has no README or docs entry, unlike the other `OBAKitConfig` keys.

- [ ] **Sheet toolbar's filter glyph doesn't reflect a non-`.all` Departure Type**, while the pushed presentation deliberately does (`StopPageViewController.swift:588-589`, with a comment stating that intent). Partly justified — in the sheet the two are *sibling* submenus, so filling the route-filter glyph would be misleading — but it should be a deliberate decision rather than an asymmetry.

- [ ] **Time-sorted no-results branch drops the trailing section**, losing the "Load More" button and walk-time row that the original code's comment says are always appended. Probably intentional alongside the new empty state, worth confirming.

diveshpatil9104:
Description:

Closes #1274 — all five follow-ups, one commit each concern.

1. **English copy** — `en.lproj` and the inline `OBALoc` value both now say "Arrival & Departure Display". They have to move together or `genstrings` reverts the strings file.
2. **VoiceOver** — selection rides on `UIAction.state` instead of a checkmark image, across all five menus in both stop pages (as you guessed, fixing one meant fixing the convention).
3. **Docs** — `DefaultArrivalDepartureFilter` documented in `WhiteLabel.md`.
4. **Sheet glyph** — kept your reasoning: the route-filter glyph stays route-only because there the two are siblings. Instead each submenu carries its own state, so Departure Type gets a filled/unfilled pair matching the route filter's. That also tells you *which* filter is on when you open the funnel in the pushed presentation.
5. **Load More** — not intentional, restored. It's the one control that can resolve a filtered-empty stop, since widening the window is how a real-time departure appears when the loaded window holds only scheduled ones.

Tests cover the three behavioral changes: menu selection state, the restored section, and the symbol pair (`Icons` force-unwraps `UIImage(systemName:)`).

## Summary by CodeRabbit

* **New Features**
  * Added documentation for configuring default arrival and departure filters.
  * Improved stop-page filtering menus with clear selected-state indicators.
  * Added dynamic icons to show whether departure-type filtering is active.
  * Renamed the settings label to “Arrival & Departure Display.”

* **Bug Fixes**
  * Improved filtered stop results, including empty states and past-departure ordering.

### DIFF
--- a/OBAKit/OBAKit.docc/Articles/WhiteLabel.md
+++ b/OBAKit/OBAKit.docc/Articles/WhiteLabel.md
@@ -146,6 +146,7 @@ These are parameters specified in your app's `project.yml` file that customizes 
     RESTServerAPIKey: org.onebusaway.iphone
     RegionsServerBaseAddress: https://regions.onebusaway.org
     RegionsServerAPIPath: /regions-v3.json
+    DefaultArrivalDepartureFilter: all
 
 ### Regions
 
@@ -158,3 +159,15 @@ There are three parameters that are associated with regions:
 The regions JSON file that you will specify with the `BundledRegionsFileName` parameter is required so that the app can show the user a list of available regions as soon as the application launches, without having to worry about having an internet connection. However, that list will eventually/inevitably become out of date, and so the app will also regularly download the remote copy of the regions file to make sure that changes (names, features, new regions, deleted regions) are handled.
 
 If you choose to leave out `RegionsServerBaseAddress` and `RegionsServerAPIPath`, then your app will rely solely on the bundled regions file for all regions that your application will know about. It is _highly_ recommended that you provide a regions server URL and path, but it is not strictly required.
+
+### Arrival and Departure Display
+
+* `DefaultArrivalDepartureFilter` - Optional. Defaults to `all`.
+
+Stop pages can hide departures that lack real-time data, or hide the real-time ones and show only the schedule. This parameter picks which of those a fresh install starts with:
+
+* `all` - Show every departure. This is the behavior you get if you omit the parameter, and the right choice for most agencies.
+* `estimatedOnly` - Show only departures the server has real-time data for. Worth considering if your feed's scheduled times are unreliable enough that riders are better off not seeing them.
+* `scheduledOnly` - Show only departures without real-time data.
+
+This is a starting value, not a lock: riders can change it in Settings under Arrival & Departure Display, or from the Departure Type menu on any stop page, and their choice is remembered from then on. An unrecognized value is ignored and treated as `all`.
--- a/OBAKit/Settings/SettingsViewController.swift
+++ b/OBAKit/Settings/SettingsViewController.swift
@@ -176,13 +176,13 @@ class SettingsViewController: FormViewController {
         store.walkingSpeedMetersPerSecond = decision.speed
     }
 
-    // MARK: - Arrival Display Section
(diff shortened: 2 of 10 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
All five follow-ups from #1274 are genuinely addressed rather than partially, which is what I was checking for. Going through them:

- **The menu-state conversion is complete.** I diffed for residual `action.image = checkmark` assignments and there are none left at the head SHA — all eight sites across both `StopViewController` and `StopPageViewController` now use `UIAction.state`. That's the substance of the accessibility fix: `.state = .on` is what VoiceOver actually announces as "selected," whereas a checkmark image announces nothing. The SwiftUI `StopPageToolbar` already carried `.accessibilityAddTraits(.isSelected)`, so it was never broken.
- **The English copy change touches both `en.lproj` and the inline `OBALoc` default**, which is the easy half to forget.
- **The `WhiteLabel.md` documentation matches reality** — the key name, the three raw values, and the "unrecognized degrades to `.all`" claim all check out against the enum.
- **The Load More restoration is strictly additive.** The route-sorted path already got Load More from its own section, so the two paths are now symmetric rather than one silently losing it behind the filter's empty state.

CodeRabbit's inline note about route-filter selection state on single-route or unloaded stops is pre-existing and byte-for-byte the same as the old image-based behavior, so it's not a regression from this PR.

A few notes for whenever you're next in these files, none blocking:

- `Selecting a different filter moves the selection` never actually performs a selection — it presets `.scheduledOnly` and re-reads state, which duplicates the previous test with a different value. Real assertions, misleading name.
- `Filtered-empty stop keeps its Load More section` asserts the section ID is present, not that the Load More item is inside it. The section could survive on other content and still pass.
- `Icons.departureType` went from a `public class var` to a `public class func`. Both in-repo call sites are updated, but this is a white-label framework, so that's a source-breaking change for anyone consuming OBAKit out of tree. Worth a release note.
- `DepartureFilterMenuTests` writes `FeatureFlags.useNewStopPageKey` into defaults, but the controller is constructed directly and never reads it — dead setup.

Approving.

PR #1285, opened 2026-08-14
Stop advertising an unsupported bundle-ID override in scripts/setup

### LINKED ISSUE #1282: scripts/setup advertises a bundle-ID override that isn't supported
`scripts/setup` tells contributors they can override the bundle identifier in `Apps/<App>/local.yml`:

```
· No Apps/${APP}/local.yml yet. Copy Apps/${APP}/local.yml.example to override DEVELOPMENT_TEAM / bundle ID without editing the shared project.yml.
```

Both other places that describe the same mechanism say the opposite, deliberately:

- `Apps/OneBusAway/local.yml.example`: "PRODUCT_BUNDLE_IDENTIFIER is intentionally not supported here. A working personal bundle ID also needs matching app groups, associated domains, Apple Pay, URL-scheme, and user-activity identifiers; create a complete white-label configuration before changing those."
- `README.markdown`: "`PRODUCT_BUNDLE_IDENTIFIER` is not a self-contained local override…"

The docs are right and the script message is stale. `append_local_overrides` in `scripts/generate_project` only reads `settings.base.DEVELOPMENT_TEAM`; a `PRODUCT_BUNDLE_IDENTIFIER` in `local.yml` reaches the generated spec only through the spliced `include`, where it loses to the hardcoded value in `Apps/OneBusAway/project.yml`. So a contributor who follows the script's advice gets a silent no-op and no explanation.

### Fix

Drop "/ bundle ID" from the message in `scripts/setup`, so it reads "…to override DEVELOPMENT_TEAM without editing the shared project.yml."

### Possibly worth doing at the same time

`generate_project` runs `rm -rf OBAKit.xcodeproj` and copies the spec *before* `append_local_overrides` parses `local.yml`. A malformed `local.yml` (an empty file makes `YAML.load_file` return `false`, so `.fetch` raises) aborts after the existing project has already been deleted. Validating `local.yml` before the destructive step would turn a confusing failure into a clear one.

Introduced in #1262.

Borisserz:
## Summary
- `scripts/setup` told contributors they could override the bundle ID via `Apps/<App>/local.yml`. `generate_project` only splices `DEVELOPMENT_TEAM`, and `local.yml.example` already says `PRODUCT_BUNDLE_IDENTIFIER` is not supported there.
- The missing-file hint now matches those two sources.

Closes #1282.

## Test plan
- [x] Diff: the new sentence names `DEVELOPMENT_TEAM` only and points at `local.yml.example` for the bundle-ID rule
- [ ] `scripts/setup OneBusAway` with no `local.yml` prints the corrected hint

### DIFF
--- a/scripts/setup
+++ b/scripts/setup
@@ -49,7 +49,7 @@ fi
 if [[ -f "Apps/${APP}/local.yml" ]]; then
   ok "Local overrides: Apps/${APP}/local.yml (gitignored)"
 else
-  say "  · No Apps/${APP}/local.yml yet. Copy Apps/${APP}/local.yml.example to override DEVELOPMENT_TEAM / bundle ID without editing the shared project.yml."
+  say "  · No Apps/${APP}/local.yml yet. Copy Apps/${APP}/local.yml.example to override DEVELOPMENT_TEAM without editing the shared project.yml. PRODUCT_BUNDLE_IDENTIFIER is not supported there — see local.yml.example."
 fi
 
 say ""


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Nice catch, and thank you for correcting your own earlier message rather than letting it sit.

I verified the claim independently before approving: `append_local_overrides` in `scripts/generate_project` only ever reads `settings.base.DEVELOPMENT_TEAM` out of `local.yml` and splices that into the copied spec. Everything else in `local.yml` reaches XcodeGen through the `include:` entry, and both `Apps/OneBusAway/project.yml` and `Apps/KiedyBus/project.yml` hardcode `PRODUCT_BUNDLE_IDENTIFIER` in target settings — which win over includes. So the override really was a silent no-op, and the old text was actively misleading.

The new wording also lines up with `local.yml.example` and the README instead of contradicting them. CI is green. Approving.

PR #1286, opened 2026-08-14
Distinguish unreachable-server copy from bad stop data

### LINKED ISSUE #1276: improve error message for bad data at a stop
Periodically the Sound Transit server has bad data for a particular stop.  When this happens, the app gives an error (which is good) but it doesn't make it clear that this is a server problem -- we regularly get questions from riders about this.  Change the message to something more informative, like "bad data for this stop on the server -- let us know so that we can notify the transit agency".

Image

### LINKED ISSUE #1281: improve error message when OneBusAway unable to reach the server
If OneBusAway is unable to reach the server, it gives a useful message that says the server is down.  This is much improved over the previous message, since before people often thought the app itself was broken.  However, a rider recently reported seeing this error, which it turns out was due to a VPN rather than the server being down.  Here is a possible rewording:

Error

Unable to reach the server for San Diego [or wherever] right now, so OneBusAway isn't able to show transit information for this region. One possible cause is that the server is down -- if that's the problem, the app should start working again once the server is back up. Another possible cause is that you are using a VPN that is blocking access to the server.

Image

Borisserz:
## Summary
- Unreachable regional hosts (timeout / cannot connect / 502–504) still use `.serverUnavailable`, but the English copy no longer asserts the server *is* down. It says we cannot reach it, and mentions a VPN or firewall as one possible cause (#1281).
- `DecodingError` is no longer classified as `.serverUnavailable`. It becomes `.invalidResponseData`, with copy that names a bad stop/agency feed (#1276). Empty HTTP 200 is unchanged (`requestNotFound`, see #1268).
- Empty-state icons use `exclamationmark.triangle` for bad payload, not `server.rack`.
- Design notes: `docs/superpowers/specs/2026-08-14-error-copy-reachability-design.md`.

Closes #1281. Closes #1276.

## Test plan
- [x] `ErrorClassifierTests` — 30 tests on iPhone 17 Pro simulator, including:
  - DecodingError + region → `.invalidResponseData`, not `.serverUnavailable`
  - `.serverUnavailable` description contains `Unable to reach` and `VPN`
  - `.invalidResponseData` description does not contain `down` or `VPN`
  - timeout / 502–504 still `.serverUnavailable`
- [ ] Dark/light: trigger a real unreachable-host alert and confirm the VPN sentence
- [ ] Bad JSON at a stop shows the feed copy, not “server is down”

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when transit data cannot be read.
  * Distinguishes invalid response data from unavailable servers.
  * Updated error screens with a warning icon and clearer retry guidance.
* **Localization**
  * Added translated invalid-data error messages across supported languages.
* **Documentation**
  * Documented error classification, messaging, icon treatment, and expected behavior.

### DIFF
--- a/OBAKit/Controls/ListView/Rows/EmptyData/OBAEmptyDataSetItem.swift
+++ b/OBAKit/Controls/ListView/Rows/EmptyData/OBAEmptyDataSetItem.swift
@@ -66,6 +66,8 @@ nonisolated struct EmptyDataSetItem: OBAListViewItem {
                 icon = UIImage(systemName: "questionmark.square.dashed")
             case .serverError, .serverUnavailable:
                 icon = UIImage(systemName: "server.rack")
+            case .invalidResponseData:
+                icon = UIImage(systemName: "exclamationmark.triangle")
             case .surveyServiceNotConfigured:
                 icon = UIImage(systemName: "doc.text.magnifyingglass")
             case .noRegionSelected:
--- a/OBAKit/Controls/ListView/Rows/EmptyData/OBAListViewEmptyDataViewModel.swift
+++ b/OBAKit/Controls/ListView/Rows/EmptyData/OBAListViewEmptyDataViewModel.swift
@@ -55,6 +55,8 @@ extension OBAListView {
                     icon = UIImage(systemName: "questionmark.square.dashed")
                 case .serverError, .serverUnavailable:
                     icon = UIImage(systemName: "server.rack")
+                case .invalidResponseData:
+                    icon = UIImage(systemName: "exclamationmark.triangle")
                 case .surveyServiceNotConfigured:
                     icon = UIImage(systemName: "doc.text.magnifyingglass")
                 case .noRegionSelected:
--- a/OBAKit/Controls/SwiftUI/ErrorView.swift
+++ b/OBAKit/Controls/SwiftUI/ErrorView.swift
@@ -51,6 +51,8 @@ struct ErrorView: View {
             return "wifi.slash"
         case .serverError, .serverUnavailable:
             return "server.rack"
+        case .invalidResponseData:
+            return "exclamationmark.triangle"
         case .captivePortal:
             return "lock.shield"
         default:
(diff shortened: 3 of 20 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is a good distinction to draw, and the copy is a real improvement — "the server appears to be down" was asserting something we don't actually know, and riders on a filtered corporate or campus network hit exactly that case. Naming VPN/firewall as a possibility is the honest version.

What I verified before approving:

- **The classification is correct.** The decoding path is only reachable after content-type and status-code checks have already diverted genuine unreachable/5xx failures, so a `DecodingError` really does mean "host responded, payload unusable." Adding `.invalidResponseData` rather than overloading `.serverUnavailable` is the right shape.
- **All four `APIError` switch sites are handled** — the two exhaustive ones got the new case, and the two with `default:` already fall through to a sensible icon.
- **Key and format-specifier parity holds across all 13 locales**, which `LocalizationTests` enforces anyway. CI is green on the head SHA.

One thing I want to flag clearly, and I went back and forth on whether to hold the merge for it: the two *revised* strings — `api_error.server_unavailable_fmt` and `api_error.decoding_failure` — were only updated in `en.lproj`. The other 12 locales still carry the old wording, so `fr` still says "semble hors service" and `es` still says "parece estar caído." That's the exact claim this PR sets out to stop making, and for non-English riders it keeps making it. It also leaves `decoding_failure` meaning two different things depending on locale.

I'm approving rather than blocking, because this regresses nothing — those locales had the old copy before — and `git log` shows localizations get refreshed in batches here rather than per-PR. But since you were already willing to machine-translate the *new* key into all 12 files in this same commit, translating the two adjacent revised strings while you're in there would finish the job. I'd like that done before the next release even if it's a separate PR.

- `en.lproj` line 7 still carries the old `/* ... indicating a likely server-side issue. */` comment above the rewritten value. You updated the `comment:` argument in the Swift source, so a `scripts/extract_strings` run will reconcile it. The new `invalid_response_data_fmt` entries have no translator comment in any locale for the same reason.
- The `Classify invalid response data passes through` test would pass even without the `.invalidResponseData` addition to the pass-through list, since `classifyAPIError`'s `default:` returns the error unchanged anyway. Same shape as the existing `.cellularDataRestricted` test, so it's consistent — just not load-bearing.

Approving.

PR #1287, opened 2026-08-14
Refresh stop-page arrival filter after a Settings change

### LINKED ISSUE #1273: Stop page keeps a stale arrival/departure filter after a Settings change
Follow-up from #1096, which added the white-label configurable arrival/departure filter.

`StopViewModel` seeds `arrivalDepartureFilter` once, in `init`:

```swift
self.arrivalDepartureFilter = environment.effectiveArrivalDepartureFilter
```

Nothing observes `UserDefaults` afterward, and `viewWillAppear` doesn't re-read it. The only other writer is the in-page Departure Type menu.

**Effect:** Settings is presented modally from the More tab, so a stop page already open in another tab survives that round trip showing both stale results *and* a stale checkmark in its own Departure Type menu. The user changes the setting, returns to the stop page, and nothing happens — and the menu claims the old value is still selected.

The feature is correct on every fresh navigation; this only affects an already-constructed `StopViewModel`.

**Why this is worth fixing rather than accepting:** this screen already propagates settings live, in two different ways —

- `@AppStorage(UserDefaultsStore.stopUIReducedColorsKey)` in `TripCardView.swift:33`
- a `UserDefaults.didChangeNotification` subscription in `MapViewModel.swift:114`

so seeding-once is out of step with the established pattern rather than consistent with it.

The notification-subscription shape is probably the better fit here, since the filter change needs to re-run the filtering pipeline rather than just re-render a value.

Borisserz:
## Summary
- `StopViewModel` seeded `arrivalDepartureFilter` once in `init`. Settings is modal from More, so an already-open stop kept a stale list and a stale Departure Type checkmark.
- Subscribe to `UserDefaults.didChangeNotification` on the same store Settings writes, same pattern as `MapViewModel` map-type sync. Equality-guarded; does not write defaults or refetch.
- Design notes: `docs/superpowers/specs/2026-08-14-stale-stop-filter-design.md`.

Closes #1273.

## Test plan
- [x] `StopViewModelTests` on iPhone 17 Pro simulator, including `Arrival departure filter syncs from an external Settings change` — writes via `application.setArrivalDepartureFilter` (Settings' path), not `updateArrivalDepartureFilter`
- [ ] Open a stop, switch More → Settings → Arrival/Departure filter, return to the stop without popping it: list and menu checkmark match the new value
- [ ] Changing the filter from the stop's own menu still persists and still updates the list

### DIFF
--- a/OBAKit/ViewModels/StopViewModel.swift
+++ b/OBAKit/ViewModels/StopViewModel.swift
@@ -158,6 +158,7 @@ class StopViewModel: ObservableObject {
     /// Cache for trip-panel approach timelines, invalidated on each refresh.
 
     private var alarmFiredCancellable: AnyCancellable?
+    private var userDefaultsCancellable: AnyCancellable?
 
     // MARK: - Init Context
 
@@ -222,6 +223,14 @@ class StopViewModel: ObservableObject {
             .publisher(for: .alarmFired)
             .receive(on: DispatchQueue.main)
             .sink { [weak self] _ in self?.rebuildAlarmIndex() }
+
+        // Settings is presented modally from More, so this VM can outlive a
+        // filter change. Re-read the persisted value; skip the write path so
+        // we don't echo defaults back at ourselves (#1273).
+        userDefaultsCancellable = NotificationCenter.default
+            .publisher(for: UserDefaults.didChangeNotification, object: environment.userDefaults)
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] _ in self?.syncArrivalDepartureFilterFromDefaults() }
     }
 
     /// Backward-compatible entry point for existing callers that pass `Application` directly.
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Good fix, and the scoping is what makes it good — you re-read the persisted value and stop there, rather than refetching arrivals or echoing a write back into defaults.

Things I checked before approving:

- **No feedback loop.** `syncArrivalDepartureFilterFromDefaults` never calls `setArrivalDepartureFilter`, and the in-page write path sets the published property before writing defaults, so the notification that comes back hits the equality guard and no-ops. That equality guard is also what keeps this cheap under the general `didChangeNotification` fan-out.
- **Observer lifecycle.** `userDefaultsCancellable` is stored on the VM and the sink is `[weak self]`, so no retain cycle and it tears down on dealloc. Same shape as the existing `alarmFiredCancellable` right above it.
- **Same store.** The `object:` filter targets `environment.userDefaults`, which is the instance `Application` inherits from `CoreApplication` and the one Settings writes through — so the notification actually matches instead of silently never firing.
- **The test earns its keep.** It drives `app.setArrivalDepartureFilter`, i.e. Settings' path rather than the view model's own mutator, so it genuinely fails without the production change. Asserting `.all` up front also means a changed white-label default fails loudly instead of passing vacuously.

Mirroring `MapViewModel`'s subscription was the right call — it keeps one pattern for this in the codebase instead of inventing a second.

One thing worth knowing, not blocking and not yours to fix here: the legacy `StopViewController` still won't repaint immediately on an external filter change, since `bindListData()` has no `$arrivalDepartureFilter` sink. It converges on the next arrivals tick, and it's behind `isNewStopPageEnabled` anyway, so it's strictly better than the permanent staleness we had before.

Approving.

PR #1288, opened 2026-08-14
Home sheet sections for the SwiftUI map panel

mosliem:
Populates the home sheet of the map-panel experience (`OBAUseMapPanelExperience`) with three preview sections — Nearby Stops, Recent Stops, Bookmarks, in that fixed order — each showing four items under a header that opens the section's full index. The classic UIKit path is untouched.

The home sheet is the entry point to those indexes, so the governing constraint was that opening it stay cheap: **zero network requests for Nearby and Recent, one per displayed bookmark, and no polling timer.**

> Bookmarks used to be capped at four requests. Pinning (below) makes the section's size user-controlled, so the bound is now per displayed row rather than a fixed four. The Bookmarks tab already fetches every bookmark in the region, so this stays the cheaper of the two surfaces.

## What's included

**Three section models.** `HomeNearbyStopsSectionModel` re-slices stops `MapStopsObserver` has already accumulated from the map's own loads, ordered by the map's viewport centre. `HomeRecentStopsSectionModel` reads `UserDataStore`. `HomeBookmarksSectionModel` drives a scoped `BookmarkDataLoader`. `HomeSheetViewModel` composes the three and publishes `visibleSections`, which is the single place the fixed order and the empty-section omission rule live.

**A scoped bookmark loader.** `BookmarkDataLoader` gains defaulted `bookmarkProvider` and `autoRefreshes` parameters, so the home sheet fetches arrivals for the bookmarks it displays and installs no timer, while the Bookmarks tab passes neither and keeps its 30-second cycle unchanged.

**Bookmark ordering by creation date.** `Bookmark` gains `dateCreated`, and the preview orders newest-first. `sortOrder` couldn't do this job: it's the manual Manage Bookmarks position, and `UserDataStore.add` appends a new bookmark to the end of its group — so ordering by it put the newest one *last*, past the cut, permanently invisible once you had more than four. It's also renumbered per group, so its values were never comparable in a whole-region sort.

**Pinning.** `Bookmark` gains `isPinned`, with a pin/unpin context menu on both the Bookmarks tab and the home sheet's rows. Pinned bookmarks lead the section and are never squeezed out by a newer bookmark — the item limit becomes a floor rather than a cap, so pinning six shows six. Pinning is placement-only: the Bookmarks tab's grouped/manual ordering is unchanged. Deliberately distinct from `isFavorite`, which despite the name is the "Show in Today View" toggle — reusing it would have silently changed what appears in the widget.

**Shared extractions**, each additive and defaulted: `Stop.nearest(_:to:limit:)` (which `MapStopsObserver`'s cap-eviction now folds onto), `UserDataStore.recentStops(in:)` (adopted by `RecentStopsViewModel`), `Strings.nearbyStops` / `.bookmarks` (adopted by `NearbyStopsListViewController` and `SearchInteractor`), and the loader seams above. `BookmarkRowViewModel`, `BookmarkCardView`, and `StopBookmarkRow` are reused as-is, so bookmark rows render identically to the Bookmarks tab.

**Observer ownership moved up.** `MapStopsObserver` is now built in `MapPanelRootController` and passed to both the sheet factory and `MapPanelRootView`, mirroring how `MapSearchDisplayModel` is already threaded. The factory is constructed first, so leaving the observer as a `@StateObject` on the view would have given the sheet a different instance from the one the map renders — the sheet listing stops the map isn't showing.

**Index routes render placeholders.** `.nearbyAll`, `.recentStopsAll`, and `.bookmarksAll` are wired but the screens are not built; they were exempted from `unimplementedView`'s DEBUG `assertionFailure` so a section header doesn't trap a debug build. Registering a real view later is a one-line factory change each.

## Notable decisions

**Section models republish through the view model.** `HomeSheetViewModel` holds the three section models as nested `ObservableObject`s, and SwiftUI subscribes only to the object a view is handed — never to children reached through it. Without forwarding, a section's own `@Published` write updated the model while the sheet kept rendering stale rows; `visibleSections` couldn't stand in for it, because it's guarded against no-op writes, so viewing a stop while Recent Stops was already on screen published nothing at all. This was reproducible in the app: view a stop from Nearby, dismiss, and it was missing from Recent until the sheet was closed and reopened.

**Sections refresh on data change, not just on activation.** Refreshing only from `.task` looked sufficient and wasn't: `.task` fires once per view identity, and tapping a stop row pushes `.stopDetails` as a *stacked* sheet, so the home sheet underneath is never torn down. Bookmarks observe `.bookmarksDidChange` — scoped to the app's own `UserDataStore`, which is the only poster — and the sheet re-activates when the stacked layer empties. Both paths run through the same staleness/region/selection gate.

**The `.bookmarksDidChange` handler goes straight to `loadIfNeeded()`.** It used to call `refreshSelection()` first, which advanced the selection *before* `loadIfNeeded()` snapshotted it — so a newly created bookmark never registered as a selection change, and its arrivals were never fetched inside the 30-second staleness window. The row appeared blank until something else re-activated the sheet.

**The stacked-layer transition is derived from `onChange`'s previous value, not stored `@State`.** A tracked flag re-seeds when the sheet system rebuilds a sheet's content view without the user leaving — the same rebuild behaviour documented in `SearchSheetViewModel` — which would silently skip the next refresh.

**`viewportCenter` is only written when the centre moves.** `CLLocationCoordinate2D` isn't `Equatable`, so `@Published` can't drop a same-value write; without an explicit guard every map settle fired `objectWillChange` and invalidated `MapPanelRootView`, breaking the no-republish invariant the rest of `MapStopsObserver` is built around.

**The section-title strings live in `OBAKit`, not `OBAKitCore`.** Putting them beside `Strings.recentStops` in Core looked right, but Core's `OBALoc` resolves against the Core bundle and neither key exists there; the titles would have fallen back to English in all 12 localised languages while `LocalizationTests` still passed, because a fallback resolves to a non-empty string.

**Nearby follows the map viewport, not the user's location.** It orders by the map's centre, matching the UIKit panel and what's visibly pinned behind the sheet, and it works with no location permission.

## Persistence

`dateCreated` and `isPinned` are both new keys on a persisted Core model shared with the widget. Both decode with `decodeIfPresent` fallbacks — `.distantPast` and `false` — so bookmarks written by earlier builds still load. `dateCreated` falls back to `.distantPast` rather than "now" on purpose: "now" would make every stored bookmark look freshly created on the first launch after upgrading and scramble the ordering. It's excluded from `isEqual`/`hash`, since `Date` loses sub-second precision through the store's encoder and a round-tripped bookmark would otherwise compare unequal to the one written.

## Verification

- 47 new tests across the branch, covering per-section capping, ordering, nil-region handling, staleness/region/selection gating, the `.bookmarksDidChange` refresh, fixed section order, empty-section omission, `activate()` idempotence, creation-date ordering, pin placement and the uncapped pin case, store persistence of a pin (including that it leaves `sortOrder` and group membership alone), and legacy decodes for both new fields.
- Two of those are regression tests verified against the bugs they guard: re-introducing the `refreshSelection()` call fails the bookmark-arrivals test, and removing the `viewportCenter` guard fails the republish test.
- `BookmarksViewModelTests`, `SearchResultsSheetViewTests`, and `SearchInteractorTests` pass unchanged, which is the evidence the extractions preserved the UIKit paths.
- SwiftLint clean — the four remaining violations are all in files this branch doesn't touch.
- `OBAKitTests` has two failures, `RentalFormatTests.rangeFallbackUsesAbbreviatedUnits` (compares `MKDistanceFormatter`'s abbreviated and spelled-out output, identical under a metric host locale) and `BackgroundAnnotationDeemphasisTests` (passes in isolation; cross-suite pollution). Both reproduce identically on the branch point with every change stashed, so they are pre-existing and unrelated. Note that the run summary prints "Test run … passed" even when suites fail — the exit code is the signal.
- Exercised on the simulator: stop and bookmark pins still render after the observer ownership move, and the sheet opens with its pinned search bar.

## Notes for review

- This branch is stacked on #1280, which is still open, so the diff against `main` includes that PR's work. #1280 needs to land first.
- The strings added for the new keys across the 12 localised languages are machine-assisted and would benefit from a native-speaker pass.

## Summary by CodeRabbit

- **New Features**
  - Added a redesigned home screen with nearby stops, recent stops, and bookmark previews.
  - Added interactive search with result counts, route and map-item details, nearby stops, and vehicle status handling.
  - Added bookmark pinning and unpinning with visible indicators and accessibility support.
  - Added route-stop views, map-item sharing, website links, and improved map result display.
- **Bug Fixes**
  - Improved loading, retry, cancellation, error messaging, navigation, and localization across search and trip details.
- **Accessibility & Localization**
  - Added translated labels, pluralized result counts, section hints, and empty-state messaging across supported languages.

### DIFF
--- a/OBAKit/Sheet/Content/Home/HomeBookmarksSectionModel.swift
+++ b/OBAKit/Sheet/Content/Home/HomeBookmarksSectionModel.swift
@@ -52,11 +52,16 @@ final class HomeBookmarksSectionModel: NSObject, ObservableObject, BookmarkDataD
 
         // `.bookmarksDidChange` may be posted off the main actor, so hop rather
         // than assume isolation. This mirrors the pattern in `MapStopsObserver`.
+        //
+        // Scoped to this application's own store rather than `object: nil`:
+        // `UserDefaultsStore` is the only poster and always posts `object: self`,
+        // so narrowing costs nothing in the app and keeps a concurrently-running
+        // test suite's store from driving this model.
         NotificationCenter.default.addObserver(
             self,
             selector: #selector(bookmarksDidChange),
             name: .bookmarksDidChange,
-            object: nil
+            object: application.userDataStore
         )
     }
 
@@ -66,17 +71,39 @@ final class HomeBookmarksSectionModel: NSObject, ObservableObject, BookmarkDataD
 
     /// Re-reads which bookmarks belong on screen and rebuilds the rows.
     ///
-    /// `findBookmarks(in:)` returns raw persisted order — only `bookmarksInGroup`
-    /// sorts — so the sort here is what makes the four shown match the user's
-    /// Manage Bookmarks ordering.
+    /// Every pinned bookmark is shown, then unpinned ones fill up to `limit`.
+    /// `limit` is therefore a *floor* on the section's size, not a cap: pinning
+    /// is the user saying "always keep this one here", so a pin is never squeezed
+    /// out by a newer bookmark. With nothing pinned this is just the `limit`
+    /// most recent.
+    ///
+    /// Within each half, most-recently-created first — a bookmark the user just
+    /// made should be visible. `sortOrder` can't do that job: `UserDataStore.add`
+    /// appends a new bookmark to the end of its group, so ordering by it puts the
+    /// newest one *last*, past the cut. It's also renumbered per group, so its
+    /// values aren't comparable across groups at all. It still breaks ties,
+    /// ascending: bookmarks stored before `dateCreated` existed all decode as
+    /// `.distantPast`, and among those the user's own Manage Bookmarks order is
+    /// the best signal left.
     func refreshSelection() {
-        selection = Array(
-            application.userDataStore
-                .findBookmarks(in: application.currentRegion)
-                .sorted { $0.sortOrder < $1.sortOrder }
-                .prefix(limit)
-        )
+        let all = application.userDataStore
+            .findBookmarks(in: application.currentRegion)
+            .sorted(by: Self.isOrderedBefore)
+
+        let pinned = all.filter(\.isPinned)
+        let unpinned = all.filter { !$0.isPinned }
+
+        selection = pinned + unpinned.prefix(max(0, limit - pinned.count))
         rebuildRows()
+    }
+
+    /// Newest first, with the user's manual order breaking same-date ties.
+    /// `static` so the ordering can be asserted directly in tests.
+    static func isOrderedBefore(_ lhs: Bookmark, _ rhs: Bookmark) -> Bool {
+        if lhs.dateCreated != rhs.dateCreated {
+            return lhs.dateCreated > rhs.dateCreated
+        }
+        return lhs.sortOrder < rhs.sortOrder
     }
 
     // MARK: - Bookmarks
@@ -84,10 +111,14 @@ final class HomeBookmarksSectionModel: NSObject, ObservableObject, BookmarkDataD
     /// `.bookmarksDidChange` may be posted off the main actor, so hop rather
     /// than assume isolation. Selector-based observation is auto-removed on
     /// dealloc, so no token/deinit needed.
+    ///
+    /// Goes straight to `loadIfNeeded()`, which re-reads the selection itself.
+    /// Calling `refreshSelection()` first would update `selection` *before*
+    /// `loadIfNeeded()` snapshots it, so the newly bookmarked stop would never
+    /// register as a selection change and its arrivals would never be fetched.
     @objc
     private nonisolated func bookmarksDidChange() {
         Task { @MainActor [weak self] in
-            self?.refreshSelection()
             self?.loadIfNeeded()
         }
     }
@@ -115,6 +146,15 @@ final class HomeBookmarksSectionModel: NSObject, ObservableObject, BookmarkDataD
         lastFetchedRegionID = regionID
         loader.loadData()
         return true
+    }
+
+    /// Flips `bookmark`'s pinned state.
+    ///
+    /// No local refresh: the store posts `.bookmarksDidChange`, which this model
+    /// already observes, so the reorder arrives through the same path as any
+    /// other bookmark edit rather than a second one that could drift from it.
+    func togglePin(_ bookmark: Bookmark) {
+        application.userDataStore.setPinned(!bookmark.isPinned, for: bookmark)
     }
 
     // MARK: - Row Building
--- a/OBAKit/Sheet/Content/Home/HomeBookmarksSectionModel.swift
+++ b/OBAKit/Sheet/Content/Home/HomeBookmarksSectionModel.swift
@@ -1,0 +1,125 @@
+//
+//  HomeBookmarksSectionModel.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import Foundation
+import OBAKitCore
+
+/// The home sheet's bookmarks preview: the first few bookmarks by the user's
+/// own ordering, with live arrivals for those few only.
+///
+/// Reuses `BookmarkDataLoader` — the same loader the Bookmarks tab uses — but
+/// scoped to the displayed bookmarks and with auto-refresh off, so this screen
+/// costs at most `limit` requests per activation and installs no polling timer.
+///
+/// Subclasses `NSObject` to adopt `BookmarkDataDelegate`.
+@MainActor
+final class HomeBookmarksSectionModel: NSObject, ObservableObject, BookmarkDataDelegate {
+
+    @Published private(set) var rows: [BookmarkRowViewModel] = []
+
+    /// The bookmarks a fetch is scoped to. Read by the loader's provider.
+    private(set) var selection: [Bookmark] = []
+
+    private let application: Application
+    private let limit: Int
+    private var loader: BookmarkDataLoader!
+
+    private var lastFetchDate: Date?
+    private var lastFetchedRegionID: Int?
+
+    init(application: Application, limit: Int = HomeSheetSection.itemLimit) {
+        self.application = application
+        self.limit = limit
+        super.init()
+
+        // The provider reads `selection` at fetch time rather than capturing a
+        // snapshot, so `refreshSelection()` before a load is enough to rescope
+        // the next batch.
+        self.loader = BookmarkDataLoader(
+            application: application,
+            delegate: self,
+            bookmarkProvider: { [weak self] in self?.selection ?? [] },
+            autoRefreshes: false
+        )
+
+        refreshSelection()
+    }
+
+    isolated deinit {
+        loader?.cancelUpdates()
+    }
+
+    /// Re-reads which bookmarks belong on screen and rebuilds the rows.
+    ///
+    /// `findBookmarks(in:)` returns raw persisted order — only `bookmarksInGroup`
+    /// sorts — so the sort here is what makes the four shown match the user's
+    /// Manage Bookmarks ordering.
+    func refreshSelection() {
+        selection = Array(
+            application.userDataStore
+                .findBookmarks(in: application.currentRegion)
+                .sorted { $0.sortOrder < $1.sortOrder }
+                .prefix(limit)
+        )
+        rebuildRows()
+    }
+
+    /// Fetches arrivals for the current selection, but only when something has
+    /// actually changed: the data is stale, the region moved, or the displayed
+    /// bookmarks differ. Returns whether a fetch was started.
+    ///
+    /// The sheet system tears sheet content down and rebuilds it without the
+    /// user navigating anywhere, so activation can fire repeatedly per visit —
+    /// this gate is what keeps that from becoming repeated network traffic.
+    @discardableResult
+    func loadIfNeeded(now: Date = Date(), staleAfter: TimeInterval = 30) -> Bool {
+        let previousIDs = selection.map(\.id)
+        refreshSelection()
+
+        let regionID = application.currentRegion?.regionIdentifier
+        let selectionChanged = previousIDs != selection.map(\.id)
+        let regionChanged = regionID != lastFetchedRegionID
+        let isStale = lastFetchDate.map { now.timeIntervalSince($0) > staleAfter } ?? true
+
+        guard selectionChanged || regionChanged || isStale else { return false }
+
+        lastFetchDate = now
+        lastFetchedRegionID = regionID
+        loader.loadData()
+        return true
+    }
+
+    // MARK: - Row Building
+
+    /// Rebuilds the row snapshots from the loader's current arrival data.
+    ///
+    /// `highlightedTripIDs` is always empty: the flash-on-change affordance
+    /// belongs to the polling Bookmarks tab, and nothing polls here.
+    private func rebuildRows() {
+        rows = selection.map { bookmark in
+            BookmarkRowViewModel(
+                bookmark: bookmark,
+                arrivalDepartures: arrivalDepartures(for: bookmark),
+                highlightedTripIDs: [],
+                hasLoadedArrivalData: loader.hasFetchedData(forStopID: bookmark.stopID)
+            )
+        }
+    }
+
+    private func arrivalDepartures(for bookmark: Bookmark) -> [ArrivalDeparture] {
+        guard let key = TripBookmarkKey(bookmark: bookmark) else { return [] }
+        return loader.dataForKey(key)
+    }
+
+    // MARK: - BookmarkDataDelegate
+
+    func dataLoaderDidUpdate(_ dataLoader: BookmarkDataLoader) {
+        rebuildRows()
+    }
+}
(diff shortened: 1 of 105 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. `MapStopsObserver.reset()` writes `viewportCenter = nil` above the `guard !stops.isEmpty` idempotence guard, so every zoomed-out map settle republishes. `@Published` has no equality check, so assigning `nil` over `nil` still fires `objectWillChange` — and `MapPanelRootView.onMapCameraChange(frequency: .onEnd)` calls `reset()` on *every* settle while zoomed out, not just on the zoom-out transition. That invalidates `MapPanelRootView` (now an `@ObservedObject` on this instance) and cascades into `HomeNearbyStopsSectionModel`'s `combineLatest` sink → `HomeSheetViewModel.objectWillChange` → `HomeSheetView`. This is the same no-republish invariant the PR carefully protects two functions below in `updateViewport(_:)` ("Comparing by hand is what keeps a re-fired settle from firing `objectWillChange`"), and the same invariant the pre-existing `guard !stops.isEmpty else { return }` has enforced since the observer was introduced (95afddc). Moving the clear below the guard, or guarding it with `if viewportCenter != nil`, restores it. `Repeat settle at the same center does not republish` covers `updateViewport` but there is no equivalent test for a repeated `reset()`.

https://github.com/OneBusAway/onebusaway-ios/blob/5d8dd3c8d3d8c203a0fe7de91b5836b613a204be/OBAKit/Sheet/Root/MapStopsObserver.swift#L108-L116

OBAKit/Sheet/Root/MapStopsObserver.swift
```diff
@@ -105,15 +110,25 @@ final class MapStopsObserver: NSObject, ObservableObject, MapRegionDelegate, Reg
     func reset() {
         accumulated.removeAll()
         viewport = nil
+        viewportCenter = nil
```
aaronbrethorst: `viewportCenter` is `@Published`, and this assignment sits above the `guard !stops.isEmpty else { return }` on the next line. `@Published` performs no equality check, so every `reset()` republishes even when the value is already nil.

`MapPanelRootView.onMapCameraChange(frequency: .onEnd)` calls `reset()` on the zoomed-out branch of every settle with no transition gate, so each pan while zoomed out fires `objectWillChange` and cascades through `MapPanelRootView` body eval → `HomeNearbyStopsSectionModel` → `HomeSheetViewModel` → `HomeSheetView`.

`updateViewport(_:)` just below hand-rolls a coordinate comparison to avoid exactly this, and its comment explains why. Move this assignment below the guard, or guard on it the same way.

aaronbrethorst (changes requested):
There's a lot of careful work here and I want to name it before the two problems, because neither is about the quality of the feature. The state ownership is deliberate and right — `@StateObject` with an autoclosure for the view model, `@ObservedObject` for the observer that's passed in, explicit child-to-parent `objectWillChange` forwarding rather than hoping it propagates. `Bookmark.dateCreated` and `isPinned` both decode with `decodeIfPresent` so existing stored bookmarks survive, and you wrote tests for the legacy-decode path. All six new keys are present in en plus all 12 locales. The `bookmarkProvider`/`autoRefreshes` seams on `BookmarkDataLoader` are a clean way to scope it without forking it.

Two things to resolve.

**1. This branch is stacked on #1280, but targets `main`.** The first dozen commits here are #1280's search-integration work, so merging this PR as it stands would silently land #1280 too — and #1280 currently has changes requested from me. The diff GitHub shows for this PR isn't the diff you're asking me to approve.

Retarget the base to `feature/map-panel-search-integration` so GitHub renders only the incremental diff, or hold this until #1280 lands and then rebase. Either is fine; what I can't do is approve a PR whose merge would bring in work that's still under revision. (I reviewed only the incremental range, so nothing below is about #1280's code.)

**2. `MapStopsObserver.reset()` breaks the no-republish invariant its sibling method goes out of its way to maintain.**

`viewportCenter` is `@Published`, and `reset()` clears it *above* the `guard !stops.isEmpty else { return }`. `@Published` does no equality check, so every call republishes even when the value is already nil. `MapPanelRootView.onMapCameraChange(frequency: .onEnd)` calls `reset()` on the zoomed-out branch of every settle, with no transition gate — so each pan while zoomed out fires `objectWillChange`, which cascades into a `MapPanelRootView` body evaluation, the `HomeNearbyStopsSectionModel` sink, `HomeSheetViewModel`, and `HomeSheetView`.

You clearly know about this failure mode: `updateViewport(_:)` hand-rolls a coordinate comparison for exactly this reason, with a comment explaining that `CLLocationCoordinate2D` isn't `Equatable` so `@Published` can't drop the write for you. `reset()` just needs the same treatment — moving the assignment below the guard, or guarding on it, is a two-line fix.

Below the blocking line, for whenever you're back in here:

- **Bookmark arrival failures have no path out.** `loadIfNeeded()` stamps `lastFetchDate = now` before the fetch resolves and regardless of outcome, and `BookmarkDataLoader.lastBatchHadError` is never surfaced. A failed batch leaves cards in "Loading…" for at least the 30s staleness window with no retry affordance and no error copy. The Bookmarks tab has polling and pull-to-refresh to bail you out; this sheet has neither.
- **Empty-state flash on cold open.** `visibleSections` starts empty before the first map settle and before recents/bookmarks load, so "Nothing Here Yet" can paint for a frame or two before the List replaces it.
- `activate()` assigns `stops`/`rows` unconditionally, firing two `objectWillChange` sends per activation even when nothing changed. Same class as the blocker, just not in a hot loop.
- `.environment(\.obaFormatters, …)` is applied to a `Section` here, where `BookmarksListView` applies it to the `List`. Propagation should work and the key has a sensible default, so worst case is default formatters — but the inconsistency is worth resolving one way or the other.
- `AppSheetRoute.homeCollapsedHeight` goes 80 → 75 with no explanation, bundled into a feature PR. The test moves in lockstep so it's clearly intentional; just say why.

Sort out the stacking and the `reset()` republish and I'll take another pass.

OBAKit/Sheet/Root/MapStopsObserver.swift
```diff
@@ -105,15 +122,39 @@ final class MapStopsObserver: NSObject, ObservableObject, MapRegionDelegate, Reg
...
+        if !hasSettledOnce {
+            hasSettledOnce = true
+        }
+        // Guarded for the same reason `updateViewport(_:)` hand-rolls its
+        // comparison: `CLLocationCoordinate2D` isn't `Equatable`, so `@Published`
+        // won't drop a nil-over-nil write. `MapPanelRootView` calls `reset()` on
+        // *every* settle while zoomed out, not just on the zoom-out transition,
+        // so an unguarded clear would fire `objectWillChange` on each pan and
+        // cascade into the nearby section and the whole home sheet.
+        if viewportCenter != nil {
```
aaronbrethorst, line 134: Verified this is the fix and not just a plausible-looking one.

`viewportCenter != nil` is an `Optional` nil-check rather than an `Equatable` comparison, so it compiles and means something despite `CLLocationCoordinate2D` not being `Equatable` — which is the trap that made the original bug possible. Walking `reset()` for the repeat-pan-while-zoomed-out case at head: `accumulated.removeAll()` isn't published, `viewport` is a plain stored property, `hasSettledOnce` is already latched, `viewportCenter` is already nil so no write, and `stops` is already empty so the guard returns. Zero `objectWillChange` sends. The cascade into `MapPanelRootView` → nearby section → home sheet is gone.

And the choice to guard rather than move is load-bearing in a way I didn't appreciate when I offered them as equivalents: your pre-existing `Reset clears the viewport center` test never loads any stops, so the move-it-below variant would fail that test outright. The constraint was already encoded in the suite.

OBAKit/Sheet/Content/Home/HomeBookmarksSectionModel.swift
```diff
@@ -0,0 +1,221 @@
...
+    /// batch left the rows stuck in their "Loading…" state for the rest of the
+    /// staleness window, and this sheet has neither the polling timer nor the
+    /// pull-to-refresh that bail the Bookmarks tab out of the same situation.
+    ///
+    /// `lastFetchedRegionID` is deliberately left alone: `isStale` is enough to
+    /// reopen the gate, and clearing the region too would make the next
+    /// successful fetch look like a region change to any future caller.
+    func dataLoader(_ dataLoader: BookmarkDataLoader, isLoadingChanged isLoading: Bool) {
+        guard !isLoading, dataLoader.lastBatchHadError else { return }
+        lastFetchDate = nil
```
aaronbrethorst, line 219: The retry works and the ordering it relies on is correct — I checked that `lastBatchHadError` is set inside the `MainActor.run` that completes before `taskFinished` schedules its Task, so it's already current when `isLoadingChanged: false` lands. Leaving `lastFetchedRegionID` alone is right too; clearing it would make the next successful fetch read as a region change.

One thing I'd think about, not blocking: clearing the stamp outright means the *very next* trigger refetches, with no backoff. The sheet system rebuilds content without the user navigating (your own comment upstream in this file says so), so on a persistently failing network a rider underground with three trip bookmarks who taps into a stop and backs out a few times gets a refetch and an `ErrorBulletin` per cycle. Stamping `lastFetchDate = now - staleAfter + 5` instead would keep the retry while bounding the rate.

I'm not holding the merge for it because the Bookmarks tab already behaves this way on its 30s poll — so this is consistent with the app rather than a new class of misbehavior. Just flagging it as a real edge rather than a hypothetical.

OBAKit/Sheet/Content/Home/HomeNearbyStopsSectionModel.swift
```diff
@@ -0,0 +1,55 @@
...
+    /// sink above rather than duplicated here.
+    private func rebuild(stops: [Stop], center: CLLocationCoordinate2D?) {
+        guard let center else {
+            // No settle yet. The observer's id ordering is arbitrary but stable,
+            // which beats rendering an empty section for the frame or two before
+            // the first camera settle lands.
+            self.stops = Array(stops.prefix(limit))
+            return
+        }
+        self.stops = Stop.nearest(stops, to: center, limit: limit)
```
aaronbrethorst, line 53: This is the same defect class you just fixed in the other two section models, still open in this one.

`rebuild(stops:center:)` assigns `self.stops` unconditionally in both branches — the `guard let center else` path at line 51 too. `self.stops` is `@Published`, and `Stop.nearest(stops, to: center, limit: 4)` can return an identical top-four when new stops arrive for an unchanged viewport: `MapStopsObserver.publish()` fires `$stops` while `$viewportCenter` stays put, `combineLatest` re-emits, and the same four stops get republished into a full `HomeSheetView` body evaluation for nothing.

`[Stop]` already has the working element-wise `Equatable` your other two guards depend on, so this is the same two-line shape as the fixes in this commit. Not blocking — just the one place the round didn't reach, and leaving it inconsistent is how the pattern erodes.

aaronbrethorst (approved):
Both things I asked for are fixed, and fixed properly — you went after the mechanism rather than the symptom in each case. Approving.

**The `reset()` republish is closed, and you made the better of the two available choices.** I'd offered "move the assignment below the guard, or guard on it" as if they were equivalent. They aren't, and your commit message says why: moving it below `guard !stops.isEmpty` would strand a non-nil `viewportCenter` on a settle that loaded no stops — a state `updateViewport` can genuinely produce, since it sets the center on every settle regardless of what loads. Guarding was correct. The comment you left pins it to the same `CLLocationCoordinate2D`-isn't-`Equatable` reasoning as its sibling method, so the two now read as one deliberate pattern instead of two accidents.

**The equality guards in item 3 were the risky part of this round, and they hold.** That was the change most likely to go quietly wrong — an equality comparison over a reference type that silently never matches (no fix) or always matches (much worse than the bug). I checked both: `Stop` is an `NSObject` with a hand-written `isEqual` over ten value fields, and `BookmarkRowViewModel.==` is hand-written too, deliberately excluding the `bookmark` reference and comparing all eleven copied fields — `hasLoadedArrivalData` among them. That last one matters: it means a fetch that returns byte-identical arrivals still republishes on the false→true flip, so the "Loading…" → real-content transition can't be suppressed by the guard. That's the failure I was worried about and it isn't there.

**The bookmark-failure retry is real** and the ordering it depends on is sound — `lastBatchHadError` is set inside the `MainActor.run` that completes before `taskFinished` schedules, so it's current when the delegate fires. I traced every caller of `loadIfNeeded()` looking for a retry loop and there's no self-sustaining path. One rate note inline.

**On the stale-CI question, which applies to this PR more than the others.** Your run is green on the current head, attempt 1 — but it was created at 19:11Z on the 20th and `main` moved at 00:26Z on the 21st. Rather than trust that, I merged #1280, #1288, and #1289 onto current `main` locally and built and ran the suite against the integrated result: **2277 tests in 246 suites passed.** So this is verified against the base it's actually landing on, not the one it was tested against.

**The stacking objection is resolved by sequencing rather than by retargeting.** I've just merged #1280, so this PR's diff is now the incremental range I reviewed — which is what I actually wanted. I confirmed `pr-1280..pr-1288` is 29 commits with zero merge commits, strictly linear, so nothing from #1280 sneaks in unreviewed.

Three inline notes, all nits, none blocking — one of them (the third) is the same defect class you just fixed twice, sitting in the one section model the round didn't touch. Worth closing for consistency whenever you're next in there.

Nice work on the regression tests, incidentally. Counting `objectWillChange` sends with a positive control alongside each guard is exactly right — it's what makes "this test would pass with the feature deleted" impossible to claim.

PR #1289, opened 2026-08-15
Home sheet section indexes: Nearby, Recent Stops, and Bookmarks

mosliem:
Builds the three index screens the home sheet's section headers navigate to — Nearby Stops, Recent Stops, Bookmarks. #1288 wired those routes and left them rendering a "coming soon" placeholder; this replaces all three with native SwiftUI sheets, consolidates a duplicate nearby-stops screen, and extracts the Bookmarks tab's row actions so the tab and the new sheet share one implementation. The classic UIKit path is untouched.

The governing constraint was reuse: every screen is built on the view model its UIKit counterpart already uses, so no index screen introduces a second way to load the same data.

## What's included

**Nearby Stops index.** `NearbyStopsSheetView` over the existing `NearbyStopsViewModel` — a fresh `getStops(coordinate:)` fetch grouped into direction sections with a search field, which is a native port of `NearbyStopsViewController`. Grouping and filtering live in `NearbyStopsIndexSection.sections(stops:filter:)`, a pure function, so the rules are testable without a view.

**One nearby screen instead of two.** `.nearbyStops(coordinate:)` already served this exact screen through `NearbyStopsSheetHost`, a `UIViewControllerRepresentable`. Both routes now render `NearbyStopsSheetView` and the host is deleted. `.nearbyAll` carries no coordinate — it's pushed from a section header, not a tapped place — so `NearbyCoordinateResolver` resolves one: the map's last settled viewport centre, then the device location, then the current region's centre. All three nil renders an empty state rather than fetching stops around (0, 0).

**Recent Stops index.** `RecentStopsSheetView` over `RecentStopsViewModel`, with `HomeStopRow` rows, a search field, swipe-to-delete, and a Delete All toolbar button behind the Recent tab's existing confirmation copy. The tab's Alarms section is deliberately absent: the home sheet's header promises recent *stops*, and alarm deep-links have no place in the sheet stack.

**Bookmarks index.** Renders the existing `BookmarksListView` with a sheet-flavoured `BookmarksNavigationHandler`, so group sections, collapse state, context menus, and row rendering cannot drift from the Bookmarks tab. Two closures differ by design: a row tap stacks `.stopDetails` on the sheet coordinator rather than pushing through `viewRouter`, and the bookmark editor is presented from a SwiftUI `.sheet` rather than `viewRouter.present`. `viewModel.start()`/`deactivate()` are bound to appear/disappear so the tab's 30-second poll never outlives the sheet.

**`BookmarkActions`.** The Bookmarks tab's row actions — delete-with-analytics, the Live Activity start path with its duplicate guard, and bookmark-editor construction — moved out of `BookmarksViewController` into a `@MainActor` type both surfaces use. `startLiveActivity` returns a `TrackResult` rather than alerting, because `showLiveActivityErrorAlert()` is a `UIViewController` extension: the tab keeps raising its `UIAlertController`, the sheet raises a SwiftUI alert. Arrivals are passed in, so the type needs no view-model reference, and `makeBookmarkEditor` builds the editor while presentation stays with the caller. `updateRunningLiveActivities`, `reloadWidget`, and the sort-menu rebuild stay on the controller — those are tab and app-lifecycle chores, not row actions.

**A recent-stops duplication bug, fixed at the source.** `addRecentStop` and `delete(recentStop:)` matched with `firstIndex(of:)`, which uses `Stop.isEqual` — a ten-field comparison covering the route list, name, code, and coordinates. The same stop re-viewed after any of those drifted server-side failed to match its stored copy, so adding stored a second entry under one `id` and deleting silently removed nothing. All three entry points now key on `id`, as bookmarks already do via `findBookmark(id:)`. This predates these screens; the UIKit Recent tab renders through `OBAListView`, whose item identity isn't `id` alone, so it drew two rows without complaint. SwiftUI's `ForEach(id: \.id)` is the first consumer that checks, and it reported `the ID 1_660 occurs multiple times within the collection`.

**Widget freshness, moved off the call sites.** `BookmarkWidgetRefresher` observes `.bookmarksDidChange` and reloads the widget's timeline from wherever the bookmark set changed. Five surfaces now perform those writes — the Bookmarks tab, this index sheet, the home sheet's bookmarks section, Manage Bookmarks, and the stop page's add-bookmark flow — and asking each to remember a `WidgetCenter` call is how they came to disagree. `Application` owns it eagerly, because the widget must still refresh when the screen that made the change has already gone. It doesn't over-fire: `upsert(bookmark:)`, the path `addRecentStop` takes, posts nothing, so browsing stops doesn't burn the reload budget. `BookmarksViewController` separately reloads on each arrival batch — that keeps the *times* current; this covers the bookmark set itself.

## Notable decisions

**All three index sheets wear the same chrome.** `NavigationStack`, inline title, a Close button driving `@Environment(\.dismiss)`, `.searchSheetBackground()` outside and `.searchListChrome()` on the list — the pattern `RouteStopsSheetView` and `SearchResultsSheetView` established. The chrome is repeated per view rather than factored into a shared modifier: three call sites with three different toolbars don't yet justify the indirection.

**`NearbyStopsSheetView` splits in two.** A `@StateObject` cannot be constructed conditionally, and there is no honest coordinate to build a view model with when the map hasn't settled, the device has no fix, and there is no current region. The outer view owns the chrome and the no-anchor case; a private inner view owns the view model, so it is only built when a coordinate exists.

**Errors on the Nearby index surface inline, not through `application.displayError`.** The UIKit controller raises an app-level alert; over a stacked sheet that is the wrong affordance, and the view holds no `Application` to raise one with. Loading and failure render as list states with a retry action.

**`BookmarksListView` gained a `Presentation` parameter rather than a flag per difference.** The tab gets a plain list with pull-to-refresh; the sheet gets inset-grouped chrome matching the other two indexes and no pull-to-refresh, because a downward drag there is already the sheet's own dismiss gesture. It refreshes from a toolbar button instead, driving the same `navigation.refresh()` closure so the fetch and its completion haptic are identical. The parameter defaults to `.tab`, so the Bookmarks tab needed no edit.

**The refresh spinner is keyed to an explicit tap, not `viewModel.isLoading`.** The background poll also raises `isLoading`; binding the toolbar to it would pop a spinner into view every 30 seconds without the user asking for anything. The spinner is held for a 300 ms minimum so a cache-served batch is still visibly acknowledged rather than blinking.

**`recentStops(in:)` dedupes on read as well as on write.** Fixing `addRecentStop` alone would only heal a store as the user revisits each affected stop. Collapsing duplicates on read makes an already-corrupted store render correctly immediately, which matters because a repeated `ForEach` id is undefined behaviour rather than a cosmetic issue.

**`AppSheetViewFactory.indexPlaceholderView` survives with no callers among the routes.** No index route dispatches to it any more, but `unimplementedView` still calls it on the release path, where it is the visible "coming soon" body rather than a blank sheet.

## Verification

- 22 new tests: direction grouping and filtering (including blank-filter and no-match cases), the coordinate fallback chain including all-nil, recent-stop filtering and both delete paths, `BookmarkActions`' delete analytics and content-state construction, and the Bookmarks sheet's handler wiring — that a row tap pushes `.stopDetails` on the coordinator, that pinning writes through the store, and that a failed Track reaches the alert callback. Two more came with the review follow-ups: the editor's Cancel button reaching its delegate, and the nearby anchor falling back to the device fix.
- Three of those are regression tests written against the duplicate-ID bug before it was fixed, and verified failing first: `recents.count → 2` for one stop id, and a delete that removed nothing.
- The `BookmarkActions` extraction was reviewed member by member against the removed bodies for semantic identity, and `OBAKitTests` passes unchanged around it — that is the evidence the Bookmarks tab is behaviourally untouched.
- SwiftLint clean; the five remaining violations are all in files this branch doesn't touch.
- `OBAKitTests`: 2272/2274, rebased onto #1288's current tip. The two failures are `RentalFormatTests.rangeFallbackUsesAbbreviatedUnits` and `BackgroundAnnotationDeemphasisTests`, both of which #1288 also documents as pre-existing. They live in Mapping and Formatters code that none of this branch's commits touch. Note that the run summary prints "Test run … passed" even when suites fail — the exit code is the signal.

## Review follow-ups

- **The bookmark editor's Cancel path.** `EditBookmarkViewController.close()` dismissed itself and told nobody, so `BookmarkEditorHost.Coordinator.bookmarkEditorCancelled` was unreachable for the edit path and `editingBookmark` stayed non-nil after UIKit had torn the sheet down. Fixed at the source: `close()` now notifies the delegate the way `AddBookmarkViewController.cancel()` does, with a dismiss fallback for a delegate that has been deallocated. All four existing UIKit delegates already respond by dismissing, so the tab, stop page, and trip page are behaviourally unchanged — and a new test fires the actual Cancel bar button to prove the callback lands.
- **The Nearby index's empty-state flash.** The spinner was guarded on `viewModel.isLoading`, which is still `false` on the first render because `.task` hasn't run — so "No Nearby Stops" painted for a frame. It now also waits on `hasCompletedFirstLoad`.
- **`BookmarksSheetView`'s doc comment** no longer claims pull-to-refresh is shared with the tab; it names `Presentation.sheet` as the second deliberate difference alongside the navigation handler.
- **`AppSheetViewFactoryTests`.** `try? #require` couldn't abort anything, and the nil-anchor test branched past the case it was named for — a test `Application` is built with a fixed region, so `currentRegion` is never nil and the nil branch was dead. Now `try #require` throughout, and two deterministic tests cover the second and third links of the chain. The all-nil case stays in `NearbyCoordinateResolverTests`, where it is reachable.

## Notes for review

- This branch is stacked on #1288, which is still open, so the diff against `main` includes that PR's work. #1288 needs to land first. Retargeting the base isn't available here: #1288's branch lives on the fork, and GitHub only offers base branches from the base repository. Reviewing the incremental range is the only way to see the real diff until the stack lands.
- Live Activities do not start in the Simulator, so the Track paths are covered only in their pure parts — identity keys, content-state construction, and the no-arrivals failure. Starting and promoting an activity still needs a device pass, as does the Bookmarks tab after the extraction.
- The three screens have not been exercised by hand yet: opening each index from its header, stacking stop details over one, the bookmark editor presenting over a stacked sheet, and the new toolbar refresh are all verified by static review and unit tests only.


aaronbrethorst:
### Code review

Found 1 issue:

1. Cancelling the bookmark editor from the Bookmarks index sheet leaves `editingBookmark` non-nil. `EditBookmarkViewController`'s Cancel button is wired to `close()`, which calls `dismiss(animated:)` directly and never notifies its delegate — only `AddBookmarkViewController` calls `bookmarkEditorCancelled`. So `Coordinator.bookmarkEditorCancelled` is unreachable for this editor: UIKit tears the presented sheet down while SwiftUI's `.sheet(item: $editingBookmark)` binding still holds the bookmark, so the state desyncs (re-tapping Edit on that same bookmark won't reopen the editor, since `.sheet(item:)` sees no identity change) and `viewModel.rebuildSections()` never runs on cancel. The Bookmarks tab is unaffected — nothing there depends on the delegate firing for the cancel path. The `onFinish` doc comment ("Fires when the editor is dismissed, whether saved or cancelled") describes the intended behaviour, not the actual one; the save path does work.

https://github.com/OneBusAway/onebusaway-ios/blob/216f07b8a3496dce889a1d10585abb55babe724d/OBAKit/Sheet/Content/Home/Index/BookmarkEditorHost.swift#L47-L51

aaronbrethorst (changes requested):
The architecture here is sound and the ports are faithful — I compared `NearbyStopsIndexSection.sections` against `NearbyStopsViewController.items(for:)` and the direction-grouping and filtering behavior match. Pulling the grouping out as a pure function is what makes that checkable, which is the point. The `UserDataStore` recent-stops fix is a genuine bug fix on its own merits: keying on `id` instead of full `Stop` equality is right, since a stop whose name or route list changed server-side was previously appended as a duplicate. Localization is clean — one new key, present in all 13 locales, and every other `OBALoc` reuses an existing key with a matching value and comment.

**1. It's stacked on #1288, which I've just requested changes on.** Same situation as that PR: this branch targets `main` but carries #1288's (and therefore #1280's) commits, so merging it as-is would land all three. I reviewed only the incremental range. Retarget the base to #1288's branch so GitHub shows the real diff, or hold until the stack lands underneath you.

**2. The bookmark editor's Cancel path leaves the SwiftUI sheet binding stuck.** `EditBookmarkViewController.close()` calls `dismiss(animated:)` and nothing else — it never invokes `delegate?.bookmarkEditorCancelled(self)`. Only `AddBookmarkViewController` does that (line 133). So `BookmarkEditorHost.Coordinator.bookmarkEditorCancelled` is dead code for the edit path, `onFinish` never runs on cancel, and `editingBookmark` stays non-nil after UIKit has already torn the sheet down.

This is a pre-existing gap in `EditBookmarkViewController` — the existing UIKit callers implement the method but don't depend on it for state cleanup, so nothing broke before. Your new host does depend on it, which is what turns a latent inconsistency into a live bug. Either have `close()` notify its delegate the way `AddBookmarkViewController` does, or have `BookmarkEditorHost` observe dismissal directly. The first is tidier and fixes the inconsistency at its source.

Smaller notes:

- **`BookmarksSheetView`'s doc comment contradicts the code below it.** It claims `BookmarksListView` is rendered "unchanged, so group sections, collapse state, pull-to-refresh, and the row context menu can't drift from the Bookmarks tab" — but this PR adds the `Presentation` parameter specifically to remove pull-to-refresh on the sheet. Drop that item from the list.
- **Three `try? #require(...)` calls in `AppSheetViewFactoryTests` discard the throw**, so they can't abort the test. And `Nearby all view passes nil when nothing can anchor the search` branches on `application.currentRegion == nil`, so only one of its two paths actually exercises the nil-anchor case it's named for.
- **Empty-state flash on the Nearby index**: `list` guards the spinner on `viewModel.isLoading && stops.isEmpty`, but on first render `isLoading` is still `false` because `.task` hasn't run — so "No Nearby Stops" paints briefly before the spinner.
- `BookmarkWidgetRefresher`, its `Application` wiring, and its five tests are new here but unmentioned in a description that otherwise documents every decision. Worth a sentence. I did check it doesn't over-fire: `upsert(bookmark:)` — the path `addRecentStop` takes — doesn't post `.bookmarksDidChange`, so browsing stops won't burn the WidgetKit reload budget.

Thank you for stating plainly in the description that the three screens haven't been hand-tested. That's the right thing to disclose and it shaped where I spent attention.

mosliem:
Thanks — all six addressed, and the branch is rebased onto #1288's current tip (`7a56a221`). The incremental diff came through the rebase byte-identical apart from line offsets, so nothing in the earlier review has silently shifted underneath you.

**1. The stack / retarget.** I can't retarget this one: #1288's branch lives on my fork, and GitHub only offers base branches from the base repository, so `feature/home-sheet-map-panel` never appears in the picker. Holding until #1288 lands is the plan — the rebase at least means the incremental range you reviewed is now the range that sits on top of #1288's tip rather than an older copy of it. I've noted the constraint in the description.

**2. The bookmark editor's Cancel path.** Fixed at the source, as you suggested. `EditBookmarkViewController.close()` now notifies `bookmarkEditorCancelled` instead of dismissing silently:

```swift
@objc private func close() {
    guard let delegate else {
        dismiss(animated: true, completion: nil)
        return
    }

    delegate.bookmarkEditorCancelled(self)
}
```

That makes `close()` consistent with `save()`, which already routes dismissal through the delegate and never calls `dismiss` itself. I checked all four existing implementors — `BookmarksViewController`, `StopViewController`, `StopPageActionPresenter`, `TripPageViewController` — and every one responds by calling `viewController.dismiss(animated:)`, which is exactly what `close()` used to do inline, so the tab, stop page, and trip page are behaviourally unchanged. The `guard` covers a delegate that has been deallocated: the weak reference means nobody would be left to dismiss us, and `AddBookmarkViewController` forwards a weak delegate through to this controller.

There's a regression test in `BookmarkActionsTests` that builds the editor through `BookmarkActions.makeBookmarkEditor` with a spy delegate and fires the Cancel button's own target/action rather than reaching for the private method — so the bar-button wiring is under test too.

**3. `BookmarksSheetView`'s doc comment.** You're right, it contradicted the `Presentation` parameter three lines down. Rewritten to name both deliberate differences instead of claiming the list is rendered unchanged.

**4. `AppSheetViewFactoryTests`.** Both parts fixed. `try? #require` is gone — the tests are `throws` and use `try #require`. On the nil-anchor branch: you're right that it never ran, and digging in, it *can't* — `buildApplication` passes `fixedRegionName`, so `currentRegion` is never nil in a test `Application` and the nil arm was dead code. Rather than fake a region-less application, I split it into two deterministic tests covering the links that are reachable from the factory: the device fix (second link) and the region centre (third). The all-nil case stays in `NearbyCoordinateResolverTests`, which is where it's honestly assertable.

One wrinkle worth flagging: the device-location test needed `Fixtures.stubAllAgencyAlerts` first. Publishing a fix sends the app off to fetch agency alerts, and `MockDataLoader` traps unstubbed requests with a `fatalError` — which took the whole test process down, not just that test, and silently truncated the run by ~97 tests. Caught it because the totals moved.

**5. Empty-state flash on the Nearby index.** Confirmed and fixed. The view now tracks `hasCompletedFirstLoad`, set after `.task` returns, and the spinner covers the window before the fetch has started:

```swift
} else if !hasCompletedFirstLoad || (viewModel.isLoading && viewModel.stops.isEmpty) {
```

The error branch is still evaluated first, so the no-API-service path — which returns before `isLoading` is ever raised — still renders its error rather than spinning forever.

**6. `BookmarkWidgetRefresher`.** Fair catch, it deserved a paragraph. Added one to the description covering why it observes the store rather than being called from each write site, why `Application` owns it eagerly, and the `upsert(bookmark:)` point you'd already verified.

**Verification.** `OBAKitTests` 2272/2274 on iPhone 16 / iOS 26. The two failures are the same pre-existing `RentalFormatTests.rangeFallbackUsesAbbreviatedUnits` and `BackgroundAnnotationDeemphasisTests` #1288 documents. SwiftLint clean — five violations, all in files this branch doesn't touch. The three screens still haven't had a hand pass, and Live Activities still need a device.

OBAKit/Bookmarks/EditBookmarkViewController.swift
```diff
@@ -170,8 +170,20 @@ class EditBookmarkViewController: FormViewController, AddGroupAlertDelegate {
...
     // MARK: - Actions
 
+    /// Cancels the editor by telling the delegate, the way `save()` and
+    /// `AddBookmarkViewController.cancel()` both do.
+    ///
+    /// Dismissing here directly would leave the delegate believing the editor is
+    /// still up, which matters to any presenter that keeps its own state — a
+    /// SwiftUI `.sheet(item:)` binding, for one. Every delegate dismisses us in
+    /// response; the fallback covers a delegate that has been deallocated.
     @objc private func close() {
```
aaronbrethorst, line 180: Notify-only rather than notify-then-dismiss is the right call here — it lets the SwiftUI `.sheet(item:)` binding own the teardown instead of racing UIKit for it — and the nil-delegate fallback keeps the old behavior when the delegate has gone away.

I verified the blast radius rather than assuming it, since this is a shared controller with four other presenters that have never received this callback for the edit path. `BookmarksViewController`, `StopViewController`, `StopPageActionPresenter`, and `TripPageViewController` all implement `bookmarkEditorCancelled` as exactly `viewController.dismiss(animated:)` on the passed-in controller — which is `self`, the same receiver the old `close()` used. So the delegate hop is call-for-call identical for all four; only `BookmarkEditorHost.Coordinator` behaves differently, which is the point of the change.

Also confirmed `save()` can't reach here: `close()` has exactly one reference (the Cancel bar button wired in `init`), and the save path goes through `notifyEdited` to `bookmarkEditor(_:editedBookmark:isNewBookmark:)`. A successful save can't report a cancel.

The one case worth naming: `AddBookmarkViewController` pushes an editor onto its own nav stack with the *outer* delegate, so cancelling there now routes the dismiss through that delegate instead of calling it on `self`. UIKit forwards a dismiss from a non-presenting controller to the presenting ancestor, so the modal goes away exactly as before. Fine — just not obvious from the diff.

OBAKit/Sheet/Content/Home/Index/NearbyStopsSheetView.swift
```diff
@@ -0,0 +1,172 @@
...
+            EmptyStateView(
+                title: Strings.error,
+                description: error.localizedDescription,
+                systemImage: AppSymbol.error
+            ) {
+                Button(Strings.retry) {
+                    Task { await viewModel.loadStops() }
+                }
+            }
+        } else if !hasCompletedFirstLoad || (viewModel.isLoading && viewModel.stops.isEmpty) {
```
aaronbrethorst, line 127: This closes the flash, and it answers my "can the spinner hang forever?" question structurally rather than with a flag, which is the better outcome.

I'd worried about the nil-anchor path — `NearbyCoordinateResolver` returning nil means no fetch runs, so a naive latch would never flip and the spinner would sit there. But `hasCompletedFirstLoad` lives in `NearbyStopsSheetContent`, which the parent only builds inside `if let coordinate`; on the nil path the parent renders `EmptyStateView` directly and this view never exists. The spinner can't hang because the code holding it isn't instantiated.

For the coordinate-present case the flag is set unconditionally after `loadStops()` returns, and `loadStops` has no non-returning path — the missing-`apiService` guard sets `operationError` and returns into the error branch above this one, and the network path is `defer`-protected. So the only way this stays up is a genuinely in-flight fetch, which is what it should mean.

aaronbrethorst (approved):
Both items are fixed, and the one I was most nervous about — reaching into a shared UIKit controller that four other surfaces already depend on — turned out to be genuinely safe. Approving.

**The `EditBookmarkViewController.close()` fix is the right shape.** You made it notify-only rather than notify-then-dismiss, which is what the SwiftUI binding actually needs: the coordinator's `onFinish()` clears `editingBookmark` and lets SwiftUI own the teardown, so there's no race where UIKit and SwiftUI both tear the sheet down. The nil-delegate fallback keeps the old behavior for a deallocated delegate. It also brings this in line with `AddBookmarkViewController.cancel()`, which is where the inconsistency started.

**I checked the blast radius myself rather than reasoning about it.** The worry with a change like this is that four presenters suddenly start receiving a callback they've never received for the edit path. I read all five conformers: `BookmarksViewController`, `StopViewController`, `StopPageActionPresenter`, and `TripPageViewController` each implement it as exactly `viewController.dismiss(animated:)` on the passed-in controller — which *is* `self`, the same receiver and the same call the old `close()` made directly. So the new delegate hop is call-for-call identical for every pre-existing caller; only `BookmarkEditorHost.Coordinator` behaves differently, which is the entire point of the change. Zero regression surface.

I also confirmed `close()` can't be reached from a successful save. It has one reference in the file — the Cancel bar button wired in `init` — and `save()` runs a separate `notifyEdited` path to `bookmarkEditor(_:editedBookmark:isNewBookmark:)`. A save cannot report a cancel.

The one subtle case is `AddBookmarkViewController`, which pushes an `EditBookmarkViewController` onto its own nav stack with the *outer* delegate. Cancelling there now routes the dismiss through that delegate instead of calling it on `self` — and since UIKit forwards a `dismiss` from a non-presenting controller up to the presenting ancestor, the whole modal goes away exactly as before. Worth knowing about; not a problem.

**The Nearby spinner latch answers my concern structurally, which is better than answering it with a flag.** I'd asked what happens when `NearbyCoordinateResolver` returns nil and no fetch ever runs. The answer is that the latch lives in `NearbyStopsSheetContent`, which the parent only constructs inside `if let coordinate` — so on the nil-anchor path that view, and the flag with it, never exist, and the parent renders the empty state directly. The spinner can't hang because the code holding it isn't there.

**Doc comment and factory tests** are both corrected as asked. The two replacement anchor tests are deterministic — I checked that `MockAuthorizedLocationManager` populates `currentLocation` synchronously in its `didSet`, and that nothing in `Application.init` starts location updates, so neither test depends on ambient state or on suite ordering. And the nil-anchor case the old branchy test never actually reached is now covered properly in `NearbyCoordinateResolverTests`, with the factory test's comment pointing at it and explaining why it isn't reachable from that level. That's the right way to close a coverage gap you can't close in place.

**Verified against the real base, not the stale one.** Your CI is green on the current head, but it was created before `main` moved on the 21st. So I merged #1280, #1288, and #1289 onto current `main` locally and built and ran the suite against the integrated result: **2277 tests in 246 suites passed.**

**Sequencing.** #1280 and #1288 are now merged, so this PR's diff is exactly the incremental range I reviewed — which resolves the stacking objection without you having to retarget anything.

No new findings from me on this round, and I'm not going to invent any — the five changed files are small, each change is the minimal one that closes its finding, and the comments accurately describe the code they sit on. Thank you again for having disclosed up front that the three screens hadn't been hand-tested; it told me where to spend the attention.

PR #1290, opened 2026-08-16
Add Bike Mode: cycling speed for stop ETAs, plus separate walk/bike header chips

vlumpkin:
## Summary

Adds a standalone **Bike Mode** to speed up walk-time-to-stop estimates for cyclists, without disturbing the existing walking-speed path:

- **Bike Mode toggle** (Settings) swaps the travel velocity used for stop ETAs, arrival estimates, and the reachable/missed departure split from walking speed to cycling speed. Speed syncs from HealthKit cycling data when available (mirroring the existing \`WalkingSpeedManager\` pattern as an independent mechanism), falling back to a fixed ~15 km/h constant otherwise.
- **Stop header** now shows two independent, always-visible chips — a walk estimate and a bike estimate, each computed at its own fixed speed — instead of a single pill that silently followed Bike Mode while still reading "walk". Bike Mode continues to drive only the reachable/missed arrival split, not these header chips.
- **Chip styling**: walk (green) and bike (blue) chips now use distinct background colors so the two estimates read as different at a glance, not just by icon.

### Known limitation

The bike chip isn't tappable — there's no "open cycling directions" action to match the walk chip's \`onWalkingDirections\`. Apple's documented Maps URL scheme only supports \`dirflg=d/w/r\`; there's no officially supported bike value, so wiring that up would mean relying on undocumented behavior. Left as a deliberate follow-up rather than guessing.

## Test plan

- [x] Full unit suite on iPhone 17 Pro simulator: 1254/1254, 0 failures
- [x] \`BikeModeManagerTests\` — HealthKit auth success/failure/exception, sample present/absent/out-of-range, active-vs-passive sync asymmetry (passive refresh never downgrades source to \`.manual\`)
- [x] \`UserDataStoreTests\` — bike speed/mode/source defaults, roundtrip, clamping, \`effectiveTravelVelocityMetersPerSecond\` in both Bike Mode states
- [x] \`StopViewModelTests\` — new coverage asserting \`headerWalkTime\`/\`headerBikeTime\` stay fixed across the Bike Mode toggle while the mode-aware \`walkTime\` (driving the chronological split) switches between them
- [x] Live-data simulator walkthrough (Puget Sound): both chips render with correct minutes and distinct colors on a real stop
- [ ] **Needs a human/device pass** (not exercisable in Simulator): live HealthKit authorization + a real \`cyclingSpeed\` sample round-trip, and the denial toast on a physical device

## Summary by CodeRabbit

- **New Features**
  - Added Bike Mode, allowing cycling speed to be used for arrival-time estimates.
  - Added optional HealthKit synchronization for cycling speed, with manual fallback.
  - Settings now includes a Bike Mode toggle and synchronization feedback.
  - Stop pages display separate walking and cycling travel-time estimates.
  - Updated HealthKit messaging to clarify that walking or cycling speed may be used.

- **Bug Fixes**
  - Improved travel-time calculations when Bike Mode is enabled.
  - Preserved walking and cycling estimates independently when switching modes.

### DIFF
--- a/Apps/Shared/app_shared.yml
+++ b/Apps/Shared/app_shared.yml
@@ -50,8 +50,8 @@ targets:
           - twitter
           - comgooglemaps
         NSLocationWhenInUseUsageDescription: See where you are in relation to transit, and help you navigate more easily.
-        NSHealthShareUsageDescription: OneBusAway uses your walking speed from the Health app to give you more accurate arrival time estimates.
-        NSHealthUpdateUsageDescription: OneBusAway uses your walking speed from the Health app to give you more accurate arrival time estimates.
+        NSHealthShareUsageDescription: OneBusAway uses your walking or cycling speed from the Health app to give you more accurate arrival time estimates.
+        NSHealthUpdateUsageDescription: OneBusAway uses your walking or cycling speed from the Health app to give you more accurate arrival time estimates.
         NSLocationTemporaryUsageDescriptionDictionary:
           MapStatusView: See where you are in relation to transit, and help you navigate more easily.
         UILaunchStoryboardName: LaunchScreen
--- a/OBAKit/Orchestration/Application.swift
+++ b/OBAKit/Orchestration/Application.swift
@@ -93,6 +93,9 @@ public class Application: CoreApplication, PushServiceDelegate {
     @MainActor
     lazy var walkingSpeedManager = WalkingSpeedManager(userDataStore: userDataStore)
 
+    @MainActor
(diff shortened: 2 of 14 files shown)


vlumpkin:
Forgot to pull before, oops lol

Closing for now — branch is 468 commits behind main with real conflicts across every touched file (StopPageHeaderView, StopViewModel, Settings, etc.), likely due to overlap with the newer stop-page and bikeshare-map-layer work. Will rebase/resolve and reopen.

Issue #1150, opened 2026-06-04
Proximity alerts: harden authorization signalling, region-limit handling, and failure filtering (PR 3 follow-ups)


aaronbrethorst:
Follow-up items surfaced during review of #1092 (PR 2 of 4 for destination proximity alerts, #455). None block PR 2 — they are deliberately deferred to the `ProximityAlertManager` work in PR 3 (and one to the PR 1 model). Tracking here so they don't get lost.

### 1. `startMonitoringProximity` gives the caller no signal when it no-ops
`startMonitoringProximity(for:)` follows the existing `LocationService` convention of `guard isLocationUseAuthorized else { return }`. That's correct for PR 2, but unlike `startUpdatingLocation`, region monitoring has **no self-healing re-arm** when authorization later changes — so a silently-dropped alert stays permanently dead. PR 3's `ProximityAlertManager` must:
- Verify authorization (ideally `.authorizedAlways`, since background delivery requires it) **before** calling, and surface a denied/insufficient state to the user.
- Treat "monitoring did not start" as a user-visible condition, not a silent no-op.

### 2. Enforce vs. document the `.authorizedAlways` requirement
The doc comment on `startMonitoringProximity` notes region monitoring needs `.authorizedAlways` for background delivery, but `isLocationUseAuthorized` also accepts `.authorizedWhenInUse`. PR 3 should ensure the manager requests/validates Always authorization so alerts actually fire in the background.

### 3. `monitoringDidFailFor` does not filter by the proximity prefix
`didEnterRegion` filters on the `oba.proximity.` prefix, but `monitoringDidFailFor` forwards **every** monitoring failure to proximity delegates. For symmetry, filter by prefix before notifying, and consider classifying `CLError.Code` (esp. `regionMonitoringFailure` from exceeding the OS limit) so the manager can distinguish fatal from transient failures.

### 4. Handle the 20-region CoreLocation monitoring cap
iOS caps an app at 20 simultaneously monitored regions. `startMonitoringProximity` is unaware of `monitoredRegions.count`; the 21st region fails asynchronously via `monitoringDidFailFor` with a generic error. PR 3 should guard against / surface this limit.

### 5. Validate `ProximityAlert.radiusMeters` against `maximumRegionMonitoringDistance`
`CLCircularRegion` silently clamps an oversize radius, so an alert can fire at a different distance than requested with no feedback. Best addressed on the `ProximityAlert` model (PR 1 lineage).

[...]

PR #1292, opened 2026-08-16
Harden proximity monitoring follow-ups from #1150

diveshpatil9104:
## Summary
- `startMonitoringProximity` returns a non-discardable `ProximityMonitoringResult` instead of no-oping silently. Region monitoring never self-heals, so a dropped failure is a permanently dead alert.
- Requires `.authorizedAlways` rather than `isLocationUseAuthorized`, and adds the missing `NSLocationAlwaysAndWhenInUseUsageDescription` to `app_shared.yml` — all twelve locales already carried the translation.
- Filters `monitoringDidFailFor` by the `oba.proximity.` prefix, mirroring `didEnterRegion`, and classifies `CLError` into `RegionMonitoringFailureKind`.
- Guards the 20-region cap up front (exempting re-arms, which replace rather than add) and clamps the radius against `maximumRegionMonitoringDistance` and on the model.
- Logs a prefixed region that is not a `CLCircularRegion` instead of returning silently.
- Refs #1150. The user-facing half of items 1 and 2 belongs to `ProximityAlertManager`, which does not exist yet; this lands the signals it will need.

## Summary by CodeRabbit

* **New Features**
  * Added support for Always location authorization for background proximity alerts.
  * Improved proximity monitoring with authorization checks, region limits, radius adjustment, and monitoring status results.
  * Added clearer reporting and classification of region-monitoring failures.
  * Added validation and safe defaults for proximity alert radii.

* **Bug Fixes**
  * Prevented unsupported, oversized, invalid, or unrelated regions from being monitored.
  * Ensured persisted proximity alerts are validated when restored.

* **Tests**
  * Expanded coverage for authorization, region limits, radius handling, monitoring results, and failure reporting.

### DIFF
--- a/Apps/Shared/app_shared.yml
+++ b/Apps/Shared/app_shared.yml
@@ -75,6 +75,10 @@ targets:
           - fb
           - twitter
           - comgooglemaps
+        # Proximity alerts geofence the destination, and Core Location only delivers
+        # region events in the background under Always. Same body as the When In Use
+        # string, matching every InfoPlist.strings translation already in the repo.
+        NSLocationAlwaysAndWhenInUseUsageDescription: See where you are in relation to transit, and help you navigate more easily.
         NSLocationWhenInUseUsageDescription: See where you are in relation to transit, and help you navigate more easily.
         # HealthKit purpose strings live in Apps/Shared/info_plist.yml (shared by every app).
         NSLocationTemporaryUsageDescriptionDictionary:
--- a/OBAKitCore/Location/Location/LocationManagerProtocol.swift
+++ b/OBAKitCore/Location/Location/LocationManagerProtocol.swift
@@ -17,6 +17,10 @@ public protocol LocationManager {
 
     func requestWhenInUseAuthorization()
 
+    /// Region monitoring only delivers events in the background under
+    /// `.authorizedAlways`, so proximity alerts must be able to ask for it.
+    func requestAlwaysAuthorization()
+
     @available(iOS 14, *)
     func requestTemporaryFullAccuracyAuthorization(withPurposeKey purposeKey: String)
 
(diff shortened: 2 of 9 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is the follow-up I wanted from #1150, and it's thorough — all six items, each with tests that actually exercise the new branch rather than just constructing the type.

The judgment calls are the right ones:

- **`isProximityMonitoringAuthorized` requiring `.authorizedAlways`** rather than reusing `isLocationUseAuthorized`. When In Use is accepted by Core Location and then silently useless for a proximity alert, which is the worst possible combination. Good that you documented why in the property itself.
- **Making the result non-`@discardableResult`.** There is no self-healing re-arm here, so a dropped failure ships an alert that never fires and never explains itself. I confirmed there are no production callers yet, so the signature change breaks nothing today and forces the eventual `ProximityAlertManager` to handle it.
- **Exempting re-arms from the cap check.** `CLRegion` hashes on identifier, so a replacement genuinely can't push you over 20, and rejecting it would have been a real bug. Easy detail to miss.
- **Checking the cap up front** because Core Location enforces it asynchronously and reports the failure with no region attached — so up-front is the only way to tell the caller *which* alert failed.

Also glad to see `NSLocationAlwaysAndWhenInUseUsageDescription` added in `app_shared.yml` with the KiedyBus override, and the doc comment warning that the call silently does nothing without it. That's the kind of thing that costs someone an afternoon.

Three things for the follow-up, none blocking:

- **Failed regions are never torn down.** `monitoringDidFailFor` logs and forwards but never calls `stopMonitoring(for:)`, so a permanently-rejected region keeps occupying one of the 20 slots your new guard counts. The guard's accuracy depends on cleanup that doesn't exist yet. You've explicitly deferred the consumer side to `ProximityAlertManager`, so I'll take it there — just flagging that it's the loose end.
- **`maximumRegionMonitoringDistance == -1` is treated as "value unknown"**, but Apple documents it as "region monitoring is unavailable on this device." You arm anyway and return `.started`, which is optimistic; the failure does surface asynchronously, so it isn't silent. An `isMonitoringAvailable(for:)` check would close it properly.
- **The clamping comment's stated reason is off.** It says `CLCircularRegion` "clamps an oversize radius without reporting it"; current docs say exceeding the maximum yields a `CLError.regionMonitoringFailure`. Your behavior — clamp deliberately and report it — is right either way; only the rationale needs a tweak.

Approving.

PR #1293, opened 2026-08-17
Bring the Map settings sheet and rental markers to the map panel via a shared registrar

mosliem:
## Summary
- Brings the Map settings sheet to the SwiftUI map panel: basemap styles, Points of Interest, layer toggles, rental range filter, and Reset. `MapSheetView` is reused unchanged rather than forked.
- Makes the layer pipeline host-agnostic so both map surfaces can drive it: `MapRegionManager` takes a viewport from either host, `RentalLayerCoordinator` publishes rentals instead of writing annotations (with `RentalAnnotationSyncer` keeping the UIKit map identical), and registration moves into a shared `MapLayerRegistrar`.
- Fixes satellite rendering as hybrid on the panel — `MapBaseType` now maps to `.imagery`.
- Draws rental bikes/scooters on the panel with screen-space clustering, since SwiftUI `Map` has no clustering API. Clusters bucket in `MKMapPoint` space so panning cannot re-bucket them, and centroids are clamped to a quarter-cell so adjacent markers cannot overlap.
- Hides "Plan a trip using this bike" on the panel: `AppSheetRoute.tripPlanner` has no registered view yet, so a dead primary action is hidden rather than shown. The UIKit map keeps the action.
- Route lines/vehicles and followed-trip layers are deliberately not registered on the panel — they draw `MKOverlay` polylines SwiftUI `Map` cannot render.

## Test plan
- [x] `xcodebuild test-without-building -only-testing:OBAKitTests` — 2001 tests in 219 suites, one pre-existing failure (`RentalFormatTests.rangeFallbackUsesAbbreviatedUnits`, a locale artifact on code this branch does not touch).
- [x] Simulator with the panel flag on: settings sheet opens from the basemap button, all three basemaps render distinctly, the layer badge tracks toggles, and switching Bus stops off clears stop pins while bookmarks remain.
- [ ] On hardware: pan and zoom a dense rental area and confirm cluster markers stay legible.

## Known follow-up
Cluster membership near the viewport edges still churns while panning: the rental fetch box is padded by only 1.1, so a small drag drops vehicles at the trailing edge and the affected clusters renumber. The fix (wider padding plus refetch gating) lives in the shared `RentalLayerCoordinator` and would change the UIKit map's fetch cadence too, so it is left out of this PR.

## Summary by CodeRabbit

* **New Features**
  * Added map settings for basemap styles, points of interest, and layer visibility.
  * Added rental-bike markers with availability details, fuel labels, and clustered markers.
  * Added rental detail and cluster views with map selection and sheet navigation.
  * Added enabled-layer count badges and localized accessibility labels.
  * Added viewport-aware map updates, search views, and optional trip-planning actions.

* **Bug Fixes**
  * Preserved selected bookmark callouts during map refreshes.
  * Improved rental annotation updates when regions or map contents change.
  * Improved map-layer refreshes when changing regions or availability.


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is a well-decomposed refactor. Making the layer pipeline host-agnostic — registration into `MapLayerRegistrar`, `RentalLayerCoordinator` publishing `visibleRentals` instead of writing `MKMapView` annotations, and `RentalAnnotationSyncer` keeping the UIKit map behaviorally identical — is the right way to bring a UIKit surface to SwiftUI without forking it. That's harder than duplicating the sheet and it leaves the codebase better.

The thing I most wanted to verify was map-type divergence, since two hosts writing the same setting is exactly where this kind of port goes wrong. It's clean: `MapPanelRootController` builds one `MapViewModel` seeded from `MapBaseType(mapRegionManager.userSelectedMapType)` and hands the *same instance* to both `MapPanelRootView` and `AppSheetViewFactory`. `selectBaseType` routes to `setMapType`, which persists through `MapRegionManager.userSelectedMapType` — the same UserDefaults key the UIKit map reads, which also assigns `mapView.mapType` on write. One source of truth, no second store. `MapBaseType ↔ MKMapType` is a lossless three-way mapping, and the deleted `MapViewModel.toggleMapType()` has no remaining callers.

Fixing satellite to render as `.imagery` on the panel is a good catch too.

Two localization things I'd like fixed, and I'm not holding the merge for them only because they're additive rather than regressions:

- **`"%1$@, %2$d layers on"` isn't pluralized.** In a non-bikeshare region the stops layer is the only one, so `badgeCount == 1` is the *common* case and VoiceOver will say "standard, 1 layers on". The repo already has `Localizable.stringsdict` and `LocalizationTests.pluralKeys` for exactly this — please use them.
- **That key isn't in any `.strings` file.** This PR touches zero localization files, so the new VoiceOver string can never be translated. Related: the plan doc added here asserts "This plan introduces no new strings — every label is reused from an existing key," which is no longer true and should be corrected.

Smaller notes:

- `MapSheetModel` is constructed inside `AppSheetViewFactory.mapSettingsView()`, which is called from `MapPanelRootView.body` — so a new model and three new Combine subscriptions get built on every panel re-render, which now includes every camera settle since `layersModel.rentalItems` is `@Published`. It's functionally correct because all state reads through `MapRegionManager`, but it falsifies `rangeFilterPresets`' own comment about being "computed once per sheet presentation." A `@StateObject`-owned model on the panel side would restore that.
- `MapPanelLayersModelTests."Publishes no rental items without a bikeshare region"` doesn't test what it says: `rentalItems` is empty at construction in every scenario, and the default test region does have bikeshare — so it'd pass with the feature removed.
- The panel still calls `stopsObserver.updateViewport` and `scheduleStopsRequest(in:)` on every settle even when the Bus stops layer is toggled off; only rendering is gated.
- Two doc comments drifted: `MapPanelLayersModel.registrar` says "Nothing on the panel reads it" (`refresh()`, `rentalFetchedAt`, `rentalStaleAfter`, `rentalUserLocation` all do), and `RentalMapLayer.freshnessWindow` claims the panel reads it directly when it goes through `registrar.rentalLayers.first?.staleAfter`.

One sequencing note for me as much as you: this touches `SheetRoute.swift`, `AppSheetViewFactory.swift`, and `AppSheetRouteTests.swift`, all of which #1280, #1288, and #1289 also modify — and `AppSheetViewFactory.init` gains different required parameters in each. Whichever of the four merges second will need a rebase. Since this is the only one of the four that doesn't touch `.strings`, landing it first minimizes the churn.

Approving.

aaronbrethorst:
Approved on the merits. Two things landed on `main` while this sat: #1272 (which touched `SheetRoute.swift`, `AppSheetViewFactory.swift`, and `MapPanelRootController.swift`) and #1266 (which added a `refreshBookmarkAnnotationLabels()` to `MapRegionManager`). Between them this branch now has conflicts.

Please merge `main` in and resolve. `AppSheetViewFactory.init` and the `AppSheetRoute` exhaustiveness switch are where the real overlap is; the `MapRegionManager` side should be a straightforward textual resolution since #1266 only added a method and a call above the early returns.

Once it's mergeable it's good to go — no re-review needed unless the resolution changes behavior. The two localization asks from my review (pluralizing the layer-count string and getting the new key into the `.strings` files) still stand, but I'm not gating the merge on them.

OBAKit/Sheet/Root/Controls/MapTypeButton.swift
```diff
@@ -41,7 +57,27 @@ struct MapTypeButton: View {
...
+    /// The basemap name, plus the layer count when the badge is showing one.
+    ///
+    /// The badge itself is `accessibilityHidden` — it is decoration sitting on
+    /// top of the button — so without folding the count in here a VoiceOver user
+    /// would have no way to learn the layer state short of opening the sheet,
+    /// which is exactly the trip the badge exists to save.
     private var accessibilityValueText: String {
+        guard badgeCount > 0 else { return baseTypeValueText }
+
+        return String(
```
aaronbrethorst: This needs to be `String.localizedStringWithFormat`, not `String(format:)` — one line, but it's what makes the rest of this commit work.

Quoting the comment you left in `SearchResultsSheetView` in #1280:

> `localizedStringWithFormat`, not `String(format:)`: the latter expands `%#@count@` but always resolves it against the root plural rule, so the `few`/`many`/`zero`/`two` forms in the ar, pl, and ru entries could never be selected. Invisible in English; wrong everywhere with more than two.

Same trap, same codebase, and it bites harder here because this key is `%2$#@count@` — a positional binding with a second argument. I checked the stringsdict entries you added and the categories are right in all 13 locales: six for ar, one/few/many/other for pl and ru, bare `other` for ko/vi/zh-Hans/zh-Hant. With `String(format:)` none of the ar/pl/ru forms are reachable, so the most careful part of the change is the part that can't run.

The original symptom — "standard, 1 layers on" in English — *is* fixed by the stringsdict move alone, which is why this looked done.

OBAKitTests/Strings/LocalizationTests.swift
```diff
@@ -143,4 +144,24 @@ final class LocalizationTests {
...
+    /// and VoiceOver read "standard, 1 layers on".
+    @Test func `Map type layer count reads its singular and plural`() {
+        let bundle = Bundle(for: DonationCell.self)
+        let format = bundle.localizedString(
+            forKey: "map_controller.map_type.accessibility_value_with_layers_fmt",
+            value: "MISSING",
+            table: nil
+        )
+
+        #expect(String(format: format, "standard", 1) == "standard, 1 layer on")
```
aaronbrethorst, line 208: This is why the `String(format:)` bug above got through: the assertion goes through `String(format:)` too, so it exercises the same wrong path as production, on an English-locale test host where the root plural rule and English agree on one/other. It passes whether or not the call site is correct.

I don't think it's a bad test — it does catch a broken `%2$#@count@` positional binding, which is what you wrote it for, and the doc comment says exactly that. It just can't see this class of bug.

One rendering asserted in Polish or Russian would have caught it, and would be the thing that keeps it caught. If pinning a locale in this suite is awkward, asserting through `String.localizedStringWithFormat` at least makes the test agree with what the call site should be doing.

OBAKit/Mapping/Layers/MapSheetView.swift
```diff
@@ -125,9 +125,19 @@ import UIKit
...
+
+    /// `@StateObject`, not `@ObservedObject`: the panel builds this view inside
+    /// `MapPanelRootView.body`, so an eager model would be rebuilt — with its
+    /// three subscriptions — on every re-render. The autoclosure runs once per
+    /// view identity. The UIKit host's eager `MapSheetModel(...)` wraps unchanged.
+    @StateObject private var model: MapSheetModel
+
     @Environment(\.dismiss) private var dismiss
 
+    init(model: @autoclosure @escaping () -> MapSheetModel) {
```
aaronbrethorst, line 137: This is the right fix and I checked both call sites rather than taking the commit message's word for it, since `@StateObject` silently ignoring a fresh model is a nasty way to get stale state.

`AppSheetViewFactory` passes `MapSheetModel(...)` directly, which the autoclosure now defers — that's the actual fix, since the factory is called from `MapPanelRootView.body` and was rebuilding the model plus its three subscriptions on every camera settle. The UIKit host passes an already-constructed `model`, so the autoclosure just captures the reference and nothing is built twice.

The staleness question resolves cleanly too: the UIKit path builds a fresh `UIHostingController` per presentation, so view identity is new each time and the autoclosure re-runs. And on the panel side a retained model is fine regardless, since all its state reads through `MapRegionManager` rather than being snapshotted at construction. No stale-model path that I can find.

aaronbrethorst (changes requested):
I'm reversing my earlier approval on this one — not because anything regressed, but because I checked the new plural work end to end and it doesn't reach the locales it was added for. It's a one-line fix, and the branch needs a rebase anyway, so both can go in the same pass.

**First, what's genuinely fixed.** All four items from my last review are closed properly:

- **`MapSheetView` owning its model via `@StateObject` + `@autoclosure`** is the right fix, and I checked both construction sites rather than assuming. On the panel path, `MapSheetView(model: MapSheetModel(...))` no longer evaluates that initializer at the call site — it's deferred into the autoclosure and runs once per view identity, so `rangeFilterPresets`' "computed once per sheet presentation" comment is true again. On the UIKit path, `MapSheetView(model: model)` captures an already-built instance, so nothing is constructed twice, and each presentation builds a fresh `UIHostingController` and therefore fresh view identity — no stale model carried across presentations. That was the failure mode I went looking for and it isn't there.
- **The tautological rental test is really replaced.** Seeding the panel and then switching to Tampa is a test that can fail.
- **Both drifted doc comments** now match the code, and the `updateViewport`/`scheduleStopsRequest` justification checks out against `MapRegionManager`.

**Now the problem.** `MapTypeButton.accessibilityValueText` builds the string with `String(format:)`. It needs to be `String.localizedStringWithFormat`.

You'll recognize the reasoning, because you wrote it. This is the comment you left in `SearchResultsSheetView` in #1280, which I merged an hour ago:

> `localizedStringWithFormat`, not `String(format:)`: the latter expands `%#@count@` but always resolves it against the root plural rule, so the `few`/`many`/`zero`/`two` forms in the ar, pl, and ru entries could never be selected. Invisible in English; wrong everywhere with more than two.

That's exactly this situation. The plural categories you added are *correct* — I checked all 13 locales and they're right per language: six for ar, one/few/many/other for pl and ru, bare `other` for ko/vi/zh. And with `String(format:)` none of the ar, pl, or ru forms can ever be selected. The most careful part of the commit is the part that can't run.

**The test can't catch it, and that's the part worth fixing properly.** `Map type layer count reads its singular and plural` asserts through `String(format:)` — the same wrong call the production code makes — on an English-locale test host, where the root plural rule and English agree on one/other. So it passes either way. It's not a bad test; it does catch a broken positional binding, which is what you wrote it for. It just can't see this class of bug. If you can assert one Polish or Russian rendering, that's the assertion that would have caught this.

I'd also gently note the placement: this is `%2$#@count@`, a positional plural binding with a second argument, which is the trickiest plural shape in the codebase. Getting the stringsdict right and the call site wrong is an easy way for that to look done.

**Everything else about the localization is right.** The key is registered in `LocalizationTests.pluralKeys`, both new keys landed in all 13 `.strings` files, and having the key in both `.strings` and `.stringsdict` is correct — I checked the existing plural keys and that's the established convention here, not a shadowing bug.

**On the rebase.** #1280, #1288, and #1289 all landed while this was open, and this branch is now conflicted against `main` in 20 files. The Swift ones:

- `OBAKit/Sheet/DI/AppSheetViewFactory.swift` — the real one. `init` now takes `application`, `onPresentTrip`, `onPresentVehicleTrip`, `presentingController`, `coordinator`, `searchDisplayModel`, `stopsObserver` on `main`; yours takes `application`, `mapViewModel`, `layersModel`, `onPresentTrip`, `presentingController`. The resolution is the union of both, not either side.
- `OBAKit/Sheet/Coordinator/SheetRoute.swift`, `OBAKit/Sheet/Root/MapPanelRootController.swift`, `OBAKit/Sheet/Root/MapPanelRootView.swift`
- `OBAKitTests/Sheet/AppSheetRouteTests.swift`, `OBAKitTests/Sheet/AppSheetViewFactoryTests.swift`, `OBAKitTests/Strings/LocalizationTests.swift`

Plus all 13 `Localizable.stringsdict` files, which should be additive on both sides — keep both sets of keys.

This is my fault as much as sequencing luck: I told you in my last review to land this one first to minimize churn. That advice was correct when this PR touched no localization files, and stopped being correct once it grew `.stringsdict` changes across every locale. At that point landing the three-deep stack first meant one branch needed a rebase instead of three. Sorry for the extra work — the ordering call was mine.

Fix the `localizedStringWithFormat` line, rebase, and I'll re-review promptly. Everything else here I've already signed off on, so this should be a quick round trip.

aaronbrethorst:
Two small additions to the review above, plus a more precise map of the conflicts so the rebase is less of a scavenger hunt.

**Supporting evidence for the `String(format:)` point.** I went looking for whether this is a repo-wide convention I'd be unfair to single you out for, and found that `LocalizationTests.pluralKeys` already documents it as a known defect class:

> Note that `String(format:)` expands `%#@…@` but always resolves it against the root plural rule, so only `one`/`other` are ever reachable through it — the call site has to use `String.localizedStringWithFormat` for a locale's `few`/`many`/`zero`/`two` entries to mean anything. Nothing here can catch that; it's a call-site property.

So the test suite itself says the assertion can't catch this, which is exactly why I'm raising it by reading rather than by CI. To be fair about scope: `StopPageView.swift:500` and `StopViewController.swift:1438` still use `String(format:)` with plural keys, so there are pre-existing instances of the same bug. **I'm not asking you to fix those** — they're not yours and they predate this work. I'm asking for the new key because you're adding hand-authored ar/pl/ru forms in this commit that can't be reached, and because you already established the right pattern in #1280.

**Nit — the translator comment.** The `comment:` on the new key ends at "%1$@ is the base map type, %2$d is the layer count." The other plural call sites append: "Plural forms live in Localizable.stringsdict; the value above is only the not-found fallback." That sentence is what a translator actually sees in the exported strings, and without it they'll treat the flat form as authoritative and never touch the stringsdict entries you just wrote. Worth adding while you're in there.

**Nit — Arabic `zero`.** `accessibilityValueText` guards on `badgeCount > 0`, so the `zero` category can never be selected. Harmless and arguably right for completeness — just noting it's unreachable so you're not surprised if it never shows up.

**Conflict detail.** The cause in the 13 `.stringsdict` files is narrower than "they all conflict": #1280 added `search_results_sheet.result_count_fmt` and this branch adds `map_controller.map_type.accessibility_value_with_layers_fmt` at the *same alphabetical slot*, between the `data_migration_bulletin.*` and `stop_controller.*` entries. Resolution is mechanical everywhere — keep both, preserve alphabetical order. The `.strings` files auto-merge cleanly in every locale; only the `.stringsdict` files collide.

On the Swift side, `SheetRoute.swift` is the busiest at five conflict regions (the enum cases plus the `id`, `isBaseLayer`, and `detentConfiguration` switches — your `.rentalDetail` / `.rentalCluster` / `.mapSettings` alongside the new search and index routes). `AppSheetViewFactory.swift` is a single conflict on the shared `unimplementedView(for:)` line, which #1289 also rewrote to implement the index routes. Both are sibling-case additions with no semantic disagreement.

All three of #1280, #1288, and #1289 are now merged, so `main` is stable — one `git merge main` and a single resolution pass should do it, rather than resolving three times.

mosliem:
All four items are in. Rundown, plus one limitation I want to be upfront about rather than let you find by reading.

**`localizedStringWithFormat`.** Fixed in `MapTypeButton.accessibilityValueText`. You're right that this was the part of the commit that couldn't run — the ar/pl/ru forms were dead on arrival behind `String(format:)`.

**Translator comment.** The "Plural forms live in Localizable.stringsdict; the value above is only the not-found fallback." sentence is appended to the `comment:` and landed in all 13 `.strings` files.

**Arabic `zero`.** Left as written. Unreachable behind the `badgeCount > 0` guard, as you say, but I'd rather the entry be complete if that guard ever moves.

**Rebase.** `main` is merged. The `.stringsdict` collisions resolved as you predicted — both keys kept, alphabetical order preserved, in every locale. `AppSheetViewFactory.init` is the union of both sides, all nine parameters. `SheetRoute`'s five conflict regions and the `unimplementedView(for:)` line were sibling-case additions with no semantic disagreement, so nothing was dropped.

**The test — and where it stops.** Added `Polish layer count reaches its few and many forms` in `LocalizationTests`, asserting `one` (1), `few` (3), and `many` (5) render as three distinct Polish strings. It's not a locale-aware call written with `localizedStringWithFormat`, and can't be: that function takes no locale, it's `String(format:locale: Locale.current, …)`, and `Locale.current` on the test host is English — where the root rule and English agree, so the Polish categories would still be unreachable. `String(format:locale:)` with an explicit `Locale(identifier: "pl")` is the same code path with the locale pinned instead of ambient. I verified it isn't vacuous by dropping the `locale:` argument and re-running: both non-`one` assertions fail, collapsing to `other` — `"standardowa, 3 warstwy włączonej"` and `"standardowa, 5 warstwy włączonej"`. That's the bug you found by reading.

The limitation: `LocalizationTests` never touches `MapTypeButton`, and `accessibilityValueText` is `private` to the view, so this asserts the pl forms are *reachable through a locale-aware call* — not that the call site makes one. Revert the one-line fix and this test still passes. Genuinely pinning the call site needs a testability seam (lifting the construction into something like `accessibilityValue(baseType:layerCount:locale:)` with `locale` defaulting to `.current`), and I didn't want to add a test-only parameter to production code without asking. Happy to do it if you'd rather have the guarantee than the smaller surface.

I also corrected the `pluralKeys` note you quoted — it said "Nothing here can catch that," which is now true of the parity checks but not of the suite, since one key is pinned.

`xcodebuild test-without-building -only-testing:OBAKitTests` on iPhone 16: 2336 tests in 252 suites, the one pre-existing `RentalFormatTests.rangeFallbackUsesAbbreviatedUnits` locale failure on code this branch doesn't touch.

No apology needed on the ordering call — the `.stringsdict` growth is what changed the math, and that happened after your advice.

aaronbrethorst:
### Code review

Found 1 issue:

1. Changing the panel `Map`'s `selection` binding from `Stop.ID?` to `MapPinSelection?` orphans the search-result stop tags. `MapPanelRootView.stopAnnotation` was updated to `.tag(MapPinSelection.stop(stop.id))`, but `searchResultMapContent` in `MapSearchOverlays.swift` still tags with the bare `stop.id`. SwiftUI matches tags by type, so those annotations no longer participate in the selection and `.onChange(of: mapSelection)` never fires for them — after a route search, tapping one of the route's stop pins does nothing instead of pushing `.stopDetails`. The comment two lines above still claims "Tagged like ambient stops so tapping one opens its details", which is no longer true. `.tag()` accepts any `Hashable`, so neither the compiler nor CI catches this.

https://github.com/OneBusAway/onebusaway-ios/blob/2f3f725a67141a091e9fa594ef102f64a878f303/OBAKit/Sheet/Root/MapSearchOverlays.swift#L42-L44

aaronbrethorst (changes requested):
This is a big, well-organized piece of work. Making the layer pipeline host-agnostic is the right call — pulling registration into `MapLayerRegistrar` and having `RentalLayerCoordinator` publish `visibleRentals` instead of writing annotations directly onto an `MKMapView` is what lets both hosts share it without one of them pretending to be the other. `RentalAnnotationSyncer` keeping UIKit behavior intact while that happens is a good seam. And both items from my last round are genuinely closed: `accessibilityValueText` goes through `String.localizedStringWithFormat` now, and the Polish few/many assertion is in `LocalizationTests`.

One blocker, and it's a one-liner.

The selection migration from `Stop.ID?` to `MapPinSelection?` updated the tag in `MapPanelRootView`, but `searchResultMapContent` in `OBAKit/Sheet/Root/MapSearchOverlays.swift` still tags with the bare `stop.id`. That file isn't in this diff, so it's an easy one to miss. SwiftUI tags are type-keyed against the selection binding, so a `String`-tagged annotation in a map whose selection is `MapPinSelection?` simply never matches — those pins drop out of selection entirely. After a route search, tapping one of the route's stop pins stops opening its details.

`.tag()` takes any `Hashable`, so this compiles clean and nothing catches it. The comment sitting directly above that `ForEach` — "Tagged like ambient stops so tapping one opens its details" — is now false, which is a good sign it wants fixing rather than documenting.

Changing it to `.tag(MapPinSelection.stop(stop.id))` should do it, moving `MapPinSelection` somewhere both files can see it if that's needed.

Two things I looked at and want to flag without blocking on them:

`MapRegionManager.regionsService(_:updatedRegion:)` still calls `setVisibleMapRect` on the `MKMapView` that panel mode never lays out. If that fires `regionDidChangeAnimated`, the new `mapLayersViewportDidChange(mapView.visibleMapRect)` line overwrites the panel's real viewport with a whole-region rect and pushes nil to every layer, so rentals would disappear from the panel until the rider next pans. I couldn't confirm whether the delegate fires for a windowless map view without building it, so I'm not holding the PR on it — but the new doc comment ("then written by whichever host is driving") describes exactly the invariant at risk, and it's worth a look.

`MapViewModelTests."Setting the same map type is a no-op"` sets `.standard` on a view model already at `.standard` and asserts it's `.standard`. That passes with or without the guard; the test it replaced actually observed the publisher.

Fix the tag and this is good to go.

mosliem:
Thanks — all three addressed in c6f6979e.

**The blocker (fixed).** `searchResultMapContent` now tags with `.tag(MapPinSelection.stop(stop.id))`. Your diagnosis was exactly right: type-keyed tags meant those pins never matched a `MapPinSelection?` binding, so route-search stops silently dropped out of selection. `MapPinSelection` is already internal to OBAKit and visible from that file, so no type move was needed. The comment above the `ForEach` is true again rather than aspirational.

**`regionsService(_:updatedRegion:)` overwriting the panel viewport.** I went looking for this and couldn't make it happen — `regionDidChangeAnimated` does not fire for an `MKMapView` with no window, so the whole-region rect never reaches the layers. Rather than assert that from a hunch, there's now a test in `MapLayerViewportForwardingTests` that sets a panel viewport, triggers the region update, waits out the 1s animation, and asserts neither `currentVisibleMapRect` nor any layer saw the region rect. I confirmed it isn't vacuous by injecting the exact bug (`mapLayersViewportDidChange(region.serviceRect)` right after the framing call) — it fails with `9485.2 != 10000`. The `currentVisibleMapRect` doc comment now records why the invariant holds and points at the test, so the next person to touch that path gets caught instead of having to re-derive it.

**The tautological map-type test.** You're right, it passed with or without the guard. Rewritten as `"Setting the same map type publishes nothing"`, observing `$mapType` with `dropFirst()`: a redundant `setMapType(.standard)` must emit nothing, and a real `.satellite` change must still emit — so it can't pass by the guard swallowing everything. Deleting `guard newType != mapType` fails it on both assertions.

Full suite: 2337 tests / 252 suites, with only the pre-existing `RentalFormatTests.rangeFallbackUsesAbbreviatedUnits` locale artifact failing (untouched by this branch, noted in the description). SwiftLint clean — the one `type_body_length` warning on `MapRegionManager.swift` reproduces on a stashed tree, so it predates this work.

The wider rental-fetch padding is still deliberately out of scope here, since it changes the UIKit map's fetch cadence too.

aaronbrethorst:
### Code review

aaronbrethorst (changes requested):
Every point from all three previous rounds is closed, and I checked each one at head rather than trusting the commit log.

`searchResultMapContent` tags with `MapPinSelection.stop(stop.id)` now, so the comment above that `ForEach` is true again. The region-change viewport concern I raised as a maybe turned out to be worth a test, and you wrote a good one — seeding a 10,000-height viewport, firing the region update, waiting out the animation, then asserting both the height and that no layer got handed nil. The invariant is recorded in the `currentVisibleMapRect` doc comment pointing at that test, which is how it should be. And `Setting the same map type publishes nothing` is a real test now: it observes `$mapType.dropFirst()` and fails if the guard is deleted, where the old one couldn't.

The refactor underneath is good work. Pulling registration into `MapLayerRegistrar` and having `RentalLayerCoordinator` publish `visibleRentals` instead of writing annotations directly — with `RentalAnnotationSyncer` keeping the UIKit map byte-identical — is the right seam, and it's what makes one settings sheet serve both hosts instead of two implementations drifting apart. I traced the teardown ordering, the syncer swap, the availability/failure handling, and viewport priming on late registration, and the UIKit behavior holds at each of them.

I'm still not going to merge it today, and the reason has nothing to do with the code you wrote.

This branch's merge base is `0956472f` and it's 37 commits behind `main`. In that gap, #1304 replaced the bikeshare first-run callout with a TipKit tip, and part of that fix was adding `showMapLayersTipIfNeeded()` to the end of `configureMapLayers()`, with a comment explaining why: on a cold launch the region resolves *after* `viewDidAppear`, so the coordinator that gates the tip doesn't exist for the first attempt, and the region-change retry is what makes the tip appear on the launch that introduces bikeshare — the only launch it exists for.

This PR changes both `regionsService(_:updatedRegion:)` and `regionsService(_:updatedRegionsList:)` to call `configureStopRouteFocusLayer()` instead of `configureMapLayers()`. Those hunks don't touch anything #1304 changed, so they merge clean and silent, and afterwards the only surviving call is the `viewDidAppear` one at `MapViewController.swift:217` — the attempt #1304's comment already documents as insufficient. The tip stops appearing, GitHub reports no conflict, and nothing in the diff shows it. I only caught it because I went looking at what landed in those 37 commits.

So: rebase onto `main`, and while you're in `MapViewController+MapLayers.swift`, put `showMapLayersTipIfNeeded()` back on a path the region change actually reaches. `attachRentalLayerHost` looks like the natural home now that it's the thing that runs once the rental layers are really registered — same timing the old call had, and it already calls `updateMapLayerBadge()` right there. I don't need to re-review the refactor; I've read it. Just the rebase and that one line.

Two things to look at while you're in there, neither blocking.

`MapPanelLayersModel.recomputeClusters()` assigns `rentalItems` unconditionally on every camera settle, and `RentalMapItem` isn't `Equatable`, so `@Published` fires and the whole panel body re-renders — including `factory.view(for:)` — even when the computed array is identical. Same family as the perf note from round one, different line. Making `RentalMapItem` `Equatable` and only assigning on a real change would settle it.

`RentalAnnotationSyncer.sync(to:)` re-assigns `view.annotation` for every entry in `visibleRentals`, where the old `syncMapView(with:)` only did it for `changes.updated`. Correctness is fine — `applyChanges` still guards on `!changes.isEmpty` — but the per-refresh cost went from O(changed) to O(visible) on a layer with a 500-vehicle budget.

Last thing, and it's a process note rather than a defect: the title says "Bring the Map settings sheet to the SwiftUI map panel," and that's about a third of what's here. The rest is a shared-pipeline refactor that changes UIKit map internals, a new 184-line clustering algorithm, new rental marker/detail/cluster surfaces, a basemap fix, and the removal of `MapViewModel.toggleMapType()`. Your PR body describes all of it honestly, so nothing is hidden — but the title is what a reviewer budgets time against, and it undersold this by a lot. Worth splitting next time; the registrar refactor would have been a clean standalone PR.

Rebase and restore the tip call and I'll merge it.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Approving. Both blockers from my August 23 review are genuinely closed, and I checked them at the current head rather than going on the replies.

**The rebase** — `compare/main...08922fac` reports ahead 38, behind 0, with the merge base at current `main`. Fully caught up.

**The #1304 TipKit regression** — `showMapLayersTipIfNeeded()` is restored at `MapViewController+MapLayers.swift:111`, and more importantly it's actually reachable on a region change: `MapLayerRegistrar` is now the `RegionsServiceDelegate`, so `regionsService(_:updatedRegion:)` → `configure()` → `onDidConfigure` → `attachRentalLayerHost` → tip, with the coordinator assigned before `onDidConfigure` fires so the `rentalLayerCoordinator != nil` gate passes. Cold-launch ordering works out too: `viewDidLoad`'s call is dropped by the `window != nil` guard, `viewDidAppear` covers region-already-known, and the registrar callback covers region-resolves-late — which is the launch the tip exists for. That's the case that silently broke last time, so I'm glad it's pinned by `MapLayerRegistrarTests`.

The two non-blocking notes from that round are fixed as well: `RentalMapItem` has a hand-written `==` with a comment explaining why synthesis won't work, and `recomputeClusters()` guards `items != rentalItems` with a test that counts `$rentalItems` emissions *and* proves a real change still publishes.

On the parts I most wanted to see for a change this size: registrar add/remove is symmetric, both `onDidConfigure` closures capture `[weak self]`, and teardown ordering is safe — the old syncer is still alive when `removeMapLayer` → `deactivate()` empties `visibleRentals`, so the previous region's annotations do come off the map before the swap. Clustering in `MKMapPoint` space with quantized cell size gives real pan stability, and those tests assert the premise rather than just the conclusion.

I checked the localization directly, since a 13-locale change with plurals is where this would most likely go wrong: both new keys carry real translations everywhere, and the `.stringsdict` plural categories are CLDR-correct — Polish `one/few/many/other`, Arabic `zero/one/two/few/many/other`, Korean bare `other`. `LocalizationTests` now pins a Polish `few`/`many` rendering, which is precisely the assertion that would have caught the earlier bug.

Filed #1350 for the leftovers, none of them blocking: `rental_cluster.title_fmt` can now render "1 vehicles here" when a cluster member leaves the feed and it has no `.stringsdict` entry; the `RentalClustering` doc overstates the marker bound (~36 cells at the production `cellSize: 100`, not ~90); one clustering test's bound holds only because the fixture fits in a viewport; and a few test helpers use `UserDefaults()` where CLAUDE.md warns about process-global defaults under concurrent suites.

Issue #1277, opened 2026-08-11
Add Toggle for “Arriving at XX:XX pm via ### - Destination” banner on stop


cskinnerr:
<img width="590" height="1278" alt="Image" src="https://github.com/user-attachments/assets/d0d5c73a-57d8-40a3-ae1a-3e1d908f850a" />

Repro steps:
1. Open a stop
2. click on any upcoming departure of a bus trip
3. in the bus trip details, select a different stop, future or past is fine. 
4. Note that times will show NOW for a time that shows the trip you selected’s expected departure from that stop. 

This view is plagued with issue:
It basically sets your expected arrival time of your bus stop at “NOW” so you can see how much time you have between buses if transferring. 
However, it does not handle early or late buses well at all. If a bus is 5 min early, it shows a -5 min not “NOW”
Additionally, if the stop you selected is 35 min later than your current time, then the page will just be empty by default unless you click show earlier departures and show later departures. 

Please disabled this feature. It sucks. 
Please at least let users toggle this off. 
Please replace this feature with simply highlighting the trip on a the stop page, or whatever is decided, please bring it back to the old functionality that didn’t have such a bad UI Experience

Borisserz:
Partial: the toggle you asked for (“at least let users toggle this off”) shipped in #1294 — Settings → Arrival & Departure Display → Show transfer arrival banner, default on. That PR used `Refs`, so this stayed open.

Still not done from this issue:
- replacing the banner with highlighting the trip on the stop page
- the new stop page still always shows transfer context (legacy page honors the switch)

Happy to close this on the toggle and open a follow-up for highlighting, or leave this open for that remaining piece.

Borisserz:
Picking up the remaining piece: the toggle shipped in #1294; still open is replacing/highlighting the transfer trip on the stop page and making the new stop page honor the switch.

Borisserz:
Update after re-reading \`main\`: the toggle + router gate from #1294 are in place — legacy honors \`displayedTransferContext\`, and with the switch off the router opens the new stop page with no transfer context. What is still missing is replacing the banner with highlighting the transfer trip on the stop page (\`TransferContext\` has no \`tripID\` yet). I'll take that remaining piece.

Borisserz:
Working on the highlight replacement now: carry `tripID` on `TransferContext`, route transfer taps to the new stop page when that flag is on, and highlight the matching departure row instead of the legacy banner.

PR #1294, opened 2026-08-18
Add a Settings toggle for the transfer arrival banner

Borisserz:
## Summary
- Settings → Arrival Display → **Show transfer arrival banner** (default **on**) so riders can turn off transfer-relative times when they open a stop from a trip.
- When off, `UserDataStore.displayedTransferContext(_:)` returns `nil` and the legacy stop page shows wall-clock times, the walk-time row, and the full departure list.
- New keys in all 13 localization catalogs. `displayedTransferContext` lives on a Swift protocol extension because `TransferContext` cannot appear on the `@objc` `UserDataStore` protocol.

Does not replace the banner with trip highlighting, and does not change the new stop page (transfers still route to the legacy screen). **Refs #1277**.

## Test plan
- [ ] Settings → Arrival Display: switch defaults on; turning it off persists after leaving Settings.
- [ ] Open a stop from a trip with the switch **on**: Arriving-at banner, relative times, earlier departures hidden.
- [ ] Same path with the switch **off**: ordinary clock times, walk-time row, full list, no banner.
- [ ] `UserDefaultsStoreTests` — transfer banner defaults on; off persists and drops displayed context.
- [ ] `LocalizationTests` — every locale has the new keys.

## Summary by CodeRabbit

* **New Features**
  * Added an Arrival & Departure Display setting to show or hide transfer-arrival banners.
  * The setting is enabled by default and persists across app launches.
  * Disabling it removes transfer-specific arrival and departure details from stop displays.
  * Improved departure filtering, menu accessibility states, and time-sorted arrival sections.
* **Localization**
  * Added translated labels, explanations, and related stop-display text across supported languages.
* **Documentation**
  * Documented banner behavior, relative-time formatting, and setting limitations.
* **Tests**
  * Added coverage for persistence, disabled-state behavior, and localization completeness.

### DIFF
--- a/OBAKit/Settings/SettingsViewController.swift
+++ b/OBAKit/Settings/SettingsViewController.swift
@@ -70,6 +70,7 @@ class SettingsViewController: FormViewController {
             walkingSpeedMetersPerSecondKey: snapToPreset(application.userDataStore.walkingSpeedMetersPerSecond),
             walkingSpeedUseHealthKitKey: application.userDataStore.walkingSpeedSource == .healthKit,
             stopUIReducedColorsTag: application.userDataStore.stopUIReducedColors,
+            transferBannerTag: application.userDataStore.showTransferArrivalBanner,
             alwaysShowFeedbackPrompt: application.reviewPromptPolicy.alwaysShowPrompt
         ])
     }
@@ -127,6 +128,10 @@ class SettingsViewController: FormViewController {
             application.setArrivalDepartureFilter(filter)
         }
 
+        if let showBanner = values[transferBannerTag] as? Bool {
+            application.userDataStore.showTransferArrivalBanner = showBanner
+        }
+
         saveWalkingSpeedValues(values)
     }
 
@@ -179,10 +184,16 @@ class SettingsViewController: FormViewController {
     // MARK: - Arrival & Departure Display Section
 
     private let arrivalFilterTag = "arrivalDepartureFilter"
+    private let transferBannerTag = UserDefaultsStore.showTransferArrivalBannerKey
 
     private lazy var arrivalDisplaySection: Section = {
         let section = Section(
-            OBALoc("settings_controller.arrival_display_section.title", value: "Arrival & Departure Display", comment: "Settings section title for controlling which arrivals/departures are shown")
+            header: OBALoc("settings_controller.arrival_display_section.title", value: "Arrival & Departure Display", comment: "Settings section title for controlling which arrivals/departures are shown"),
+            footer: OBALoc(
+                "settings_controller.arrival_display_section.transfer_banner.footer",
+                value: "Show transfer arrival banner: when on, opening a stop from a trip shows times relative to when you arrive. Turn that switch off for ordinary clock times and every departure.",
+                comment: "Settings > Arrival & Departure Display > Footer for the Show transfer arrival banner switch only, not the Show Departures filter"
+            )
         )
 
         section <<< AlertRow<ArrivalDepartureFilter> {
@@ -192,6 +203,15 @@ class SettingsViewController: FormViewController {
             $0.options = Array(ArrivalDepartureFilter.allCases)
             $0.displayValueFor = { $0?.displayTitle }
             $0.value = self.application.effectiveArrivalDepartureFilter
+        }
+
+        section <<< SwitchRow {
+            $0.tag = transferBannerTag
+            $0.title = OBALoc(
+                "settings_controller.arrival_display_section.transfer_banner",
+                value: "Show transfer arrival banner",
+                comment: "Settings > Arrival Display > Toggle that shows the Arriving at XX:XX via route banner when opening a stop from a trip"
+            )
         }
 
         return section
--- a/OBAKit/Settings/SettingsViewController.swift
+++ b/OBAKit/Settings/SettingsViewController.swift
@@ -210,7 +210,7 @@ class SettingsViewController: FormViewController {
             $0.title = OBALoc(
                 "settings_controller.arrival_display_section.transfer_banner",
                 value: "Show transfer arrival banner",
-                comment: "Settings > Arrival Display > Toggle that shows the Arriving at XX:XX via route banner when opening a stop from a trip"
+                comment: "Settings > Arrival & Departure Display > Toggle that shows the Arriving at XX:XX via route banner when opening a stop from a trip"
             )
         }
 
--- a/OBAKit/Stops/StopViewController.swift
+++ b/OBAKit/Stops/StopViewController.swift
@@ -96,6 +96,13 @@ public class StopViewController: UIViewController,
         set { viewModel.transferContext = newValue }
     }
 
+    /// Honors Settings > Arrival Display > transfer banner. When the rider
+    /// turns that off, the stop page ignores the trip-sourced transfer context
+    /// so times are wall-clock and the full list is shown (#1277).
+    private var displayedTransferContext: TransferContext? {
+        application.userDataStore.displayedTransferContext(transferContext)
+    }
+
     var stop: Stop? { viewModel.stop }
     var stopArrivals: StopArrivals? { viewModel.stopArrivals }
     var stopPreferences: StopPreferences { viewModel.stopPreferences }
@@ -807,7 +814,7 @@ public class StopViewController: UIViewController,
             arrivalDeparture: arrivalDeparture,
             isAlarmAvailable: alarmAvailable,
             highlightTimeOnDisplay: highlightTimeOnDisplay,
-            transferContext: transferContext,
+            transferContext: displayedTransferContext,
             onSelectAction: onSelectAction,
             alarmAction: addAlarmAction,
             bookmarkAction: bookmarkAction,
@@ -855,7 +862,7 @@ public class StopViewController: UIViewController,
         // Filter departures before transfer arrival time, unless user opted to show all.
         // Only insert the "Show earlier" button when sorting by time (groupRoute == nil)
         // to avoid duplicate item IDs across route groups. See #409.
-        if let transferContext = transferContext, !showAllTransferDepartures {
+        if let transferContext = displayedTransferContext, !showAllTransferDepartures {
             let (visible, hidden) = partitionTransferDepartures(items: items, arrivalTime: transferContext.arrivalTime)
             items = visible
             if !hidden.isEmpty && groupRoute == nil {
@@ -987,7 +994,7 @@ public class StopViewController: UIViewController,
 
     /// Inserts either a transfer arrival banner or the GPS-based walk time row.
     private func addWalkTimeOrTransferBanner(to items: inout [AnyOBAListViewItem]) {
-        if let transferContext = transferContext {
+        if let transferContext = displayedTransferContext {
             let bannerItem = TransferArrivalItem(
                 id: "transfer_arrival_banner",
                 arrivalTime: transferContext.arrivalTime,
--- a/OBAKit/Stops/StopViewController.swift
+++ b/OBAKit/Stops/StopViewController.swift
@@ -96,7 +96,7 @@ public class StopViewController: UIViewController,
         set { viewModel.transferContext = newValue }
     }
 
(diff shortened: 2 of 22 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The toggle isn't consulted at the routing gate, so turning it off strands trip→stop taps on the legacy stop screen. `Router.makeStopController` picks the legacy `StopViewController` whenever the raw `transferContext` is non-nil, and `FeatureFlags.isNewStopPageEnabled` defaults to `true` — so in the default configuration a rider who switches "Show transfer arrival banner" off still gets the old stop page from a trip (now with no banner, wall-clock times, full list), while every other stop tap opens the new page. The comment on that line justifies the fallback as "TransferContext UX ... is not yet built on the new stop page", but with the setting off there is no transfer UX left to preserve; the condition would need to be something like `application.userDataStore.displayedTransferContext(transferContext) == nil` for the setting to actually restore normal behavior. #1277 asked to "bring it back to the old functionality", and this leaves half of that undelivered for default users. (bug due to `OBAKit/ViewRouting/Router.swift:108` combined with `FeatureFlags.isNewStopPageEnabled` returning `true` by default)

https://github.com/OneBusAway/onebusaway-ios/blob/b7be6f467bc35e72893602be61a8b48b3767e9db/OBAKit/ViewRouting/Router.swift#L105-L115

OBAKit/Stops/StopViewController.swift
```diff
@@ -96,6 +96,13 @@ public class StopViewController: UIViewController,
         set { viewModel.transferContext = newValue }
     }
 
+    /// Honors Settings > Arrival Display > transfer banner. When the rider
+    /// turns that off, the stop page ignores the trip-sourced transfer context
+    /// so times are wall-clock and the full list is shown (#1277).
+    private var displayedTransferContext: TransferContext? {
+        application.userDataStore.displayedTransferContext(transferContext)
```
aaronbrethorst, line 103: This helper is the right abstraction — the problem is that it is only applied *inside* the page, not at the routing boundary.

`Router.makeStopController` (`OBAKit/ViewRouting/Router.swift:108`) still tests the raw `transferContext` to decide between `StopPageViewController` and the legacy `StopViewController`. With the toggle off, that context is still non-nil, so the rider gets routed to the legacy screen *and* sees no transfer UI there — strictly worse than either intended state, in the default configuration (`isNewStopPageEnabled` defaults to true).

Run the router branch through this same helper and assign the effective value to `stopController.transferContext`.

OBAKit/Settings/SettingsViewController.swift
```diff
@@ -179,10 +184,16 @@ class SettingsViewController: FormViewController {
...
 
     private let arrivalFilterTag = "arrivalDepartureFilter"
+    private let transferBannerTag = UserDefaultsStore.showTransferArrivalBannerKey
 
     private lazy var arrivalDisplaySection: Section = {
         let section = Section(
-            OBALoc("settings_controller.arrival_display_section.title", value: "Arrival Display", comment: "Settings section title for controlling which arrivals/departures are shown")
+            header: OBALoc("settings_controller.arrival_display_section.title", value: "Arrival Display", comment: "Settings section title for controlling which arrivals/departures are shown"),
+            footer: OBALoc(
+                "settings_controller.arrival_display_section.transfer_banner.footer",
```
aaronbrethorst, line 193: This footer hangs off the whole Arrival Display section, but it only describes the transfer-banner switch. The section also contains the "Show Departures" filter row, so as written it reads as if it is explaining that row too. Either name the toggle explicitly in the copy or split it into its own section.

aaronbrethorst (changes requested):
The mechanism here is right, and I checked the parts that usually go wrong on a new preference: the default is registered as `true` so nobody's behavior changes until they opt out, the key is dot-free per the `stopUIReducedColors` precedent, `UserDefaultsStore` is the only conformer, all three `transferContext` consumption sites in `StopViewController` are converted with no fourth site hiding anywhere, and all 13 locales are covered. Putting the logic behind a `displayedTransferContext` helper on the store rather than sprinkling `if showBanner` at each site is the right call.

One thing to fix before this lands.

**Turning the toggle off downgrades you to the legacy stop page for nothing.** `Router.makeStopController` (`OBAKit/ViewRouting/Router.swift:108`) branches on the *raw* `transferContext`:

```swift
if transferContext == nil, FeatureFlags.isNewStopPageEnabled(userDefaults: application.userDefaults) {
```

That branch exists solely because the transfer UX was never built on the new stop page — the comment right above it says so. But with your toggle off, the transfer context is still non-nil at routing time, so a trip→stop tap still lands on `StopViewController`, which then suppresses the transfer UI you just routed there to get. Since `isNewStopPageEnabled` defaults to `true`, the rider's net result from turning the banner off is: no banner, *and* the old stop page. That's a worse outcome than either state they could have picked.

You already have the right helper for this — apply it at the routing boundary too, so the router sees the same effective context the page will:

```swift
let effective = application.userDataStore.displayedTransferContext(transferContext)
if effective == nil, FeatureFlags.isNewStopPageEnabled(userDefaults: application.userDefaults) { ... }
```

and assign `effective` to `stopController.transferContext`.

- The new footer is attached to the whole Arrival Display section but only describes the transfer-banner switch. The section also holds the unrelated "Show Departures" filter row, so as written the footer reads as if it explains both. Consider a sub-section, or wording that names the toggle.
- Flipping this setting posts no notification, so a `StopViewController` already on the nav stack won't re-render until its next refresh. Low impact — Settings is modal from More and the stop page reloads on appear — and #1287 is doing the equivalent thing for the filter, so it may be worth matching that pattern eventually.

Fix the router gate and I'll re-review promptly.

Borisserz:
`makeStopController` now gates on `displayedTransferContext`, not the raw transfer payload. With the banner off, a trip→stop tap stays on `StopPageViewController` and `transferContext` is nil — the same helper the stop page already used.

The Arrival Display footer names that switch so it is not read as applying to Show Departures.

Test that fails if the gate goes back to the raw context: `Make stop controller ignores transfer context when the banner toggle is off` (`StopViewModelTests`, 32/32 passed locally on iPhone 17 Pro).

aaronbrethorst:
### Code review

The router gate from the previous round is fixed correctly: `makeStopController` now computes `displayedTransferContext` once and uses it for both the branch and the assignment, the new `StopViewModelTests` case is a real regression test (revert the router and it fails), the default is registered `true`, the key is dot-free, and all three `transferContext` consumption sites in `StopViewController` plus both trip→stop entry points (tap and long-press preview) go through the helper.

Found 1 issue:

1. The footer's leading noun phrase is supposed to name the switch above it — that is what makes it read as documentation for the toggle rather than for the whole Arrival Display section. In 7 of the 13 locales the footer names something different from the switch title, so it reads as a second, nonexistent setting. `fil` also leaves the phrase in English: title is "banner ng pagdating sa transfer", footer is "transfer arrival banner". Same mismatch in `it` ("banner di arrivo della coincidenza" vs "banner arrivo trasferimento"), `pt-BR` ("chegada da conexão" vs "chegada da transferência"), `vi` ("chuyển tuyến" vs "điểm trung chuyển"), and more loosely in `fr`, `pl`, `ru`. `ar`, `es`, `ko`, `zh-Hans`, `zh-Hant` and `en` are consistent — the fix is to make the other seven match their own switch title verbatim. (bug due to `OBAKit/Strings/fil.lproj/Localizable.strings`)

https://github.com/OneBusAway/onebusaway-ios/blob/b56e0c601acc4fc3bfa93951d5afc6200982b211/OBAKit/Strings/fil.lproj/Localizable.strings#L987-L992

aaronbrethorst (changes requested):
The blocking item from my last round is fixed exactly as asked: `makeStopController` computes the effective transfer context once through `displayedTransferContext(_:)` and uses it for both the branch condition and the value handed to the controller, so the router can't disagree with itself anymore. The regression test that fails if it's reverted is a good addition. I also went looking for a fourth `transferContext` read and there isn't one — all three are converted, and both trip-to-stop entry points go through `makeStopController`.

The `@objc` reasoning in your comment is right, too. `UserDataStore` is `@objc(OBAUserDataStore)`, so a `TransferContext`-typed protocol requirement genuinely isn't possible, and the protocol extension is the correct way around it.

Two things to sort out.

It needs a rebase — the branch is conflicting because it predates `e9463271`, which renamed that Settings section to "Arrival & Departure Display" and changed the `// MARK:` above it. When you resolve, please keep main's title rather than reverting to "Arrival Display".

The other is the section footer. Naming the toggle in the footer copy was the fix for my earlier note, and it works in English — but the mechanism only holds if the footer names the *localized* switch. In `fil` the footer leaves "transfer arrival banner" in English while the switch title is translated; `it`, `pt-BR`, and `vi` use a materially different noun than their own switch title; `fr`, `pl`, and `ru` drift more loosely. `ar`, `es`, `ko`, `zh-Hans`, and `zh-Hant` are consistent. Worth a pass to make the footer and the switch agree in each locale, since a footer pointing at a control by a name that doesn't appear on screen is worse than a generic footer.

Non-blocking, and mostly carried over:

Flipping the toggle still posts no notification, so a `StopViewController` already on the nav stack won't re-render until its next refresh. Low impact and #1287 is establishing the pattern for the filter, so I'm fine leaving it.

The English footer says the banner off state gives you "ordinary clock times and every departure", but with the toggle off a trip-to-stop tap now also lands on the new stop page rather than the legacy one. That's the behavior I asked for; the copy just doesn't mention the screen swap.

`#expect((controller as? StopPageViewController)?.transferContext == nil)` passes vacuously if the cast fails — the `#expect(controller is StopPageViewController)` above it is doing the real work, so that second line isn't load-bearing.

Rebase and a localization consistency pass and this is ready.

Borisserz:
Rebased onto current `main`. The Settings section title stays **Arrival & Departure Display** (`e9463271`); this branch does not revert it to “Arrival Display”.

Every locale footer for `settings_controller.arrival_display_section.transfer_banner.footer` now starts with that locale’s switch title verbatim (`fil`, `it`, `pt-BR`, `vi`, `fr`, `pl`, `ru` were the mismatches). `LocalizationTests` has `Transfer banner footer names the switch title in every locale` — it fails if a translation leaves the English phrase in the footer.

Router gate (`displayedTransferContext`) is unchanged. One-line `import CoreLocation` on `TripPageViewController` so the target builds after the rebase.

`LocalizationTests` (5) passed on iPhone 17 Pro.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Both blockers from the last two rounds are fixed as asked, and I checked each one at head rather than taking the commit message for it.

The router gate was the one that mattered. `makeStopController` now computes `effective` once and uses it for both the branch and the assignment, so a trip-to-stop tap with the toggle off no longer detours to the legacy screen for a context it's about to discard. The regression test says so in the way I like — "leave the router on the raw value and this fails" — and that's true, not decoration.

The footer/title mismatch is gone in all thirteen locales, and you did better than fixing the seven I listed: `LocalizationTests` now asserts `footer.hasPrefix(title)` for every localization in the bundle, so the next translator can't reintroduce it. That's the right shape for a class of bug that only shows up when someone reads a language you don't.

The rest holds up. `register(defaults:)` with `true` means every existing rider keeps the current flow untouched, which is the thing to get right on an opt-out. The protocol extension is the correct workaround — `UserDataStore` is `@objc(OBAUserDataStore)` so a `TransferContext`-typed requirement genuinely can't go on the protocol — and the comment explains why, so nobody undoes it later. All three read sites in `StopViewController` route through `displayedTransferContext` rather than one or two, and the Eureka wiring matches the `stopUIReducedColorsTag` precedent exactly: seeded in `setValues`, written back in `saveFormValues`.

Two cosmetic things, neither worth another round. The translator comments in the `.strings` files still say "Settings > Arrival Display" while the source `comment:` argument says "Arrival & Departure Display" — rebase drift that `scripts/extract_strings` will sort out next time it runs. And `import CoreLocation` in `TripPageViewController.swift` looks like it rode along from a different branch; main uses `CLLocationCoordinate2D` in that file without it.

Merging.

PR #1295, opened 2026-08-18
Coalesce bookmark list writes on move and delete

### LINKED ISSUE #548: Editing bookmarks has bad performance with > 10 bookmarks
This is based upon a couple user complaints. I think that the worst issues are addressed, but there's more to do.

example:

> Thanks so much! I've got the update. I will note there is a roughly 5 second freeze after hitting delete before it deletes, and about an 8 second freeze for sliding a bookmark up or down the list. I don't mind it, but just wanted to let you know. Thanks again!!

Borisserz:
## Summary
- `add(_:to:index:)` and `delete(bookmark:)` used to decode and encode the full bookmark plist once per remaining row. That is the 5–8 second freeze in #548.
- Both methods now mutate an in-memory `[Bookmark]` and write the plist once.
- Existing sort-order tests still define grouping/`sortOrder`. Two new tests count writes to `UserDataStore.bookmarks` and fail if the per-row loop is restored.

**Closes #548.** (The comment asking for a per-group edit mode is a separate feature.)

## Test plan
- [ ] Seed ~12 bookmarks. Drag one to a new index: list order is correct, no multi-second freeze.
- [ ] Delete one: remaining `sortOrder` values compact, no multi-second freeze.
- [ ] Move a bookmark between groups: source and destination groups reindex independently.
- [ ] `UserDefaultsStore_BookmarksTests` — existing sort-order tests plus the two write-count tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bookmark insertion and deletion reliability, including safe handling of missing bookmarks.
  * Preserved bookmark ordering when moving or deleting items.
* **Performance**
  * Bookmark edits now save changes more efficiently, reducing unnecessary storage operations.
* **Tests**
  * Added coverage for bookmark order, deletion behavior, item counts, and save frequency.
* **Documentation**
  * Documented bookmark editing performance improvements and ordering behavior.

### DIFF
--- a/OBAKitCore/Models/UserData/UserDataStore.swift
+++ b/OBAKitCore/Models/UserData/UserDataStore.swift
@@ -570,60 +570,52 @@ public class UserDefaultsStore: NSObject, UserDataStore, StopPreferencesStore {
 
         bookmark.groupID = group?.id ?? nil
 
-        if let existing = findBookmark(id: bookmark.id) {
-            delete(bookmark: existing, reorderGroup: true)
-        }
+        // Decode once, mutate in memory, encode once. The previous loop called
+        // `bookmarks` get/set (full plist round-trip) per row, which froze the
+        // UI for several seconds with a dozen bookmarks. See #548.
+        var all = bookmarks.filter { $0.id != bookmark.id }
+        let newGroupID = bookmark.groupID
 
-        var newGroupBookmarks = bookmarksInGroup(group)
-        newGroupBookmarks.insert(bookmark, at: min(index, newGroupBookmarks.count))
-
-        for (idx, elt) in newGroupBookmarks.enumerated() {
-            if let existing = findBookmark(id: elt.id) {
-                delete(bookmark: existing, reorderGroup: false)
-            }
-
+        var destination = all
+            .filter { $0.groupID == newGroupID }
+            .sorted { $0.sortOrder < $1.sortOrder }
+        destination.insert(bookmark, at: min(index, destination.count))
+        for (idx, elt) in destination.enumerated() {
             elt.sortOrder = idx
-
-            bookmarks.append(elt)
         }
 
-        if oldGroupID != bookmark.groupID {
-            let oldGroupBookmarks = bookmarksInGroup(findGroup(id: oldGroupID))
-            for (idx, elt) in oldGroupBookmarks.enumerated() {
-                if let existing = findBookmark(id: elt.id) {
-                    delete(bookmark: existing, reorderGroup: false)
-                }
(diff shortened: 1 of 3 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is a good fix for a genuinely miserable bug. The root cause is worth stating plainly for the record: `bookmarks` is a computed property that decodes the whole plist on every `get` and re-encodes it on every `set`, and the old code called both once per row in a loop. Decode once, mutate in memory, encode once is exactly right.

I traced the coalescing rather than taking it on faith, since a reorder bug here silently corrupts everyone's bookmark list:

- **Nothing is dropped.** `all.removeAll { destinationIDs.contains($0.id) }` followed by `all.append(contentsOf: destination)` is exactly complementary, and `all` already had the moved bookmark filtered out at the top, so there's no double-remove.
- **The in-memory reindex actually lands.** `Bookmark` is a `final class`, so `elt.sortOrder = idx` on the filtered `destination`/`remaining` slices mutates the same instances `all` holds. This would be a silent no-op if it were a struct — worth a comment if you ever touch it again.
- **`sortOrder` semantics are preserved.** For an in-group move, the old code compacted the group during the delete and then reinserted; you sort the uncompacted list instead, which yields the same relative order and — importantly — the same `min(index, count)` clamp, because both versions exclude the moved bookmark before counting. For a cross-group move, the destination reindex plus the `oldGroupID != newGroupID` remaining-reindex reproduces the old two-pass result.
- **Notification count is unchanged** for real callers: `add` still posts exactly once, and the removed private overload never posted.

The new code is also quietly more correct than what it replaces: the old path reindexed `bookmarksInGroup(findGroup(id: oldGroupID))`, which silently reindexed the ungrouped bucket when the old group had already been deleted. `all.filter { $0.groupID == oldGroupID }` doesn't have that failure mode.

Two notes, neither blocking:

- **`Moving a bookmark encodes the bookmark list once` doesn't move anything.** It inserts a *new* bookmark at index 3, so it never exercises the existing-bookmark-removal branch a real drag takes. The assertions are correct and it does fail against the old loop, so it isn't vacuous — but the name, the PR body, and the doc all claim move coverage that isn't there. Cheap fix: retain one of the 12 seeded bookmarks, pass it to `add(_:to:index:)`, and assert the count stays 12.
- **`delete(bookmark:)` no longer posts `.bookmarksDidChange` when the bookmark isn't in the store.** Both production callers resolve the bookmark from the store first, and a no-op delete has nothing to refresh, so I think this is the better behavior — just noting it's an unremarked semantic change rather than a pure refactor.

Approving.

Issue #1198, opened 2026-07-17
Audit and unwind tactical concurrency escape hatches from the Swift 6 migration


aaronbrethorst:
Follow-up to #1192. The migration used `@unchecked Sendable` / `nonisolated(unsafe)` deliberately, each with a documented invariant. Most are permanent by design (the REST models under the `HasReferences` contract can't get checked conformances while they subclass NSObject), but several could become compiler-checked with modest refactors:

- `ArrivalDeparture.route/stop/trip` (and other `HasReferences` conformers): convert post-`loadReferences` stored `var`s to `private(set)` per the plan doc's own recommendation, so the "read-only after handoff" contract is partially enforced.
- `AnalyticsOrchestrator.userDefaults` (`nonisolated(unsafe) let`): could become checked if/when UserDefaults gains a Sendable annotation, or by routing `reportingEnabled()` through a Sendable wrapper.
- `DecodingErrorReporter._reportHandler` (`nonisolated(unsafe)` + lock): a candidate for `Mutex` (iOS 18+) which is checked.
- `UncheckedSendableBox` (`OBAKit/Utilities`): re-evaluate both MapKit call sites (`MapRegionManager.handleMapFeatureSelection`, `MapItemViewModel.fetchScene`) whenever the deployed Swift compiler floor rises — newer region analysis accepts the direct `await` and the boxes can be deleted.
- `AgencyAlertsStore` notify methods use `Task { @MainActor }`, which doesn't guarantee FIFO relative to `DispatchQueue.main.async` work; no delegate currently depends on ordering, but note it if delegate contracts grow.
- `Icons.iconCache` / other `nonisolated(unsafe)` NSCache statics: checked automatically once NSCache is annotated Sendable.

Rule of thumb from the plan: prefer deleting an escape hatch over re-justifying it whenever a toolchain or dependency update makes the checked form possible.

Borisserz:
Taking a small first slice of the concurrency-hatch audit (not the whole issue).

Borisserz:
Looks like the first unwind slice already landed on \`main\` (\`docs/swift6-escape-hatches.md\`: \`private(set)\` on ArrivalDeparture/StopArrivals refs + \`Mutex\` for \`DecodingErrorReporter\`). Leaving the remaining hatches (UncheckedSendableBox, AnalyticsOrchestrator, Trip.route) for a later PR unless you want them bundled.

PR #1296, opened 2026-08-18
Tighten a few Swift 6 concurrency escape hatches

Borisserz:
## Summary
- `ArrivalDeparture.route`, `.stop`, and `.trip`, plus `StopArrivals.stop`, are `public private(set)`. `loadReferences` remains the only writer.
- `DecodingErrorReporter.reportHandler` is a compiler-checked `Mutex` instead of `nonisolated(unsafe)` plus `NSLock`.
- Left alone on purpose: `Trip.route` (tests assign it; `References` has no public factory), `UncheckedSendableBox` (MapKit hops), `UserDefaults` / `NSCache` marks waiting on Apple.

Does not finish the whole audit. **Refs #1198.**

## Test plan
- [ ] `ArrivalDepartureTests` `Load references` still wires route/stop/trip.
- [ ] `DecodingErrorReporterTests` still set and invoke `reportHandler`.
- [ ] App builds in the Swift 6 language mode.

### DIFF
--- a/OBAKitCore/Models/REST/ArrivalDeparture.swift
+++ b/OBAKitCore/Models/REST/ArrivalDeparture.swift
@@ -50,7 +50,7 @@ public final class ArrivalDeparture: NSObject, Identifiable, Decodable, HasRefer
     public let routeID: RouteID
 
     /// the route for the arriving vehicle
-    public var route: Route!
+    public private(set) var route: Route!
 
     /// the route long name that potentially overrides the route long name in the referenced `Route` element
     private let _routeLongName: String?
@@ -79,7 +79,7 @@ public final class ArrivalDeparture: NSObject, Identifiable, Decodable, HasRefer
     public let stopID: StopID
 
     /// The stop the vehicle is arriving at
-    public var stop: Stop!
+    public private(set) var stop: Stop!
 
     /// The index of the stop into the sequence of stops that make up the trip for this arrival
     public let stopSequence: Int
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is exactly the right shape for unwinding escape hatches: only take the ones the compiler can check for you, and write down why you left the rest alone.

I verified both halves before approving:

- **`private(set)` on `ArrivalDeparture.route/.stop/.trip` and `StopArrivals.stop`** — I grepped the whole tree for assignments to those properties, and every hit is on a different type (`Bookmark.stop`, `StopViewModel.stop`, `Trip.route`, various view controllers). `loadReferences` really is the only writer, which matches the `HasReferences` read-only-after-handoff contract. And of course a green build is the strongest evidence here, since the compiler would reject any external write outright.
- **`Mutex` in place of `nonisolated(unsafe)` + `NSLock`** — semantically identical, and it converts a comment-enforced invariant into a compiler-enforced one, which is a clear win. `Synchronization` needs iOS 18.0 and `app_shared.yml` pins the deployment target to exactly that, so we're fine.

`docs/swift6-escape-hatches.md` is a genuinely useful addition — the "left alone on purpose" section with reasons is what makes the next pass at this tractable.

One tiny thing, not worth holding the merge over: in `report(...)` the local `guard let handler = reportHandler` now shadows the new static `handler` mutex. Correct, just slightly confusing to read. Rename one of them if you touch the file again.

Approving.

Issue #1278, opened 2026-08-12
Add a ‘compact’ mode for stops and trip details


aaronbrethorst:
A request from a user for a more “space saving” mode came in. I think this would be a great addition to the new stop and trip interfaces. Opt in, off by default.

PR #1297, opened 2026-08-18
Add an opt-in compact mode for stop and trip pages

Borisserz:
## Summary
- Settings → Accessibility → **Compact stop and trip pages**. Off by default.
- New stop departure rows, grouped cards, trip header card, and trip stop timeline read tighter spacing from `StopTripSpacing`. Accessibility-size stacked layouts are unchanged. No arrivals or occupancy are hidden.
- `@AppStorage` key is `stopTripCompactMode` (dot-free, same KVO constraint as `stopUIReducedColors`). New keys in all 13 catalogs.

**Refs #1278.**

## Test plan
- [ ] Setting defaults off. Turning it on persists after leaving Settings.
- [ ] New stop page (chronological and grouped) and new trip page look unchanged with the switch off.
- [ ] Switch on: rows and the trip timeline are tighter. Accessibility sizes still stack.
- [ ] `UserDefaultsStoreTests` compact-mode persistence. `StopTripSpacingTests` pins regular vs compact numbers.

## Summary by CodeRabbit

- **New Features**
  - Added an Accessibility setting for compact stop and trip pages.
  - Stop and trip screens now support reduced spacing while preserving content and tap-target requirements.
  - The preference is saved and applied across supported screens.

- **Localization**
  - Added translations for the new setting across supported languages.

- **Documentation**
  - Documented compact layout behavior and accessibility considerations.

- **Tests**
  - Added coverage for preference persistence, compact spacing, and app-storage behavior.


aaronbrethorst:
### Code review

Found 1 issue:

1. Compact mode does reach the accessibility-size stacked layouts, contrary to the invariant this PR states. `StopTripSpacing` says "Accessibility-size stacked layouts do not consult this table", and `docs/compact-stop-trip-mode.md` says they are "left alone", but the outer `VStack` that wraps the `isAccessibilitySize` branch takes its spacing from `StopTripSpacing.stack(compactMode)`, so at AX sizes headsign / time / status / occupancy go from 3pt apart to 1pt. Same shape in `GroupedListView.headerPrimaryRow`; `TripCardView`'s outer `VStack` (10 to 6) and `.padding` (14 to 10) also apply inside its AX branch, and `TripStopRow`'s vertical padding (11 to 6) has no AX branch at all. Either gate the table on `dynamicTypeSize.isAccessibilitySize` at those call sites, or correct the comment and the doc to say that compact tightens AX layouts too.

https://github.com/OneBusAway/onebusaway-ios/blob/8ac6486b93ba7a7de6b84997ec1b5a17592354e2/OBAKit/Stops/StopPage/Departures/DepartureRowView.swift#L54-L56

OBAKit/Trip/TripPage/TripStopListView.swift
```diff
@@ -77,7 +81,7 @@ private struct TripStopRow: View {
                     Spacer(minLength: 8)
                     time
                 }
-                .padding(.vertical, 11)
+                .padding(.vertical, StopTripSpacing.stopRowVertical(compactMode))
```
aaronbrethorst, line 84: This is the blocker. `TripStopRow` is tappable (`.onTapGesture(perform: onSelect)` a few lines down), and its height is essentially one subheadline line plus this padding. At 11pt it is already ~42pt; at compact 6pt it drops to roughly ~32pt, well under the 44pt minimum.

Adding `.frame(minHeight: 44)` on the row keeps the visual density compact without shrinking the hit target.

OBAKit/Stops/StopPage/StopTripSpacing.swift
```diff
@@ -0,0 +1,34 @@
...
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import CoreGraphics
+
+/// Spacing table for the new stop and trip screens. Compact values are the
+/// opt-in space-saving mode (#1278); regular values are the layout those
+/// screens shipped with. Accessibility-size stacked layouts do not consult
```
aaronbrethorst: This says accessibility-size stacked layouts do not consult the table, but they do: in `DepartureRowView`, `GroupedListView.headerPrimaryRow`, and `TripCardView`, the `VStack` that reads `StopTripSpacing.stack(_:)` / `.card(_:)` wraps the `if dynamicTypeSize.isAccessibilitySize` branch rather than sitting inside the non-AX side. `TripCardView.padding(StopTripSpacing.cardPadding(_:))` applies at AX sizes too.

Either move the table reads inside the non-AX branches, or reword this comment (and the matching claim in `docs/compact-stop-trip-mode.md`).

OBAKit/Stops/StopPage/Departures/DepartureRowView.swift
```diff
@@ -51,7 +52,7 @@ struct DepartureRowView: View {
         // fast path holds. At accessibility sizes the row "stacks" (the guide's
         // committed layout): badge + countdown share the first line as glance
         // tokens, destination + status flow below — nothing is dropped.
-        VStack(alignment: .leading, spacing: 3) {
+        VStack(alignment: .leading, spacing: StopTripSpacing.stack(compactMode)) {
```
aaronbrethorst, line 58: Concrete instance of the mismatch above: this `spacing:` applies to the accessibility-size stacked branch on the very next line, taking it from 3pt to 1pt in compact mode.

aaronbrethorst (changes requested):
The mechanism here is clean, and I want to be clear about what's working before I get to what isn't. One key, one place it's read, registered default `false`, `@AppStorage` default `false`, both covered by tests — the toggle genuinely can't change behavior for anyone who doesn't opt in. Pulling the numbers into `StopTripSpacing` instead of scattering `compactMode ? x : y` across five views is the right structural call, and you got all 13 locale catalogs. That's the tedious part and you didn't skip it.

Two things to fix before this lands, both small.

**1. The trip-stop row's tap target drops below 44pt in compact mode.** `TripStopRow` is tappable via `.onTapGesture`, and its only height comes from a subheadline line plus the vertical padding. At 11pt that's already borderline (~42pt); at compact's 6pt it lands around ~32pt. A row you can tap should stay at 44pt regardless of the spacing table — a `.frame(minHeight: 44)` on the row keeps compact mode visually tighter without shrinking the target. This is the one that bothers me most, because the toggle lives under Settings → Accessibility, so an accessibility-section setting shouldn't quietly make a control harder to hit.

**2. The "accessibility sizes don't consult this table" invariant isn't true as written.** `StopTripSpacing`'s doc comment says AX-size stacked layouts "do not consult this table," and `docs/compact-stop-trip-mode.md` says they're left alone — but in `DepartureRowView`, `GroupedListView.headerPrimaryRow`, and `TripCardView` the stack that reads `StopTripSpacing` *encloses* the `if dynamicTypeSize.isAccessibilitySize` branch, so AX layouts do get the compact numbers. `TripCardView`'s `.padding(StopTripSpacing.cardPadding(...))` applies in the AX branch too.

I don't think the behavior is necessarily wrong — someone who opts into compact mode at an accessibility text size may well want tighter spacing. But the comment and the doc need to match the code, or the next person to touch these views will reason from an invariant that doesn't hold. Either gate the table on `isAccessibilitySize`, or correct both the comment and the doc. Your call which.

- The new `.strings` entries translate the developer comments in the non-English catalogs. Every one of the existing comments in those files carries the English source comment, and `scripts/extract_strings` will revert yours on the next run — translators also lose the context those comments exist to provide.
- `StopTripSpacingTests` restates the literal table values, and your own comment concedes it stays green if a call site stops using `StopTripSpacing`. It pins constants rather than verifying behavior. Not worth expanding now, just worth knowing it isn't buying much.

Happy to re-review as soon as the tap target and the doc/code mismatch are sorted.

Borisserz:
Tappable `TripStopRow`s now use `.frame(minHeight: StopTripSpacing.stopRowMinHeight)` (44pt) before `.contentShape` / `.onTapGesture`, so compact 6pt vertical padding cannot shrink the hit target.

Docs and the `StopTripSpacing` comment match the wrapping: accessibility-size stacked layouts **do** consult the table (outer `VStack`s wrap the AX branch). Compact-at-AX stays — the comment was the lie, not the layout.

`Tappable trip-stop rows keep a 44pt minimum height` in `StopTripSpacingTests` (3/3 passed locally).

aaronbrethorst:
### Code review

Found 1 issue:

1. The new `.frame(minHeight:)` on `TripStopRow` fixes the hit target but takes the row's content with it: a flexible frame with the default `.center` alignment stretches the connector (good — the line stays continuous) while the inflexible name/divider `VStack` gets centered in the 44pt box. Rendering the row structure at a 15pt subheadline gives: regular row 41.7pt → 44pt with the divider 2pt off the row boundary; **compact** row 31.7pt → 44pt with the divider 7pt off the boundary, so the separator sits 6pt under its own stop name and ~13pt above the next one. Two consequences: (a) for single-line stop names at default Dynamic Type, compact and regular trip-stop rows are both exactly 44pt, so the setting saves nothing on the stop timeline — the 6 vs. 11 `stopRowVertical` split only shows up once names wrap or text scales up; (b) the minimum is unconditional, so it also grows the *non-compact* rows (41.7 → 44), which contradicts "New stop page ... and new trip page look unchanged with the switch off" in the test plan and `docs/compact-stop-trip-mode.md`. Applying the minimum to the inner content (or giving the frame a `.top`/stretching alignment so the divider stays on the boundary) keeps the 44pt target without the centered dead space.

https://github.com/OneBusAway/onebusaway-ios/blob/1ad875fb48993fbd3b8411792dc1679c1365b049/OBAKit/Trip/TripPage/TripStopListView.swift#L92-L96

aaronbrethorst (changes requested):
Both of my blocking items from the last round are properly addressed — the 44pt minimum is there, and the `StopTripSpacing` doc comment and `docs/compact-stop-trip-mode.md` both say the right thing about AX layouts now. The mechanism underneath is solid: the defaults key, the registration, and the settings round-trip are a faithful copy of the `stopUIReducedColors` pattern, and every regular-mode constant in the table matches the literal it replaced. Centralizing the spacing was the right idea.

The problem is where the 44pt floor landed.

`.frame(minHeight: StopTripSpacing.stopRowMinHeight)` sits on the whole row, outside the padding. With a default subheadline line at roughly 20pt, compact rows come out around 32pt and regular rows around 42pt — both under 44, so both get clamped to exactly 44. Two consequences:

Compact mode does nothing to the trip stop list at default Dynamic Type. That's the one screen the setting is named after. Your own comment in the table concedes this ("Tappable trip-stop rows keep a 44pt minimum height regardless"), so I don't think it's a surprise to you — but it does mean the feature's headline case is a no-op, which is worth resolving rather than documenting.

The non-compact path also changed, and that one is a surprise. Regular rows grow a couple of points, and because a flexible frame centers its child, the content gets centered inside the 44pt box. The rule at the bottom of the `VStack` is no longer flush with the row boundary — it ends up a few points above it, with more space below than above. The comment right there says the rule stops short of the connector "so the line reads as continuous through the whole list", and off-boundary dividers are precisely what breaks that read. This affects every user, not just the ones who opt in, which is why I'd rather not land it as-is.

Putting the minimum on the tappable content — or on the inner `HStack` before the divider joins it — should keep the hit target honest without moving the divider or nudging the default layout.

One leftover from last round, still non-blocking: the non-English `.strings` files carry translated `/* */` comments. `scripts/extract_strings` regenerates `en.lproj` via `genstrings`, so those won't survive a regen cleanly. Low stakes, but it's the second round it's come up.

Also worth a look: `StopTripSpacingTests` asserts `44 == 44` and `44 > 12`, which is a test of the constant rather than of anything the constant does — it wouldn't have caught the centering above. Not a blocker, just noting that the coverage isn't buying much here.

One housekeeping note: `StopTripSpacingTests.swift` has a comment reading "the thing Aaron blocked #1297 on." I'd rather not carry reviewer names and PR numbers in the source tree permanently — a plain description of the constraint reads better in six months.

Fix the frame placement and I'll take another look.

Borisserz:
`.frame(minHeight: 44)` is off the whole `TripStopRow`. At default Dynamic Type that clamped compact (~32pt) and regular (~42pt) to the same 44pt row and, with default center alignment, floated the divider off the connector.

The 44pt floor is now canceling vertical padding around the tappable row (`+slop` / `contentShape` / `−slop`). Laid-out height stays `text + 2 * stopRowVertical`. Compact still tightens the trip list; the hit target is ≥ 44pt. Regular layout is unchanged.

`StopTripSpacingTests` now asserts compact layout < 44 < hit height (not `44 == 44`). That suite passed on iPhone 17 Pro. Putting `minHeight` back on the outer row makes compact layout equal the floor and fails the new test.

aaronbrethorst:
### Code review

Found 1 issue:

1. The canceling-padding trick does not actually deliver a 44pt hit target on the trip stop list, and it now offsets each row's tap area from its own pixels. `TripStopListView` lays the rows out in `LazyVStack(spacing: 0)`, and the `-stopRowHitSlop` at line 97 cancels the `+slop` at line 94, so consecutive rows sit exactly `stopRowLayoutHeight` apart (compact: ~32pt at default Dynamic Type). N rows therefore occupy 32·N points of screen, and no arrangement of per-row rectangles can give each of N rows 44pt of *disjoint* tappable area inside 32·N points — every row's 44pt rect overlaps each neighbour's by 12pt, so the average disjoint target per interior row is still 32pt. Worse, SwiftUI resolves an overlap to the later-drawn sibling, so the 6pt band that visually belongs to row *N* (it contains row *N*'s own `Divider`) dispatches to row *N+1*: the effective target isn't enlarged, it's shifted 6pt upward off the content it labels. Tapping a stop's separator selects the next stop. Two supporting details: `stopRowHitSlop` is computed from the hard-coded `stopRowDefaultTextHeight = 20` rather than from the row's real height, so the full 6pt of overlap is applied even when the name wraps to two lines or the text scales up and the row already clears 44pt — the "extra slop there is harmless" note at `StopTripSpacing.swift:43-44` doesn't hold once the slop is overlapping a neighbour; and the same mechanism adds a (negligible, but non-zero) ±1pt of overlap in *non*-compact mode. This also makes `StopTripSpacing.swift:19` ("Tappable trip-stop rows keep a 44pt *hit* target") and the matching claim in `docs/compact-stop-trip-mode.md` inaccurate. Giving the row real height for the target — or keeping the row 44pt and moving the compact savings somewhere that isn't the tappable row — is the shape that survives adjacent rows; overlapping rects can't.

https://github.com/OneBusAway/onebusaway-ios/blob/d0cb9d77917557a79c31a3e83e5598d2dd80f89d/OBAKit/Trip/TripPage/TripStopListView.swift#L93-L98

aaronbrethorst (changes requested):
Two of the three things I blocked on last round are genuinely settled, and I want to be clear about that before the rest.

The AX invariant now matches the code. The comment used to claim accessibility-size stacked layouts didn't consult the spacing table, and they did; you corrected the comment rather than gating the table, which was the call I left you, and `docs/compact-stop-trip-mode.md` says the same thing. The divider is back on the row boundary — regular slop works out to 1pt and the `+1`/`-1` cancels, so the non-compact path lays out at the same 42pt it always did. Key registration, the settings round-trip, the spacing centralization, and 13-locale parity on the new string are all in good shape. Compact rows really are tighter now, which is the point of the feature.

The 44pt tap target still isn't real, though, and the current shape makes things slightly worse than a plain small row would.

`TripStopListView` puts the rows in a `LazyVStack(spacing: 0)`, and `TripStopRow` applies `+slop` / `contentShape` / `-slop`. The trailing negative padding cancels the leading positive one, so what the stack lays out is `stopRowLayoutHeight` — 20 + 6*2 = 32pt in compact at default type. But `contentShape` is measured on the padded frame, so each row claims 44pt. Rows 32pt apart with 44pt rects overlap their neighbours by 12pt, and you can't fit N disjoint 44pt rectangles into 32*N points no matter how the padding is arranged.

Where the overlap resolves is what makes it a bug rather than a no-op. SwiftUI gives an overlapping region to the later sibling, so row N wins the 6pt it steals from row N-1's content and loses the 6pt it lends to row N+1. Net disjoint target is still 32pt, and it sits 6pt above the row's own pixels — the band containing a row's own `Divider` dispatches to the next row down. A tap near the bottom of a stop name opens the stop below it. On `main` the hit area equals the visual row and is aligned, so this is new.

Two smaller things feeding the same knot: `stopRowHitSlop` derives from the hard-coded `stopRowDefaultTextHeight = 20` rather than the row's actual height, so the full 6pt offset applies even when a name wraps to two lines and the row already clears 44pt on its own — which is the case `StopTripSpacing.swift:43-44` says is harmless, and it isn't. And `stopRowHitHeight(compact:) >= stopRowMinHeight` in `StopTripSpacingTests` can't fail: given `hitSlop = max(0, (44 - layout)/2)`, the expression is >= 44 for every input, so it would have passed on this exact bug.

The fix I'd take is to stop trying to buy the 44pt back. Delete `stopRowHitSlop`, `stopRowHitHeight`, `stopRowMinHeight` and both padding modifiers, and let a compact row be a 32pt row whose tap area is its own content. Compact mode is opt-in and the rider asked for tighter rows; a smaller target is the trade they chose, and an honest 32pt aligned target beats a 32pt one shifted off its content. Then correct `StopTripSpacing.swift:19` and the docs page — both currently promise a 44pt hit target — to say what compact actually does. If you'd rather keep 44pt, the only way to get it is to stop compact from going below it, which means compact stops tightening this particular row, and that's a legitimate choice too. Either way the code and the docs need to agree with the geometry.

Not blocking, and I've mentioned both before, so treat them as yours to take or leave: the twelve non-English `.strings` files carry translated `/* */` comments where every neighbouring entry has the English source comment, and the spacing tests mostly pin constants rather than behavior — the tautology above is the sharp end of that.

The feature itself is right and I'd like it in. Fix the tap target, square the comment and docs with it, and I'll re-review promptly.

Borisserz:
Removed `stopRowHitSlop` / `stopRowHitHeight` / `stopRowMinHeight` and both canceling paddings. A compact trip-stop row is now its own tap target (~32pt at default Dynamic Type). `LazyVStack(spacing: 0)` plus overlapping 44pt rects was giving the later sibling the band that contains the previous row's divider.

Docs and the `StopTripSpacing` comment say that: compact is opt-in, a smaller aligned target is the trade, do not put a 44pt overlay back on.

`StopTripSpacingTests` (3) passed on iPhone 17 Pro. Compact layout `20 + 6*2` stays under 44; regular stays the shipped 11pt vertical pad.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Everything I blocked on across the previous three rounds is addressed, and addressed in the shape I asked for: the slop/`minHeight`/`stopRowHitHeight` machinery is gone entirely, the compact row is its own aligned target, and both `StopTripSpacing.swift` and the docs say that rather than promising 44pt. I checked every regular-mode constant against the literal it replaced (3/13/3, 10, 14, 11, 14) — switch-off is byte-for-byte identical to main.

Two leftovers I am explicitly *not* holding this on, but would take in a follow-up:

- `docs/compact-stop-trip-mode.md` gives the wrong reason for banning `minHeight`. The overlap argument is true of the canceling slop but not of `.frame(minHeight:)`, which reports 44pt to the stack and lays rows out 44 apart with no overlap. The directive is right; the rationale is half wrong, which is the same class of misleading-invariant thing I blocked rounds 1 and 2 on.
- The twelve non-English catalogs still carry translated `/* */` comments. Third time I have mentioned it; `scripts/extract_strings` will not preserve them.

Borisserz:

aaronbrethorst (changes requested):
The merge of `main` you pushed to resolve conflicts didn't land cleanly, and the branch no longer compiles — that's what the red build is.

`OBAKitTests/Modeling/UserData/UserDataStoreTests.swift` has a doubled closing brace at lines 256-257: line 256 closes the transfer-banner test, and line 257 closes `UserDefaultsStoreTests` itself. Everything from line 258 onward — the survey properties, walking speed, and default alarm lead time tests, roughly 120 lines — ends up at file scope instead of inside the suite, and the file's real final brace becomes the error the compiler reports:

```
UserDataStoreTests.swift:380:1: error: Extraneous '}' at top level
```

Deleting the stray `}` on line 257 should be the whole fix.

Two other things while you're in there. Worth re-reading the rest of that merge resolution rather than just this one file — a stray brace usually means the conflict was resolved by hand under time pressure, and the same thing may have happened somewhere the compiler is happy but the behavior changed. And `GroupedListView.swift` / `DepartureRowView.swift` both moved on `main` since you branched (#1382 landed a context menu on the grouped rows), so you'll want a fresh merge anyway.

The compact-mode work itself looked fine at the last review — this is purely the merge. Ping me once it's building and I'll take another pass.

Borisserz:
Fixed the compile break: deleted the stray \`}\` after the transfer-banner test so the rest of \`UserDefaultsStoreTests\` is inside the suite again. Also merged current \`main\` (including #1382 context menu) cleanly.

Ready for another pass.

aaronbrethorst:
### Code review

Found 1 issue:

1. Compact mode is a no-op on the trip page — half of what the setting's label promises. The Settings toggle writes through `application.userDataStore`, which in the shipping app is `UserDefaultsStore(userDefaults:)` built on the **app-group suite** (`Apps/OneBusAway/AppDelegate.m:36`, `[[NSUserDefaults alloc] initWithSuiteName:appGroup]`). A bare `@AppStorage` with no `store:` reads `UserDefaults.standard`, which is a different domain — nothing in the repo calls `addSuite(named:)`. The stop page works around exactly this: `StopPageRootView` exists solely to apply `.defaultAppStorage(userDefaults)` ("so the page's `@AppStorage` reads and writes the app-group suite ... rather than `UserDefaults.standard`"), and `StopDetailsSheetRootView` does the same for the sheet. The trip page has no such wrapper — `TripPageViewController` is `UIHostingController<TripPageView>` and hands `TripPageView` to `super.init(rootView:)` unmodified, so the three new readers (`TripPageView` L67, `TripCardView` L34, `TripStopListView` L25) resolve against `UserDefaults.standard`, which has neither the registered default nor the written value. They fall back to `false` forever: `tripPage`, `card`, `cardPadding`, `hStack` and `stopRowVertical` never see `true`, and the trip header card and stop timeline stay at regular spacing no matter what the switch says. The `stopTripCompactModeKey` doc comment ("Dot-free so `@AppStorage` on the stop and trip pages can observe it") and `docs/compact-stop-trip-mode.md` both assert the trip half works. Worth noting the pre-existing `@AppStorage(stopUIReducedColorsKey)` on `TripCardView` has the same defect, so the fix — a thin root view around `TripPageView` applying `.defaultAppStorage(application.userDefaults)`, remembering that the modifier only reaches `@AppStorage` *below* the view it is applied to — repairs both.

https://github.com/OneBusAway/onebusaway-ios/blob/ddbb7a903dabe38e29089d1f56aefbc9a1ba630a/OBAKit/Trip/TripPage/TripPageView.swift#L66-L68

aaronbrethorst (changes requested):
The merge damage from last round is properly cleaned up — braces balance, the compact-mode tests sit inside `UserDefaultsStoreTests` where they belong, and CI is green. I also diffed every changed file against `main` and confirmed switching compact off is byte-identical to today, which is what I wanted. The tap-target saga from rounds 1-3 is settled.

But CI being green hid something, and it's blocking: **compact mode doesn't do anything on the trip page.**

The Settings toggle writes through `application.userDataStore`, which in the shipping app is the app-group suite — `Apps/OneBusAway/AppDelegate.m:36` builds it with `initWithSuiteName:appGroup`. A bare `@AppStorage` reads `UserDefaults.standard`, a different domain, and nothing calls `addSuite(named:)`.

The stop page compensates for exactly this: `StopPageRootView` exists only to apply `.defaultAppStorage(userDefaults)` (`StopPageView.swift:115`), and `StopDetailsSheetView.swift:51` and `BookmarksListView.swift:51` do the same, each with a comment explaining why.

`TripPageViewController` has no such wrapper — it passes `TripPageView` straight to `super.init(rootView:)` at line 61 and reassigns it at 191. So your three new readers in `TripPageView`, `TripCardView` and `TripStopListView` resolve against `.standard`, which holds neither the written value nor the registered default. `tripPage`, `card`, `cardPadding`, `hStack` and `stopRowVertical` never see `true`.

The fix is a thin `TripPageRootView` applying `.defaultAppStorage(application.userDefaults)`, modelled on `StopPageRootView`. Worth knowing that `.defaultAppStorage` only reaches `@AppStorage` *below* the view it decorates, which is why these are separate root views rather than a modifier on the hosting controller.

Useful side effect: `TripCardView`'s existing `@AppStorage(stopUIReducedColorsKey)` has the identical bug on `main` today, so the same wrapper fixes both. I didn't score that one since it's not your line, but it's evidence the pattern is easy to miss and worth a comment when you add the wrapper.

Three leftovers, all previously non-blocking and all unchanged — fine to carry, just noting they're still there:

`docs/compact-stop-trip-mode.md` still gives the overlap argument as the reason to avoid `minHeight`. That reasoning applies to the canceling slop, not to `.frame(minHeight:)`, which reports 44pt and lays rows out 44 apart with no overlap.

Twelve non-English catalogs carry translated `/* */` developer comments. `scripts/extract_strings` regenerates those from the `OBALoc` `comment:` argument, so they won't survive a regen and translators lose the English context.

`StopTripSpacingTests` asserts the table's own literals back at it — including `42 < 44`, which is arithmetic on constants. It wouldn't have caught any of the three layout defects from rounds 1-3, nor the store mismatch above. If you want one test that earns its keep, assert that a compact-mode view actually reads `true` from the app-group suite.

Borisserz:
Compact mode on the trip page was reading `UserDefaults.standard` while Settings writes the app-group suite — same gap `StopPageRootView` already closes for the stop page.

Added `TripPageRootView` applying `.defaultAppStorage(application.userDefaults)` and pointed `TripPageViewController` at it. That also fixes `TripCardView`'s existing `@AppStorage(stopUIReducedColorsKey)` bug.

Also corrected the docs/`StopTripSpacing` note on why not to use `.frame(minHeight: 44)` (clamps compact=regular at default Dynamic Type; the overlap argument is for canceling slop only), and added `TripPageAppStorageTests` so a suite write of `true` is visible under `.defaultAppStorage` and invisible on `.standard`.

`TripPageAppStorageTests` / `TripPageCollapseTests` / `StopTripSpacingTests`: 6/6 passed on iPhone 17 Pro.

aaronbrethorst:
### Code review

_Re-review at 000c844, covering commits since the previous review._

Borisserz:
Addressed the trip-page store mismatch in `000c844a`: `TripPageRootView` applies `.defaultAppStorage(application.userDefaults)` (same pattern as `StopPageRootView`), plus `TripPageAppStorageTests` that assert compact mode reads `true` from the app-group suite.

Ready for another look.

aaronbrethorst (approved):
Verified at 000c844a. Thanks for this.

Issue #1222, opened 2026-07-25
Feature Request - Shortcut notification trigger


nbarnard:
The live notifications enhancement is much appreciated. 

It would be excellent if a shortcut could trigger a live notification for a specific bus stop and routes. This would allow people to have an automation that would trigger at specific locations and/or times to bring up their needed realtime transit information.

aaronbrethorst:
Thanks for the feature request! I think this is a great idea.

aryn-rjpt:
Hi @aaronbrethorst — I’ve also been working on this feature over the past few weeks. It’s been a little slow going since I’m fairly new to iOS, so I’ve been spending some time getting familiar with the codebase, frameworks, etc.

I’ve pushed my work here:
https://github.com/aryn-rjpt/onebusaway-ios/tree/Feature/LiveActivityFromShortcuts

You can either diff it against `main` or look at the latest commit for the changes.

I’d love to get some feedback on the approach before I take this further. Here’s what I’ve done/proposed so far:

* **Live Activity creation:** Right now, the Live Activity is created directly from `BookmarksViewController`, which means a UI layer is also handling business logic. I’ve introduced a `LiveActivityManager` to provide a single place for this logic, and the Shortcuts → Live Activity flow goes through that manager.

* **Single entry point for Live Activity creation:** My thought is that all Live Activity creation should go through something like `LiveActivityManager`. This seems particularly useful because when an App Intent runs in the background, all the objects marked with `@Dependency` (and their required dependencies) are instantiated.

* **Bookmarks / data access:** I think bookmark access should ideally go through a single source of truth, such as `Application.userDataStore`, rather than creating a separate `UserDefaults` instance using `Bundle.main.appGroup`. However, trying to use `Application.userDataStore` here runs into concurrency issues and may require changes to `Application.swift`.

* **Intent Target Membership** Also, the intent files need to have a target membership of OBA app and not OBAKit. As of now, I have kept them inside OBAkit, but with target membership of app, which feels awkward. So should I move these files to OBA app? WIll that be right because as far as I have gone through the project, app is supposed to be totally made from OBAkit and OBAKitCore, so anyone else can use them. What are your thoughts on this?

I’m not entirely sure what the best approach is for the last point, so I’d really appreciate your thoughts on it.

I’ve also added an `OpenBookmarkIntent`, in case we want to support opening bookmarks directly from Shortcuts as well.

[...]

PR #1298, opened 2026-08-18
Add a Shortcut that starts a bookmark Live Activity

Borisserz:
## Summary
- New **Track Bookmark** App Intent (`openAppWhenRun`) queues a trip-bookmark UUID in the app-group defaults. The Bookmarks tab then starts (or promotes) the Live Activity on the existing Track path once arrivals load. No `Activity` is captured into a `Task`.
- Whole-stop bookmarks are omitted from the entity query: they have no single route to Track. Arbitrary stop IDs that are not bookmarked are out of scope.
- Intents live in OBAKit. The app target includes them through `OBAAppIntentsPackage` → `OBAKitAppIntentsPackage`; without that chain, Shortcuts never sees Track.

**Refs #1222.**

## Test plan
- [ ] Shortcuts: **Track Bookmark** appears, lists trip bookmarks only, and opens the app.
- [ ] After arrivals load, a Live Activity starts (or an existing one is promoted). Cold launch still works.
- [ ] A whole-stop bookmark is not offered. Running the Shortcut twice does not mint a second activity.
- [ ] `LiveActivityShortcutRequestTests` (take clears; peek does not; garbage → nil). `BookmarkIntentMappingTests` (trip vs whole-stop).

## Summary by CodeRabbit

* **New Features**
  * Added Shortcuts support for starting Live Activities from eligible bookmarked trips.
  * Requests are processed when the app launches, returns to the foreground, or bookmark data becomes available.
  * Only trip bookmarks in the current region are offered.
  * Live Activity requests expire automatically and are cleaned up when invalid or unavailable.

* **Bug Fixes**
  * Pending requests no longer require navigating to the Bookmarks screen.

* **Documentation**
  * Documented the Shortcut-to-Live-Activity workflow, filtering, expiration, and cleanup behavior.

### DIFF
--- a/OBAKit/Bookmarks/BookmarksViewController.swift
+++ b/OBAKit/Bookmarks/BookmarksViewController.swift
@@ -71,6 +71,7 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         viewModel.start()
+        consumePendingLiveActivityShortcut()
 
         application.notificationCenter.addObserver(self, selector: #selector(applicationEnteredBackground), name: UIApplication.didEnterBackgroundNotification, object: nil)
         application.notificationCenter.addObserver(self, selector: #selector(applicationWillEnterForeground), name: UIApplication.willEnterForegroundNotification, object: nil)
@@ -196,6 +197,26 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
 
     // MARK: - Live Activity Management
 
+    /// Starts a Live Activity queued by the Track Bookmark Shortcut (#1222).
+    /// Waits until arrivals for that stop have loaded so we reuse the Track
+    /// path instead of hitting its empty-data error. Peek until arrivals are
+    /// in; clear only when starting or when the request cannot succeed.
+    func consumePendingLiveActivityShortcut() {
+        guard let id = LiveActivityShortcutRequest.peek(userDefaults: application.userDefaults) else { return }
+        guard let bookmark = application.userDataStore.findBookmark(id: id) else {
+            _ = LiveActivityShortcutRequest.take(userDefaults: application.userDefaults)
+            return
+        }
+
+        guard viewModel.hasFetchedArrivals(for: bookmark) else { return }
+
+        _ = LiveActivityShortcutRequest.take(userDefaults: application.userDefaults)
+        let arrivals = viewModel.arrivalDepartures(for: bookmark)
+        if bookmarkActions.startLiveActivity(for: bookmark, arrivalDepartures: arrivals) == .failed {
+            showLiveActivityErrorAlert()
+        }
+    }
+
     func updateRunningLiveActivities() {
         let activities = Activity<TripAttributes>.activities
         for activity in activities {
@@ -258,6 +279,7 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
 
     @objc private func applicationWillEnterForeground() {
         viewModel.start()
+        consumePendingLiveActivityShortcut()
     }
 
     // MARK: - Bookmark Groups
@@ -301,6 +323,7 @@ private extension BookmarksViewController {
         viewModel.didUpdate
             .sink { [weak self] _ in
                 self?.updateRunningLiveActivities()
+                self?.consumePendingLiveActivityShortcut()
             }
             .store(in: &cancellables)
 
--- a/OBAKit/Bookmarks/BookmarksViewController.swift
+++ b/OBAKit/Bookmarks/BookmarksViewController.swift
@@ -71,7 +71,6 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         viewModel.start()
-        consumePendingLiveActivityShortcut()
 
         application.notificationCenter.addObserver(self, selector: #selector(applicationEnteredBackground), name: UIApplication.didEnterBackgroundNotification, object: nil)
         application.notificationCenter.addObserver(self, selector: #selector(applicationWillEnterForeground), name: UIApplication.willEnterForegroundNotification, object: nil)
@@ -197,12 +196,6 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
 
     // MARK: - Live Activity Management
 
-    /// Starts a Live Activity queued by the Track Bookmark Shortcut (#1222).
-    /// `Application` owns the drain so a pushed stop page cannot hide it.
-    func consumePendingLiveActivityShortcut() {
-        application.consumePendingLiveActivityShortcut()
-    }
-
     func updateRunningLiveActivities() {
         let activities = Activity<TripAttributes>.activities
         for activity in activities {
@@ -265,7 +258,6 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
 
     @objc private func applicationWillEnterForeground() {
         viewModel.start()
-        consumePendingLiveActivityShortcut()
     }
 
     // MARK: - Bookmark Groups
@@ -309,7 +301,6 @@ private extension BookmarksViewController {
         viewModel.didUpdate
             .sink { [weak self] _ in
                 self?.updateRunningLiveActivities()
-                self?.consumePendingLiveActivityShortcut()
             }
             .store(in: &cancellables)
 
--- a/OBAKit/Bookmarks/BookmarksViewController.swift
+++ b/OBAKit/Bookmarks/BookmarksViewController.swift
@@ -210,11 +210,13 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
 
         let currentRegionID = application.currentRegion?.regionIdentifier
         if bookmark.regionIdentifier != currentRegionID {
+            Logger.warn("Track Bookmark Shortcut: bookmark \(id) is region \(String(describing: bookmark.regionIdentifier)); current region is \(String(describing: currentRegionID)). Dropping the queued request.")
             LiveActivityShortcutRequest.clear(application.userDefaults)
             return
         }
 
         guard ActivityAuthorizationInfo().areActivitiesEnabled else {
+            Logger.warn("Track Bookmark Shortcut: Live Activities are disabled. Dropping the queued request \(id). Enable Live Activities for this app in iOS Settings to use Track Bookmark.")
             LiveActivityShortcutRequest.clear(application.userDefaults)
             return
         }
--- a/OBAKit/Bookmarks/BookmarksViewController.swift
+++ b/OBAKit/Bookmarks/BookmarksViewController.swift
@@ -204,13 +204,24 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
     func consumePendingLiveActivityShortcut() {
         guard let id = LiveActivityShortcutRequest.peek(userDefaults: application.userDefaults) else { return }
         guard let bookmark = application.userDataStore.findBookmark(id: id) else {
-            _ = LiveActivityShortcutRequest.take(userDefaults: application.userDefaults)
+            LiveActivityShortcutRequest.clear(application.userDefaults)
+            return
+        }
+
+        let currentRegionID = application.currentRegion?.regionIdentifier
+        if bookmark.regionIdentifier != currentRegionID {
+            LiveActivityShortcutRequest.clear(application.userDefaults)
+            return
+        }
+
+        guard ActivityAuthorizationInfo().areActivitiesEnabled else {
+            LiveActivityShortcutRequest.clear(application.userDefaults)
             return
         }
 
         guard viewModel.hasFetchedArrivals(for: bookmark) else { return }
 
-        _ = LiveActivityShortcutRequest.take(userDefaults: application.userDefaults)
+        LiveActivityShortcutRequest.clear(application.userDefaults)
         let arrivals = viewModel.arrivalDepartures(for: bookmark)
         if bookmarkActions.startLiveActivity(for: bookmark, arrivalDepartures: arrivals) == .failed {
             showLiveActivityErrorAlert()
--- a/OBAKit/Bookmarks/BookmarksViewController.swift
+++ b/OBAKit/Bookmarks/BookmarksViewController.swift
@@ -198,36 +198,9 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
     // MARK: - Live Activity Management
 
     /// Starts a Live Activity queued by the Track Bookmark Shortcut (#1222).
-    /// Waits until arrivals for that stop have loaded so we reuse the Track
-    /// path instead of hitting its empty-data error. Peek until arrivals are
-    /// in; clear only when starting or when the request cannot succeed.
+    /// `Application` owns the drain so a pushed stop page cannot hide it.
     func consumePendingLiveActivityShortcut() {
-        guard let id = LiveActivityShortcutRequest.peek(userDefaults: application.userDefaults) else { return }
-        guard let bookmark = application.userDataStore.findBookmark(id: id) else {
-            LiveActivityShortcutRequest.clear(application.userDefaults)
-            return
-        }
-
-        let currentRegionID = application.currentRegion?.regionIdentifier
-        if bookmark.regionIdentifier != currentRegionID {
-            Logger.warn("Track Bookmark Shortcut: bookmark \(id) is region \(String(describing: bookmark.regionIdentifier)); current region is \(String(describing: currentRegionID)). Dropping the queued request.")
-            LiveActivityShortcutRequest.clear(application.userDefaults)
-            return
-        }
-
-        guard ActivityAuthorizationInfo().areActivitiesEnabled else {
-            Logger.warn("Track Bookmark Shortcut: Live Activities are disabled. Dropping the queued request \(id). Enable Live Activities for this app in iOS Settings to use Track Bookmark.")
-            LiveActivityShortcutRequest.clear(application.userDefaults)
-            return
-        }
-
-        guard viewModel.hasFetchedArrivals(for: bookmark) else { return }
-
-        LiveActivityShortcutRequest.clear(application.userDefaults)
-        let arrivals = viewModel.arrivalDepartures(for: bookmark)
-        if bookmarkActions.startLiveActivity(for: bookmark, arrivalDepartures: arrivals) == .failed {
-            showLiveActivityErrorAlert()
-        }
+        application.consumePendingLiveActivityShortcut()
     }
 
     func updateRunningLiveActivities() {
--- a/OBAKit/LiveActivities/TrackBookmarkIntent.swift
+++ b/OBAKit/LiveActivities/TrackBookmarkIntent.swift
@@ -16,6 +16,9 @@ import OBAKitCore
 /// the app (`openAppWhenRun`). `Application` starts the activity on the
 /// existing Track path once arrivals load — ActivityKit is not called from
 /// the intent, and no `Activity` is captured into a `Task`. See #1222.
+///
+/// Not `nonisolated`: `@Parameter` property wrappers are incompatible with a
+/// nonisolated AppIntent type on this SDK (entity/query/provider stay opted out).
 struct TrackBookmarkIntent: AppIntent {
     static let title: LocalizedStringResource = "Track Bookmark"
     static let description = IntentDescription("Start a Live Activity for a bookmarked trip.")
--- a/OBAKit/LiveActivities/TrackBookmarkIntent.swift
+++ b/OBAKit/LiveActivities/TrackBookmarkIntent.swift
@@ -17,9 +17,9 @@ import OBAKitCore
 /// on the existing Track path once arrivals load — ActivityKit is not called
 /// from the intent, and no `Activity` is captured into a `Task`. See #1222.
 struct TrackBookmarkIntent: AppIntent {
-    static var title: LocalizedStringResource = "Track Bookmark"
-    static var description = IntentDescription("Start a Live Activity for a bookmarked trip.")
-    static var openAppWhenRun = true
+    static let title: LocalizedStringResource = "Track Bookmark"
+    static let description = IntentDescription("Start a Live Activity for a bookmarked trip.")
+    static let openAppWhenRun = true
 
     @Parameter(title: "Bookmark")
     var bookmark: BookmarkEntity
@@ -34,9 +34,10 @@ struct TrackBookmarkIntent: AppIntent {
     }
 }
 
-struct BookmarkEntity: AppEntity {
-    static var typeDisplayRepresentation = TypeDisplayRepresentation(name: "Bookmark")
-    static var defaultQuery = BookmarkEntityQuery()
+/// App Entity types are extracted off the main actor; OBAKit defaults to `@MainActor`.
+nonisolated struct BookmarkEntity: AppEntity {
+    static let typeDisplayRepresentation = TypeDisplayRepresentation(name: "Bookmark")
+    static let defaultQuery = BookmarkEntityQuery()
 
     var id: UUID
     var name: String
@@ -46,7 +47,7 @@ struct BookmarkEntity: AppEntity {
     }
 }
 
-struct BookmarkEntityQuery: EntityQuery {
+nonisolated struct BookmarkEntityQuery: EntityQuery {
     func entities(for identifiers: [UUID]) async throws -> [BookmarkEntity] {
         tripBookmarkEntities().filter { identifiers.contains($0.id) }
     }
@@ -64,7 +65,7 @@ struct BookmarkEntityQuery: EntityQuery {
     }
 }
 
-enum BookmarkIntentMapping {
+nonisolated enum BookmarkIntentMapping {
     static func entities(from bookmarks: [Bookmark]) -> [BookmarkEntity] {
         bookmarks.filter(\.isTripBookmark).map {
             BookmarkEntity(id: $0.id, name: $0.name)
@@ -72,7 +73,12 @@ enum BookmarkIntentMapping {
     }
 }
 
-struct OBAAppShortcuts: AppShortcutsProvider {
+/// Exports this module's App Intents so the app target can include them.
+/// Without an `AppIntentsPackage` chain, `appintentsmetadataprocessor` never
+/// indexes intents compiled into OBAKit, and Shortcuts would not see Track.
+nonisolated public struct OBAKitAppIntentsPackage: AppIntentsPackage {}
+
+nonisolated struct OBAAppShortcuts: AppShortcutsProvider {
     static var appShortcuts: [AppShortcut] {
         AppShortcut(
             intent: TrackBookmarkIntent(),
--- a/OBAKit/LiveActivities/TrackBookmarkIntent.swift
+++ b/OBAKit/LiveActivities/TrackBookmarkIntent.swift
@@ -27,6 +27,7 @@ struct TrackBookmarkIntent: AppIntent {
     func perform() async throws -> some IntentResult {
         guard let suite = Bundle.main.appGroup,
               let defaults = UserDefaults(suiteName: suite) else {
+            Logger.error("Track Bookmark Shortcut: app group UserDefaults unavailable; cannot queue a Live Activity.")
             return .result()
         }
         LiveActivityShortcutRequest.store(bookmark.id, userDefaults: defaults)
@@ -56,18 +57,25 @@ nonisolated struct BookmarkEntityQuery: EntityQuery {
         tripBookmarkEntities()
     }
 
-    /// Trip bookmarks only. A whole-stop bookmark has no single route to Track.
+    /// Trip bookmarks in the current region only. A whole-stop bookmark has
+    /// no single route to Track; an out-of-region trip bookmark never gets
+    /// arrivals loaded, so offering it would queue a request that cannot start.
     private func tripBookmarkEntities() -> [BookmarkEntity] {
         guard let suite = Bundle.main.appGroup,
               let defaults = UserDefaults(suiteName: suite) else { return [] }
         let store = UserDefaultsStore(userDefaults: defaults)
-        return BookmarkIntentMapping.entities(from: store.bookmarks)
+        let regionID = defaults.object(forKey: BookmarkIntentMapping.currentRegionIdentifierKey) as? Int
+        return BookmarkIntentMapping.entities(from: store.bookmarks, regionIdentifier: regionID)
     }
 }
 
 nonisolated enum BookmarkIntentMapping {
-    static func entities(from bookmarks: [Bookmark]) -> [BookmarkEntity] {
-        bookmarks.filter(\.isTripBookmark).map {
+    /// Same string as `RegionsService.currentRegionIdentifierUserDefaultsKey`.
+    /// Duplicated because `RegionsService` is `@MainActor` and this query is not.
+    static let currentRegionIdentifierKey = "OBACurrentRegionIdentifierUserDefaultsKey"
+
+    static func entities(from bookmarks: [Bookmark], regionIdentifier: Int?) -> [BookmarkEntity] {
+        bookmarks.filter { $0.isTripBookmark && $0.regionIdentifier == regionIdentifier }.map {
             BookmarkEntity(id: $0.id, name: $0.name)
         }
     }
--- a/OBAKit/LiveActivities/TrackBookmarkIntent.swift
+++ b/OBAKit/LiveActivities/TrackBookmarkIntent.swift
@@ -64,15 +64,13 @@ nonisolated struct BookmarkEntityQuery: EntityQuery {
         guard let suite = Bundle.main.appGroup,
               let defaults = UserDefaults(suiteName: suite) else { return [] }
         let store = UserDefaultsStore(userDefaults: defaults)
-        let regionID = defaults.object(forKey: BookmarkIntentMapping.currentRegionIdentifierKey) as? Int
+        let regionID = defaults.object(forKey: BookmarkIntentMapping.regionIdentifierUserDefaultsKey) as? Int
         return BookmarkIntentMapping.entities(from: store.bookmarks, regionIdentifier: regionID)
     }
 }
 
 nonisolated enum BookmarkIntentMapping {
-    /// Same string as `RegionsService.currentRegionIdentifierUserDefaultsKey`.
-    /// Duplicated because `RegionsService` is `@MainActor` and this query is not.
-    static let currentRegionIdentifierKey = "OBACurrentRegionIdentifierUserDefaultsKey"
+    static let regionIdentifierUserDefaultsKey = RegionsService.currentRegionIdentifierUserDefaultsKey
 
     static func entities(from bookmarks: [Bookmark], regionIdentifier: Int?) -> [BookmarkEntity] {
         bookmarks.filter { $0.isTripBookmark && $0.regionIdentifier == regionIdentifier }.map {
@@ -91,6 +89,7 @@ nonisolated struct OBAAppShortcuts: AppShortcutsProvider {
         AppShortcut(
             intent: TrackBookmarkIntent(),
             phrases: [
+                "Track \(\.$bookmark) with \(.applicationName)",
                 "Track \(.applicationName) bookmark",
                 "Start \(.applicationName) Live Activity"
             ],
--- a/OBAKit/LiveActivities/TrackBookmarkIntent.swift
+++ b/OBAKit/LiveActivities/TrackBookmarkIntent.swift
@@ -13,9 +13,9 @@ import OBAKitCore
 /// Starts a Live Activity for a trip bookmark via Shortcuts.
 ///
 /// The intent only queues the bookmark id in the app-group defaults and opens
-/// the app (`openAppWhenRun`). `BookmarksViewController` starts the activity
-/// on the existing Track path once arrivals load — ActivityKit is not called
-/// from the intent, and no `Activity` is captured into a `Task`. See #1222.
+/// the app (`openAppWhenRun`). `Application` starts the activity on the
+/// existing Track path once arrivals load — ActivityKit is not called from
+/// the intent, and no `Activity` is captured into a `Task`. See #1222.
 struct TrackBookmarkIntent: AppIntent {
     static let title: LocalizedStringResource = "Track Bookmark"
     static let description = IntentDescription("Start a Live Activity for a bookmarked trip.")
--- a/OBAKit/LiveActivities/TrackBookmarkIntent.swift
+++ b/OBAKit/LiveActivities/TrackBookmarkIntent.swift
@@ -1,0 +1,87 @@
+//
+//  TrackBookmarkIntent.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import AppIntents
+import OBAKitCore
+
+/// Starts a Live Activity for a trip bookmark via Shortcuts.
+///
+/// The intent only queues the bookmark id in the app-group defaults and opens
+/// the app (`openAppWhenRun`). `BookmarksViewController` starts the activity
+/// on the existing Track path once arrivals load — ActivityKit is not called
+/// from the intent, and no `Activity` is captured into a `Task`. See #1222.
+struct TrackBookmarkIntent: AppIntent {
+    static var title: LocalizedStringResource = "Track Bookmark"
+    static var description = IntentDescription("Start a Live Activity for a bookmarked trip.")
+    static var openAppWhenRun = true
+
+    @Parameter(title: "Bookmark")
+    var bookmark: BookmarkEntity
+
+    func perform() async throws -> some IntentResult {
+        guard let suite = Bundle.main.appGroup,
+              let defaults = UserDefaults(suiteName: suite) else {
+            return .result()
+        }
+        LiveActivityShortcutRequest.store(bookmark.id, userDefaults: defaults)
+        return .result()
+    }
+}
+
+struct BookmarkEntity: AppEntity {
+    static var typeDisplayRepresentation = TypeDisplayRepresentation(name: "Bookmark")
+    static var defaultQuery = BookmarkEntityQuery()
+
+    var id: UUID
+    var name: String
+
+    var displayRepresentation: DisplayRepresentation {
+        DisplayRepresentation(title: "\(name)")
+    }
+}
+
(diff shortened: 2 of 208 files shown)


aaronbrethorst:
### Code review

Found 2 issues:

1. The Shortcut entity query offers trip bookmarks from *every* region, but the Bookmarks tab only ever fetches arrivals for the current region (`BookmarkDataLoader.eligibleBookmarks()` filters on `regionIdentifier == currentRegion?.id`, and `BookmarksViewModel.buildSection` does the same for display). So if the user picks a bookmark from a region they aren't currently in, `hasFetchedArrivals(for:)` can never become true, the queued id is never consumed, and no Live Activity ever starts. `OBAWidget`'s `WidgetDataProvider` already filters trip bookmarks by `regionIdentifier`; the same filter belongs here (bug due to `store.bookmarks` being unfiltered in `tripBookmarkEntities()`).

https://github.com/OneBusAway/onebusaway-ios/blob/cded9b006dbfea4020735e0eb8407058df07c2ce/OBAKit/LiveActivities/TrackBookmarkIntent.swift#L59-L65

2. Nothing ever gives up on a queued request. `revealBookmarksTabIfLiveActivityShortcutPending()` runs on every `applicationDidBecomeActive` (and on `rootUserInterfaceDidLoad`), and the id is only cleared when the bookmark is missing or when the activity actually starts. A request that can't be satisfied — the cross-region case in issue 1, or an arrivals fetch that keeps failing — therefore force-selects the Bookmarks tab on every single launch, indefinitely, overriding whatever tab the user was on. Even when it does eventually resolve, there's no staleness bound, so a request queued hours earlier can start a Live Activity out of the blue. A timestamp on the stored request plus a short expiry window would bound both.

https://github.com/OneBusAway/onebusaway-ios/blob/cded9b006dbfea4020735e0eb8407058df07c2ce/OBAKit/Orchestration/Application.swift#L582-L587

OBAKit/LiveActivities/TrackBookmarkIntent.swift
```diff
@@ -0,0 +1,93 @@
...
+    func suggestedEntities() async throws -> [BookmarkEntity] {
+        tripBookmarkEntities()
+    }
+
+    /// Trip bookmarks only. A whole-stop bookmark has no single route to Track.
+    private func tripBookmarkEntities() -> [BookmarkEntity] {
+        guard let suite = Bundle.main.appGroup,
+              let defaults = UserDefaults(suiteName: suite) else { return [] }
+        let store = UserDefaultsStore(userDefaults: defaults)
+        return BookmarkIntentMapping.entities(from: store.bookmarks)
```
aaronbrethorst: Needs a region filter. `BookmarkDataLoader.eligibleBookmarks()` only fetches bookmarks matching `regionsService.currentRegion?.id`, and it is the only thing that populates `fetchedStopIDs` — so `hasFetchedArrivals(for:)` can never return true for a bookmark from another region. As written, Siri offers those bookmarks and the resulting request can never be satisfied.

Match the widget: `.filter { $0.isTripBookmark && $0.regionIdentifier == app.regionsService.currentRegion?.id }` (see `OBAWidget/Provider/WidgetDataProvider.swift:102`).

OBAKit/Orchestration/Application.swift
```diff
@@ -572,6 +575,17 @@ public class Application: CoreApplication, PushServiceDelegate {
...
 
+    /// A Track Bookmark Shortcut queued a Live Activity. Select the Bookmarks
+    /// tab so `BookmarksViewController` loads (it is lazy until first shown)
+    /// and can start the activity on the existing Track path once arrivals
+    /// arrive. Does not consume the pending id — that is the VC's job.
+    @MainActor
+    private func revealBookmarksTabIfLiveActivityShortcutPending() {
+        guard !isOnboardingRoot else { return }
+        guard LiveActivityShortcutRequest.peek(userDefaults: userDefaults) != nil else { return }
+        (viewRouter.rootController as? UITabBarController)?.selectedIndex = SelectedTab.bookmarks.rawValue
```
aaronbrethorst: This runs on every foreground and on every root install, and nothing here bounds how long the pending id can survive. Since the id is only cleared on "bookmark missing" or "activity started", any request that cannot be satisfied — a cross-region bookmark, or arrivals that keep failing to load — force-selects the Bookmarks tab on every launch indefinitely, with no user-visible way to clear it.

The request needs an expiry (stamp it and drop it after a short window), and ideally the reveal should stop retrying after the first real attempt.

OBAKit/Bookmarks/BookmarksViewController.swift
```diff
@@ -205,6 +206,23 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
...
+    /// path instead of hitting its empty-data error. `take()` runs only when
+    /// we are about to start, so a still-loading foreground does not drop the id.
+    func consumePendingLiveActivityShortcut() {
+        guard let id = LiveActivityShortcutRequest.peek(userDefaults: application.userDefaults) else { return }
+        guard let bookmark = application.userDataStore.findBookmark(id: id) else {
+            _ = LiveActivityShortcutRequest.take(userDefaults: application.userDefaults)
+            return
+        }
+
+        guard viewModel.hasFetchedArrivals(for: bookmark) else { return }
```
aaronbrethorst: This is the line that makes the pending id immortal: if `hasFetchedArrivals(for:)` never becomes true, we return without consuming. Combined with the unfiltered entity query, that is reachable for any out-of-region bookmark.

Worth also gating on `ActivityAuthorizationInfo().areActivitiesEnabled` here the way `makeNavigationHandler` does — otherwise the shortcut path surfaces a raw error alert when Live Activities are disabled system-wide.

aaronbrethorst (changes requested):
The design here is good and I want to say so before the list of problems: routing the shortcut through the existing Track path means you inherit the `Activity<TripAttributes>.running(matching:)` promote-instead-of-duplicate guard from #1215 for free, rather than reimplementing it. Deferring to `rootUserInterfaceDidLoad` as well as `applicationDidBecomeActive` shows you thought about the cold-launch race. Writing the request through the app-group suite is correct — I confirmed `application.userDefaults` really is that suite.

Three things to fix.

**1. CI is red — the new test file doesn't compile.**

```
OBAKitTests/Models/BookmarkIntentMappingTests.swift:40:19: error: value of optional type 'Stop?' must be unwrapped to a value of type 'Stop'
OBAKitTests/Models/BookmarkIntentMappingTests.swift:44:30: error: cannot infer key path type from context; consider explicitly specifying a root type
** TEST BUILD FAILED **
```

**2. The entity query has no region filter, so the shortcut is inert for out-of-region bookmarks.** `tripBookmarkEntities()` maps `store.bookmarks` unfiltered, but `BookmarkDataLoader.eligibleBookmarks()` only fetches bookmarks whose `regionIdentifier` matches the current region — and `fetchedStopIDs` is only ever populated there. So `hasFetchedArrivals(for:)` can never return true for a bookmark from another region. Siri will happily offer that bookmark, the app will open, and nothing will happen. `OBAWidget/Provider/WidgetDataProvider.swift:102` has the filter you want.

**3. The pending request never expires, and it drives an unbounded tab hijack.** `revealBookmarksTabIfLiveActivityShortcutPending()` runs on *every* `applicationDidBecomeActive` and on `rootUserInterfaceDidLoad`, and the id is only cleared on "bookmark missing" or "activity started." Combine that with #2 — or just a persistently failing arrivals fetch — and the app force-selects the Bookmarks tab on every single launch, forever, with no way for the user to clear it. This is the one that would generate support mail.

Fixing #2 shrinks the blast radius but doesn't close it. The request needs a give-up: stamp it with a timestamp and drop it after some short window, and clear it once you've had a real chance to satisfy it rather than only on success.

Smaller things, none blocking on their own:

- **New strings aren't localized.** "Track Bookmark", the `IntentDescription`, "Bookmark", the `shortTitle`, and the Siri phrases are bare literals. Note the trap here: a `LocalizedStringResource` declared in the OBAKit *framework* resolves against `Bundle.main` by default, so adding entries to `OBAKit/Strings` won't be enough without naming the bundle explicitly.
- **The shortcut path skips the `areActivitiesEnabled` gate** that the in-app menu applies in `makeNavigationHandler`. With Live Activities off system-wide the user gets an error alert instead of an actionable "turn this on."
- **Tab selection bypasses `ViewRouter.rootNavigateTo(page:)`.** You cast an already-typed `ClassicApplicationRootController?` to `UITabBarController` and index it with `SelectedTab.bookmarks.rawValue`. `SelectedTab` and `Page` happen to agree at index 2 today; that's a coincidence, not a contract. It's also a silent no-op in map-panel mode where `rootController` is nil, whereas `rootNavigateTo` deliberately logs that case.
- **`perform()` silently returns `.result()`** when the app group is unavailable — app opens, nothing happens, no log.
- Test suites create random `UserDefaults` suites without `removePersistentDomain(forName:)` in teardown.

Happy to re-review once the build is green and the region filter and expiry are in.

Borisserz:
CI unwrap is `#require(arrivalDep.stop)`.

The Shortcut entity query only offers trip bookmarks in the current region (`nil` region → none). Out-of-region bookmarks never get arrivals, so they cannot start a Live Activity.

A queued request expires after 90s (and a leftover id without a timestamp is dropped). `consumePending` also clears on a missing bookmark, a region mismatch, or Live Activities disabled, instead of leaving the Bookmarks tab hijacked. Tab selection is `viewRouter.rootNavigateTo(page: .bookmarks)`. Missing app-group defaults in `perform()` is logged.

`BookmarkIntentMappingTests` + `LiveActivityShortcutRequestTests`: 7/7 passed locally.

aaronbrethorst:
### Code review

Found 1 issue:

1. The queue-then-drain handoff is inverted: nothing re-checks for the pending id after `perform()` writes it, so on a warm launch the Shortcut is a no-op. `openAppWhenRun = true` is the legacy spelling of `supportedModes = .foreground(.immediate)`, and Apple defines `.immediate` as "bring the app to the foreground immediately after the system resolves the intent's parameters" — i.e. the app is already frontmost *before* `perform()` runs. Both drain points (`applicationDidBecomeActive`, Application.swift:520, and `rootUserInterfaceDidLoad`, Application.swift:535) have therefore already fired and peeked an empty request by the time `LiveActivityShortcutRequest.store` executes. `revealBookmarksTabIfLiveActivityShortcutPending()` never runs again for that request, so unless the user happened to already be sitting on the Bookmarks tab (where the 30s `didUpdate` sink would pick it up), the request just sits in defaults and silently expires 90 seconds later. This is the difference from `pendingStopID`, which the same drain-point pattern serves correctly: URL-scheme and push handlers write that stash *before* activation, so the drain sees it. A cold launch may work by luck, since the async root install can push `rootUserInterfaceDidLoad` past `perform()` — but that is a race, not a contract. The fix is a real handoff rather than a poll: post a notification from `perform()` after storing (it runs in the app process, so an in-process `NotificationCenter` hop reaches `Application`) and drive the reveal from that, or reach the app directly via `@Dependency`. `.foreground(.deferred)` would also give the ordering you want, but it is iOS 26+ and this app deploys to 18.

https://github.com/OneBusAway/onebusaway-ios/blob/fb920ff586bb0ee9af2046450847fe4b3cab497a/OBAKit/LiveActivities/TrackBookmarkIntent.swift#L25-L36

aaronbrethorst (changes requested):
All three blockers from the last round are properly fixed — the region filter on the entity query, the timestamp and expiry on the stashed request, and `rootNavigateTo(page:)` instead of the raw tab-index cast. The `areActivitiesEnabled` gate and the log in `perform()` are in too. The overall design is right: putting the request in app-group defaults and letting the Bookmarks path start the activity once arrivals load reuses the existing machinery instead of duplicating it, and `LiveActivityShortcutRequest` staying Foundation-only keeps OBAKitCore extension-safe.

First, it needs a rebase — it's conflicting against main. `startLiveActivity(for:)` and `liveActivityKeys` moved from `BookmarksViewController` to `BookmarkActions` in `0df90879`, and `consumePendingLiveActivityShortcut` still calls the old view-controller-level method.

Second, and worth fixing in the same pass: I think the queue-and-drain handoff is inverted, which would make the Shortcut do nothing on its most common path. `openAppWhenRun = true` is the older spelling of `supportedModes = .foreground(.immediate)`, and that brings the app frontmost *before* `perform()` runs. So `applicationDidBecomeActive` and `rootUserInterfaceDidLoad` both peek an empty request, then `perform()` stores the id afterward with nothing left to drain it. Unless the rider already happens to be sitting on the Bookmarks tab — where the 30s `didUpdate` sink picks it up — the id just expires after 90 seconds.

The contrast with `pendingStopID` is what convinced me: the URL-scheme and push handlers write that stash *before* activation, which is why the same drain-point pattern works there. Cold launch may appear to work, but only because the async root install can land after `perform()` — that's a race, not a contract.

The fix is probably to have `perform()` itself kick the reveal — post a notification or call into `Application` directly — rather than relying on a lifecycle hook that has already fired by then.

Smaller things, none blocking:

`LiveActivityShortcutRequest.take()` has no production caller; `consumePendingLiveActivityShortcut` uses `peek` + `clear`. Either wire it up or drop it.

`BookmarkIntentMapping.currentRegionIdentifierKey` duplicates `RegionsService.currentRegionIdentifierUserDefaultsKey` as a string literal with nothing pinning them equal. If that key ever moves, the entity query silently offers zero bookmarks. Making the `RegionsService` constant public would be safer than the copy.

The `areActivitiesEnabled == false` path clears the request with no log and nothing user-facing, so the rider runs the Shortcut and gets silence. I asked for an actionable message last round; the docs file says the silence is intentional, but "nothing happened and I don't know why" is a bad outcome even when it's deliberate. Same shape on the region-mismatch clear.

`AppShortcut` uses the non-parameterized init even though `TrackBookmarkIntent` has a required `bookmark` parameter — no `$bookmark` token in either phrase and no `parameterPresentation` — so Siri has to fall back to a disambiguation prompt. `updateAppShortcutParameters()` also never gets called when the bookmark set changes.

The test suites still create random `UserDefaults` suites without `removePersistentDomain(forName:)` in teardown; that was in the last round too, and it leaves plists behind in the test host container.

Rebase plus the intent-ordering fix and I'll re-review.

Borisserz:
Rebased onto current `main` so consume uses `BookmarkActions.startLiveActivity` (the VC method is gone after `0df90879`).

`openAppWhenRun` is equivalent to `supportedModes = .foreground(.immediate)`: the app is already frontmost before `perform()`. `applicationDidBecomeActive` / `rootUserInterfaceDidLoad` therefore peeked an empty queue. `LiveActivityShortcutRequest.store` now posts `.liveActivityShortcutRequestDidStore`; `Application` observes that and selects Bookmarks. Cold launch still peeks from `rootUserInterfaceDidLoad` if the observer was not registered yet.

Also in this push:
- `take()` removed (consume is peek-until-arrivals, then `clear`).
- Entity query reads `RegionsService.currentRegionIdentifierUserDefaultsKey` (`public nonisolated`), not a copied string.
- Region mismatch and Live Activities disabled log a warning, then drop the request.
- App Shortcut phrase includes `$bookmark`; `updateAppShortcutParameters()` runs at launch and on `.bookmarksDidChange`.
- Isolated UserDefaults suites call `removePersistentDomain` in `defer`.

`LiveActivityShortcutRequestTests` (6) and `BookmarkIntentMappingTests` (3) passed on iPhone 17 Pro. The store-posts test fails if `store` writes without posting.

aaronbrethorst:
### Code review

Found 2 issues:

1. CI is red at this head. `scripts/concurrency_ratchet` fails the build with `error: concurrency warning count grew from 0 to 5`, and all five are the same diagnostic in this file: `no calls to throwing functions occur within 'try' expression [#UnnecessaryEffectMarker]` at lines 53, 63, 72, 81, and 90. `withDefaults` is `rethrows`, the closures passed to it never call a throwing function, and `LiveActivityShortcutRequest` exposes no throwing API at all — so the `try` (and the `throws` on those five `@Test` functions) is vestigial and should just come off. Nothing here wants a throwing call it isn't making (bug due to `try withDefaults { ... }` over non-throwing closure bodies).

https://github.com/OneBusAway/onebusaway-ios/blob/cc80ea7129d368a60e1d439748dbdfc9d4bd551b/OBAKitTests/Models/LiveActivityShortcutRequestTests.swift#L52-L54

2. `Garbage stored value is treated as empty` never reaches the code path it names, so the UUID-parsing guard in `peek` has no coverage. The test writes only `userDefaultsKey` and leaves `storedAtKey` unset, but `peek`'s first guard is the timestamp one — `guard let storedAt = userDefaults.object(forKey: storedAtKey) as? TimeInterval else { clear(...); return nil }` — which returns `nil` before `string(forKey:).flatMap(UUID.init(uuidString:))` is ever evaluated. That makes this test an exact duplicate of `Legacy id without a timestamp is treated as empty` on the line below it, and it would still pass if the `flatMap(UUID.init(uuidString:))` guard were deleted outright. Storing a timestamp alongside the garbage string would exercise the branch the name promises (bug due to guard ordering in `LiveActivityShortcutRequest.peek`).

https://github.com/OneBusAway/onebusaway-ios/blob/cc80ea7129d368a60e1d439748dbdfc9d4bd551b/OBAKitTests/Models/LiveActivityShortcutRequestTests.swift#L70-L76

aaronbrethorst (changes requested):
The round-two blocker is genuinely closed, and it was the hard one. The handoff is no longer inverted: `store` posts `.liveActivityShortcutRequestDidStore`, `Application.observeShortcutLifecycle()` observes it from `init`, and I checked that the hop actually connects — `CoreApplication.notificationCenter` is `NotificationCenter.default`, and there's no AppIntents extension target, so `perform()` runs in the app process and the notification reaches a live observer. The cold-launch path through `rootUserInterfaceDidLoad` is still there as the fallback. `take()` is gone, the region key is aliased off `RegionsService.currentRegionIdentifierUserDefaultsKey` instead of being retyped as a literal, the `areActivitiesEnabled` gate is in, tab selection goes through `rootNavigateTo(page:)`, and the test suites clean up their `UserDefaults` domains. The 90-second expiry is a good answer to the unbounded-hijack problem — short enough that a failed arrivals fetch can't force the tab forever.

CI is red, though, and it's red on this PR's own new test file rather than on anything flaky.

`scripts/concurrency_ratchet` failed the build for `cc80ea71` with `concurrency warning count grew from 0 to 5`. All five are `no calls to throwing functions occur within 'try' expression`, in `OBAKitTests/Models/LiveActivityShortcutRequestTests.swift` at lines 53, 63, 72, 81 and 90. I checked whether the `try` was pointing at something real that had gone missing, because that's the failure mode worth catching — it isn't. `store`, `peek` and `clear` are all non-throwing, none of those five closures calls any other throwing API, and every assertion in them is `#expect` rather than `#require`. So the `try` is vestigial and the `throws` on those five `@Test` functions should come off with it. (`BookmarkIntentMappingTests` is the contrast — its `throws` is load-bearing because of `try #require(arrivalDep.stop)`. Leave that one alone.)

While you're in that file, one of the tests isn't testing what its name says. `Garbage stored value is treated as empty` at line 71 sets only `userDefaultsKey` and leaves `storedAtKey` unset. But `peek`'s *first* guard is the timestamp guard at line 47, which clears and returns nil before the `UUID.init(uuidString:)` guard at line 55 is ever evaluated. So it's an exact behavioral duplicate of `Legacy id without a timestamp is treated as empty` below it, and deleting the UUID-parsing guard entirely would leave both tests green. Store a valid timestamp alongside the bad UUID string and it'll exercise the branch it's named for.

Three things I'm not blocking on, in descending order of how much I'd like them.

The localization point from round one is still open: `title`, `IntentDescription`, `@Parameter(title:)`, `shortTitle` and all three Siri phrases in `TrackBookmarkIntent.swift` are bare literals. In a framework that matters more than usual, because `LocalizedStringResource` resolves against `Bundle.main` unless you name the bundle. This is a white-label codebase with thirteen locales; the Shortcut is going to be the one English string in a Polish app.

`peek` isn't the pure read its name promises — the empty-queue path calls `clear()`, so every `applicationDidBecomeActive`, every Bookmarks `viewWillAppear`, and every thirty-second `didUpdate` tick issues two `removeObject` writes against the app-group suite. Cheap individually, but it's a lot of writes for a function called peek. An early return when both keys are already absent would settle it.

And `BookmarkEntityQuery` conforms to `EntityQuery` rather than `EntityStringQuery` while the shortcut phrase carries a `$bookmark` token, so Siri has no way to resolve a spoken bookmark name to an entity. I couldn't find a doc that makes this a hard requirement, so I'm flagging it as something to try on device rather than asserting it's broken.

Drop the five `try`s to get CI green, fix the garbage-value test, and I'll re-review quickly — the architecture here is in good shape now.

Borisserz:
CI fail was vestigial `try` / `throws` on five `LiveActivityShortcutRequestTests` functions. `store` / `peek` / `clear` do not throw; those annotations are gone. `BookmarkIntentMappingTests` still throws — `#require` is load-bearing there.

`Garbage stored value is treated as empty` now writes a fresh timestamp next to `"not-a-uuid"`, so it reaches the `UUID.init` guard. Drop that guard and the leftover string plus a still-fresh timestamp would be treated as a live request.

`peek` returns immediately when both keys are absent, so become-active / `viewWillAppear` / the 30s tick do not `removeObject` on an empty suite.

`LiveActivityShortcutRequestTests` (7) and `BookmarkIntentMappingTests` (3) passed on iPhone 17 Pro.

aaronbrethorst:
### Code review

Found 1 issue:

1. The reveal-then-drain handoff still has a hole: `rootNavigateTo(page: .bookmarks)` only switches the tab, it does not pop that tab's own navigation stack. `ClassicApplicationRootController.navigate(to:)` is `navigationController?.popToViewController(self, …)` plus `selectedIndex = …`, and the tab bar controller *is* the window root (`AppDelegate.m:104`), so its `navigationController` is nil and the pop is a no-op. Each tab is its own `UINavigationController`, and stop pages are pushed onto it (`Router.navigate(to:from:)`). So if the rider has a stop page pushed inside the Bookmarks tab — the normal result of tapping a bookmark — running Track Bookmark selects a tab that is already selected, the pushed stop page stays on top, `BookmarksViewController.viewWillAppear` never fires, and the view model is still deactivated from `viewWillDisappear` (`viewModel.deactivate()` → `dataLoader.cancelUpdates()`), so no `didUpdate` ever arrives either. All three `consumePendingLiveActivityShortcut()` call sites are therefore dead on that path and the queued request just expires 90 seconds later with nothing shown to the rider. `drainPendingUIPresentations()` right above this avoids the problem by pushing from `topViewController` instead of relying on a specific view controller becoming visible; popping the Bookmarks nav stack to root (or draining from `Application` rather than from the VC's lifecycle) would close it (bug due to `viewRouter.rootNavigateTo(page: .bookmarks)` never reaching `BookmarksViewController` when its tab has a pushed child).

https://github.com/OneBusAway/onebusaway-ios/blob/15bee6c57f118a341bdc2c7eb1b60947bc44ea6b/OBAKit/Orchestration/Application.swift#L604-L609

OBAKit/Orchestration/Application.swift
```diff
@@ -592,6 +597,42 @@ public class Application: CoreApplication, PushServiceDelegate {
...
 
+    /// A Track Bookmark Shortcut queued a Live Activity. Select the Bookmarks
+    /// tab so `BookmarksViewController` loads (it is lazy until first shown)
+    /// and can start the activity on the existing Track path once arrivals
+    /// arrive. Does not consume the pending id — that is the VC's job.
+    @MainActor
+    private func revealBookmarksTabIfLiveActivityShortcutPending() {
+        guard !isOnboardingRoot else { return }
+        guard LiveActivityShortcutRequest.peek(userDefaults: userDefaults) != nil else { return }
+        viewRouter.rootNavigateTo(page: .bookmarks)
```
aaronbrethorst: `rootNavigateTo(page: .bookmarks)` switches the tab index but doesn't clear the Bookmarks tab's own navigation stack, so when a stop page is pushed there the reveal never actually surfaces `BookmarksViewController`.

The chain:

```swift
// ClassicApplicationRootController.swift:77
func navigate(to destination: Page) {
    navigationController?.popToViewController(self, animated: true)  // no-op
    selectedIndex = destination.rawValue
}
```

`ClassicApplicationRootController` is a `UITabBarController` set as `window.rootViewController` (`Apps/OneBusAway/AppDelegate.m:104`), so it has no `navigationController` of its own and that pop does nothing. Each tab holds its own `UINavigationController`, and stop pages are pushed into it.

By then `BookmarksViewController.viewWillDisappear` has already run — it removes the `willEnterForeground` observer and calls `viewModel.deactivate()` → `dataLoader.cancelUpdates()` — so the `viewWillAppear`, foreground-notification, and `didUpdate` drain sites are all dead. The stored request sits until the 90s expiry and is dropped with nothing shown.

Two ways out. Pop the *selected tab's* stack:

```swift
(selectedViewController as? UINavigationController)?.popToRootViewController(animated: true)
```

or drain from `Application` the way `drainPendingUIPresentations()` does immediately above, resolving the target from `topViewController` instead of depending on the view controller's lifecycle callbacks. The second matches the existing precedent and doesn't yank the rider off their stop page as a side effect.

aaronbrethorst (changes requested):
This has come a long way. I went back through all three earlier rounds item by item, and every one of the seven things I raised is genuinely fixed at `15bee6c5` — not merely replied to:

- the region filter is real (`entities(from:regionIdentifier:)`, nil region → empty, and it lines up with `BookmarkDataLoader.eligibleBookmarks()`)
- the unbounded tab hijack now has a 90s expiry, with `peek` clearing expired, garbage, and timestamp-less values, and consume clearing on missing bookmark, region mismatch, and Live Activities being off
- the rebase onto `BookmarkActions.startLiveActivity` is done
- the inverted queue/drain handoff now goes through `.liveActivityShortcutRequestDidStore` — I checked that `CoreApplication.notificationCenter` really is `NotificationCenter.default`, so that in-process hop does connect
- the `concurrency_ratchet` `try`/`throws` cleanup is done and CI is green
- and the garbage-value test now writes a fresh `storedAtKey` so it actually reaches the UUID guard instead of short-circuiting on the timestamp

That last one is the kind of fix I like seeing: you didn't just make the test pass, you made it exercise the thing it was named for.

One thing from round 2 isn't fully closed, and it's why I'm not approving yet. Flagged inline: the reveal can't reach `BookmarksViewController` when the Bookmarks tab already has a stop page pushed onto it — which is an ordinary state, not an edge case. I traced it and confirmed each link:

1. `ClassicApplicationRootController.navigate(to:)` is `navigationController?.popToViewController(self,…)` then `selectedIndex = …`
2. that controller is a `UITabBarController` installed as `window.rootViewController` (`AppDelegate.m:104`), so its *own* `navigationController` is nil and the pop is a no-op
3. each tab is its own `UINavigationController` and stop pages are pushed into it
4. `BookmarksViewController.viewWillDisappear` already removed the `willEnterForeground` observer and called `viewModel.deactivate()` → `dataLoader.cancelUpdates()`

So for a rider who taps a bookmark, backgrounds the app from the stop page, and runs Track Bookmark: the tab index flips, the stop page stays on screen, and all three `consumePendingLiveActivityShortcut()` sites are unreachable. The request just expires silently after 90 seconds with nothing shown.

Worth noting `drainPendingUIPresentations()` sitting right above it already solves this shape by working from `topViewController` — so the pattern is in the file, just not on this path.

While you're in there, I'd fold in the stale-donation gap: `OBAAppShortcuts.updateAppShortcutParameters()` runs at launch and on `.bookmarksDidChange`, but a region switch posts neither, so Shortcuts keeps offering the previous region's bookmarks until relaunch — and running one then hits the region-mismatch drop, which is log-only. A hook on the regions-service region-changed delegate would close it.

Two more I'm noting rather than blocking on. The user-facing strings ("Track Bookmark", "Start a Live Activity for a bookmarked trip.", and the Siri phrases) are bare literals with no catalog entries, so they're English in all 13 locales. And `BookmarkEntityQuery` is a plain `EntityQuery` rather than an `EntityStringQuery`, which likely means spoken bookmark names won't resolve in the parameterized phrase — the in-app Shortcuts picker works either way.

Please take a look at the inline comment. This is close.

OBAKit/Orchestration/Application.swift
Borisserz: Consume now lives on `Application` via `LiveActivityShortcutDrain`, same shape as `drainPendingUIPresentations`: missing/mismatch/disabled drop immediately; a valid bookmark fetches arrivals and starts on the existing Track path. Alerts go to `topViewController`. No tab switch. `updatedRegion` also calls `OBAAppShortcuts.updateAppShortcutParameters()`. Tests: store notification and cold-load `rootUserInterfaceDidLoad` both clear a missing-bookmark request without a Bookmarks tab. RED then GREEN, 2/2.

aaronbrethorst:
### Code review

Found 2 issues:

1. `BookmarksViewModel.hasFetchedArrivals(for:)` has no callers anywhere at this head, and its doc comment describes a design this push replaced. The comment claims "Shortcut Live Activity start waits on this so it does not fire the empty-data error path before the loader returns," but `LiveActivityShortcutDrain.start` no longer consults the view model at all — it builds its own `BookmarkDataLoader` and awaits `loadDataAndWait()`. This is the same shape as the `take()` finding from the last round (dead API left behind by a redesign); grep for `hasFetchedArrivals` across the branch returns only this definition and a doc comment in `BookmarkIntentMappingTests` (bug due to `LiveActivityShortcutDrain` owning its own loader).

https://github.com/OneBusAway/onebusaway-ios/blob/e9d50e93c473b67b9d70b716ffa584da281294c9/OBAKit/ViewModels/BookmarksViewModel.swift#L168-L175

2. `docs/shortcut-live-activity.md` documents tab selection that the final implementation deliberately removed, so the one file explaining this feature is wrong about its central design decision. Line 37 states "`rootNavigateTo(page: .bookmarks)` selects the tab" — that call appears nowhere in this feature; the only `rootNavigateTo` at this head is the pre-existing `.map` navigation in `Application.swift:724`. Line 15 has the same problem: "The store notification is what selects the Bookmarks tab on a warm launch," while `Application.consumePendingLiveActivityShortcut()`'s own comment says the opposite — "Consume it here from `topViewController` rather than switching to the Bookmarks tab" — and `LiveActivityShortcutDrain.consume` is documented as "Does not switch tabs." Both sentences describe the round-3 design, not this one (bug due to the drain moving to `Application` with no tab switch).

https://github.com/OneBusAway/onebusaway-ios/blob/e9d50e93c473b67b9d70b716ffa584da281294c9/docs/shortcut-live-activity.md#L36-L38

Borisserz:
Removed dead `hasFetchedArrivals(for:)` and corrected `docs/shortcut-live-activity.md` — drain does not switch tabs and owns its own loader.

aaronbrethorst:
### Code review — correction

Disregard my previous comment: it reviewed `e9d50e9`, and you'd already pushed `ddcc12e4` ("Drop dead hasFetchedArrivals and fix Shortcut Live Activity docs") while it was in flight. Both findings are fixed there — `hasFetchedArrivals` has no references left anywhere in the tree, and the docs no longer claim the drain selects the Bookmarks tab.

Nothing outstanding from me on the current head. The one thing still worth doing before this lands is the rebase — the branch is a long way behind `main`.

aaronbrethorst (changes requested):
Every blocker from rounds 1–4 is genuinely closed, including the one that mattered most: the queue-then-drain handoff is now the right way round — `store` posts `.liveActivityShortcutRequestDidStore`, `Application.observeShortcutLifecycle()` observes from `init`, and `rootUserInterfaceDidLoad` stays as the cold-launch fallback. The region filter, the 90s expiry, the `areActivitiesEnabled` gate, the region-key alias, `updateAppShortcutParameters()` actually being called, and the defaults-domain cleanup are all in. The two items I flagged earlier today were already fixed by your newer commit; retracted above.

Two cheap things before it lands:

**1. The branch is ~85 commits behind `main`.** For a feature this size touching app-group defaults and the Application lifecycle, I want CI on a current base.

**2. Drop the drive-by `// swiftlint:disable:next type_body_length` on `MapViewController`.** Your own #1325 extracted the trip planner and put that type back under the warning line, so main doesn't need the disable — merging this re-adds a superfluous one plus a comment claiming the type is "over the 1000-line body budget", which is no longer true.

Still open from earlier rounds, and I'd like at least the first before this ships:

- **Localization of the Shortcut-facing strings.** `title`, `IntentDescription`, the `@Parameter(title:)`, `TypeDisplayRepresentation`, `shortTitle` and all three Siri phrases are bare literals. I take the point that `OBAWidget` already does this, so if you'd rather fix both together in a follow-up, say so and I'll take that — but it shouldn't just keep rolling.
- The region-mismatch and Live-Activities-disabled paths now log a warning but still say nothing to the rider. Third time I've raised it; if the silence is deliberate, put that in the doc rather than leaving it looking like an oversight.

One I'm not asking you to fix: overlapping callers can each spawn a `BookmarkDataLoader` for the same stop while `start` awaits, since the request stays stored across the await. The post-await `peek == id` guard correctly prevents a second Live Activity, so it costs redundant fetches in a short window, nothing worse.

Borisserz:
Rebased onto main (~85 commits). Dropped \`swiftlint:disable:next type_body_length\` on \`MapViewController\` (trip planner is in \`MapViewController+TripPlanner\` now; main passes without the disable).

Localization follow-up for any new strings can be handled in a separate pass.

aaronbrethorst:
### Code review

Found 2 issues:

1. Every user-facing string on this new Shortcuts surface ships English-only. `title: "Track Bookmark"`, `IntentDescription("Start a Live Activity for a bookmarked trip.")`, `@Parameter(title: "Bookmark")`, `TypeDisplayRepresentation(name: "Bookmark")`, `shortTitle: "Track bookmark"`, and all three Siri phrases are raw literals with no entry in any of the 13 `.lproj/Localizable.strings` files, and there is no `.xcstrings` catalog or `AppShortcuts.strings` anywhere in the tree to pick them up. `LocalizedStringResource` does not fix this on its own here: the repo has no `SWIFT_EMIT_LOC_STRINGS` extraction, its strings are hand-maintained via `OBALoc`, and a `LocalizedStringResource` declared inside a framework resolves against `Bundle.main` rather than `OBAKit`'s bundle — so adding keys to `OBAKit/Strings/*.lproj` would not reach these without an explicit `bundle:`. A Polish or Spanish rider gets an English tile in Shortcuts and English-only Siri phrases. The two pre-existing `LocalizedStringResource` titles in `OBAWidget` are the same gap at a much smaller scale; this adds a top-level Siri/Shortcuts entry point (bug due to unlocalized `LocalizedStringResource`/`AppShortcut` literals).

https://github.com/OneBusAway/onebusaway-ios/blob/ba094a68735cb24915efd52f6a671a1a023d7b33/OBAKit/LiveActivities/TrackBookmarkIntent.swift#L18-L26

2. `LiveActivityShortcutRequest`'s comments still describe the tab-switching design this PR replaced — the same defect you flagged in `docs/shortcut-live-activity.md` last round, fixed in the `.md` but not in the source. The only rationale given for the `expiration = 90` constant is "How long a queued request may hijack the Bookmarks tab. Short enough that a failed arrivals fetch cannot force the tab on every launch" — but the drain never touches the tab, and this PR says so twice in the opposite direction: `LiveActivityShortcutDrain.consume` is documented "Does not switch tabs" and `Application.consumePendingLiveActivityShortcut()` says "Consume it here from `topViewController` rather than switching to the Bookmarks tab". The type-level comment eight lines up has the same problem: "ActivityKit `request` lives on the in-app start path (`BookmarksViewController`)" — `Activity.request` moved to `BookmarkActions` (`BookmarksViewController` only calls `bookmarkActions.startLiveActivity`), and `LiveActivityShortcutDrain` is now a second start path, so the sentence is wrong on both counts. A future reader gets no correct reason for the 90-second window (bug due to the drain moving to `Application` with no tab switch).

https://github.com/OneBusAway/onebusaway-ios/blob/ba094a68735cb24915efd52f6a671a1a023d7b33/OBAKitCore/LiveActivities/LiveActivityShortcutRequest.swift#L32-L37

aaronbrethorst (changes requested):
Six rounds in, and the architecture has landed in the right place. Queueing the UUID with a 90s expiry and draining from `Application` rather than the tab controller is the correct answer to the "a pushed stop page hides the Bookmarks tab" problem, and every functional concern from the earlier rounds is genuinely fixed — region-filtered entities, the queue/drain ordering, the `try`/`throws` ratchet, and the garbage-UUID test that now actually reaches the guard. `openAppWhenRun` is also the right call here rather than `supportedModes`, since that's iOS 26+ and this app deploys to 18. Duplicate activities, ActivityKit authorization, and entity ID stability all check out.

Two things before it lands.

**1. The drain clears the request before it has succeeded, so a transient failure silently eats the Shortcut.**

In `LiveActivityShortcutDrain.start(...)`, `LiveActivityShortcutRequest.clear(...)` runs immediately after the `peek == id` guard, several lines before `startLiveActivity` is attempted. If `loadDataAndWait()` came back with nothing — a network blip, or `application.apiService` still nil, in which case `BookmarkDataLoader.loadData` silently no-ops — then `arrivals` is empty, `buildContentState` returns nil, and the result is `.failed`. The request is already gone, so none of the six drain entry points can retry it and the 90s window buys nothing.

It's worse on the cold-launch path this feature exists for: `presenter` is `topViewController`, which can still be nil that early, so `presenter?.showLiveActivityErrorAlert()` is dropped too. A rider taps the Shortcut, the app opens, and nothing happens with no explanation.

Clearing after a successful start (or clearing only on outcomes that aren't worth retrying) fixes both halves.

**2. The comments still describe the design from round 3.**

`LiveActivityShortcutRequest.expiration` is documented as "How long a queued request may hijack the Bookmarks tab" — there's no tab hijacking any more, that's exactly what moving the drain to `Application` removed. And the enum's doc comment says the ActivityKit `request` lives in `BookmarksViewController`; it's `BookmarkActions.startLiveActivity`. Two other comments in this same PR describe the current design correctly, so the file now contradicts itself. This is the same thing that came up last round on the `.md` — fixed there, missed in the source.

Smaller items, none blocking:

- The Shortcuts and Siri strings are the app's first Siri surface and they ship untranslated. I don't think that's fair to hold you to in this PR — a framework-declared `LocalizedStringResource` resolves against `Bundle.main`, and there's no extraction path in the repo for it, so it needs infrastructure that doesn't exist yet. Worth filing as its own issue rather than bolting onto this.
- `#expect(app.viewRouter.rootController == nil)` in `LiveActivityShortcutDrainTests` can't fail — `buildApplication` never installs a root controller and the drain never touches `viewRouter` — so it doesn't back up the "without the Bookmarks tab" claim the tests are making.
- `consumePendingLiveActivityShortcut()` on `BookmarksViewController` is now a pure passthrough whose own doc comment argues the VC shouldn't be a drain site. Three call sites left over from the old design, one of which fires every 30s from the `didUpdate` sink.
- `observeShortcutLifecycle()` discards both observer tokens. Harmless for the production singleton, but every test-built `Application` leaves two blocks registered on `NotificationCenter.default`.
- `TrackBookmarkIntent` is the only App Intents type in that file not marked `nonisolated`, so it picks up OBAKit's `@MainActor` default while the entity, query, and shortcuts provider all opt out.

Sequencing note: merge #1381 before this one — it changes `BookmarkActions`, which your drain calls straight through, so you'd inherit the trip-pinned behavior rather than having to redo it. (#1381 has its own blocker to clear first.)

Borisserz:
Addressed the two blockers from the latest review:

1. **Clear after success** — \`LiveActivityShortcutDrain\` only clears the queued UUID on \`.started\` / \`.promotedExisting\`. Empty arrivals or ActivityKit failure leave the request so another drain entry can retry within 90s (and still shows the Track error alert when a presenter exists). Permanent blockers (missing bookmark / region mismatch / Activities disabled) still clear immediately.
2. **Stale comments** — \`LiveActivityShortcutRequest\` docs no longer mention tab hijacking or \`BookmarksViewController\` owning \`request\`. Docs updated to match.

Also removed the Bookmarks VC \`consumePendingLiveActivityShortcut\` passthrough and its three call sites (\`Application\` owns the drain). Left \`TrackBookmarkIntent\` on the main actor — \`nonisolated\` + \`@Parameter\` does not compile on this SDK.

Test: \`failed Track keeps the queued shortcut for retry\`.

Sequencing: still prefer landing #1381 first so this inherits trip-pinned Track.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Both round-6 blockers are fixed. The drain now switches on the result and clears only on `.started` / `.promotedExisting`, so a transient failure keeps the request for retry instead of eating it — that was the important one. And removing the `BookmarksViewController` passthrough entirely is better than what I asked for.

Six rounds is a lot to ask of anyone and you worked through all of it. Merging.

`TrackBookmarkIntent` is still main-actor isolated by OBAKit's default, and it's the only App Intents type in the file not marked `nonisolated`. Your new comment says `@Parameter` is incompatible with a nonisolated AppIntent — that's the canonical Apple pattern, so I'd like to know what the actual compiler error was. Everything `perform()` does here is main-actor-safe anyway, so nothing is broken; I just want the comment to be right if it's going to explain a deviation.

The comment in `LiveActivityShortcutRequestTests` still lists "become-active, Bookmarks `viewWillAppear`, and the 30s arrivals tick" — two of those three call sites were deleted by this PR's own commit. Comment drift is the thing I've flagged most in this batch, so worth a sweep.

`#expect(app.viewRouter.rootController == nil)` in the first test still can't fail; you removed it from the second.

Merging.

moggerrescure:
> ## Summary
> * New **Track Bookmark** App Intent (`openAppWhenRun`) queues a trip-bookmark UUID in the app-group defaults. The Bookmarks tab then starts (or promotes) the Live Activity on the existing Track path once arrivals load. No `Activity` is captured into a `Task`.
> * Whole-stop bookmarks are omitted from the entity query: they have no single route to Track. Arbitrary stop IDs that are not bookmarked are out of scope.
> * Intents live in OBAKit. The app target includes them through `OBAAppIntentsPackage` → `OBAKitAppIntentsPackage`; without that chain, Shortcuts never sees Track.
> 
> **Refs #1222.**
> 
> ## Test plan
> * [ ]  Shortcuts: **Track Bookmark** appears, lists trip bookmarks only, and opens the app.
> * [ ]  After arrivals load, a Live Activity starts (or an existing one is promoted). Cold launch still works.
> * [ ]  A whole-stop bookmark is not offered. Running the Shortcut twice does not mint a second activity.
> * [ ]  `LiveActivityShortcutRequestTests` (take clears; peek does not; garbage → nil). `BookmarkIntentMappingTests` (trip vs whole-stop).
> 
> ## Summary by CodeRabbit
> * **New Features**
>   
>   * Added Shortcuts support for starting Live Activities from eligible bookmarked trips.
>   * Requests are processed when the app launches, returns to the foreground, or bookmark data becomes available.
>   * Only trip bookmarks in the current region are offered.
>   * Live Activity requests expire automatically and are cleaned up when invalid or unavailable.
> * **Bug Fixes**
>   
>   * Pending requests no longer require navigating to the Bookmarks screen.
> * **Documentation**
>   
>   * Documented the Shortcut-to-Live-Activity workflow, filtering, expiration, and cleanup behavior.

Issue #1221, opened 2026-07-25
Bookmarks from iOS Shortcuts do not work.


nbarnard:
When a bookmark from iOS shortcuts is activated, it only opens OneBusAway, it does not open the appropriate bookmark page.

aaronbrethorst:
thanks for the head's up. I'll investigate.

nbarnard:
I have a video of this problem. It doesn’t appear that I can upload it here. If it would be useful, I can put it on YouTube. 
(That is why I skipped the replication steps…)

aryn-rjpt:
> I have a video of this problem. It doesn’t appear that I can upload it here. If it would be useful, I can put it on YouTube. (That is why I skipped the replication steps…)

Hi. Can you please share the mentioned video ( either by uploading on youtube or drive ) or share the replication steps?

nbarnard:
> > I have a video of this problem. It doesn’t appear that I can upload it here. If it would be useful, I can put it on YouTube. (That is why I skipped the replication steps…)
> 
> Hi. Can you please share the mentioned video ( either by uploading on youtube or drive ) or share the replication steps?

Sure. Here is the video on Youtube. https://youtube.com/shorts/wakRyrbD7X8?is=ZHbDNLR_XTpi_Cro

Borisserz:
Fixed on `main` by #1301 (merged 2026-08-22): donated stop Shortcuts/`NSUserActivity` now stash `pendingStopID` / parse `regionID` as `NSNumber`/`String` and fall back to `webpageURL`. The PR used `Refs`, so GitHub left this open.

@aaronbrethorst okay to close? (Reporter’s “bookmark” shortcut here is the donated stop activity — not Live Activity Track.)

PR #1301, opened 2026-08-21
Open donated stop shortcuts instead of a no-op launch

Borisserz:
## Summary
- Donated stop `NSUserActivity`s (Siri/Shortcuts “Show me my bus”) used to return `false` from `routeStop` unless `apiService` and `currentRegion` were already live, and `showStopHandler` then dropped the `Stop` when `topViewController` was still nil. The app opened; the stop did not.
- Routing now reads the stop from `userInfo` (`Int` / `NSNumber` / `String` region IDs) or from the donated `webpageURL` (`/regions/{id}/stops/{stopID}`). Trip URLs (`.../trips`) stay on the existing trip decoder.
- The stop ID is queued on the same `pendingStopID` stash the alarm-push and `viewStop` URL-scheme paths already use, and drained once the root UI exists **and** `currentRegion` matches. A mismatched region is dropped rather than fetched against the wrong API.

Refs #1221. This opens the **stop** those donated activities already describe (same as Handoff). It is not the Track Bookmark Live Activity path (#1298 / #1222).

## Test plan
- [x] `Stop destination reads a Swift Int region ID`
- [x] `Stop destination reads an NSNumber region ID restored by Shortcuts`
- [x] `Stop destination reads a region ID restored as a string` — fails if production only accepts `as? Int`
- [x] `Stop destination uses the stop webpage URL when userInfo is missing the keys`
- [x] `Stop destination ignores a trip deep link URL`
- [x] `Application continues a stop user activity on cold launch` — asserts `pendingStopID` is stashed, not just `true`
- [x] `Application continues a browsing-web stop URL`
- [ ] Manual: add a Siri Shortcut from a stop (“Show me my bus”), kill the app, run the shortcut — the stop page should open, not just the map.

## Summary by CodeRabbit

* **New Features**
  * Added deep-link support for transit stop URLs and stop-related web links.
  * Stop links now work with multiple region ID formats and user activity metadata.
  * Stop navigation is deferred until the app and requested region are ready, including during cold launch.

* **Bug Fixes**
  * Prevented trip links from being incorrectly treated as stop links.
  * Discarded pending stop navigation requests for mismatched regions.

* **Tests**
  * Added coverage for cold-launch activities, URL formats, region identifiers, and routing fallbacks.

### DIFF
--- a/OBAKit/DeepLinks/AppLinksRouter.swift
+++ b/OBAKit/DeepLinks/AppLinksRouter.swift
@@ -42,6 +42,9 @@ public class AppLinksRouter: NSObject {
 
     private let deepLinkPathFormat = "/regions/%d/stops/%@/trips"
     private let deepLinkPattern = "/regions/(?<region>.*)/stops/(?<stop>.*)/trips"
+    /// Stop pages are the same path without the `/trips` suffix. Named groups are
+    /// `[^/]+` so a trip URL cannot masquerade as a stop whose ID contains `/trips`.
+    private static let stopPathPattern = "/regions/(?<region>[^/]+)/stops/(?<stop>[^/]+)$"
 
     /// Encodes an `ArrivalDeparture` into an `URL` so that it can be shared as a deep link with others.
     /// - Parameters:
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Nice fix — this is the right shape. Splitting the parse out into a pure `stopDestination` and routing everything through `queueOrOpenStop` means a donated shortcut no longer depends on `apiService` being live and the region already matching, which was the whole reason these launched to nothing. Anchoring the stop pattern with `$` so trip URLs can't masquerade as stops is a good catch, and covering all three region-ID representations Shortcuts can hand back is exactly the kind of thing that bites six months later.

One follow-up, not blocking this merge: `pendingStopRegionID` is only written by `queueOrOpenStop`. The alarm-push path and the `viewStop` URL-scheme path both assign `pendingStopID` without clearing it, so a region-bearing stash followed by a regionless one — before any drain — leaves the second stop gated on the first's region. It needs two cold-launch stop navigations racing with no drain between them, so it's narrow, and it degrades to "stop doesn't open" rather than anything worse. But the doc comment says "Nil for the URL-scheme path" and nothing actually nils it, which is the kind of drift that turns into a real bug later. A `pendingStopRegionID = nil` alongside those two assignments closes it.

Also minor: the new `guard let current = currentRegion?.regionIdentifier else { return }` returns from all of `drainPendingUIPresentations`, so a stop waiting on an unresolved region also defers the donation prompt and add-region alert below it. Self-corrects on the next drain, so it's a deferral rather than a loss.

Merging as-is.

PR #1305, opened 2026-08-22
De-emphasize the travelled half of the trip polyline

Borisserz:
## Summary
- `TripViewController`'s map drew one route-colored `MKPolyline` for the entire shape. The SwiftUI trip page already splits that shape (`TripShapeSplit` + `TripFocusMapLayer`); the classic UIKit trip did not.
- The UIKit map now draws two `TripShapeOverlay`s from the same fraction: gray/thinner behind the vehicle, route color ahead. No progress (schedule-only / missing total distance) keeps the whole shape ahead — the same choice `TripPageViewController` already documents. There is no invented split point.
- Stroke is `TripRouteOverlayAppearance`, so tests do not load `TripViewController.view` (that path reads an unresolved `ArrivalDeparture.route`).

Refs #444.

## Test plan
- [x] `No reported progress draws the whole shape as ahead`
- [x] `Mid-trip progress yields a spent overlay and an ahead overlay`
- [x] `A trip that has not started has no spent overlay`
- [x] `A finished trip has no ahead overlay`
- [x] `Spent stroke is gray and thinner than the live half`
- [x] Existing `TripShapeSplitTests` still pass
- [ ] Manual (feature flag off / transfer path that still pushes `TripViewController`): open a live trip — behind the vehicle is gray, ahead is the route color. Schedule-only trip stays one color.

## Summary by CodeRabbit

* **New Features**
  * Trip routes now visually distinguish completed portions from upcoming portions.
  * Route styling adapts to accessibility visibility settings and trip progress.
  * Map framing updates to include the full combined trip route and details.

* **Bug Fixes**
  * Improved route overlay updates when trip progress or route details change.

### DIFF
--- a/OBAKit/Mapping/Layers/TripFocus/TripRouteOverlays.swift
+++ b/OBAKit/Mapping/Layers/TripFocus/TripRouteOverlays.swift
@@ -1,0 +1,59 @@
+//
+//  TripRouteOverlays.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import CoreLocation
+import MapKit
+import UIKit
+
+/// Builds the two polylines `TripViewController` draws on its map: gray behind
+/// the vehicle, route-colored ahead of it. The SwiftUI trip page already does
+/// this via `TripFocusMapLayer`; this is the same split for the UIKit map
+/// `TripViewController` still owns.
+///
+/// See: https://github.com/OneBusAway/onebusaway-ios/issues/444
+enum TripRouteOverlays {
+
+    /// - Parameter fraction: `nil` when the trip reports no progress (schedule-
+    ///   only, or a feed that omits total distance). The whole shape is then
+    ///   ahead — the same choice `TripPageViewController` makes — rather than
+    ///   inventing a split point.
+    static func make(coordinates: [CLLocationCoordinate2D], fraction: Double?) -> [TripShapeOverlay] {
+        let split = fraction.map {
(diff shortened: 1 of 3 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is the right way to do it. The UIKit trip map needed the same travelled/ahead split the SwiftUI trip page already had, and rather than writing a second split you reused `TripShapeSplit` and `TripShapeOverlay` outright — so the two surfaces can't drift, and the vertex-tolerance and degenerate-half handling that file already worked out come along for free.

I checked the parts that usually go wrong here. `TripShapeSplit.fraction` returns nil on a zero total distance and the `flatMap` over `details?.status` carries that through, so a schedule-only trip draws the whole shape as ahead instead of inventing a split at zero. The `count >= 2` guards line up on both sides — `normalized()` already drops a short half, and `make` re-checks because the `fraction == nil` branch builds its `Result` directly. And the `MKMapRect.null` reduce with the `isNull` guard is a real fix, not ceremony: the old code framed on a single polyline's `boundingMapRect` and this keeps framing identical while surviving an empty overlay list.

Pulling the stroke into `TripRouteOverlayAppearance` so a test can assert spent-is-gray-and-thinner without standing up `TripViewController.view` is exactly the testability move I'd want.

One thing to keep an eye on rather than fix now: the spent stroke returns before `needsIncreasedVisibility` is consulted, so it's `.systemGray3` in every appearance, while `TripFocusMapLayer` pairs that same gray with a white casing underneath. In dark mode the travelled half may read closer to invisible than de-emphasized. De-emphasis is the goal so I'm not going to hold the PR on it, but if it looks washed out on device, the casing is where to start.

Merging.

Issue #563, opened 2021-10-25
Annotations being added extraneously, causing weird "reset" animation


ualch9:
See attached video. The stop annotations appear to "reset" to a default image before actually presenting the correct icon whenever the map is scrolled.

~I originally observed this on 120hz iphone screen and thought it was related, but running OBA on macOS monterey (60hz) also presents this issue, suggesting that this is a bug introduced with this release cycle.~
see below comment.

# Video
(screen recording is from Monterey, but same visual bug exists on iOS)

https://user-images.githubusercontent.com/22162410/138784698-18ec3dea-596b-4dff-83b4-6a764a420f5f.mov

# Logs
(maybe related? i didn't notice these until Monterey)
```
2021-10-25 16:27:09.095086-0700 App[43106:1890208] [AXVectorKit] Filtered because not within map bounds.
2021-10-25 16:27:09.096118-0700 App[43106:1890208] [AXVectorKit] Raw accessibility elements: (
    "VKPointFeatureAccessibilityElement:0x600000e49d60: Kobe Terrace - Paths: 1 [en]",
    "VKPointFeatureAccessibilityElement:0x600000ea57c0: Harborview Medical Center - View Park Garage - Paths: 1 [en]",
    "VKPointFeatureAccessibilityElement:0x600000e8a9e0: US Post Office - Paths: 1 [en]",
    "VKPointFeatureAccessibilityElement:0x600000e8bc00: Tsukushinbo - Paths: 1 [en]",
    "VKPointFeatureAccessibilityElement:0x600000ea5860: Kim Hair Salon - Paths: 1 [en]",
    "VKPointFeatureAccessibilityElement:0x600000ea59a0: Angel Spa - Paths: 1 [en]",
    "VKPointFeatureAccessibilityElement:0x600000e8b520: Bank of America - Paths: 1 [en]",
    "VKPointFeatureAccessibilityElement:0x600000ea6940: Fuji Sushi - Paths: 1 [en]",
    "VKPointFeatureAccessibilityElement:0x600000e8b5c0: Hing Hay Park - Paths: 1 [en]",
    "VKPointFeatureAccessibilityElement:0x600000e8ae40: Jade Garden Restaurant - Paths: 1 [en]",
    "VKPointFeatureAccessibilityElement:0x600000e888c0: International Model Toys - Paths: 1 [en]",
    "VKPointFeatureAccessibilityElement:0x600000e89e00: Harborview Medical Center - Paths: 1 [en]",
[...]

PR #1306, opened 2026-08-22
Stop the map pin reset flash on pan

Borisserz:
## Summary
- Introduced in #519 (always-on bookmarks). Every camera settle added newly fetched stops and **never removed** ones that left the viewport. MapKit then recycled annotation views; pins flashed to the default image before the stop icon came back. Alan's note on #563: the AXVectorKit logs are unrelated.
- `StopAnnotationSync` is the add/remove set. A pan that returns the same stop IDs adds and removes nothing. Stops that leave the viewport are removed unless they are selected (same exemption #1266 already gives bookmarks). Bookmarks still replace the sibling `Stop` for that stop ID.
- `viewFor` sets the icon via `applyPresentation()` before the view is returned, so MapKit never displays a nil-image placeholder. `refreshAnnotationViews` no longer calls `prepareForReuse` on a live pin — that method is for views leaving the map.

Refs #563.

## Test plan
- [x] `A pan that returns the same stop IDs adds and removes nothing` — fails if production re-adds the current set
- [x] `A pan removes stops that left the viewport and adds only new IDs`
- [x] `A selected stop that left the viewport is not removed`
- [x] `A bookmark replaces the Stop annotation for that stop ID`
- [x] `Stops layer off adds no Stop annotations and keeps bookmarks`
- [x] `Display unique stop annotations preserves selected bookmark` (#1266) still passes — `prepareForReuseCount == 0`
- [x] `BackgroundAnnotationDeemphasisTests` still pass
- [ ] Manual: pan the map at stop-visible zoom — existing pins should not flash to a default marker. Bookmark a stop, tap it, pan — callout stays open.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved map annotation updates when panning or changing the visible area.
  * Preserved selected bookmark callouts while map views refresh.
  * Prevented bookmarked stops from appearing as duplicate stop annotations.
  * Ensured annotation views retain the correct presentation during updates and reuse.
* **Enhancements**
  * Stop annotations now update more efficiently by adding and removing only items affected by map changes.
  * Improved behavior when the stops layer is enabled or disabled.

### DIFF
--- a/OBAKit/Mapping/MapRegionManager.swift
+++ b/OBAKit/Mapping/MapRegionManager.swift
@@ -649,70 +649,46 @@ public class MapRegionManager: NSObject,
 
         let existingAnnotations = mapView.annotations
         let existingStopIDs = Set(existingAnnotations.compactMap { ($0 as? Stop)?.id })
-        var affectedStopIDs: Set<StopID> = []
-
-        // Remove Stop annotations that now have a corresponding Bookmark
-        let stopAnnotationsToRemove = existingAnnotations.compactMap { annotation -> MKAnnotation? in
-            guard
-                let stop = annotation as? Stop,
-                bookmarksHash[stop.id] != nil
-            else {
-                return nil
-            }
-
-            affectedStopIDs.insert(stop.id)
-            return stop
+        let existingBookmarks = existingAnnotations.compactMap { annotation -> StopAnnotationSync.BookmarkRef? in
+            guard let bookmark = annotation as? Bookmark else { return nil }
+            return StopAnnotationSync.BookmarkRef(id: bookmark.id, stopID: bookmark.stopID)
         }
-
-        // Remove Bookmark annotations that are stale:
-        //   - The bookmark's stop no longer has ANY bookmark (deleted), OR
-        //   - The bookmark on the map is a different object than the current one (replaced)
-        let bookmarkAnnotationsToRemove = existingAnnotations.compactMap { annotation -> MKAnnotation? in
-            guard let bookmark = annotation as? Bookmark else {
-                return nil
-            }
-
-            guard let currentBookmark = bookmarksHash[bookmark.stopID] else {
-                // No bookmark exists for this stop anymore, remove the stale annotation
-                affectedStopIDs.insert(bookmark.stopID)
-                return bookmark
-            }
-
-            // If a different bookmark now represents this stop, remove the old one
-            if currentBookmark.id != bookmark.id {
-                affectedStopIDs.insert(bookmark.stopID)
-                return bookmark
-            }
-
+        let selectedStopIDs = Set(existingAnnotations.compactMap { annotation -> StopID? in
+            guard mapView.selectedAnnotations.contains(where: { $0 === annotation }) else { return nil }
+            if let stop = annotation as? Stop { return stop.id }
+            if let bookmark = annotation as? Bookmark { return bookmark.stopID }
             return nil
-        }
+        })
 
-        let allAnnotationsToRemove = stopAnnotationsToRemove + bookmarkAnnotationsToRemove
-        for annotation in allAnnotationsToRemove {
+        let changes = StopAnnotationSync.changes(
+            existingStopIDs: existingStopIDs,
+            existingBookmarks: existingBookmarks,
+            incomingStopIDs: Set(stops.map(\.id)),
+            bookmarksByStopID: Dictionary(uniqueKeysWithValues: bookmarksHash.map { ($0.key, $0.value.id) }),
+            selectedStopIDs: selectedStopIDs,
+            isStopsLayerEnabled: isStopsLayerEnabled
+        )
+
+        let stopsToRemove = existingAnnotations.compactMap { $0 as? Stop }
+            .filter { changes.stopIDsToRemove.contains($0.id) }
+        let bookmarksToRemove = existingAnnotations.compactMap { $0 as? Bookmark }
+            .filter { changes.bookmarkIDsToRemove.contains($0.id) }
+        let annotationsToRemove: [MKAnnotation] = stopsToRemove + bookmarksToRemove
+
+        for annotation in annotationsToRemove {
             guard mapView.selectedAnnotations.contains(where: { $0 === annotation }) else { continue }
             mapView.deselectAnnotation(annotation, animated: false)
         }
-        mapView.removeAnnotations(allAnnotationsToRemove)
+        mapView.removeAnnotations(annotationsToRemove)
 
-        // Add new Bookmark annotations that aren't already on the map
-        // Check by bookmark ID (not just stopID) to correctly add replacements
-        let existingBookmarkIDs = Set(mapView.annotations.compactMap { ($0 as? Bookmark)?.id })
-        let bookmarksToAdd = bookmarksHash.values.filter {
-            !existingBookmarkIDs.contains($0.id)
-        }
-        mapView.addAnnotations(Array(bookmarksToAdd))
+        mapView.addAnnotations(bookmarksHash.values.filter { changes.bookmarkIDsToAdd.contains($0.id) })
+        mapView.addAnnotations(stops.filter { changes.stopIDsToAdd.contains($0.id) })
 
-        // Re-add Stop annotations for stops that no longer have bookmarks.
-        // With the stops layer toggled off, nothing is added — bookmarks are
-        // user content and stay visible regardless.
-        let stopsToAdd = isStopsLayerEnabled ? stops.filter {
-            !bookmarksHash.keys.contains($0.id) &&
-            !existingStopIDs.contains($0.id)
-        } : []
-        mapView.addAnnotations(stopsToAdd)
-
-        // Removing a sibling Stop for a bookmarked stopID shouldn't refresh a
-        // selected Bookmark — rebinding its view dismisses an open callout.
+        // A bookmark↔stop identity change still needs the surviving view rebound.
+        // Skip selected bookmarks: rebinding dismisses an open callout (#1266).
+        var affectedStopIDs = changes.stopIDsToRemove.union(
+            Set(bookmarksToRemove.map(\.stopID))
+        )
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. Exempting selected bookmarks from `bookmarkIDsToRemove` leaves stale/duplicate bookmark pins. The removal set here isn't a viewport diff — it's the identity diff that `8eea1a99` ("Fix: Map fails to dynamically refresh bookmark annotations") added for two cases: the bookmark was **deleted**, or a **different** bookmark now represents that stop. Subtracting `selectedBookmarkIDs` from it turns off both. Because `.bookmarksDidChange` fires while the pin is still selected (that's exactly the #1266/#782 scenario), the reachable outcomes are: (a) delete a bookmark whose pin is selected → the stale `Bookmark` annotation stays on the map, and since `bookmarkedStopIDs` no longer contains that stop, `wantedStopIDs` re-adds a `Stop` for it → two pins at one coordinate; (b) bookmark replaced (a second bookmark for the same stop, `dedupedByStopID` picks the new one) → old ID isn't removed while `bookmarkIDsToAdd` adds the new one → two `Bookmark` pins showing the old and new names. The `#1266` exemption belongs on the *stop*-viewport removal (and on `refreshAnnotationViews`, which already has it), not on bookmark identity churn. None of the new tests cover a selected bookmark that was deleted or replaced.

https://github.com/OneBusAway/onebusaway-ios/blob/3dd73d53b81a1d0d1aaab2982573a34c377757be/OBAKit/Mapping/StopAnnotationSync.swift#L47-L50

aaronbrethorst (changes requested):
The diagnosis here is right, and the shape of the fix is the one I'd want. `displayUniqueStopAnnotations` was accumulating — it re-added every stop the fetch returned and only ever removed bookmark-shadowed ones, so a pan that returned the same IDs still handed MapKit a pile of adds and MapKit recycled the views. Pulling the decision out into a pure `StopAnnotationSync.changes(...)` and letting the caller just apply the result is the correct move, and it's what makes the five tests possible. Those tests are real ones: the first case fails loudly if the extra-add ever comes back.

I also checked the `prepareForReuse` → `applyPresentation` swap, because that's the kind of change that quietly strands state. It doesn't. `prepareForReuse` cleared `labelStack.isHidden` and `titleLabel.text`, and both `prepareForDisplay(bookmark:delegate:)` and `prepareForDisplay(stop:delegate:)` set the label stack, the title, and the image. Calling `applyPresentation()` after rebinding covers everything the old call cleared, and setting the delegate before it is the right order.

What I can't merge is the selected-bookmark exemption, because it's applied to the wrong removal set.

`bookmarkIDsToRemove` at `StopAnnotationSync.swift:48` is not the viewport diff. It's the *identity* diff — the bookmark was deleted, or a different bookmark now represents this stop — which is what `8eea1a99` added to fix bookmark pins not refreshing. Line 49 then suppresses it for anything selected, and that reintroduces the bug in two reachable ways. Both fire through `bookmarks.didSet` → `displayUniqueStopAnnotations()`, which runs with the selection live; that's the same path as the #1266 callout case.

Delete a bookmark whose pin is selected. `bookmarkIDsToRemove` starts as `{B}`, line 49 empties it, and the stale pin stays. Then, because `bookmarkedStopIDs` no longer contains that stop, line 57's `wantedStopIDs.formUnion(selectedStopIDs.subtracting(bookmarkedStopIDs))` puts the stop ID into `wantedStopIDs`, and since the map holds a `Bookmark` for it rather than a `Stop`, it isn't in `existingStopIDs` — so line 62 adds a `Stop` too. Two annotations on one coordinate, one of them a bookmark the rider just deleted, with its callout still open.

Replace a bookmark — a second bookmark for the same stop, where `dedupedByStopID` picks the new one — and the old ID is suppressed from removal while the new one is added. Two bookmark pins for one stop, showing the old name and the new one.

On main both cases deselect and remove unconditionally, so this is new.

The exemption belongs on the stop-viewport removal, not on the identity diff. A selected bookmark that still exists and still represents its stop never lands in `bookmarkIDsToRemove` in the first place — the identity diff already leaves it alone — and the open-callout protection you actually want is the `affectedStopIDs.subtract(selectedBookmarkStopIDs)` in `MapRegionManager`, which is still there and still doing its job. Dropping line 49 should leave the #1266 behavior intact; the existing test for it will tell you. Worth adding a case for a selected bookmark that gets deleted, and one for a selected bookmark that gets replaced.

Two smaller things while you're in here.

The new comment at `MapRegionManager.swift:720-721` says calling `prepareForReuse` on a live pin "clears the icon and is what the rider sees as a reset." It doesn't — `StopAnnotationView.prepareForReuse` only hides `labelStack` and nils `titleLabel.text`, and never touches `image`; `MKAnnotationView`'s default implementation does nothing. The icon flash came from the re-add churn this PR removes, which is the more interesting story anyway. The change is right, the explanation isn't, and that comment is going to be load-bearing for whoever reads this next.

`affectedStopIDs` is now seeded with `changes.stopIDsToRemove`, which on a pan can be hundreds of IDs, and `refreshAnnotationViews` does a linear `mapView.annotations.first { ... }` scan with dynamic casts for each one. Every one of those is a guaranteed miss, since the annotation was removed moments earlier. It's wasted main-thread work on every camera settle, in a PR whose whole point is that panning feels smooth. The set that actually needs rebinding is the bookmark-shadow/identity one.

Fix the exemption and I'll take another pass quickly — the rest of this is in good shape.

Borisserz:
Addressed the review:

1. Dropped the selected-bookmark exemption from `bookmarkIDsToRemove` (identity diff). Open-callout protection stays in `MapRegionManager` via `selectedBookmarkStopIDs`.
2. Added tests for deleted and replaced selected bookmarks.
3. Fixed the `prepareForReuse` comment; `affectedStopIDs` for rebinding is now only the identity-diff bookmark stops (not viewport removals).

aaronbrethorst:
### Code review

_Re-review at 4e6bf69, covering commits since the previous review._

Found 1 issue:

1. `stopIDsToRemove` is now "every `Stop` on the map that isn't in `stops` and isn't selected", but `stops` is not the only source of `Stop` annotations. A route search runs `displaySearchResult(stopsForRoute:)`, which puts the route's stops on the map directly, and `searchResponseOverridesStopLoading()` then blocks region fetches, so `stops` still holds the stops from before the search. `displayUniqueStopAnnotations()` has no search guard, and `bookmarks.didSet` still calls it: on `.bookmarksDidChange`, which `MapViewController` observes, and on every `viewWillAppear`, which a tab switch triggers. So if a rider searches a route, taps one of its stops and bookmarks it, every route stop that isn't selected and isn't in the stale `stops` gets removed. Tapping the callout chevron calls `show(stop:deselecting:)`, so the tapped stop isn't selected either. The pre-search stops are added back in their place, and the route polyline is left with no pins on it. On main, stops were only removed when a bookmark shadowed them, so route-search pins survived this. None of the new tests cover `Stop` annotations that didn't come from the region fetch. The fix could be to skip the viewport removal while `searchResponseOverridesStopLoading()` is true.

https://github.com/OneBusAway/onebusaway-ios/blob/4e6bf69ab8ad8b8f22a86c52a89477e157deee79/OBAKit/Mapping/StopAnnotationSync.swift#L50-L62

https://github.com/OneBusAway/onebusaway-ios/blob/4e6bf69ab8ad8b8f22a86c52a89477e157deee79/OBAKit/Mapping/MapRegionManager.swift#L847-L857

Borisserz:
`displayUniqueStopAnnotations()` now passes `preserveStopsOutsideIncoming` while `searchResponseOverridesStopLoading()` is true. Route-search pins that did not come from the region fetch stay on the map. A bookmark still replaces the one stop it shadows, and the stale pre-search set is not added back.

`StopAnnotationSyncTests` covers that case. The suite passed (8 tests).

aaronbrethorst:
### Code review

_Re-review at `ae16aed`, covering commits since the previous review._

Both earlier findings are addressed at this head:

- The selected-bookmark exemption is gone from the identity diff (`bookmarkIDsToRemove = existingBookmarkIDs.subtracting(wantedBookmarkIDs)`), with regression tests for the deleted and replaced cases.
- Route-search `Stop` pins now survive a bookmark refresh via `preserveStopsOutsideIncoming: searchResponseOverridesStopLoading()`, which is the fix the last review suggested, plus a test for it.

aaronbrethorst (approved):
Both rounds of feedback are addressed, and addressed with the exact remedies rather than something adjacent to them.

The bookmark diff is now identity-only — `existingBookmarkIDs.subtracting(wantedBookmarkIDs)` with no selection exemption — and the open-callout protection moved up to `MapRegionManager`. That was the right split; conflating "this bookmark was deleted" with "this stop left the viewport" was what produced the stale and duplicate pins. The two regression tests for a selected bookmark being deleted and replaced pin it properly.

The route-search fix is the part I looked hardest at, and it's correct: `preserveStopsOutsideIncoming: searchResponseOverridesStopLoading()`, with that branch computing `wantedStopIDs = existingStopIDs.subtracting(bookmarkedStopIDs)` so nothing stale gets added back and only bookmark-shadowed stops drop. Selected stops are unioned in unconditionally, so a rider with an open callout never loses the pin under them. And `stopIDsToAdd = wantedStopIDs.subtracting(existingStopIDs)` is the line that actually kills the flash — existing pins are left alone instead of being re-added for MapKit to recycle.

Pulling the whole decision into a pure `StopAnnotationSync.changes(...)` is a real improvement over the inline slab. It's the kind of thing that's testable without a map, which matters here given annotations aren't drivable by UI automation.

Three things worth a follow-up, none blocking:

- `affectedStopIDs.subtract(selectedBookmarkStopIDs)` is effectively dead now — `affectedStopIDs` only holds stops of just-removed-and-deselected bookmarks, so the subtraction can only matter in a case that can't occur. The `#1266` comment above it describes protection that no longer has anything to protect.
- The local `selectedStopIDs` shadows the existing private computed property of the same name and recomputes the same value a different way.
- `preserveStopsOutsideIncoming` silently overrides `isStopsLayerEnabled`, so a route search with the stops layer off keeps its pins. I think that's right — search is explicit intent — but it's undocumented and untested.

Issue #1303, opened 2026-08-22
Local builds of main fail on a SwiftLint error, and CLAUDE.md's simulator destination is stale


aaronbrethorst:
Two unrelated papercuts surfaced while rebasing the `ghost-bus` branch onto `main`. Neither is caused by that branch; both affect local/agent development only.

## 1. `main` fails a local build: SwiftLint `type_body_length` error in `MapViewController.swift`

`OBAKit/Mapping/MapViewController.swift` trips SwiftLint's `type_body_length` **error** threshold, and because SwiftLint runs as a build phase, this fails the build outright:

```
OBAKit/Mapping/MapViewController.swift:24:1: error: Type Body Length Violation:
Type body should span 1000 lines or less excluding comments and whitespace:
currently spans 1004 lines (type_body_length)
```

Verified against a pristine worktree of `origin/main` at `fd7f6812`, so this is not branch-local.

### Why nobody noticed

`scripts/swiftlint.sh` no-ops entirely in CI:

```sh
if [ "$CI" = true ]; then
  echo "skipping swiftlint because in CI environment"
```

So `main` is green on GitHub and red on every developer's machine. The lint result is effectively unreviewed.

### How it got here

`.swiftlint.yml` sets `type_body_length` to `warning: 900` / `error: 1000`. The type has been over the *warning* threshold for a long time and crept steadily toward the error line, one PR at a time:

| Commit | Date | Type body lines |
|---|---|---|
| `1fd5f917` | 2026-08-01 | 979 |
| `b9d2280c` | 2026-08-04 | 981 |
| `e264c56a` | 2026-08-08 | 991 |
| `2f3e5f61` | 2026-08-13 | 995 |
| `33c208e2` | 2026-08-20 | 997 |
| `3a4e7689` | 2026-08-21 | 999 |
| **`fd7f6812`** (merge of #1302) | 2026-08-21 | **1004** ← crosses |

The merge of #1302 is where it crossed, but that PR is not really at fault — it added the last 5 lines to a type that had been sitting one line under the limit. The warning tier did its job for weeks and there was no mechanism to act on it.

### Suggested fix

Two parts, and the second matters more than the first:

1. **Unblock the build.** `MapViewController` is 1653 lines on disk. There is already an established split pattern in `MapViewController+MapLayers.swift`; moving a cohesive chunk into another extension is the low-risk move.
[...]

PR #1307, opened 2026-08-23
Resolve the simulator by UDID in CLAUDE.md, not a hardcoded device name

diveshpatil9104:
Part 2 of #1303 — docs only.

The three `xcodebuild` examples hardcoded `name=iPhone 17 Pro`, which only works if you happen to have that exact simulator installed. It's also the second device name these lines have carried (`iPhone 16` before it), so renaming it again would just push the problem to the next Xcode release.

Switched them to select by UDID instead — the first iPhone on the newest installed runtime, which is what `tests.yml` already does. Added a short "Picking a Simulator" section so there's one place that resolves `$SIMULATOR_UDID`, and pointed the xcodebuildmcp examples at it too, since they asked for a `--simulator-id <UDID>` that nothing in the file explained how to get.

Easy to miss in review: `-destination` went from single to double quotes so the variable actually expands.

Tested by running the snippet straight out of the committed file — resolves in both zsh and bash, and `xcodebuild` accepts the destination.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance for selecting an iOS simulator and using its unique identifier with build commands.
  * Updated simulator examples to resolve the appropriate device automatically instead of relying on fixed device details.

* **CI Improvements**
  * Standardized simulator selection across test workflows and local guidance.
  * Automatically selects the first available iPhone simulator on the newest installed iOS runtime.
  * Runs linting independently from builds.
  * Provides clearer errors when simulator discovery fails or no compatible device is available.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -17,6 +17,35 @@ scripts/generate_project                # Defaults to OneBusAway if no app speci
 
 **Available Apps**: OneBusAway, KiedyBus
 
+### Picking a Simulator
+
+Both sections below need a simulator UDID. Resolve one first — the newest installed
+iOS runtime's first iPhone, which is the same selection `.github/workflows/tests.yml`
+makes:
+
+```bash
+SIMULATOR_UDID=$(xcrun simctl list devices available -j | python3 -c '
+import json, re, sys
+devices = json.load(sys.stdin)["devices"]
+def ios_version(runtime):
+    match = re.search(r"iOS-(\d+)-(\d+)", runtime)
+    return (int(match.group(1)), int(match.group(2))) if match else (0, 0)
+for runtime in sorted(devices, key=ios_version, reverse=True):
+    phones = [d for d in devices[runtime] if d["name"].startswith("iPhone")]
+    if phones:
+        print(phones[0]["udid"])
+        break
+')
+: "${SIMULATOR_UDID:?No available iPhone simulator found — install a runtime in Xcode > Settings > Components}"
+```
+
+Address the device by `id=` rather than `name=`. A hardcoded `name=iPhone 17 Pro`
+fails two different ways behind the *same* misleading error — "Unable to find a device
+matching the provided destination specifier" — when the named device is not installed,
+and when it is ambiguous across multiple installed runtimes. `OS=latest` fixes only the
+second. It also goes stale on every Xcode release: this line has already been `iPhone 16`
+and `iPhone 17 Pro`.
+
 ### Building, Testing, and Driving the App: use `xcodebuildmcp`
 
 **Reach for `xcodebuildmcp` first.** It wraps `xcodebuild`/`xcrun simctl` and — critically —
@@ -33,10 +62,10 @@ Typical loop for verifying a change in the running app:
 
 ```bash
 scripts/generate_project OneBusAway                      # required before any build
-xcodebuildmcp simulator build-and-run --project-path OBAKit.xcodeproj --scheme App --simulator-id <UDID>
-xcodebuildmcp ui-automation snapshot-ui  --simulator-id <UDID> --style minimal
-xcodebuildmcp ui-automation touch        --simulator-id <UDID> --element-ref e26 --down --up --delay 0.08
-xcodebuildmcp ui-automation screenshot   --simulator-id <UDID> --return-format path --style minimal
+xcodebuildmcp simulator build-and-run --project-path OBAKit.xcodeproj --scheme App --simulator-id "$SIMULATOR_UDID"
+xcodebuildmcp ui-automation snapshot-ui  --simulator-id "$SIMULATOR_UDID" --style minimal
+xcodebuildmcp ui-automation touch        --simulator-id "$SIMULATOR_UDID" --element-ref e26 --down --up --delay 0.08
+xcodebuildmcp ui-automation screenshot   --simulator-id "$SIMULATOR_UDID" --return-format path --style minimal
 ```
 
 `snapshot-ui` returns semantic `elementRef` handles; drive those refs instead of
@@ -81,13 +110,13 @@ Use these only when `xcodebuildmcp` genuinely cannot do the job.
 
 ```bash
 # Build for testing
-xcodebuild clean build-for-testing -scheme 'App' -destination 'platform=iOS Simulator,name=iPhone 17 Pro'
+xcodebuild clean build-for-testing -scheme 'App' -destination "platform=iOS Simulator,id=$SIMULATOR_UDID"
 
 # Run unit tests
-xcodebuild test-without-building -only-testing:OBAKitTests -project 'OBAKit.xcodeproj' -scheme 'App' -destination 'platform=iOS Simulator,name=iPhone 17 Pro'
+xcodebuild test-without-building -only-testing:OBAKitTests -project 'OBAKit.xcodeproj' -scheme 'App' -destination "platform=iOS Simulator,id=$SIMULATOR_UDID"
 
 # Run specific test class
-xcodebuild test-without-building -only-testing:OBAKitTests/SpecificTestClass -project 'OBAKit.xcodeproj' -scheme 'App' -destination 'platform=iOS Simulator,name=iPhone 17 Pro'
+xcodebuild test-without-building -only-testing:OBAKitTests/SpecificTestClass -project 'OBAKit.xcodeproj' -scheme 'App' -destination "platform=iOS Simulator,id=$SIMULATOR_UDID"
 ```
 
 ### Code Quality
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -32,9 +32,17 @@ the two cannot drift apart.
 **Run it in the same shell invocation as the command that uses it.** It sets an
 ordinary shell variable, and this file's audience is Claude Code, whose Bash tool
 starts a fresh shell per call — the working directory persists, environment does
-not. Resolve it in one call and build in the next and you get
-`-destination "platform=iOS Simulator,id="`, which fails with exactly the error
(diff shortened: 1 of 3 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. `$SIMULATOR_UDID` is an ordinary shell variable set in one fenced block and consumed in two others 25 and 90 lines further down. That works for a human who keeps one terminal session open, but this file's stated audience is Claude Code, and its Bash tool starts a fresh shell per call — working directory persists, shell/env state does not. An agent that runs the "Picking a Simulator" block as one call and the `xcodebuild` block as the next gets `-destination "platform=iOS Simulator,id="`, whose failure is *"Unable to find a device matching the provided destination specifier"* — the exact misleading error lines 42-47 set out to eliminate. `tests.yml` doesn't have this problem because it writes the UDID to `$GITHUB_ENV`, which does persist across steps; the doc has no equivalent. A sentence saying the resolution must run in the same shell invocation as the command (or be `export`ed into the session), or inlining it into each block, closes it.

https://github.com/OneBusAway/onebusaway-ios/blob/a55851d31f3873ec3a995f6dd1f8c08ec2bdd66f/CLAUDE.md#L19-L27

CLAUDE.md
```diff
@@ -17,6 +17,35 @@ scripts/generate_project                # Defaults to OneBusAway if no app speci
...
+def ios_version(runtime):
+    match = re.search(r"iOS-(\d+)-(\d+)", runtime)
+    return (int(match.group(1)), int(match.group(2))) if match else (0, 0)
+for runtime in sorted(devices, key=ios_version, reverse=True):
+    phones = [d for d in devices[runtime] if d["name"].startswith("iPhone")]
+    if phones:
+        print(phones[0]["udid"])
+        break
+')
+: "${SIMULATOR_UDID:?No available iPhone simulator found — install a runtime in Xcode > Settings > Components}"
```
aaronbrethorst: `$SIMULATOR_UDID` is set here but consumed 25 and 75 lines further down, in separate fenced blocks (L65-68, L113-119). For this file's stated audience — Claude Code — those are separate Bash tool calls, and each one starts a fresh shell: cwd persists, environment does not.

Verified directly:

```
call 1:  export SWEEP_PROBE=set-in-first-call   →  "set-in-first-call"
call 2:  echo "${SWEEP_PROBE:-<EMPTY>}"          →  "<EMPTY — did not persist>"
```

An agent following the doc across two calls therefore runs `-destination "platform=iOS Simulator,id="` and gets "Unable to find a device matching the provided destination specifier" — the error this section exists to prevent. CI dodges it via `$GITHUB_ENV`; there's no equivalent here.

Either add an explicit note that the resolution and the build must happen in **one** shell invocation, or make each usage block self-contained, e.g.:

```bash
SIMULATOR_UDID=$(scripts/resolve_simulator_udid)   # or inline the snippet
xcodebuild clean build-for-testing -scheme 'App' \
  -destination "platform=iOS Simulator,id=$SIMULATOR_UDID"
```

Factoring the snippet into a small script under `scripts/` would also let `tests.yml` and the doc share one implementation instead of maintaining the same Python twice.

aaronbrethorst (changes requested):
This is fixing something that genuinely bites — I've hit the stale hardcoded device name on this repo myself, and `id=` over `name=` is the right call. I ran your snippet verbatim against real `simctl` output here and it works: it resolved a UDID cleanly, the `iOS-(\d+)-(\d+)` regex matches the real runtime keys, `watchOS-27-0` correctly doesn't false-match, and empty runtime buckets are skipped. I also diffed the algorithm against `.github/workflows/tests.yml`'s "Prepare iOS Simulator" step and your CI-parity claim is accurate — same regex, same reverse sort, same `startswith('iPhone')`, same `phones[0]`. Nice touch confirming the staleness claim is real, too; on this machine it resolves to an iPhone Air, not the iPhone 17 Pro the doc used to name.

One thing to fix before this lands, flagged inline: `$SIMULATOR_UDID` doesn't survive from the block that sets it to the blocks that use it.

CLAUDE.md's audience is Claude Code, whose Bash tool starts a **fresh shell for every call** — working directory persists, shell and environment state does not. I verified it just now:

```
call 1:  export SWEEP_PROBE=set-in-first-call   →  "set-in-first-call"
call 2:  echo "${SWEEP_PROBE:-<EMPTY>}"          →  "<EMPTY — did not persist>"
```

So an agent that runs the "Picking a Simulator" block and then, as a separate call, the build command at L113 gets `-destination "platform=iOS Simulator,id="` — and xcodebuild answers with *"Unable to find a device matching the provided destination specifier"*, which is precisely the error L42-47 is written to eliminate. CI avoids this only because it writes to `$GITHUB_ENV`, which persists across steps; the doc has no equivalent.

It fails loudly rather than silently, so it costs a round trip rather than producing a wrong result — but since this file is the thing steering every future session, I'd rather it be right. A sentence telling the reader to resolve the UDID **in the same shell invocation** as the command that uses it (or to inline the resolution into each block) would do it.

- The "same misleading error" framing at L42-45 is probably not accurate for the ambiguity case. An ambiguous `name=` historically reports *"The requested device could not be found because multiple devices matched the request"* — different text — and `tests.yml`'s own comment calls it "the 'multiple matching destinations' ambiguity". The conclusion (use `id=`) is right either way; it's the "two failures behind the same error" claim that's shaky.
- No `pipefail` in the snippet, so if `simctl` itself fails, python3 dumps a `JSONDecodeError` traceback and the guard then says "No available iPhone simulator found — install a runtime", which misdiagnoses the real cause. That's not hypothetical: `xcrun simctl` fails outright on this machine because `xcode-select` points at CommandLineTools, and the message sends you to the wrong place.

Please take a look at the inline comment — small fix and I'll re-review right away.

aaronbrethorst:
### Code review

Found 1 issue:

1. The new "same shell invocation" paragraph names the wrong error for the split-call case. It says an empty `$SIMULATOR_UDID` yields `-destination "platform=iOS Simulator,id="`, "which fails with exactly the error below" — i.e. the table's *not installed* row, "Unable to find a device matching the provided destination specifier". It doesn't. `id=` with no value is rejected at argument-parse time, before any destination lookup. Verified on Xcode 27.0 beta 6 against this repo's generated project, using the exact command the section documents:

   ```
   $ xcodebuild clean build-for-testing -scheme 'App' -project OBAKit.xcodeproj \
       -destination "platform=iOS Simulator,id="
   xcodebuild: error: missing value for key 'id' of option 'Destination'
   ```

   The table's first row is accurate for its own case — `name=iPhone 16` (not installed) does report "Unable to find a device matching the provided destination specifier", confirmed the same way. So the two are genuinely distinct errors, and the paragraph maps the wrong one to the empty-variable cause. That matters more than usual here, because this section exists to tell the reader which error means what, and this is the same class of mis-attribution flagged in the last round. The advice itself (resolve and use in one invocation) is right and the repeated line in each block is the correct fix; only the sentence attributing the failure to the table's error is wrong. Swapping "which fails with exactly the error below" for the actual `missing value for key 'id' of option 'Destination'` text is enough — it also happens to be the friendlier, self-diagnosing error, which is worth saying plainly.

https://github.com/OneBusAway/onebusaway-ios/blob/3fa6b91ae55dfe87bdcc553c58306244d78db76e/CLAUDE.md#L31-L40

aaronbrethorst (changes requested):
This is in good shape — the shared `scripts/resolve_simulator_udid` is exactly the right answer to #1303, and having `tests.yml` call the same script is what stops the docs and CI drifting apart, which was the whole problem. The `pipefail`, the explicit `simctl` failure branch with the `xcode-select` hint, and the "run it in the same shell invocation" paragraph all address what I raised last round.

One thing to fix, and it's one sentence:

The paragraph above the table says an empty `$SIMULATOR_UDID` "fails with exactly the error below," pointing at the not-installed row ("Unable to find a device matching the provided destination specifier"). I ran both on this machine against the generated project — an empty `id=` actually fails earlier, at argument parsing:

```
xcodebuild: error: missing value for key 'id' of option 'Destination'
```

The table's not-installed row *is* accurate (a bogus `name=` does give "Unable to find a device matching…"), so these are two genuinely different errors and the paragraph points at the wrong one. Given this file's whole job is telling the reader which error means what, and given last round turned on the same kind of inaccuracy, worth getting right.

Smaller: "Every command below needs a simulator UDID" overstates it — `generate_project`, `swiftlint.sh`, `docs`, `version` and the `xcodebuildmcp --help` block all appear below that line and need no UDID.

Separately: #1380 proposes substantially the same change to `CLAUDE.md` from a different author. Yours is the better treatment — it sorts by newest runtime, guards the empty case, covers the `xcodebuildmcp` block as well as raw `xcodebuild`, and de-duplicates against CI — so I plan to keep this one and close that.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
I ran `scripts/resolve_simulator_udid` from this branch on my machine before approving: it returned iPhone Air on the newest installed runtime, correctly skipping an empty older-runtime bucket. That's the thing I wanted proven rather than described.

Addressing a UDID by `id=` is genuinely more robust than a fresher device name, not just less stale — a UDID is unique across runtimes, so it needs no `OS=` companion, which is the failure I kept hitting. Sharing one implementation between CI and the documented commands is the right call, and the `simctl`-failure branch now points at `xcode-select`, which is exactly the case that bit me.

One leftover from my last review, small enough that I'm not holding on it: line 22's "Every command below needs a simulator UDID" still isn't true — `scripts/swiftlint.sh`, `scripts/docs`, `scripts/version` and the `--help` blocks need nothing. "Every command that touches a simulator" would fix it. Worth a one-line follow-up.

Merging.

PR #1308, opened 2026-08-23
Show arrival clock times in the region's time zone

### LINKED ISSUE #332: Displayed times does not use region's time zone

Borisserz:
## Summary
- Opt-in (default **off**): Settings → Arrival & Departure Display → “Show times in region time zone”. Matches Aaron’s #1102 requirement that this not ship default-on.
- When enabled, clock times use the region’s dominant agency IANA zone (`agencies-with-coverage`), with a short badge (`PST`/`CET`/`GMT+1`) only when the phone offset differs. Never uses `TimeZone.NameStyle.shortGeneric`.
- Timetable `Date`s are synthesized with a `Calendar` pinned to `formatters.timeZone` so agency-local seconds-since-midnight stay on the agency clock.
- `formattersTimeZoneDidChange` redraws stop/timetable UI when the zone lands after first paint.

Closes #332.

## Test plan
- [x] `TimeZoneScheduleBadgeTests` — Warsaw must not contain “Poland”/“Time”; Taipei vs Puget Sound shows Pacific noon + `(PST)`; non-GMT device/region pair keeps timetable at agency-local 08:00; timezone assignment posts redraw notification
- [x] `UserDefaultsStoreTests` — region time zone defaults off and persists
- [x] `ScheduleForRouteViewModelTests`
- [ ] Enable the Settings toggle; on a device set to a non-Pacific zone, open a Puget Sound stop: clock times are Pacific, badge present
- [ ] Same device, switch the phone to Pacific: badge disappears
- [ ] Toggle off: clocks return to the phone zone with no badge
- [ ] KiedyBus / Warsaw: badge is `CET`/`CEST` or `GMT+1`, never “Poland Time”

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in setting to display arrival and departure times in the transit region’s time zone.
  * Added schedule time-zone badges when regional and device time zones differ.
  * Updated timetable and stop displays to refresh when time-zone settings change.
  * Added localized settings and time-zone formatting text across supported languages.

* **Bug Fixes**
  * Improved timetable date and time calculations across differing device and regional time zones.

* **Documentation**
  * Added documentation for regional time-zone display behavior.


aaronbrethorst:
### Code review

Found 3 issues:

1. The route timetable is now *formatted* in the region's zone while its `Date`s are still *built* in the device's zone, so every cell shifts by the device↔region offset. `ScheduleForRoute.StopTime.departureDate(for:)` (unchanged) synthesizes `Calendar.current.startOfDay(for: scheduleDate).addingTimeInterval(TimeInterval(departureTime))`, and `departureTime` is seconds-since-midnight in *agency-local* time — the old `formatter.timeZone = TimeZone.current` cancelled that out and rendered the correct agency clock time regardless of where the phone was. With device `Asia/Taipei` and region `America/Los_Angeles`, an 08:00 agency departure now renders `16:00`. The shift appears exactly when the badge does, i.e. in the scenario this PR targets, and it is intermittent: if the `agencies-with-coverage` response has not landed when the view model's `lazy` formatter is first touched, the same screen renders correctly. Fix by pinning the `Calendar` used in `departureDate(for:)`/`arrivalDate(for:)` to `formatters.timeZone`, or by leaving these two formatters on `TimeZone.current`. The GMT pin in `OBAKitTests/Helpers/OBATestCase.swift` means neither the new suite nor CI can observe this.

https://github.com/OneBusAway/onebusaway-ios/blob/b950dae3ac82f6ac11bedc36b335e973da04bba4/OBAKit/Schedules/ScheduleForRouteViewModel.swift#L133-L148

2. Nothing invalidates clock text that is already on screen when the zone lands. `refreshFormattersTimeZone()` mutates `formatters.timeZone` from an unstructured `Task` once `agencies-with-coverage` returns, but `Formatters` is a plain `NSObject` — not observable, and it posts no notification — so every row rendered before that response keeps the device zone (and no badge) until something unrelated redraws it. `ScheduleForRouteViewModel`'s `lazy` formatters are worse: they snapshot `application.formatters.timeZone` once and never track a later change or a region switch. This is the same first-paint race flagged on #1102 ("which fetch completes first decides what the first paint looks like"), except there is no call ordering left to fix — the launch paint always wins, and the clocks flip mid-session.

https://github.com/OneBusAway/onebusaway-ios/blob/b950dae3ac82f6ac11bedc36b335e973da04bba4/OBAKitCore/Orchestration/CoreApplication.swift#L203-L220

3. Region-zone display ships default-on with no opt-out, which is the one change #1102 was asked to make before it could land: "I'm open to adding this, but I think it needs to be an opt-in option instead of being the default behavior," together with a concrete launch-time picker mock. `docs/region-timezone-display.md` declines the toggle on the issue's wording alone and does not mention that request.

https://github.com/OneBusAway/onebusaway-ios/blob/b950dae3ac82f6ac11bedc36b335e973da04bba4/docs/region-timezone-display.md#L15-L18

aaronbrethorst (changes requested):
This is a real gap and the agency-timezone resolution is a reasonable approach, but three things need to change before it can land.

**1. The route timetable double-shifts.** `ScheduleForRoute.StopTime.departureDate(for:)` builds dates as `Calendar.current.startOfDay(scheduleDate) + departureTime` seconds, where `departureTime` is seconds-since-midnight in *agency-local* time. The old `formatter.timeZone = TimeZone.current` cancelled that out, so the timetable was right on any device. Moving only the formatter to the region zone breaks the cancellation: device in Taipei, region in Los Angeles, an 08:00 agency departure renders as 16:00. It shows up exactly when the badge does — i.e. in the cross-zone case this PR exists to fix. The arrivals path is fine because those dates are absolute instants. The GMT pin in `OBATestCase` means the suite can't see this.

**2. The zone lands too late to affect first paint.** `formatters.timeZone` is set from an unstructured `Task`, and `Formatters` isn't observable, so anything rendered before the response keeps the device zone; `ScheduleForRouteViewModel`'s lazy formatters snapshot it once and never update. Arrival lists self-heal on the ~30s refresh, but the schedule VM doesn't, and the mid-session flip is visible.

**3. Default-on with no opt-out.** I asked for this to be opt-in on #1102 — "I'm open to adding this, but I think it needs to be an opt-in option instead of being the default behavior" — and the PR and `docs/region-timezone-display.md` decline a toggle citing the issue's "maybe" rather than that request. I'd still like the setting.

Also worth fixing: on a failed `agencies-with-coverage` or a region whose agencies resolve to nothing, `refreshFormattersTimeZone()` only assigns inside `if let`, so the *previous* region's zone sticks — Warsaw stops rendering Pacific times labelled `(PST)`. There's no reset to `.current`.

CI is red, and it's this PR: `refreshFormattersTimeZone()` fires `getAgenciesWithCoverage()` unconditionally from `refreshRESTAPIService()`, every `ApplicationTests` fixture triggers that, and `MockDataLoader` `fatalError`s on the unmocked URL, taking the test host down —

```
OBAKitTests/MockDataLoader.swift:126: Fatal error: Missing response to URL:
https://api.pugetsound.onebusaway.org/api/where/agencies-with-coverage.json?...
```

Registering that fixture in the shared setup (or gating the lookup) fixes it.

Borisserz:
Addressed the three review findings on the latest push:

1. **Timetable Date synthesis** — `ScheduleForRoute.RouteScheduleStopTime.departureDate/arrivalDate` now take an explicit `timeZone` (same as `formatters.timeZone`). Agency-local seconds-since-midnight stay on the agency clock when the phone is in another zone. Covered by `Route timetable clock stays agency-local when device zone differs` (does not rely on `TimeZone.current` / the GMT pin).

2. **Redraw when the zone lands** — `Formatters.timeZone` posts `.formattersTimeZoneDidChange`; `StopViewModel` / `StopViewController` and `ScheduleForRouteViewModel` + timetable views bump a generation and re-apply/rebuild.

3. **Opt-in, default off** — Settings → Arrival & Departure Display → “Show times in region time zone” (`showRegionTimeZone`, default `false`). Docs updated accordingly.

Borisserz:
Addressed the three review items in `0bb2187e`:

1. **Schedule date/zone mismatch** — `RouteScheduleStopTime.arrivalDate` / `departureDate` (and timetable sorting) now take an explicit `timeZone` and pin `Calendar` to it (call sites pass `formatters.timeZone`). Added a Taipei-device / LA-region test that asserts an 08:00 agency cell stays `8:00 … (PST)` and diverges from the device-calendar synthesis path.

2. **No redraw when zone lands** — `Formatters.timeZone` posts `.formattersTimeZoneDidChange`. `StopViewModel` / `ScheduleForRouteViewModel` bump a generation; UIKit stop list and SwiftUI timetables rebuild so clocks/badge update after a late `agencies-with-coverage` response.

3. **Opt-in, not default-on** — Settings → Arrival & Departure Display → “Show times in region time zone” (default **off**). `CoreApplication.refreshFormattersTimeZone()` only applies the region zone when enabled. Docs and PR body updated; strings in all 13 locales.

**Tests:** `TimeZoneScheduleBadgeTests` + `UserDefaultsStoreTests` + `ScheduleForRouteViewModelTests` — 74 passed on simulator `FA364C8B-D8FF-4900-AE73-FA9FC3A537F4`.

aaronbrethorst:
### Code review

Found 1 issue:

1. Localization regression in the Settings footer. The section footer moves from `settings_controller.arrival_display_section.transfer_banner.footer` — translated in all 13 locales since 55a6b523 — to a new key `settings_controller.arrival_display_section.footer` whose value is the **English** sentence in every one of the 13 `.strings` files. Twelve locales therefore go from a translated footer to English text on a screen every rider can reach (Settings → Arrival & Departure Display); the new switch title `settings_controller.arrival_display_section.region_time_zone` ships English-only the same way, and the old translated key is left behind as an orphan. `LocalizationTests` cannot see this — `Transfer banner footer names the switch title in every locale` still reads the *old* key, which is still present in every file, so it passes while guarding a string the app no longer renders (and the new English footer would fail that invariant: it starts with "Transfer banner:", not with the switch title "Show transfer arrival banner"). The prior commit that introduced these strings landed real translations for all 13 locales; this one should too.

https://github.com/OneBusAway/onebusaway-ios/blob/0bb2187ee6870f93f2c03f5346c1e176deaee6ce/OBAKit/Strings/ru.lproj/Localizable.strings#L1023-L1031

aaronbrethorst (changes requested):
The time-zone work here is genuinely good, and all three findings from the last round are properly fixed. Pinning the `Calendar` used to synthesize `ScheduleForRoute` dates to the same zone the formatter uses is exactly right — agency-local seconds-since-midnight cancel out, and I checked the new Taipei-device/LA-region test actually diverges between the buggy and fixed paths rather than being masked by the GMT pin in `OBATestCase`. The `formattersTimeZoneDidChange` generation bump for the late-arriving zone is a nice touch, and keeping "minutes until arrival" on elapsed-time math (zone-independent) is correct.

One blocker, and it's the recurring one in this repo.

**The Settings footer regresses from translated to English in twelve locales.**

`SettingsViewController` now renders `settings_controller.arrival_display_section.footer`, a new key. In every locale file that key holds the English sentence — `OBAKit/Strings/ru.lproj/Localizable.strings:1027` is the English string verbatim, sitting three lines below `...transfer_banner.footer:1024` which has a perfectly good Russian translation that the app no longer reads. Same story in the other eleven. The new switch title `...region_time_zone` (ru.lproj:1030) is English-only too.

`LocalizationTests` won't catch this — it only diffs against `en`, and its `Transfer banner footer names the switch title in every locale` case still passes because it reads the old key, which is now orphaned. So that test is currently guarding a string the app doesn't render, and the new English footer would actually fail its invariant if it were pointed at the right key (the footer starts with "Transfer banner:", not the switch title).

Either translate the two new keys across the twelve locales and drop the orphan, or keep the existing `transfer_banner.footer` key and reword it — whichever you prefer, but the app shouldn't ship a Settings screen that goes English for non-English riders.

A few non-blocking notes for the same pass:

- `Formatters.timeZone` defaults to `.current` rather than `.autoupdatingCurrent`, so those three formatters now freeze at launch where before they tracked the system. For a rider crossing a time-zone boundary mid-session the clocks stay in the old zone — and because `formattedClockTime`'s `deviceTimeZone: .current` default *is* re-evaluated per call, a badge would appear on every row even with the toggle off.
- `preferredScheduleTimeZone` tie-breaks with `max(by:)` over an unordered dictionary, so a region whose agencies split evenly between two zones can pick a different one each launch.
- The badge itself is unlocalized — `TimeZone.gmtOffsetLabel` hand-builds `"GMT+5:30"` and `appendingScheduleBadge` hardcodes `"\(clock) (\(badge))"`. Worth routing the wrapper through `OBALoc` at least.
- Coverage gaps worth knowing about even if you don't fix them here: `ScheduleForStopView` still formats with `.formatted(date:.omitted, time:.shortened)` (device zone), so the stop-schedule page contradicts the stop-arrivals page when the toggle is on; `StopArrivalView`'s VoiceOver clock label reads the raw `timeFormatter` and never speaks the badge; and the widget and `TripActivityPresenter` build their own `Formatters` and stay on the device zone.

The localization fix is small — ping me when it's in and I'll re-review.

Borisserz:
Addressed the localization blocker in 2534fbb3:

- Translated `settings_controller.arrival_display_section.footer` and `settings_controller.arrival_display_section.region_time_zone` across all 12 non-English locales (ar, es, fil, fr, it, ko, pl, pt-BR, ru, vi, zh-Hans, zh-Hant), using the existing transfer-banner footer translations as the base for the transfer half and adding a region time zone clause.
- Updated `LocalizationTests` to assert on the section footer and region time zone keys the app actually renders, instead of the orphaned `transfer_banner.footer` key.

Skipped the non-blocking items (autoupdatingCurrent, badge OBALoc) for now.

aaronbrethorst:
### Code review

Found 1 issue:

1. `refreshFormattersTimeZone()` never resets `formatters.timeZone` when the region lookup fails or resolves nothing, so the *previous* region's zone sticks — and the badge then actively mislabels it. The zone is only assigned inside `if let timeZone = TimeZone.preferredScheduleTimeZone(...)`; the `catch` only logs. With the toggle on, switching from Puget Sound to Warsaw and having `agencies-with-coverage` fail (network error, a server that doesn't serve the endpoint — the codebase already carries `supportsScheduleForRoute` because OBA deployments differ, or agencies whose `timezone` fields are blank/unparseable) leaves every Warsaw clock rendering Pacific time labelled `(PST)`, for the rest of the session. The function's own doc comment two lines above claims it "keeps (or resets to) the device zone", which is what the code does *not* do on this path. Same block: the `Task` is unstructured and re-checks nothing before applying, so a request that was in flight when the rider turned the setting **off** (or when the region changed again) still writes its result afterwards — `setShowRegionTimeZone(false)` sets `.current` and then loses to the late completion. An `else`/`catch` that assigns `.current`, plus a guard on `showRegionTimeZone` and the current region identifier before the write, covers both.

https://github.com/OneBusAway/onebusaway-ios/blob/2534fbb397458437bfd11642f662d7456717f50e/OBAKitCore/Orchestration/CoreApplication.swift#L211-L231

aaronbrethorst (changes requested):
All three of my numbered blockers are fixed, and fixed properly. The timetable double-shift in particular — requiring an explicit `timeZone` on `arrivalDate`/`departureDate` so the compiler forces every call site is a better answer than threading a default through, and I hand-checked the new test genuinely diverges rather than being masked by `OBATestCase`'s GMT pin. The redraw notification works (I confirmed `ArrivalDepartureItem`'s UUID identity makes the diffable snapshot reload), and defaulting the toggle off addresses the opt-out concern.

I'm holding on the one item I listed as "also worth fixing" last round, because looking at it again it's worse than I framed it.

`refreshFormattersTimeZone()` assigns the zone only inside `if let timeZone = TimeZone.preferredScheduleTimeZone(...)`. The `catch` logs and returns; a nil resolution falls through; `guard let apiService else { return }` returns. None of those reset anything. Since `formatters` is a `lazy var` created once per app lifetime, a region change whose lookup fails leaves the *previous* region's zone in place — and the badge then actively mislabels it. A rider switching to Warsaw sees Pacific times tagged `(PST)` for the rest of the session. That's worse than not having the feature, which is why I'd rather it not ship.

The doc comment right above says the function "keeps (or resets to) the device zone" — that's true of the guard path and not of the opt-in path, so the comment is describing the fix rather than the code.

Same block, same fix shape: the unstructured `Task` re-checks nothing before writing, so turning the setting off while a request is in flight loses to the late completion. Re-read `showRegionTimeZone` before assigning, or hold and cancel the task.

Should be about four lines. Everything else here is ready.

Non-blocking, and I'm happy for these to be follow-ups:

`Formatters.timeZone` defaults to `.current` rather than `.autoupdatingCurrent`. With the toggle off, a device that changes zone mid-session keeps the launch-time zone *and* gets a spurious badge on every row, because `formattedClockTime`'s `deviceTimeZone: .current` default *is* re-evaluated per call.

`GhostBusReportView.swift:77` uses `.formatted(date: .omitted, time: .shortened)` — device zone, no badge — so it contradicts the region-zone rows beside it when the toggle is on. I'd flagged `ScheduleForStopView` and the widget path as known gaps; this one I hadn't noticed.

`StopArrivalView.swift:288`'s VoiceOver label reads the raw `timeFormatter` and never speaks the badge, so VoiceOver users hear an unqualified time.

The badge wrapper `"\(clock) (\(badge))"` and the hand-built `"GMT+5:30"` bypass `OBALoc` entirely.

`preferredScheduleTimeZone`'s `max(by:)` over an unordered `Dictionary` can tie-break differently per launch for an evenly-split region.

Last thing, cosmetic but public: `docs/region-timezone-display.md` refers to me by first name in prose ("Aaron required this to be opt-in"). Rephrase to something role-neutral — it reads oddly in a repo doc.

Borisserz:
Addressed the review:

1. `refreshFormattersTimeZone()` resets to `.current` before the async lookup (and on empty/failed resolution), so a prior region zone cannot stick.
2. The unstructured `Task` re-checks `showRegionTimeZone` before assigning.
3. Docs no longer name a reviewer in prose.

Added `FormattersTimeZoneRefreshTests` for the reset-on-empty-coverage path.

aaronbrethorst:
### Code review

_Re-review at 17b92b6, covering commits since the previous review._

Borisserz:
VoiceOver on the arrival row now uses `formattedClockTime`, so the badge is spoken with the clock.

`preferredScheduleTimeZone` breaks an even split by lexicographic identifier, so the winner does not depend on `Dictionary` order. The clock wrapper and `GMT±` labels go through `OBALoc`.

`TimeZoneScheduleBadgeTests` passed (15 tests), including a stable even-split case.

aaronbrethorst:
### Code review

_Re-review at 1ac2c87, covering the commit pushed since the last review._

Found 1 issue:

1. The three new `timezone.*` keys land in `OBAKitCore/Strings/en.lproj/Localizable.strings` only — the other twelve OBAKitCore locales don't get them, so `LocalizationTests.Every locale has same keys as english` now fails twelve times. That test iterates *both* framework bundles (`frameworks` is `[("OBAKit", …), ("OBAKitCore", Bundle(for: Strings.self))]`), and OBAKitCore's `project.yml` pulls in `path: "."`, so all thirteen `.lproj` directories ship inside the framework; en and the other twelve are at exact key parity on `main` today (198 keys each), and this commit moves en to 201. This is the same localization-regression shape that blocked the 2026-09-05 round, just in Core rather than in OBAKit: the OBAKit half of this PR correctly translated `settings_controller.arrival_display_section.footer` and `.region_time_zone` into all twelve locales, and then the badge wrapper and `GMT±` labels that the feature actually renders on every row were added English-only. The rider-visible damage is small because `%1$@ (%2$@)` and `GMT+5:30` are near-locale-neutral, but the red test is not — and `OBALoc`'s `value:` fallback is what silently papers over it. Also in the same hunk: the block leaves a doubled blank line at L506-507.

https://github.com/OneBusAway/onebusaway-ios/blob/1ac2c8706c19f45ad47045c47e64750bf8b14c37/OBAKitCore/Strings/en.lproj/Localizable.strings#L496-L507

aaronbrethorst (changes requested):
The core time-zone work here is sound, and the three structural blockers from the earlier rounds are properly fixed. I re-derived the timetable arithmetic rather than trusting the tests: an LA region on a Taipei device lands at 08:00 PST via the new path, and the old buggy path lands at 16:00 — so the test genuinely diverges, and because `TimeZoneScheduleBadgeTests` passes `deviceTimeZone:` explicitly throughout, the GMT pin in `OBATestCase` can't mask it. That's the right way to test this. Requiring an explicit `timeZone` on `arrivalDate`/`departureDate` is also the correct shape — agency-local seconds-since-midnight should never be synthesized in a different zone than they're rendered in.

Two things to fix.

**1. The three new `timezone.*` keys are English-only in OBAKitCore, which fails CI.**

`OBAKitCore/Strings/en.lproj/Localizable.strings` gains `timezone.clock_with_badge_fmt`, `timezone.gmt_offset_hours_fmt` and `timezone.gmt_offset_hours_minutes_fmt`, and this PR touches no other OBAKitCore `.lproj` file. That matters because `LocalizationTests` iterates *both* bundles — `frameworks` includes `("OBAKitCore", Bundle(for: Strings.self))` — and OBAKitCore is at exact parity today: 198 keys in `en`, `ru`, `pl`, all thirteen. So `Every locale has same keys as english` records 12 failures.

`OBALoc`'s `value:` fallback hides this at runtime, which is precisely why it needs catching here rather than after it lands. These strings are close to locale-neutral, so translating them is quick.

Also a stray doubled blank line around L506 in that file.

**2. It needs a rebase** — the branch conflicts with `main`.

Two more I'd like addressed since they're the other half of asks from previous rounds:

- **`refreshFormattersTimeZone()` still has no region guard on the write.** The `Task` re-checks `showRegionTimeZone` after the `await` but not the current region identifier. Two overlapping region changes can land A's zone after B's and leave A's clocks and badge in place for the rest of the session. This was the second half of what I asked for on 09-06.
- **`Formatters.timeZone` defaults to `.current`, not `.autoupdatingCurrent`,** while `formattedClockTime`'s `deviceTimeZone: .current` is re-evaluated per call. A phone that changes zone mid-process — someone landing, auto-update kicking in — leaves `formatters.timeZone` pinned to the launch zone while the comparison side sees the new one, so every row grows a `(PST)` badge and renders in the stale zone *with the feature turned off*. Nothing resets it short of a region change. I flagged this as non-blocking before and you skipped it; now that the rest is solid I'd rather it went in.

Lower priority, noting so they're not lost:

- The VoiceOver change landed on `configureView(for traitCollection:)`, which the stop *list* doesn't reach — it goes through `StopArrivalCell.apply` → `configureView(for contentConfiguration:)` → `accessibilityTimeLabelText`, still raw `timeFormatter`. Spoken output is fine because `accessibilityValue` is badged; the gap is the *visible* clock at accessibility text sizes, where `accessibilityTimeLabel` replaces `fullExplanationLabel`.
- `.id(viewModel.formattersTimeZoneGeneration)` on the timetable containers is redundant — `@EnvironmentObject` already invalidates `body` — and its one observable effect is dumping a scrolled rider back to the top when the zone lands.
- `contextualDateTimeString` decides "is this today?" in the device zone but formats in the region zone, so near a cross-zone midnight the branch and the rendered date disagree.
- The orphaned `settings_controller.arrival_display_section.transfer_banner.footer` key is still in all 13 OBAKit files.

Fix the Core strings and the rebase and I'll take another pass.

Borisserz:
Addressed the review notes from 1ac2c87:

1. **`timezone.*` keys in every OBAKitCore locale** — not only `en.lproj` (the 12 non-English `.lproj` files now have the same three keys).
2. **Region race** — `refreshFormattersTimeZone` captures `regionIDAtStart` and bails if the region changed before the async write.
3. **Autoupdating default** — `Formatters.timeZone` and the failure / region-off paths use `.autoupdatingCurrent` instead of a fixed `TimeZone.current` snapshot.

Also rebased onto current `main` (resolved `LocalizationTests` merge with the Transfer-banner + Arrival-display tests; stripped a leftover conflict marker).

Ready for another look when you have a minute.

Borisserz:

Borisserz:
CI was red on one suite: `FormattersTimeZoneRefreshTests` still expected `TimeZone.current` after the production reset moved to `.autoupdatingCurrent`. Updated the assertions; should go green on this push.

Borisserz:
CI red was `FormattersTimeZoneRefreshTests` still asserting `.current` after production switched the reset path to `.autoupdatingCurrent`. Updated the expectations in `b745a5f6` — should go green on this tip.

aaronbrethorst:
### Code review

_Re-review at b745a5f, covering commits since the previous review._

Borisserz:
Thanks for the re-review at `b745a5f` — glad it came back clean.

GitHub still shows **Changes requested** from the earlier formal reviews (the “No issues found” notes land as issue comments, so `reviewDecision` does not clear). Happy to wait on an Approve when you are ready to merge.

Also: tip SHA already had a green `build` on run [35112975506](https://github.com/OneBusAway/onebusaway-ios/actions/runs/35112975506); a later run failed only on `ENOENT: OBAKitTests.xcresult` after tests (infra flake). Pushing an empty commit to retrigger checks.

aaronbrethorst:
### Code review

_Re-review at b7d0028, covering the commit pushed since the last review._

aaronbrethorst (approved):
Six rounds in, and this one earned it. Approving and merging.

Everything I asked for across the previous five rounds is fixed at this head, and I re-verified the three things most likely to have regressed rather than taking the earlier rounds on trust:

- Localization is at exact key parity — 658 OBAKit keys and 201 OBAKitCore keys in every one of the 13 catalogs, no diffs, no duplicates, and the new Settings strings carry real translations rather than English copies. On a 44-file diff that touches 26 `.strings` files, that is the outcome I was most worried about and it came back clean.
- The timetable date synthesis is zone-pinned everywhere it needs to be, and `Route timetable clock stays agency-local when device zone differs` proves it with an explicit Taipei calendar instead of leaning on the suite's GMT pin.
- No new test mutates `NSTimeZone.default` — the new suites pass `deviceTimeZone:` explicitly, so they can't corrupt suites running concurrently alongside them.

The failure-path handling from round five is also right: resetting to `.autoupdatingCurrent` up front and re-guarding both the setting and the region identifier on each side of the `await` is what keeps a stale zone from sticking.

One thing I'm landing as a follow-up rather than a sixth round, because it's a one-line change on a file this PR doesn't touch — I've filed it separately so it doesn't get lost.

`StopArrivalView` sets `accessibilityTimeLabel` from two different places. The `arrivalDeparture` didSet path goes through `formattedClockTime` and gets the badge. The content-configuration path — the stop list — goes through `StopArrivalItem.accessibilityTimeLabelText`, which still calls `formatters?.timeFormatter.string(from:)` directly. Since `timeFormatter` is now region-zoned globally, that label renders a region-zone time with no badge. And at accessibility content sizes `normalInfoStack` is hidden while `accessibilityInfoStack` is shown, so on a stop-list row that label is the *only* visible clock. Large-text users would get exactly the unlabelled-misleading-time case the badge exists to prevent.

It's gated behind a default-off setting, which is why it isn't blocking, but it should be routed through `formattedClockTime` like its sibling.

Nice work seeing this through.

PR #1309, opened 2026-08-23
Show Arrives vs Departs on the stop-page countdown

### LINKED ISSUE #447: Clarify arrival/departure listings at the beginning stops of routes
### Expected behavior

When looking at a stop that is the layover/first stop for a route, I should be able to tell at a glance when a particular vehicle is departing the stop. 

### Actual behavior

I have to very carefully parse the text that is displayed for that stop to understand what is an arrival vs. a departure.

### Steps to reproduce the behavior

1. Select the Puget Sound region.
2. Open Stop ID 13200

### Screenshot of the issue (if applicable)

img_a83781f9ee0b-1

### App Version/Hardware

18.2

Borisserz:
## Summary
- At first/layover stops (`ArrivalDepartureStatus.departing`, stop sequence 0) the minutes badge was just `5m` / `NOW`. The explanation line already said Arrives/Departs; you had to read the fine print.
- Stop-page countdown (chrono + grouped) now captions **Arrives** or **Departs** from `arrivalDepartureStatus`. Trip cards and bookmarks stay caption-less — those surfaces are already one-sided.
- VoiceOver for upcoming rows says `arrives in` vs `departs in`. Grouped cards say `next arrival` vs `next departure`. English fallbacks only; no new `.strings` keys, so locale parity tests stay green.

Closes #447.

## Test plan
- [x] `FormattersTests` — caption is `Arrives` / `Departs`
- [x] `StopPageAccessibilityCopyTests` (4) — row and grouped card VoiceOver verbs
- [ ] Puget Sound stop 1_13200: first-stop trips show **Departs** on the minutes; mid-route trips show **Arrives**
- [ ] VoiceOver on those rows: "departs in" vs "arrives in"
- [ ] Bookmark and trip cards still show minutes only (no caption)

## Summary by CodeRabbit

* **New Features**
  * Stop departure countdowns now clearly indicate whether a vehicle **arrives** or **departs**.
  * VoiceOver announcements use matching arrival and departure wording, including grouped route cards.
* **Accessibility**
  * Improved spoken details for upcoming vehicles, including route, destination, timing, and grouped departure counts.
* **Documentation**
  * Added guidance covering arrival and departure captions and accessibility behavior.
* **Tests**
  * Added coverage for captions and arrival/departure VoiceOver wording.

### DIFF
--- a/OBAKit/Stops/StopPage/Shared/StopPageAccessibilityCopy.swift
+++ b/OBAKit/Stops/StopPage/Shared/StopPageAccessibilityCopy.swift
@@ -1,0 +1,69 @@
+//
+//  StopPageAccessibilityCopy.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import Foundation
+import OBAKitCore
+
+/// Spoken identity for an upcoming stop-page row. The minutes badge is
+/// "5m" either way; first/layover stops are `.departing` and every other
+/// stop is `.arriving`. VoiceOver used to say "departs" for both (#447).
+enum StopPageAccessibilityCopy {
+    static func upcomingIdentity(
+        routeShortName: String,
+        headsign: String,
+        minutes: Int,
+        arrivalDepartureStatus: ArrivalDepartureStatus,
+        adherence: String
+    ) -> String {
+        let fmt: String
+        switch arrivalDepartureStatus {
+        case .arriving:
+            fmt = OBALoc(
+                "stop_page.row.a11y_arrives_fmt",
+                value: "Route %@ to %@, arrives in %d minutes, %@",
+                comment: "VoiceOver for a stop-page row whose vehicle is arriving at this stop. Route, headsign, minutes, adherence."
+            )
+        case .departing:
+            fmt = OBALoc(
+                "stop_page.row.a11y_fmt",
+                value: "Route %@ to %@, departs in %d minutes, %@",
+                comment: "VoiceOver for a stop-page row whose vehicle is departing this stop (first stop or layover). Route, headsign, minutes, adherence."
+            )
+        }
+        return String(format: fmt, routeShortName, headsign, minutes, adherence)
+    }
+
+    /// VoiceOver for a grouped route card. The next trip is arriving or
+    /// departing; saying "next departure" for both is the same #447 gap.
+    static func groupedCardIdentity(
+        routeShortName: String,
+        headsign: String,
+        minutes: Int,
+        arrivalDepartureStatus: ArrivalDepartureStatus,
+        adherence: String,
+        moreCount: Int
+    ) -> String {
+        let fmt: String
+        switch arrivalDepartureStatus {
+        case .arriving:
+            fmt = OBALoc(
+                "stop_page.grouped.a11y_arrives_fmt",
+                value: "Route %@ to %@, next arrival in %d minutes, %@. %d more departures loaded.",
+                comment: "VoiceOver for a grouped route card whose next vehicle is arriving at this stop."
+            )
+        case .departing:
+            fmt = OBALoc(
+                "stop_page.grouped.a11y_fmt",
+                value: "Route %@ to %@, next departure in %d minutes, %@. %d more departures loaded.",
+                comment: "VoiceOver for a grouped route card whose next vehicle is departing this stop (first stop or layover)."
+            )
+        }
+        return String(format: fmt, routeShortName, headsign, minutes, adherence, moreCount)
+    }
+}
--- a/OBAKitCore/Utilities/Formatters.swift
+++ b/OBAKitCore/Utilities/Formatters.swift
@@ -268,6 +268,18 @@ public class Formatters: NSObject {
         }
     }
 
+    /// One-word caption under the minutes countdown so a layover/first stop
+    /// is readable at a glance (#447). The explanation line already says
+    /// Arrives/Departs; the minutes badge did not.
+    public func arrivalDepartureCaption(for status: ArrivalDepartureStatus) -> String {
+        switch status {
+        case .arriving:
+            return OBALoc("formatters.caption.arrives", value: "Arrives", comment: "Short caption under a stop-page countdown for a vehicle arriving at this stop.")
+        case .departing:
+            return OBALoc("formatters.caption.departs", value: "Departs", comment: "Short caption under a stop-page countdown for a vehicle departing this stop (first stop or layover).")
+        }
+    }
+
     // MARK: - ArrivalDeparture Schedule Deviation
 
     /// Creates a formatted string representing the deviation from schedule described by `arrivalDeparture`
--- a/OBAKitCore/Utilities/Formatters.swift
+++ b/OBAKitCore/Utilities/Formatters.swift
@@ -271,11 +271,15 @@ public class Formatters: NSObject {
     /// One-word caption under the minutes countdown so a layover/first stop
     /// is readable at a glance (#447). The explanation line already says
     /// Arrives/Departs; the minutes badge did not.
-    public func arrivalDepartureCaption(for status: ArrivalDepartureStatus) -> String {
-        switch status {
-        case .arriving:
+    public func arrivalDepartureCaption(for status: ArrivalDepartureStatus, temporalState: TemporalState = .future) -> String {
+        switch (temporalState, status) {
+        case (.past, .arriving):
+            return OBALoc("formatters.caption.arrived", value: "Arrived", comment: "Short caption under a past stop-page countdown for a vehicle that already arrived.")
+        case (.past, .departing):
+            return OBALoc("formatters.caption.departed", value: "Departed", comment: "Short caption under a past stop-page countdown for a vehicle that already left.")
+        case (_, .arriving):
             return OBALoc("formatters.caption.arrives", value: "Arrives", comment: "Short caption under a stop-page countdown for a vehicle arriving at this stop.")
-        case .departing:
+        case (_, .departing):
             return OBALoc("formatters.caption.departs", value: "Departs", comment: "Short caption under a stop-page countdown for a vehicle departing this stop (first stop or layover).")
         }
     }
--- a/OBAKit/Stops/StopPage/Departures/DepartureRowView.swift
+++ b/OBAKit/Stops/StopPage/Departures/DepartureRowView.swift
@@ -169,7 +169,7 @@ struct DepartureRowView: View {
             minutes: departure.arrivalDepartureMinutes,
             isRealTime: status.isRealTime,
             color: dimmed ? Color(uiColor: .tertiaryLabel) : Color(uiColor: status.color),
-            caption: formatters.arrivalDepartureCaption(for: departure.arrivalDepartureStatus)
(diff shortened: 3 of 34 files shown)


aaronbrethorst:
### Code review

Found 2 issues:

1. Splitting the row/grouped VoiceOver labels keeps the *translated* key for the rare case and gives the common case a brand-new untranslated one, so localized builds regress. `stop_page.row.a11y_fmt` and `stop_page.grouped.a11y_fmt` are present and translated in all 13 `.lproj` files; they are now reserved for `.departing`, which `ArrivalDeparture.arrivalDepartureStatus` only returns at `stopSequence == 0`. Every other stop — i.e. nearly every row — takes the new `stop_page.row.a11y_arrives_fmt` / `stop_page.grouped.a11y_arrives_fmt`, which exist nowhere in `.strings`, so `NSLocalizedString` falls back to the English `value:`. Net effect in the 12 non-English locales: stop-page rows go from "Ruta %@ hacia %@, sale en %d minutos" to English. Same for the new `formatters.caption.arrives` / `formatters.caption.departs` — the caption this PR is about renders "Arrives"/"Departs" in Polish, Arabic, and Chinese. `LocalizationTests` only diffs en against the other locales, so keeping the keys out of `.strings` doesn't satisfy that guard, it just steps around it — its own doc comment says "A key missing from a locale silently renders in English. Nothing else catches it." Worth noting `Formatters.explanationForArrivalDeparture` already carries a fully translated arrive/depart vocabulary (`formatters.arrives_in_x_min_fmt`, `formatters.departs_in_x_min_fmt`, …) that this could draw on.

https://github.com/OneBusAway/onebusaway-ios/blob/b1057970a65b32b68b485437883ddc8a4a4603db/OBAKit/Stops/StopPage/Shared/StopPageAccessibilityCopy.swift#L25-L39

2. The caption is passed unconditionally, so already-departed rows read in the present tense. `DepartureRowView.countdown` is shared by all three styles, including `style: .past` (`ChronologicalListView` renders `rows(partition.past, style: .past)` behind the "Past · N" disclosure). A bus that left four minutes ago will show `-4m` with **Departs** under it, while the same row's VoiceOver label correctly says "departed 4 minutes ago". `arrivalDepartureCaption(for:)` takes only `ArrivalDepartureStatus` and drops `temporalState`; `Formatters.explanationForArrivalDeparture` models the full `(temporalState, arrivalDepartureStatus)` matrix and already has "Arrived %d min ago" / "Departed %d min ago" for exactly this case.

https://github.com/OneBusAway/onebusaway-ios/blob/b1057970a65b32b68b485437883ddc8a4a4603db/OBAKit/Stops/StopPage/Departures/DepartureRowView.swift#L167-L174

OBAKit/Stops/StopPage/Shared/StopPageAccessibilityCopy.swift
```diff
@@ -0,0 +1,69 @@
...
+        headsign: String,
+        minutes: Int,
+        arrivalDepartureStatus: ArrivalDepartureStatus,
+        adherence: String
+    ) -> String {
+        let fmt: String
+        switch arrivalDepartureStatus {
+        case .arriving:
+            fmt = OBALoc(
+                "stop_page.row.a11y_arrives_fmt",
```
aaronbrethorst, line 28: This is the regression. `.arriving` is the common case — every stop except the first — and it now points at `stop_page.row.a11y_arrives_fmt`, which is in no `.strings` file. Meanwhile `.departing` below keeps `stop_page.row.a11y_fmt`, which *is* translated in all 13 locales but now only fires when `stopSequence == 0`.

Net effect in ar/es/fil/fr/it/ko/pl/pt-BR/ru/vi/zh-Hans/zh-Hant: the stop-page row's VoiceOver label was translated before this PR and is English after it, on nearly every row. Same swap applies to `stop_page.grouped.a11y_arrives_fmt` below.

Running `scripts/extract_strings` and adding translations for the two new a11y keys (plus the two caption keys) resolves it. If translations for the new strings will lag, the safer interim shape is to keep the existing translated key as the `.arriving` default and introduce the new key for `.departing` instead — that way the untranslated path is the rare one rather than the common one.

OBAKitCore/Utilities/Formatters.swift
```diff
@@ -268,6 +268,18 @@ public class Formatters: NSObject {
         }
     }
 
+    /// One-word caption under the minutes countdown so a layover/first stop
+    /// is readable at a glance (#447). The explanation line already says
+    /// Arrives/Departs; the minutes badge did not.
+    public func arrivalDepartureCaption(for status: ArrivalDepartureStatus) -> String {
```
aaronbrethorst: This takes only `ArrivalDepartureStatus` and discards `temporalState`, so the caption is always present tense. `DepartureRowView.countdown` passes `caption:` unconditionally and is shared by all three styles including `style: .past` (behind the "Past · N" disclosure), so a bus that already left renders:

```
-4m
Departs
```

while that row's own VoiceOver label says "departed 4 minutes ago".

`Formatters.explanationForArrivalDeparture` right nearby already handles the full `(temporalState, status)` matrix and has "Arrived %d min ago" / "Departed %d min ago" — threading `temporalState` through here would let the caption follow the same rule. (New past-tense captions would need `.strings` entries too.)

aaronbrethorst (changes requested):
The decision logic here is right, and I checked it carefully because it's the part that's easy to get wrong: `stopSequence == 0 → .departing` is pre-existing model logic and you consume it faithfully, the last stop and mid-route cases both correctly read as arriving, and `arrivalDepartureMinutes` switches on the same status so the number and the caption can't disagree. Splitting the VoiceOver copy instead of bolting the word onto the visible label is the right shape too.

Two things to fix before this lands, both flagged inline.

The blocking one is a localization regression on the dominant path. `stop_page.row.a11y_fmt` and `stop_page.grouped.a11y_fmt` are translated in all 13 locales today. This PR reserves those for `.departing` — which is only the first stop of a trip — and routes `.arriving`, i.e. essentially every row a rider ever sees, through two brand-new keys that exist in no `.strings` file. So VoiceOver on the stop page goes from translated to English in 12 of 13 locales. I verified the key counts directly:

| key | locales |
|---|---|
| `stop_page.row.a11y_fmt` | 13 |
| `stop_page.grouped.a11y_fmt` | 13 |
| `stop_page.row.a11y_arrives_fmt` | 0 |
| `stop_page.grouped.a11y_arrives_fmt` | 0 |
| `formatters.caption.arrives` | 0 |
| `formatters.caption.departs` | 0 |

I want to be fair about this, because there's real ambient drift in this repo — roughly 80 source keys are already missing from `en.lproj`, and I'm explicitly *not* holding other PRs in this batch to a stricter standard for adding untranslated keys. What makes this one different is that it takes copy that is currently translated and makes it untranslated for most users. That's a regression, not drift.

The PR body says the keys were omitted "so locale parity tests stay green" — worth knowing that `LocalizationTests` only diffs the other locales against `en`, so omitting from `en` sidesteps the check rather than satisfying it. That test's own comment says it: "A key missing from a locale silently renders in English. Nothing else catches it." Running `scripts/extract_strings` and adding the four translations is the fix.

The second is smaller but user-visible: the caption is present-tense on past rows. A bus that left four minutes ago renders `-4m` with **Departs** under it, while that same row's VoiceOver label correctly says "departed 4 minutes ago".

One minor thing, not blocking: the arriving variant of the grouped-card label still ends "...3 more departures loaded", so it mixes arrival and departure vocabulary in one sentence. Reads acceptably, but you may want to split that trailing count too.

Please have a look at the inline comments — glad to re-review once the keys are in.

OBAKit/Stops/StopPage/Shared/StopPageAccessibilityCopy.swift
Borisserz: Added `stop_page.row.a11y_arrives_fmt` and `stop_page.grouped.a11y_arrives_fmt` to all 13 locales so the common arriving path is no longer English-only. Departing still uses the existing translated keys.

OBAKitCore/Utilities/Formatters.swift
Borisserz: `arrivalDepartureCaption` now takes `temporalState`. Past rows say Arrived/Departed; present/future stay Arrives/Departs. Call sites on the stop page pass `departure.temporalState`.

aaronbrethorst:
### Code review

Both previously-flagged issues are fixed: all six new keys (`stop_page.row.a11y_arrives_fmt`, `stop_page.grouped.a11y_arrives_fmt`, `formatters.caption.arrives|departs|arrived|departed`) are now present and genuinely translated in all 13 locales, every `.strings` hunk is a pure addition (no previously-translated value was replaced), format-specifier parity with `en` holds for every new key, and `arrivalDepartureCaption` now takes `temporalState` so past rows read Arrived/Departed.

Found 1 issue:

1. The Filipino present/future arrival caption uses the completed aspect, so an arriving bus reads as one that already came. `formatters.caption.arrives` is `"Dumating"` — the same verb form this file already uses for the *past* (`formatters.arrived_x_min_ago_fmt` = "Dumating %d min ang nakalipas", `trip_problem_code.user_description.came_early` = "Dumating nang maaga"). The future form the file uses elsewhere is "Darating" (`formatters.arrives_in_x_min_fmt` = "Darating sa loob ng %d min", and this PR's own `stop_page.row.a11y_arrives_fmt` = "…darating sa loob ng %3$d minuto…"). The departure pair right beside it is correctly tensed ("Aalis" future / "Umalis" past), which makes the arrival pair stand out: `arrives` = "Dumating" and `arrived` = "Dumating na" are near-identical, so in `fil` the caption cannot distinguish a bus arriving in 5 minutes from one that already arrived — and arriving is the dominant path this PR captions. "Darating" (or "Dumarating", matching `formatters.arriving_now` = "Dumarating na") is the form that matches the rest of the file.

https://github.com/OneBusAway/onebusaway-ios/blob/0a92b2e0a03ae86103f066762efc13f7363ce7ee/OBAKitCore/Strings/fil.lproj/Localizable.strings#L219-L225

Borisserz:
Fixed: `fil` `formatters.caption.arrives` is now `Darating` (future aspect), matching `arrives_in_x_min_fmt` / a11y copy.

aaronbrethorst:
### Code review — correction

Disregard my previous comment: it reviewed `0a92b2e`, and you'd already pushed the fix while it was in flight. On the current head `formatters.caption.arrives` is "Darating" in `fil`, which is the right future aspect and reads correctly against "Dumating na" for the past. That was the only thing I was holding on.

Nothing outstanding from me.

aaronbrethorst (approved):
Both blocking issues from my last review are fixed. The localization is thorough and correct — I confirmed every `.strings` hunk is a pure addition, so nothing previously translated got displaced, all six new keys are in 13 locales with real translations rather than English dumps, and format specifiers match `en` throughout. The `fil` verb-aspect slip I flagged earlier is fixed on this head ("Darating"); retracted above.

The past-row caption fix is right too — `StopPageListBuilder` partitions on `temporalState == .past` and `DepartureRowView.Style.past` is applied only to that partition, so caption and row style can't disagree.

Two follow-ups, neither blocking: `stop_page.grouped.a11y_arrives_fmt` still ends "more departures loaded" in all 13 locales (I called this non-blocking last time too), and `stop_page.grouped.expanded_row.a11y_fmt` is now orphaned — translated in 13 locales and referenced by nothing.

PR #1310, opened 2026-08-23
Open the map on the selected region, not GPS

### LINKED ISSUE #615: App doesn't take me to selected region
When OBA app opens, it's using user's current location and centering/zooming map on it. App should ask the user to be brought to automatically/manually selected region instead after a moment (or when panning/zooming).
It doesn't happen.
Using 22.3.0 OneBusAway and location is turned on.

Borisserz:
## Summary
- Launch (and the first location fix) zoomed to GPS even when that fix was outside the selected region. A rider in Taipei with Puget Sound selected opened on Taipei. The mismatch bulletin existed on the UIKit map and did not move the camera.
- `LaunchMapCamera` decides the frame: GPS inside the region → zoom to the user (nearby stops). GPS outside → last viewport if it still intersects the region, otherwise the region's `serviceRect`, plus the existing mismatch bulletin. No GPS → last viewport or `serviceRect`.
- UIKit `MapViewController` and SwiftUI `MapPanelRootView` share that function. The locate button still centers on the user. Returning from background after ten minutes does not yank to GPS when the device is still outside the region.

Closes #615.

## Test plan
- [x] `LaunchMapCameraTests` (6) — Puget Sound + Seattle GPS → user; Puget Sound + Tampa GPS → Puget Sound rect; out-of-region last viewport is ignored
- [ ] Select Puget Sound, simulate a location in Taipei (or any other region): map opens on Puget Sound, mismatch bulletin appears
- [ ] Locate button still jumps to the simulated GPS
- [ ] Device actually in Puget Sound: map still zooms to the user so nearby stops appear
- [ ] Background the app for >10 minutes while GPS is outside the region: map does not jump to GPS on resume

## Summary by CodeRabbit

* **New Features**
  * Improved map startup and recentering based on the selected region, current location, and previous viewport.
  * Added region-mismatch notifications when the device is outside the selected service area.
  * Preserved the current map view when returning from the background in another city.

* **Documentation**
  * Documented launch-camera behavior and region mismatch scenarios.

### DIFF
--- a/OBAKit/Sheet/Root/MapPanelRootView.swift
+++ b/OBAKit/Sheet/Root/MapPanelRootView.swift
@@ -49,6 +49,7 @@ struct MapPanelRootView: View {
     @State private var needsInitialRecenter = false
     @State private var regionMismatchBulletin: RegionMismatchBulletin?
     @State private var didPromptRegionMismatch = false
+    @StateObject private var mismatchCamera = RegionMismatchCameraActions()
 
     @Environment(\.scenePhase) private var scenePhase
     @Environment(\.colorScheme) private var colorScheme
@@ -203,6 +204,18 @@ struct MapPanelRootView: View {
         // from `reloadStopAnnotations`, i.e. on a region change.
         .onChange(of: mapViewModel.mapType) { _, _ in
             recomputeStopLabels()
+        }
+        .onChange(of: mismatchCamera.applyLaunch) { _, flag in
+            guard flag else { return }
+            mismatchCamera.applyLaunch = false
+            applyLaunchCamera()
+        }
+        .onChange(of: mismatchCamera.showSelectedServiceRect) { _, flag in
+            guard flag else { return }
+            mismatchCamera.showSelectedServiceRect = false
+            guard let selected = application.currentRegion else { return }
+            cameraPosition = .rect(selected.serviceRect)
+            viewportRecorder.record(selected.serviceRect)
         }
         .onChange(of: selectedStopID) { _, id in
             guard let id else { return }
@@ -508,7 +521,15 @@ extension MapPanelRootView {
         guard !didPromptRegionMismatch else { return }
         didPromptRegionMismatch = true
         guard
-            let bulletin = RegionMismatchBulletin(application: application),
+            let bulletin = RegionMismatchBulletin(
+                application: application,
+                onChangedPhysicalRegion: { [mismatchCamera] in
+                    mismatchCamera.applyLaunch = true
+                },
+                onShowSelectedRegionOnMap: { [mismatchCamera] in
+                    mismatchCamera.showSelectedServiceRect = true
+                }
+            ),
             let uiApp = application.delegate?.uiApplication
         else { return }
         regionMismatchBulletin = bulletin
@@ -591,3 +612,12 @@ extension MapPanelRootView {
     }
 
 }
+
+/// Holds one-shot camera requests from `RegionMismatchBulletin`. The bulletin
+/// cannot write `MapCameraPosition` itself — that lives in SwiftUI `@State` —
+/// so buttons flip these flags and the view applies them in `.onChange`.
+private final class RegionMismatchCameraActions: ObservableObject {
+    @Published var applyLaunch = false
+    @Published var showSelectedServiceRect = false
+}
+
--- a/OBAKit/Sheet/Root/MapPanelRootView.swift
+++ b/OBAKit/Sheet/Root/MapPanelRootView.swift
@@ -47,6 +47,8 @@ struct MapPanelRootView: View {
     /// first reported (non-zero) size, in which case the recenter must be
     /// retried when the size lands.
     @State private var needsInitialRecenter = false
+    @State private var regionMismatchBulletin: RegionMismatchBulletin?
+    @State private var didPromptRegionMismatch = false
 
     @Environment(\.scenePhase) private var scenePhase
     @Environment(\.colorScheme) private var colorScheme
@@ -102,10 +104,26 @@ struct MapPanelRootView: View {
         self.factory = factory
         self.viewportRecorder = MapViewportRecorder(application: application)
 
-        // With no location fix, frame the current transit region rather than
-        // letting `.automatic` frame the bookmark annotations.
+        // Frame the selected region unless GPS is already inside it.
+        // `.userLocation` follows the device, which is what opened Taipei
+        // over Puget Sound (#615).
         let fallback: MapCameraPosition = application.currentRegion.map { .rect($0.serviceRect) } ?? .automatic
-        _cameraPosition = State(initialValue: .userLocation(fallback: fallback))
+        let initialPosition: MapCameraPosition
+        if let selected = application.currentRegion {
+            switch LaunchMapCamera.target(
+                selectedRegion: selected,
+                userLocation: application.locationService.currentLocation,
+                lastVisibleMapRect: application.mapRegionManager.lastVisibleMapRect
+            ) {
+            case .userLocation:
+                initialPosition = .userLocation(fallback: fallback)
+            case .mapRect(let rect, _):
+                initialPosition = .rect(rect)
+            }
+        } else {
+            initialPosition = fallback
+        }
+        _cameraPosition = State(initialValue: initialPosition)
 
         // A returning user already has a fix here, so the `.onChange` below
         // never fires — the flag is already `true` and never transitions. Seed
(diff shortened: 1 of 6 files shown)


aaronbrethorst:
### Code review

Found 2 issues:

1. `MapPanelRootView` now presents `RegionMismatchBulletin`, but neither of the bulletin's buttons can move the SwiftUI map. Its alternative handler does `application.mapRegionManager.mapView.visibleMapRect = currentRegion.serviceRect`, and `MapViewportRecorder` documents that "the manager's own `MKMapView` is never installed in the SwiftUI panel" — nothing under `OBAKit/Sheet` touches `mapRegionManager.mapView`. So "Show Puget Sound on the map" is a no-op there, and "Change to Tampa" changes the region without repositioning `cameraPosition` either. On the UIKit surface both buttons work, which is why this only shows up on the newly-wired path.

https://github.com/OneBusAway/onebusaway-ios/blob/36f5979d793e07731e337da25bfd314ff09fe762/OBAKit/Sheet/Root/MapPanelRootView.swift#L500-L516

2. `GPS outside prefers a last viewport that still sits in the region` passes regardless of behavior. `last` is set to `pugetSound.serviceRect`, which is exactly what the other branch of `LaunchMapCamera.target` returns, so `#expect(MKMapRectEqualToRect(rect, last))` holds whether the function prefers the last viewport or falls back to the service rect. The "prefer an in-region last viewport" branch — the one genuinely new behavior here — has no coverage. Using a smaller in-region rect (e.g. `TestData.seattleMapRect`, as the no-GPS test does) would make it discriminating.

https://github.com/OneBusAway/onebusaway-ios/blob/36f5979d793e07731e337da25bfd314ff09fe762/OBAKitTests/Application/Mapping/LaunchMapCameraTests.swift#L47-L61

OBAKit/Sheet/Root/MapPanelRootView.swift
```diff
@@ -456,14 +474,45 @@ extension MapPanelRootView {
...
+        case .mapRect(let rect, let showMismatch):
+            cameraPosition = .rect(rect)
+            viewportRecorder.record(rect)
+            if showMismatch {
+                presentRegionMismatchIfNeeded()
+            }
+        }
+    }
+
+    private func presentRegionMismatchIfNeeded() {
```
aaronbrethorst, line 584: The bulletin this presents has no working actions on the SwiftUI surface.

`RegionMismatchBulletin`'s alternative handler writes `application.mapRegionManager.mapView.visibleMapRect`, and the primary button sets `currentRegion`, which `MapRegionManager.regionsService(_:updatedRegion:)` applies to that same `MKMapView`. But that map view isn't in this hierarchy — `OBAKit/Sheet/Root/MapViewportRecorder.swift:18` states it directly:

> the manager's own `MKMapView` is never installed in the SwiftUI panel, so nothing here keeps it current

and `grep` finds no reference to `mapRegionManager.mapView` anywhere under `OBAKit/Sheet`. So the rider gets a "Change Region?" card whose buttons dismiss it and leave the camera exactly where it was.

The SwiftUI path needs to drive `cameraPosition` itself — e.g. have the handlers set `cameraPosition = .rect(newRegion.serviceRect)` (and re-run `applyLaunchCamera()`) rather than relying on the UIKit map-view write.

OBAKitTests/Application/Mapping/LaunchMapCameraTests.swift
```diff
@@ -0,0 +1,106 @@
...
+        )
+        guard case .mapRect(let rect, let showMismatch) = target else {
+            Issue.record("expected mapRect, got \(target)")
+            return
+        }
+        #expect(MKMapRectEqualToRect(rect, pugetSound.serviceRect))
+        #expect(showMismatch)
+    }
+
+    @Test func `GPS outside prefers a last viewport that still sits in the region`() {
```
aaronbrethorst, line 48: This test passes whether or not the branch it's named for exists. It sets `last = pugetSound.serviceRect`, which is the *same* rect the fallback returns when the last viewport is rejected — so both paths produce an identical result and the assertion can't distinguish them. (`Fixtures.pugetSoundRegion` decodes deterministically from a fixture, so this isn't instance-identity noise either.)

That leaves "prefer an in-region last viewport" — the one genuinely new branch in `LaunchMapCamera` — without real coverage.

Using a last viewport that intersects the region but is clearly distinct from the whole service rect would make it a real lock, e.g. a downtown-Seattle-sized rect inside Puget Sound: then `.mapRect(last)` and `.mapRect(serviceRect)` are visibly different and only the correct branch passes.

aaronbrethorst (changes requested):
The core idea here is right, and `LaunchMapCamera` as a pure decision function that both surfaces call is exactly the shape I'd want — the precedence (GPS inside the region → follow the user; GPS outside → last viewport if it still intersects, else the service rect) reads correctly and the UIKit wiring is sound. I also checked the launch-time nil-region worry and there's no ocean/(0,0) risk: `RegionsService.currentRegion` resolves synchronously against bundled regions in `init`, and `lastVisibleMapRect`'s getter already falls back to the region's service rect.

Two things to fix, both inline and both small.

The first is that the region-mismatch bulletin you newly wire into `MapPanelRootView` has two buttons that can't do anything. Its alternative handler writes `application.mapRegionManager.mapView.visibleMapRect`, and that `MKMapView` is never installed in the SwiftUI panel — `MapViewportRecorder`'s own doc comment says so in as many words, and I confirmed nothing under `OBAKit/Sheet` touches it. The primary button routes through `MapRegionManager.regionsService(_:updatedRegion:)` to the same uninstalled map view. Neither one moves `cameraPosition`. This works correctly on the UIKit surface; it's specifically the newly-wired SwiftUI path. It's behind `OBAUseMapPanelExperience`, which has no registered `true` default, so it only reaches people who flipped the toggle — but it's still a modal that does nothing when they tap it.

The second is that the test named for the new branch can't fail. Worth fixing because it's the only coverage the genuinely new behavior has.

One more worth your thought, not blocking: `locationService(_:locationChanged:)` used to call `programmaticallyUpdateVisibleMapRegion` (latched behind `initialMapChangeMade`) and now calls `applyLaunchCamera`, which can reach the bulletin block on *any* location update. A rider who starts inside the region never consumes `promptUserOnRegionMismatch`, so crossing a `serviceRect` boundary later in the session can pop "Change Region?" over whatever screen they happen to be on. It's capped at once per `MapViewController` lifetime, and with auto-select on the common travel case switches region first — so it's an edge, not a likely path. Still, "launch camera" logic firing mid-session is worth a latch.

Please have a look at the inline comments.

OBAKit/Sheet/Root/MapPanelRootView.swift
Borisserz: SwiftUI now passes callbacks that flip a small ObservableObject; MapPanelRootView applies them to `cameraPosition` (change region → `applyLaunchCamera()`, show selected → `.rect(serviceRect)`). UIKit still writes `mapView` when the callbacks are nil.

OBAKitTests/Application/Mapping/LaunchMapCameraTests.swift
Borisserz: `last` is now `TestData.seattleMapRect`: it intersects Puget Sound and is not equal to `serviceRect`. The test asserts both, so the fallback path fails this case.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Both blocking issues from my last review are genuinely fixed. The SwiftUI bulletin buttons now do something — `RegionMismatchBulletin` takes the two callbacks, the UIKit path is unchanged because they default to nil, and I followed the observation chain through `@StateObject` → `objectWillChange` → `.onChange` without finding a loop or a retain cycle. The non-discriminating test now uses a real Seattle rect with explicit intersects/not-equal preconditions, so the fallback branch actually fails it.

I also specifically checked the regression I was worried about: the granted-permission-follows-me path holds on both surfaces. `didReceiveInitialLocation` is a one-shot latch whose `.onChange` recenters, so grant-after-launch still works.

Still outstanding, still non-blocking: the launch-camera logic can now fire mid-session from `locationChanged`, so crossing the service-area boundary can pop "Change Region?" — capped once per VC lifetime, and `viewWillAppear` could already reach that branch before this PR, so it is a new trigger rather than a new class of bug. A latch would still be nicer.

Worth a second look after the rebase specifically in `MapPanelRootView.swift` — main has moved a lot under it (rentals, `MapPinSelection`, `MapPanelLayersModel`), so that resolution is not mechanical.

Borisserz:

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Pulling the decision out into `LaunchMapCamera.target(selectedRegion:userLocation:lastVisibleMapRect:)` is what makes this reviewable — a pure function over three inputs, with both the UIKit and SwiftUI surfaces calling the same thing. I traced the precedence across first launch, no permission, permission granted after launch, and out-of-region, and it holds: an in-region fix still wins and still zooms to nearby stops, and an out-of-region fix no longer drags a rider off the region they picked.

Both blockers from the earlier round are properly fixed. The SwiftUI bulletin buttons actually move the camera now via the `RegionMismatchCameraActions` flags rather than being dead, and the previously-vacuous test has real preconditions (`intersects` plus a `!equal` against the service rect) so the fallback branch would fail it.

Two for later, neither blocking. The mismatch path can still fire mid-session — `locationChanged` routes into `applyLaunchCamera`, so a rider who launches in-region and later crosses the service-rect boundary can get "Change Region?" over whatever they're doing; it's capped once per VC lifetime, but a launch latch would be cleaner. And in the no-GPS branch the last viewport is used without the `intersects` check the GPS branch applies, so a rider who switched regions and relaunched with no fix could open on the old region's viewport. Narrow in practice, and it's the pre-existing behavior, but the check is free.

PR #1311, opened 2026-08-24
Add ProximityAlertManager to fire destination alerts

diveshpatil9104:
## Add ProximityAlertManager to fire destination alerts

Adds the manager that arms a geofence for a saved destination alert, posts a local notification when the rider crosses it, and keeps Core Location's monitored regions in agreement with the alerts in `UserDataStore`.

Refs [#1150](https://github.com/OneBusAway/onebusaway-ios/issues/1150). PR 3 of 4 for [#455](https://github.com/OneBusAway/onebusaway-ios/issues/455).

PR 2 shipped the geofencing API with no consumer — `ProximityMonitoringResult`, `RegionMonitoringFailureKind`, and the 20-region guard have had nothing reading them. This is that consumer.

### What changed

**New file — `ProximityAlertManager.swift` (`OBAKit`)**

`@MainActor` manager owning the alert lifecycle. `createProximityAlert(for:radiusMeters:)` returns a `ProximityAlertActivationResult` rather than half-succeeding — nothing is persisted unless the geofence actually armed, and notification authorization is checked first, since an armed geofence whose notification can't be shown does nothing. On a crossing it posts the notification and deletes the alert (one-shot). A single `reconcileMonitoredRegions()` is the only repair path, called from construction, `.proximityAlertsDidChange`, `applicationDidBecomeActive`, and the grant of Always authorization — that last one closes the "no self-healing re-arm" gap PR 2 documented but couldn't fix alone.

**Modified — `Application.swift`**

Builds the manager in `init`, not lazily: a geofence crossing relaunches a terminated app, and nothing on that launch would ever touch a lazy property. Adds `pushService(_:receivedProximityAlertForStopID:)`, which stashes into `pendingStopID` when no region has loaded yet rather than dropping the tap.

**Modified — `PushService.swift`**

Routes the notification tap. `OBACloudPushService` owns the app's only `UNUserNotificationCenter` delegate, and its fallback for an unrecognized payload re-presents the notification's own text in a modal — for this one, telling the rider they're approaching their stop a second time.

**Modified — `LocationService.swift` (`OBAKitCore`)**

Three accessors the manager needs: `proximityAlertID(forRegionIdentifier:)`, `monitoredProximityAlertIDs`, and `stopMonitoringProximityAlert(id:)` — the last for a region whose alert no longer exists to name it.

### How to test

No UI in this PR — the stop-page entry point is PR 4, so there's nothing to record on screen.

- `OBAKitTests/ProximityAlertManagerTests` — 32 tests: creation refusals, all four reconciliation triggers, the fire path, authorization changes, both failure classifications
- `OBAKitTests/PushServiceTests` — 3 tests for the new routing branch
-  `OBAKitTests/ApplicationTests` — the tap is stashed, never dropped
-  Full `OBAKitTests` target passes (2343 tests)
-  Strings added to all 13 locales

---

> **PR Roadmap for [#455](https://github.com/OneBusAway/onebusaway-ios/issues/455)**
>
> | # | PR | Status |
> |---|-----|--------|
> | 1 | `ProximityAlert` model + `UserDataStore` persistence | `Merged` |
> | 2 | `LocationService` geofencing + hardening | `Merged` |
> | 3 | **`ProximityAlertManager` + local notifications** | `Current` |
> | 4 | Stop page UI to create and cancel alerts | `Next` |

## Summary by CodeRabbit

* **New Features**
  * Added destination proximity alerts that notify riders as they approach a selected stop.
  * Alerts support custom distances, cancellation, authorization checks, and automatic monitoring cleanup.
  * Tapping a proximity alert opens the associated destination when possible, or completes navigation when the app is ready.
* **Localization**
  * Added proximity-alert notification titles and messages in multiple supported languages.
* **Bug Fixes**
  * Improved handling of expired, missing, or invalid monitored alerts and notification data.

### DIFF
--- a/OBAKit/Orchestration/Application.swift
+++ b/OBAKit/Orchestration/Application.swift
@@ -106,6 +106,17 @@ public class Application: CoreApplication, PushServiceDelegate {
     @MainActor
     lazy var walkingSpeedManager = WalkingSpeedManager(userDataStore: userDataStore)
 
+    /// Owns the rider's destination proximity alerts.
+    ///
+    /// `lazy` only because it needs a fully-initialized `self`; `init` forces it
+    /// immediately, and the comment there explains why it must not wait for a
+    /// first use that may never come.
+    @MainActor
+    public private(set) lazy var proximityAlertManager = ProximityAlertManager(
+        locationService: locationService,
+        userDataStore: userDataStore
+    )
+
     @objc lazy var userActivityBuilder = UserActivityBuilder(application: self)
 
     /// Handles all deep-linking into the app.
@@ -169,6 +180,15 @@ public class Application: CoreApplication, PushServiceDelegate {
         bookmarkWidgetRefresher = BookmarkWidgetRefresher()
 
         super.init(config: config)
+
+        // Force the proximity alert manager now instead of leaving it to a first
+        // use that may never come. A geofence crossing relaunches a terminated
+        // app, and Core Location delivers the queued region event to whatever
+        // `LocationService` delegates exist once launch finishes — a launch on
+        // which no screen is ever built and nothing touches this property.
+        // Constructing it here registers the delegate, and re-arms the regions,
+        // before that event lands.
+        _ = proximityAlertManager
 
         configureAppearanceProxies()
     }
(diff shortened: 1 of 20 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. `Entering the geofence of an expired alert delivers nothing` never reaches the expiry guard it names, so that guard ships untested. The second `store.add(proximityAlert: expired)` posts `.proximityAlertsDidChange`, which the manager observes with the selector form — `NotificationCenter` invokes it inline, `reconcileMonitoredRegions()` runs (`isReconciling` is false at this point, since `init`'s pass already returned), and `deleteExpiredProximityAlerts()` reaps the alert a second time before `enterRegion(for:)` is ever called. The crossing therefore lands in the "alert is no longer stored" branch of `didEnterMonitoredRegion`, which is already covered by `Entering a region whose alert is gone disarms it silently`. All three assertions still hold on that path, so deleting the `guard !alert.isExpired` check entirely would leave this test green — and the inline comment ("put it back to reach the guard under test") asserts the opposite of what happens. Assigning `store.proximityAlerts = [expired]` directly after `makeManager()` bypasses the notification and reaches the guard.

https://github.com/OneBusAway/onebusaway-ios/blob/030a3f3b0713f24eaba3d22d539c6007711ee9fd/OBAKitTests/ProximityAlerts/ProximityAlertManagerTests.swift#L465-L474

aaronbrethorst (approved):
This is a well-built subsystem. Funnelling every drift-correcting path — construction, `.proximityAlertsDidChange`, `applicationDidBecomeActive`, and the grant of Always — through a single `reconcileMonitoredRegions()` is the right architecture for something that has to stay in agreement with Core Location across app launches, and the `isReconciling` re-entrancy guard is load-bearing and actually tested.

Given what this touches, I checked the things that would be serious rather than annoying:

- **Battery.** No `startUpdatingLocation`, no `allowsBackgroundLocationUpdates`, no accuracy tuning anywhere in the diff — this is pure region monitoring, and `applicationWillResignActive`'s existing `stopUpdates()` is untouched.
- **The 20-region cap.** Checked up front by `startMonitoringProximity` and surfaced as a typed `.regionLimitReached(limit:)` rather than failing silently, and reconcile disarms orphans before arming to return slots.
- **Authorization.** No new prompt (correctly deferred), and `NSLocationAlwaysAndWhenInUseUsageDescription` already exists in `app_shared.yml:81`.
- **Repeat firing.** One-shot — delivered then deleted — and `deleteExpiredProximityAlerts` skips its notification post when nothing changed, so there's no loop.
- **Retain cycles.** `LocationService.delegates` is an `NSHashTable.weakObjects()`, and the selector-form observer needs no `deinit`.
- **Localization.** All 13 catalogs carry real translations, correctly placed. No English placeholders.

Persisting only after the geofence actually arms — rather than the reverse order — is exactly right, and the comment explaining why is the kind I want in this file.

Filed #1347. The one I'd like closed soonest: `Entering the geofence of an expired alert delivers nothing` doesn't actually reach `guard !alert.isExpired`. The second `store.add` posts `.proximityAlertsDidChange`, reconcile runs inline and reaps the alert, so the crossing hits the "no longer stored" branch instead. Delete the expiry guard and that test stays green — and that guard is the last thing standing between a rider and a 2am notification for yesterday's trip.

PR #1312, opened 2026-08-24
Give loop-route trip stops unique list identifiers

### LINKED ISSUE #538: Gracefully handle duplicate or illogical transit data
Context: #531 and #537

> For these types of GTFS issues (duplicate stops, etc.), OBA iOS will need to be a little more tolerant and not crash. AFAIK there isn't actually anything in the GTFS spec that prevents this specific instance of two stops in a row with the same time. I'll be the first to admit that it's not (typically) logical, but for a variety of reasons (primarily related to transit agency export software) it's historically been allowed. So it won't be an unusual case for these issues to pop up from time-to-time.
>
> Obviously, the data should be fixed server-side if it is indeed a data problem, but iOS should allow the display of the data without crashing.

_Originally posted by @barbeau in https://github.com/OneBusAway/onebusaway-ios/issues/537#issuecomment-932564829_

In relation to App Store crash reports, OBAListView throwing duplicate identifier exceptions tops the list.
Fix requires a rethinking of how to identify items so UIDiffableDataSource does not crash.

Borisserz:
## Summary
- A loop visits the same stop twice. UIKit `TripStopViewModel.id` was the bare `stop.id`, so `NSDiffableDataSource` crashed. That is the App Store crash in #538.
- Ids are now `"\(stopIndex)-\(stop.id)"`, the same format `TripStopListModel` already uses on the SwiftUI trip page.
- `Equatable` includes `id`, matching the list-item protocol. Two visits to one stop are no longer equal.

Closes #538.

## Test plan
- [x] `TripStopViewModelIdentityTests` (2) — two visits to one stop have distinct ids; a full trip's ids are unique
- [x] `TripStopTemporalStateTests` (7) — existing classification still holds
- [ ] Open a loop trip (same stop twice) on the classic trip list: both visits appear, no crash

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed crashes and display issues on loop routes where the same stop appears multiple times.
  * Stop entries now remain distinct based on their position in the trip, even when they share the same stop record.

* **Tests**
  * Added coverage ensuring repeated stops receive unique identifiers and remain distinct.

* **Documentation**
  * Documented handling of duplicate trip-stop identifiers across UIKit and SwiftUI.

### DIFF
--- a/OBAKit/Trip/TripStopListItem.swift
+++ b/OBAKit/Trip/TripStopListItem.swift
@@ -38,7 +38,14 @@ nonisolated struct TripStopListItemRowConfiguration: OBAContentConfiguration {
 }
 
 nonisolated struct TripStopViewModel: OBAListViewItem {
-    var id: String { stop.id }
+    /// Position-qualified: a loop visits the same stop twice, and
+    /// `NSDiffableDataSource` crashes on duplicate item identifiers (#538).
+    /// Same format as `TripStopListModel.Row.id`.
+    var id: String { "\(stopIndex)-\(stop.id)" }
+
+    /// Index of this visit on the trip. Carried so `id` can stay unique
+    /// without parsing the identity string.
+    let stopIndex: Int
 
     var configuration: OBAListViewItemConfiguration {
         return .custom(TripStopListItemRowConfiguration(viewModel: self))
@@ -101,6 +108,7 @@ nonisolated struct TripStopViewModel: OBAListViewItem {
         routeType = stopTime.stop.prioritizedRouteTypeForDisplay
 
         self.onSelectAction = onSelectAction
+        self.stopIndex = stopIndex
     }
 
     func hash(into hasher: inout Hasher) {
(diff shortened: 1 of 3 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Correct fix, and I like that you matched `TripStopListModel.Row.id`'s existing `"\(index)-\(stopID)"` format rather than inventing a second scheme — the two trip pages now agree on identity.

I checked the stability question, since an index-based id is only safe if the index is stable: `stopIndex` enumerates `tripDetails.stopTimes`, which is the trip's fixed schedule order and doesn't reorder between reloads. So this is unique for duplicate visits *and* stable, not the fragile "index into a mutable array" case.

Also confirmed nothing else keys off the old id — `OBAListViewItem`'s own doc says `id` is reserved for future use, and every call site (`highlightStopInList`, `onSelectTripStop`, `buildTransferContext`, `showStopOnMap`, the context menu) goes through `tripStop.stop` or `stopTime`. And like #1327, adding `id` to `==` fixes a latent Equatable/Hashable mismatch rather than creating one.

Filed #1346: "Closes #538" auto-closed a broader tracking issue that asks for a rethink across the whole list framework, so you may want to reopen it. That issue also notes the second test passes against the old implementation (the fixture has 53 distinct stop IDs), and that `highlightStopInList` still resolves the wrong occurrence on a loop.

Issue #908, opened 2025-12-11
Showing the schedule feature on iPadOS 18 crashing


aaronbrethorst:
Reported by a user: the app is crashing when they try bringing up the schedule for the 1 line light rail line in seattle.

PR #1313, opened 2026-08-24
Ignore cancelled schedule fetches when the date changes

Borisserz:
## Summary
- Changing the schedule date cancels the in-flight fetch; treat `URLError.cancelled` / `CancellationError` as non-errors so the sheet does not flash an error or clear a newer fetch's spinner.
- Uses the existing `Error.isCancellation` helper (NSError bridge included) rather than a duplicate type.
- Does **not** change sheet presentation style — still waiting on a symbolicated trace for the iPad DatePicker crash.

Refs #908.

## Test plan
- [x] Build / existing schedule tests
- [ ] On device: open stop schedule, scrub the date quickly — no error flash, spinner stays coherent


aaronbrethorst:
### Code review

Found 2 issues:

1. The new `Task.isCancelled` guard doesn't do what its comment says: a cancelled fetch still publishes. `defer { isLoading = false }` runs unconditionally on the cancelled path, and `URLSession.data(for:)` throws `URLError(.cancelled)` when the enclosing task is cancelled, so the `catch` sets `self.error` too — both are `@Published` on the same `ObservableObject` and fire `objectWillChange` exactly like `scheduleData` does. If publishing after cancellation were really the crash, the guard would not prevent it. There is also a concrete user-visible consequence: `error` is cleared only at the *top* of `fetchSchedule()`, never on success, so a fetch cancelled by a date change can leave `error` non-nil after the replacement fetch succeeds — and `content` checks `else if let error` before `else if scheduleData != nil`, so the sheet sticks on "Unable to load schedule" with good data loaded. The guard should cover the cancellation error path (and the `isLoading` reset), not just the success assignment. Same shape in `ScheduleForStopViewModel.swift` (L107–L125).

https://github.com/OneBusAway/onebusaway-ios/blob/f2a7198339be507724cb6a144efd731c755eaa92/OBAKit/Schedules/ScheduleForRouteViewModel.swift#L166-L197

2. Nothing in this PR verifies the crash fix. The two tests assert `modalStyle(for: .phone) == .pageSheet` and `modalStyle(for: .pad) == .formSheet` — a restatement of the one-line ternary they call, which cannot fail unless someone edits that ternary. They don't touch the two lines that actually carry the fix (`ScheduleViewControllers.swift` L25 and L45); reverting either of those to `.pageSheet` leaves both tests green. The author's own test plan leaves both device-verification boxes unchecked, and `docs/ipad-schedule-crash.md` states the mechanism ("iPadOS 18 cannot attach" the calendar in a `.pageSheet`) as fact without a stack trace to support it — on iPad both `.pageSheet` and `.formSheet` are `UISheetPresentationController`-backed and differ in size, not in how a child popover attaches. Worth noting that #1111 already shipped a speculative fix for #908 and the issue is still open; a symbolicated crash report would be worth more here than a third guess. Separately, `.datePickerStyle(.compact)` looks like a no-op — `.automatic` already resolves to compact for a `[.date]` picker on iOS — so the doc's claim that it stops the picker "expand[ing] into a graphical calendar" doesn't hold either.

https://github.com/OneBusAway/onebusaway-ios/blob/f2a7198339be507724cb6a144efd731c755eaa92/OBAKitTests/Schedules/ScheduleSheetPresentationTests.swift#L19-L27

OBAKit/Schedules/ScheduleForStopViewModel.swift
```diff
@@ -110,6 +110,7 @@ class ScheduleForStopViewModel: ObservableObject {
 
         do {
             let response = try await apiService.getScheduleForStop(stopID: stopID, date: selectedDate)
+            guard !Task.isCancelled else { return }
```
aaronbrethorst: This guard can only run when the `await` **completed**. When cancellation is what interrupts the await, `URLSession` throws `URLError(.cancelled)` and control jumps straight to the `catch` below, which sets `self.error`. So the guard doesn't actually suppress the cancellation path it's aimed at.

That matters because `error` is only cleared at the top of `fetchSchedule()`. Sequence on a date change:

1. `.task(id: selectedDate)` cancels fetch A
2. fetch B starts, sets `error = nil`, `isLoading = true`
3. fetch A's `URLError(.cancelled)` lands in `catch` → `self.error = <cancelled>`
4. fetch A's `defer` fires → `isLoading = false`, spinner drops while B is still loading
5. B succeeds and sets `scheduleData`, but `error` is still non-nil

Since `content` checks `else if let error` before `else if scheduleData != nil`, the sheet can sit on "Unable to load schedule" over a perfectly good payload.

Suggest filtering cancellation explicitly rather than guarding after the fact:

```swift
} catch is CancellationError {
    return
} catch let urlError as URLError where urlError.code == .cancelled {
    return
} catch {
    self.error = error
}
```

and making the `isLoading = false` conditional on not having been cancelled. Same applies in `ScheduleForRouteViewModel`.

OBAKitTests/Schedules/ScheduleSheetPresentationTests.swift
```diff
@@ -0,0 +1,27 @@
...
+/// `UIHostingController` (#908, Link light rail schedule). Phone keeps
+/// `.pageSheet`; iPad uses `.formSheet`.
+@Suite(.serialized)
+struct ScheduleSheetPresentationTests {
+
+    @Test func `Phone presents the schedule as a page sheet`() {
+        #expect(ScheduleSheetPresentation.modalStyle(for: .phone) == .pageSheet)
+    }
+
+    @Test func `iPad presents the schedule as a form sheet`() {
```
aaronbrethorst: These two tests assert the ternary inside `modalStyle(for:)` — `idiom == .pad ? .formSheet : .pageSheet` — which is the implementation restated rather than a behavior locked down.

Nothing here reaches `ScheduleViewControllers.swift:25` / `:45`, so if someone reverted those call sites to `.pageSheet` tomorrow, both tests would still pass and the crash fix would be silently gone. A test that asserted the hosting controllers actually receive the right `modalPresentationStyle` would be the real regression lock.

Worth noting this is also the only verification in the PR — both device-verification boxes in the test plan are unchecked, and given #1111's fix for the same issue didn't hold, a device confirmation is what would carry weight here.

aaronbrethorst (changes requested):
Thanks for taking another run at #908 — it's a nasty one and I appreciate you going after it. Before this lands I want to slow down on the approach, because I think we're about to guess a third time.

#1111 already shipped a fix for this exact crash back in March, with a completely different mechanism ("Combine-spawned Tasks survived view teardown, causing a race condition"). I approved it saying it fixed a real crash. #908 is still open. This PR now proposes a third theory — that `.pageSheet` is the culprit and `.formSheet` isn't — and I don't think the evidence supports it yet:

- On iPad both `.pageSheet` and `.formSheet` are backed by `UISheetPresentationController`. They differ in size, not in how the picker is attached or presented, so it isn't obvious why one would crash and the other wouldn't.
- `.datePickerStyle(.compact)` is very likely a no-op here: `.automatic` already resolves to compact for a `[.date]` picker on iOS.
- `docs/ipad-schedule-crash.md` states the mechanism as settled fact, but there's no stack trace behind it. If the theory is wrong, that file will send the next person down the same path.

Could we get a symbolicated crash report on the issue first? Two speculative fixes in and still open is the signal to stop guessing and get the trace. I'd rather leave #908 open with a real diagnosis attached than close it a second time on a plausible-sounding change.

Separately, there is a **concrete, verifiable bug** in here worth fixing on its own merits, independent of the crash — flagged inline. The `Task.isCancelled` guard sits after the `await`, so it never runs when cancellation is what *interrupts* the await. `URLSession` throws `URLError(.cancelled)` in that case, which lands in your `catch` and sets `self.error`. Since `error` is cleared only at the top of `fetchSchedule()`, a rider changing the date can end up staring at "Unable to load schedule" even though the replacement fetch succeeded. The unconditional `defer { isLoading = false }` has the matching problem — the cancelled task's defer fires after the new task set `isLoading = true`, dropping the spinner mid-load.

That one is real, reproducible by reasoning, and worth its own PR even if the sheet-style change goes away.

Two smaller notes:

- `.formSheet` is a genuine iPad UX regression for the route timetable — `GridTimetable` is a wide multi-column grid, and `.formSheet` pins it to a fixed ~540pt box where `.pageSheet` gave it much more room. That's a real cost to pay for an unproven benefit.
- `UIDevice.current.userInterfaceIdiom` — the two other idiom checks in the repo (`MoreViewController.swift:430`, `AgenciesViewController.swift:95`) use `traitCollection.userInterfaceIdiom`, which is the multi-window-correct form.

I did confirm there's no iPhone behavior change: `.phone` still maps to `.pageSheet`, and the cancellation guard is the only line that touches the phone path.

Please take a look at the inline comments.

OBAKit/Schedules/ScheduleForStopViewModel.swift
Borisserz: Cancellation now filters `CancellationError` and `URLError.cancelled` in `catch`, and each fetch carries a generation so a cancelled task cannot clear `isLoading` or set `error` after a newer fetch starts. Tests cancel mid-`fetchSchedule()` on both view models; RED would have left `error != nil`.

OBAKitTests/Schedules/ScheduleSheetPresentationTests.swift
Borisserz: Dropped `ScheduleSheetPresentation` and the ternary tests. Both hosting controllers are back to `.pageSheet` on every idiom until we have a symbolicated trace for #908.

aaronbrethorst:
### Code review

Found 2 issues:

1. `ScheduleFetchError.isCancellation` duplicates the existing `Error.isCancellation` helper in `OBAKitCore/Extensions/FoundationExtensions.swift:296`, which is `public`, already used from OBAKit in `StopViewModel.swift:319`, `TripViewModel.swift:120`, `MapRegionManager.swift:1528`/`:1721` and `DestinationStopPickerController.swift:162`, and carries a doc comment describing exactly the same two shapes. The whole new file can go; `guard !error.isCancellation else { return }` reads the same and matches the established idiom. The existing helper is also slightly broader — it matches via the `NSError` bridge (`domain == NSURLErrorDomain && code == NSURLErrorCancelled`) specifically so it covers Objective-C `NSError`s as well as Swift `URLError` values, which matters here because `APIService.data(for:)` catches `let error as NSError` and rethrows the bridged value.

https://github.com/OneBusAway/onebusaway-ios/blob/5cd61334ea7bdd96eafed1643aea9e770eea56b3/OBAKit/Schedules/ScheduleFetchError.swift#L11-L20

2. The PR body still says `Closes #908`, but this revision no longer touches the crash at all. `ScheduleViewControllers.swift` is back to `.pageSheet` on both hosting controllers (the only change left in that file is a deleted comment), and `ScheduleSheetPresentation`, `.datePickerStyle(.compact)` and `docs/ipad-schedule-crash.md` are all gone. That is the right call — and it matches the note left inline ("back to `.pageSheet` on every idiom until we have a symbolicated trace for #908") — but merging as-is will auto-close #908, which is precisely what the trace is still needed for. The title ("Stop the iPad schedule sheet from crashing on DatePicker") and the Summary/Test-plan sections describing `.formSheet`, `ScheduleSheetPresentation.modalStyle(for:)` and `ScheduleSheetPresentationTests` no longer describe any change in the diff either. Retitle around the cancellation fix and drop the `Closes` line (`Refs #908` instead).

https://github.com/OneBusAway/onebusaway-ios/blob/5cd61334ea7bdd96eafed1643aea9e770eea56b3/OBAKit/Schedules/ScheduleViewControllers.swift#L23-L26

Borisserz:
Addressed both: schedule cancels go through `Error.isCancellation` (deleted `ScheduleFetchError`); retitled and switched to `Refs #908` so we do not auto-close the crash without a trace.

aaronbrethorst:
### Code review — correction

My previous comment on this PR was wrong, and I'm retracting both findings. It reviewed `5cd6133`, but you'd already pushed `77cf473` ("Use shared Error.isCancellation for schedule fetch cancels"), which deletes `ScheduleFetchError.swift` outright and routes both view models through the existing helper. That was exactly the fix I asked for, and you'd made it before I commented.

The second finding was also wrong on its own terms: the body says `Refs #908.`, not `Closes #908.` — so nothing auto-closes. You'd handled that too.

Sorry for the noise. No outstanding issues from me on the current head.

aaronbrethorst (approved):
Read the diff. This is the right fix and it's the one I asked for: the generation counter plus explicit cancellation filtering replaces the unconditional `defer { isLoading = false }`, so a cancelled fetch can no longer clear a newer fetch's spinner or strand an error. Using the existing `Error.isCancellation` helper rather than a new type is right too — and thank you for withdrawing the presentation-style change and leaving #908 open with `Refs` instead of closing it on a guess. That was the correct call.

I checked the tests aren't vacuous: `MockDataLoader` throws `URLError(.cancelled)` when `Task.isCancelled`, so removing the filter genuinely turns them red.

PR #1314, opened 2026-08-24
Address trip-sharing follow-ups from #1120

diveshpatil9104:
Three of the four follow-ups named at the #1120 merge. The fourth (route-grouped mode has no context menu) is its own issue — it costs Show Trip, Alarm, Schedule and Bookmark too, not just Share Trip.

**One `Strings.shareTrip`.** `stop_controller.share_trip` and `stop_page.row.share_trip` are gone. Both call sites now use the existing `Strings.shareTrip` (`common.share_trip`), which was already in OBAKitCore, already translated in all 13 catalogs, and referenced by nothing. Two keys deleted, none added.

**`.error` gets a share escape.** The destination is optional, so a failed fetch shouldn't block sharing. It's a right bar button item reusing the empty state's string and the same `destinationStopPickerDidSkipDestination` call.

Two calls worth your eye:

- `StandardEmptyDataViewModel` holds one `buttonConfig` and `.error` has to keep Try Again, so the escape went in the nav bar rather than the body. Adding a secondary-button slot would change an app-wide control. Happy to move it if you'd rather have it there.
- It shows in `.error` only, not `.empty`. `.empty` already offers that same action as its body button, and the identical control twice on one screen reads like a bug.

**`TripSharingCoordinator` tests.** The `destination_stop_id` encoding through `AppLinksRouter.encode` — the existing deep-link tests build `URLComponents` by hand and never touch it — plus both `presentShareError()` guards and the dismiss-then-present ordering.

While consolidating the keys I found none of #1120's strings had reached the catalogs, so every locale was falling back to English. All 12 now land in all 13 with real translations.

## Testing

Full suite: 2318 tests, one pre-existing unrelated failure (`RentalFormatTests.rangeFallbackUsesAbbreviatedUnits`). SwiftLint clean in the changed files.

Refs #1120, #449.

## Summary by CodeRabbit

* **New Features**
  * Added an option to share a trip without selecting a destination when destination lookup fails.
  * Added loading, empty-state, retry, accessibility, and trip-sharing error messages across supported languages.
  * Improved trip-sharing error alerts when trip-link preparation fails.

* **Bug Fixes**
  * Standardized “Share Trip” labels across stop and departure menus.

* **Tests**
  * Added coverage for destination selection states, sharing behavior, URL handling, cancellation, and failure scenarios.

### DIFF
--- a/OBAKit/Stops/StopPage/Departures/DepartureRowView.swift
+++ b/OBAKit/Stops/StopPage/Departures/DepartureRowView.swift
@@ -243,7 +243,7 @@ extension View {
                     Label(Strings.addBookmark, systemImage: "bookmark")
                 }
                 Button(action: actions.onShareTrip) {
-                    Label(OBALoc("stop_page.row.share_trip", value: "Share Trip", comment: "Context menu action that shares the trip after choosing a destination stop"), systemImage: "square.and.arrow.up")
+                    Label(Strings.shareTrip, systemImage: "square.and.arrow.up")
                 }
             }, preview: {
                 actions.makePreview()
--- a/OBAKit/Stops/StopViewController.swift
+++ b/OBAKit/Stops/StopViewController.swift
@@ -928,7 +928,7 @@ public class StopViewController: UIViewController,
                 actions.append(schedule)
             }
 
-            let shareTrip = UIAction(title: OBALoc("stop_controller.share_trip", value: "Share Trip", comment: "Context menu button that allows the user to share their trip status."), image: Icons.share) { [weak self] _ in
+            let shareTrip = UIAction(title: Strings.shareTrip, image: Icons.share) { [weak self] _ in
                 self?.shareTripStatus(viewModel: viewModel)
             }
             actions.append(shareTrip)
--- a/OBAKit/Strings/ar.lproj/Localizable.strings
+++ b/OBAKit/Strings/ar.lproj/Localizable.strings
@@ -217,6 +217,36 @@
 /* Button title to let the user view a data migration report. */
 "data_migration_bulletin.view_report_button" = "عرض التقرير";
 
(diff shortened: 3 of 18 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Every follow-up from #1120 is addressed substantively rather than nominally, which is what I was checking for.

- The duplicate "Share Trip" keys now both route through `Strings.shareTrip` / `common.share_trip`, which exists in all 13 OBAKitCore catalogs with real translations. The two dead keys were never in any `.strings` file, so nothing is orphaned — and non-English locales that used to fall back to English here no longer do.
- The `.error` share escape is wired to the same `destinationStopPickerDidSkipDestination` the empty state's button calls, driven from `state.didSet` and seeded in `viewDidLoad`. I checked `ViewRouter.buildNavigation` sets no bar button items, so nilling `rightBarButtonItem` in the other states clobbers nothing.
- The `TripSharingCoordinator` suite covers all three named gaps. The encoding tests go through the real `AppLinksRouter.encode` with hardcoded expectations, and the ordering test parks the dismiss completion rather than running it inline — so it genuinely discriminates against a present-alongside-dismiss implementation rather than passing either way. That's the detail that makes it a real test.
- Route-grouped mode correctly deferred, with the same reasoning I gave.

I verified the localization independently rather than trusting the diff: ru, pl, zh-Hans and ar all carry genuine translations for the trip-sharing keys, not English pasted into every file.

Filed #1348: "Share Without Destination" is 31 characters in ru and 27 in pl, sitting opposite a "Cancel" and under a "Select Destination" title — that bar will crowd or truncate on a narrow device. You'd flagged the `StandardEmptyDataViewModel` single-`buttonConfig` constraint that pushed it there; either lifting that constraint or shortening the bar title would do it.

PR #1316, opened 2026-08-24
Make the whole onboarding region row tappable

### LINKED ISSUE #1315: I can't choose a different region on the "Your region" setup page
# Expected

I can tap on Davis, Puget Sound, or Tampa and the selection will change.

# Actual

Nothing happens when I tap on any of those three rows.

Image

diveshpatil9104:
### Summary
The rows in "Or choose another" are plain-styled buttons with no .contentShape, so only the region name text was tappable — anywhere else in the row did nothing. Added .contentShape(.rect), the same thing SearchListRowView does for its rows.
One extra thing: the card always said "Detected near you", even for a region you picked yourself. You only notice once tapping works, so I fixed it here — new onboarding.region.selected_label ("Your selection"), translated in all 13 catalogs. Say the word if you'd rather I split it out.
 Closes #1315.

### Video

https://github.com/user-attachments/assets/7253ce66-9413-4ee2-bf8e-5ed8970a50ca

## Summary by CodeRabbit

* **New Features**
  * Region cards now indicate whether a region was automatically detected or manually selected.
  * Alternate-region options respond when tapping anywhere across the full row.

* **Localization**
  * Added translated labels for manually selected regions across supported languages.

### DIFF
--- a/OBAKit/Onboarding/Views/OnboardingRegionView.swift
+++ b/OBAKit/Onboarding/Views/OnboardingRegionView.swift
@@ -57,7 +57,7 @@ struct OnboardingRegionView<Provider: RegionProvider>: View {
         ) {
             VStack(spacing: 0) {
                 if let region = resolved {
-                    selectedCard(for: region)
+                    selectedCard(for: region, isDetected: region.id == detected?.id)
                         .padding(.top, 22)
                 }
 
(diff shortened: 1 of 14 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Both halves of this are good. `.contentShape(.rect)` as the last modifier — after `.frame(maxWidth: .infinity)`, the padding, and the 48pt height — is the correct placement to cover the full row rather than the text glyphs, and it's verbatim the pattern already shipped in `SearchListRowView.swift:49`.

I checked the accessibility angle since making a whole row tappable can flatten a11y structure: it doesn't here. The rows were already single elements (a `Button` wrapping one `Text`), and `contentShape` doesn't alter the a11y frame. The card footer keeps its `.accessibilityElement(children: .combine)` + `.isSelected` grouping.

I also verified the localization properly rather than assuming: all 13 locales got real translations for `onboarding.region.selected_label` — ar "اختيارك", ru "Ваш выбор", ko "내가 선택함", zh-Hans "你的选择", vi "Lựa chọn của bạn" — with none of them silently carrying the English string, and `en.lproj` sits in the correct alphabetical slot with the `comment:` text intact. That's the thing that most often goes wrong in a 13-locale change and it's clean here.

The "Your selection" vs "Detected near you" fix is scope creep, but it's the right scope creep — the mislabel was real.

Filed #1345 for what's left: `detectedRegion` can still say "Detected near you" for a region that came from `fixedRegionName` or the sole-region fallback, and the screen's body text now contradicts the new eyebrow.

PR #1317, opened 2026-08-25
 Offer Try Again when Current Trip has no location

diveshpatil9104:
### Summary

` .noResults `and .error both offer retryButtonConfig; `.noLocation` didn't, so the only way to re-run the search after granting location access was to leave the screen and come back. Added the same button — `findVehicle()` re-reads currentLocation, so it recovers properly.
Left `.noRealtime` alone since retrying can't fix a route with no real-time data
Follow-up from your review on #1066.

## Summary by CodeRabbit

* **New Features**
  * Added a retry button to the no-location state, allowing riders to try again after enabling location access in Settings.

### DIFF
--- a/OBAKit/Trip/CurrentTripViewController.swift
+++ b/OBAKit/Trip/CurrentTripViewController.swift
@@ -134,10 +134,15 @@ class CurrentTripViewController: UIViewController,
             ))
 
         case .noLocation:
+            // Recoverable like `.noResults` and `.error`: the rider can grant
+            // location access in Settings and come back, and `findVehicle()`
+            // re-reads `currentLocation` on the way in. Without the button the
+            // only way to re-run it is to leave the screen and return.
             return .standard(.init(
                 alignment: .center,
                 title: noLocationTitle,
-                body: nil
+                body: nil,
+                buttonConfig: retryButtonConfig
             ))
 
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Good catch — this was an inconsistency, not a deliberate omission. `git blame` shows `.noResults` and `.error` got `retryButtonConfig` in the same original commit and `.noLocation` just didn't, same author, same sitting.

I verified the mechanics: `retryButtonConfig`'s action calls `viewModel.findVehicle()`, which re-reads `application.locationService.currentLocation` on every invocation, so the button genuinely can recover. And `findVehicle(resetState: true)` sets `.loading` first, so a tap with permission still denied visibly flips to "Finding your vehicle…" and back rather than silently doing nothing — no stuck spinner, no loop.

One thing for the comment when you next touch this file: the claim that "without the button the only way to re-run it is to leave the screen and return" isn't accurate — `startRefreshTimer()` already re-runs `findVehicle(resetState: false)` every 20 seconds. The button's real justification is better than that: `shouldSkipProgrammaticRefresh` disables that timer under VoiceOver, so for VoiceOver users this button is the only in-screen recovery. Filed as #1343.

PR #1318, opened 2026-08-25
Show weather condition alongside temperature on the map

Borisserz:
## Summary
- The map weather control was temperature-only (`71°`). It now shows the condition SF Symbol plus that temperature on both the UIKit hover bar and the SwiftUI map-panel pill.
- VoiceOver value is temperature + condition (`71°, Clear`). The visible title stays compact — a phrase like "Partly Cloudy" will not fit the 42pt hover bar.
- Hour-by-hour forecast is unchanged (`WeatherDetailPopup`). The UIKit button is updated through `UIButton.Configuration` only; `setTitle`/`setImage` are ignored once a configuration exists.

Closes #71.

## Test plan
- [x] `WeatherDisplayTests` (15) — `buttonTitle` stays `71°`; `buttonAccessibilityValue` contains `71°` and `Clear`; fixture icon maps to `sun.max.fill`
- [x] `WeatherFormatterTests` (14)
- [ ] On device, map hover bar: condition icon stacked above the temperature; tap still opens the hourly popup
- [ ] VoiceOver: "Show Weather Forecast", value includes both temperature and condition
- [ ] Map-panel experience: same icon + temp on the floating pill

## Summary by CodeRabbit

* **Enhancements**
  * Updated the map weather button with a weather icon, bold temperature display, and improved layout.
  * Weather conditions are now available through VoiceOver while the button visibly shows the temperature.
  * Added a fallback for unknown weather conditions.

* **Documentation**
  * Documented the weather button’s icon, temperature presentation, accessibility behavior, and distinction from hourly weather details.

* **Tests**
  * Added coverage confirming temperature, weather condition, and weather icon accessibility behavior.

PR #1319, opened 2026-08-25
Add Directions to/from Here on the stop page

### LINKED ISSUE #883: Improve ways to discover the trip planner inputs
There doesn't seem to be a way to discover the trip planner and its origin/destination inputs without searching for an address first or long-clicking on the map, which isn't obvious.  Other methods could be added after searching for a stop ID, for example, or clicking on a stop on the map and begin with "Directions to here/from here".

Borisserz:
## Summary
- The trip planner was only reachable from search or a long-press pin. The stop Location menu (legacy, Stop page, and toolbar) now offers **Directions to Here** / **Directions from Here** when OTP is on and the rider has not disabled trip planning for the region.
- Those actions pop to the map tab and open the existing planner with the stop prefilled as destination or origin. Walking directions (Apple / Google Maps) are unchanged.
- Rows stay hidden in the experimental map-panel shell: that UI has no `MapViewController` to present on. `showTripPlanner` takes an optional `origin` so current location is not overwritten on “from here.”

Closes #883.

## Test plan
- [x] `TripPlannerEndpointsTests` (6) — stop → `MKMapItem`; explicit origin wins over GPS; nil destination stays nil
- [x] `StopTripPlannerActionTests` (3) — available when OTP is on; hidden when disabled / no OTP / no classic root
- [x] `StopPageActionPresenterTests` — presenting with `rootController == nil` does not crash
- [ ] Puget Sound, classic map: open a stop → Location menu → Directions to Here; planner opens with the stop as destination
- [ ] Same stop: Directions from Here; origin is the stop, destination empty
- [ ] Region with trip planning off (or no OTP): those rows are absent; walking directions still work

## Summary by CodeRabbit

- **New Features**
  - Added “Directions to Here” and “Directions from Here” actions to stop menus when trip planning is available.
  - Open trip planning with a selected stop as either the origin or destination.
  - Preserve current-location fallback when no origin is specified.
  - Added safeguards for unsupported regions and unavailable map presentation contexts.

- **Documentation**
  - Documented trip planning from stops, including map-tab behavior and map-panel limitations.

- **Bug Fixes**
  - Prevented crashes when trip planning cannot be presented.

### DIFF
--- a/OBAKit/TripPlanning/StopTripPlannerAction.swift
+++ b/OBAKit/TripPlanning/StopTripPlannerAction.swift
@@ -53,6 +53,9 @@ enum StopTripPlannerAction {
 
         let mapController = rootController.mapController
         mapController.navigationController?.popToRootViewController(animated: false)
+        // Map-pin stops live in `stopSheet`, not the nav stack. Same teardown
+        // as opening a map item or another panel (#883).
+        mapController.dismissStopSheetForReplacement()
 
         let stopMapItem = TripPlannerEndpoints.mapItem(from: stop)
         switch action {
--- a/OBAKit/TripPlanning/StopTripPlannerAction.swift
+++ b/OBAKit/TripPlanning/StopTripPlannerAction.swift
@@ -1,0 +1,47 @@
+//
+//  StopTripPlannerAction.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import OBAKitCore
+
+/// Stop-page trip-planner menu actions. Menus and toolbars can gate on
+/// `isAvailable` without instantiating a view controller.
+enum StopTripPlannerAction {
+    /// Prefill destination as the stop; origin stays current location.
+    case directionsToStop
+    /// Prefill origin as the stop; destination left empty for the rider to pick.
+    case directionsFromStop
+
+    /// `true` when OTP trip planning is running for the current region and the
+    /// rider has not disabled it for that region.
+    static func isAvailable(application: Application) -> Bool {
+        guard application.features.tripPlanning == .running,
+              let region = application.regionsService.currentRegion
+                ?? application.currentRegion,
+              application.userDataStore.isTripPlanningEnabled(for: region) else {
+            return false
+        }
+        return true
+    }
+
+    static var directionsToHereTitle: String {
+        OBALoc(
+            "stops_controller.directions_to_here",
+            value: "Directions to Here",
+            comment: "Stop Location menu action that opens the trip planner with this stop as the destination."
+        )
+    }
+
+    static var directionsFromHereTitle: String {
+        OBALoc(
+            "stops_controller.directions_from_here",
+            value: "Directions from Here",
+            comment: "Stop Location menu action that opens the trip planner with this stop as the origin."
+        )
+    }
+}
--- a/OBAKit/TripPlanning/StopTripPlannerAction.swift
+++ b/OBAKit/TripPlanning/StopTripPlannerAction.swift
@@ -10,7 +10,7 @@
 import OBAKitCore
 
 /// Stop-page trip-planner menu actions. Menus and toolbars can gate on
-/// `isAvailable` without instantiating a view controller.
+/// `canPresent` without instantiating a view controller.
 enum StopTripPlannerAction {
     /// Prefill destination as the stop; origin stays current location.
     case directionsToStop
@@ -18,7 +18,8 @@ enum StopTripPlannerAction {
     case directionsFromStop
 
     /// `true` when OTP trip planning is running for the current region and the
-    /// rider has not disabled it for that region.
+    /// rider has not disabled it for that region. Does not imply the classic
+    /// map tab can present the planner — see `canPresent`.
     static func isAvailable(application: Application) -> Bool {
         guard application.features.tripPlanning == .running,
               let region = application.regionsService.currentRegion
@@ -27,6 +28,39 @@ enum StopTripPlannerAction {
             return false
         }
         return true
+    }
+
+    /// Hide the stop-page actions unless the classic tab root can host
+    /// `MapViewController.showTripPlanner`. Map-panel mode leaves
+    /// `viewRouter.rootController` nil; showing the rows there would be a
+    /// dead button.
+    static func canPresent(application: Application) -> Bool {
+        isAvailable(application: application) && application.viewRouter.rootController != nil
+    }
+
+    /// Pops to the map tab and opens the existing trip planner. No-ops (with
+    /// a log) when `canPresent` is false.
+    static func present(_ action: StopTripPlannerAction, stop: Stop, application: Application) {
+        guard canPresent(application: application),
+              let rootController = application.viewRouter.rootController else {
+            if isAvailable(application: application) {
+                Logger.error("StopTripPlannerAction: present dropped — no classic root controller (map-panel mode is active)")
+            }
+            return
+        }
+
+        application.viewRouter.rootNavigateTo(page: .map)
+
+        let mapController = rootController.mapController
+        mapController.navigationController?.popToRootViewController(animated: false)
+
+        let stopMapItem = TripPlannerEndpoints.mapItem(from: stop)
+        switch action {
+        case .directionsToStop:
+            mapController.showTripPlanner(destination: stopMapItem)
+        case .directionsFromStop:
+            mapController.showTripPlanner(origin: stopMapItem, destination: nil)
+        }
     }
 
     static var directionsToHereTitle: String {
--- a/OBAKit/Mapping/MapViewController.swift
+++ b/OBAKit/Mapping/MapViewController.swift
@@ -579,45 +579,36 @@ class MapViewController: UIViewController,
 
     /// Presents the trip planner.
     /// - Parameters:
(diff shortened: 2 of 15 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. `StopTripPlannerAction.present` pops the map tab's navigation stack but never dismisses the classic map's stop *sheet*, which is the primary way this feature will be reached. On the classic map a stop opens as a `StopSheetPresenter` FloatingPanel child of `MapViewController` (`MapViewController.show(stopID:)` → `stopSheet.present(stopPageVC, from: self)`), not as a pushed view controller — so `mapController.navigationController?.popToRootViewController(...)` does not clear it. The trip-planner semi-modal is then added on top of a stop sheet that is still presented, and `StopSheetPresenter` keeps the tab bar hidden until *its* dismissal runs. Every other in-file caller of `showTripPlanner` tears its presenting sheet down first — the map-item "Plan Trip" handler calls `dismissExistingMapItemController(animated: true)` immediately before `showTripPlanner(destination:)`, and `showSemiModalPanel` calls `dismissStopSheetForReplacement()` before `addPanel(toParent:)`. `MapViewController` already exposes `dismissStopSheetForReplacement()` for exactly this (its neighbouring comment notes "the stop sheet and the semi-modal panels are FloatingPanel *children* of this controller, not modal presentations"), so this looks like a one-line omission rather than a design choice. Note the "Puget Sound, classic map: open a stop → Location menu → Directions to Here" box in the test plan is still unchecked.

https://github.com/OneBusAway/onebusaway-ios/blob/98997526651f26bee68de6af70e9620f32fcdc22/OBAKit/TripPlanning/StopTripPlannerAction.swift#L53-L64

OBAKit/TripPlanning/StopTripPlannerAction.swift
```diff
@@ -0,0 +1,81 @@
...
+            if isAvailable(application: application) {
+                Logger.error("StopTripPlannerAction: present dropped — no classic root controller (map-panel mode is active)")
+            }
+            return
+        }
+
+        application.viewRouter.rootNavigateTo(page: .map)
+
+        let mapController = rootController.mapController
+        mapController.navigationController?.popToRootViewController(animated: false)
```
aaronbrethorst, line 55: This handles the pushed stop page but not the sheet-presented one, which is the primary flow for #883.

When the rider taps a stop pin on the classic map, `MapViewController` presents the stop page via `stopSheet.present(stopPageVC, from: self)` (`MapViewController.swift:855`) — a FloatingPanel child of the map controller, not a nav-stack push. `popToRootViewController` doesn't dismiss it, and neither does `showTripPlanner` (I checked: it moves `floatingPanel` to `.tip` and adds a semi-modal panel, but never touches `stopSheet`).

Result: the trip planner opens with the stop sheet still stacked behind it, and the tab bar stays hidden because `StopSheetPresenter.hostTabBarController` is still holding it down.

Suggest matching the existing map-item precedent and adding the teardown before presenting:

```swift
mapController.navigationController?.popToRootViewController(animated: false)
mapController.dismissStopSheetForReplacement()
```

`dismissStopSheetForReplacement()` is already `internal` on `MapViewController` at line 1064 and does exactly this (`stopSheet.dismiss(animated: false)` plus cancelling the scheduled feedback prompt), so it should be a one-line change.

aaronbrethorst (changes requested):
This is well put together. Extracting `TripPlannerEndpoints` out of `showTripPlanner` rather than duplicating the Stop→`MKMapItem`→OTPKit conversion is the right instinct, and threading `origin:` through so "from here" doesn't get clobbered by GPS is exactly the detail that would have been easy to miss. I also checked the availability gating end to end: `Region.supportsOTP` is defined as `openTripPlannerGraphQLURL != nil || openTripPlannerURL != nil`, which is precisely the condition under which `buildTripPlanner` returns non-nil, so the gate and the presentation genuinely agree — there's no path where the menu item appears and then dead-ends. Retain semantics in the new closures match the surrounding code too.

One blocker, flagged inline: the flow the issue is actually about — tap a stop pin on the map, then "Directions to Here" — leaves the stop sheet behind the trip planner.

`present` calls `popToRootViewController`, which covers the *pushed* stop page. But on the classic map a pin tap presents `StopPageViewController` through `StopSheetPresenter` as a FloatingPanel **child** of `MapViewController` (`MapViewController.swift:855`), not on the nav stack, so popping doesn't touch it. I checked whether `showTripPlanner` tears it down itself and it doesn't — it moves `floatingPanel` to `.tip` and adds a semi-modal, but never touches `stopSheet`. The sheet stays stacked underneath, and `StopSheetPresenter.hostTabBarController` keeps the tab bar hidden until it's separately dismissed.

The codebase already has the pattern for this: the map-item Plan Trip handler does `dismissExistingMapItemController(animated: true)` before `showTripPlanner` (`MapViewController.swift:1273-1275`), and the helper you'd want, `dismissStopSheetForReplacement()`, is right there at `MapViewController.swift:1064`. Should be about one line.

Worth actually walking the manual test-plan boxes on this one — the unchecked item is the exact flow that breaks.

Two minor things, neither blocking:

- `StopTripPlannerAction.isAvailable`'s `application.regionsService.currentRegion ?? application.currentRegion` — the `features.tripPlanning == .running` guard above already returns `.off` when `regionsService.currentRegion` is nil, so the `??` branch is unreachable.
- `waitForRegion`'s doc comment says `fixedRegionName` selects the region during Application init, but the third test builds its own `AppConfig` without it and seeds `OBACurrentRegionIdentifierUserDefaultsKey` instead.

On the two new `OBALoc` keys not being in any `.strings` file — I'm deliberately not holding you to that, since roughly 80 keys already on `main` have the same gap. It's a repo-wide `scripts/extract_strings` sweep, not yours to fix here.

Please take a look at the inline comment; happy to re-review as soon as the sheet teardown is in.

aaronbrethorst:
Heads up: I merged a batch of PRs just now and this branch picked up conflicts — #1325 moved the trip planner out of `MapViewController` into a `MapViewController+TripPlanner.swift` extension, and #1318 also touched `MapViewController.swift`.

That's a rebase cost, not anything wrong with your work. Since you're already coming back for the stop-sheet dismissal I flagged, please merge `main` in and resolve while you're there. The extraction may actually help you — `showTripPlanner` now lives in the extension alongside the rest of the trip-planner code.

OBAKit/TripPlanning/StopTripPlannerAction.swift
Borisserz: `present` now calls `dismissStopSheetForReplacement()` after `popToRoot`, same teardown as a map-item panel. Hosted test puts a sheet on `MapViewController.stopSheet`, runs Directions to Here, and asserts the sheet is gone and the tab bar is back. RED: sheet stayed up; GREEN: 4/4.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
The blocker is fixed exactly as specified — `dismissStopSheetForReplacement()` right after `popToRootViewController` — and it comes with a hosted regression test that stages a sheet, runs the action, and asserts both the sheet teardown and the tab bar returning. That is the right kind of test for it.

I also chased the "isAvailable omits region.supportsOTP" concern that was raised and it is a false positive: `Application.FeatureAvailability.tripPlanning` already returns `.off` unless `currentRegion != nil && region.supportsOTP`, which is precisely what `buildTripPlanner` needs, so the gate and `showTripPlanner`'s guard agree.

Two cosmetic leftovers, neither worth another round trip: the dead `?? application.currentRegion` branch in `StopTripPlannerAction.isAvailable`, and `waitForRegion`'s doc comment still describing `fixedRegionName` when the third test seeds the defaults key instead.

One thing to know: the `OBAKitTests` workflow last ran on your *first* commit and has not run on the current head, so there is no CI signal on the sheet-dismissal commit or its new test. The rebase below will trigger one.

Borisserz:

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Good feature, and the refactor earns its keep — pulling the `Stop`/`MKMapItem`/`CLLocation` conversions out of `showTripPlanner` into `TripPlannerEndpoints` makes both testable, and `StopTripPlannerAction` owning the availability gate means menus can ask without instantiating a view controller.

I checked the gating rather than taking it on faith: `isAvailable` is `features.tripPlanning == .running` (which is exactly `currentRegion != nil && region.supportsOTP`) ANDed with `isTripPlanningEnabled(for:)` — provably the same condition `showTripPlanner` guards on itself, so a visible row can't lead to a dead tap. Worth saying explicitly since CodeRabbit flagged the opposite ("in regions without OTP support, a Directions action can dismiss the stop sheet without opening the trip planner") — that's incorrect, OTP support is a precondition of `.running`.

The sheet-dismissal fix from the last round is right too. `showTripPlanner` builds its panel via `createSemiModalPanel` + `addPanel(toParent:)` directly rather than through `showSemiModalPanel`, so it never hit the existing `dismissStopSheetForReplacement` path and the map-pin sheet really was left up. The added call runs the sheet's `onDismiss` synchronously, so the bookkeeping (stopSheetStopID, route focus, tab bar) all unwinds.

On the red build: that was `StopSheetPresenterTests / "Dismiss slides the sheet straight down at the size it had"`, and it was a flake — I re-ran the same SHA and it went green. Nothing for you to chase.

One dead branch worth tidying whenever you're next in there: the `?? application.currentRegion` fallback in `isAvailable` can't be reached, since `.running` already requires `regionsService.currentRegion` to be non-nil.

PR #1321, opened 2026-08-27
Use … for overflowing map stop route labels

### LINKED ISSUE #514: Extra Bus Route Consistency (UX)

Borisserz:
## Summary
- Map pin labels already used `formattedMapRoutes` (`"10, 12, 49..."`). The callout still used `Stop.subtitle` (`"Routes: 10, 12, 49"` with every route), so the two map surfaces disagreed — the leftover of #514 after #1267.
- Overflow now uses a single ellipsis glyph (`…`, U+2026) inside `OBALoc`, matching UIKit's own truncation mark. Three ASCII periods next to UIKit's `…` was a new flavor of the same inconsistency.
- `Stop.mapCalloutText` is the pin callout. `Stop.subtitle` is unchanged: Home / Recent / Nearby still list every route.

Closes #514.

## Test plan
- [x] `Formatted map routes over the limit appends ellipsis` — `"10, 20, 30…"`, no `"..."`, no `"more"`
- [x] `Map callout uses the overflowing map route list, not plus-more` — callout is `"#code\n10, 20, 30…"`; `subtitle` still has `"Routes:"` and `"62"`
- [x] `FormattersTests` (15)
- [ ] On device: stop with 4+ routes — pin label and callout both end in `…`; Home row still lists every route

## Summary by CodeRabbit

- **New Features**
  - Map stop callouts now show the stop code alongside a concise list of route names.
  - Route lists display up to three routes, with additional routes indicated by a localized ellipsis.

- **Improvements**
  - Map route labels now use the proper Unicode ellipsis character and support localized overflow text.
  - Stop subtitles and non-map route lists continue to display complete route information.

- **Documentation**
  - Added guidance describing route-label behavior across map pins, callouts, subtitles, and list views.

### DIFF
--- a/OBAKit/Extensions/ModelExtensions.swift
+++ b/OBAKit/Extensions/ModelExtensions.swift
@@ -59,6 +59,13 @@ nonisolated extension Stop: @retroactive MKAnnotation {
     public var mapTitle: String? {
         Formatters.formattedMapRoutes(routes)
     }
+
+    /// Callout text for a stop pin. Same overflowing route list as `mapTitle`
+    /// (`"10, 12, 49…"`), plus the rider-visible stop code. Distinct from
+    /// `subtitle`, which Home/Recent/Nearby use and which lists every route.
+    public var mapCalloutText: String {
+        ["#\(code)", mapTitle].compactMap { $0 }.joined(separator: "\n")
+    }
 }
 
 // MARK: - TripStatus/MKAnnotation
--- a/OBAKit/Mapping/StopAnnotationView.swift
+++ b/OBAKit/Mapping/StopAnnotationView.swift
@@ -146,7 +146,7 @@ class StopAnnotationView: MKAnnotationView {
         labelStack.isHidden = delegate.shouldHideExtraStopAnnotationData
         image = delegate.iconFactory.buildIcon(for: bookmark.stop, isBookmarked: true, traits: traitCollection)
         titleLabel.attributedText = strokedText(bookmark.name)
-        detailCalloutAccessoryView = buildDetailLabel(text: bookmark.stop.subtitle)
+        detailCalloutAccessoryView = buildDetailLabel(text: bookmark.stop.mapCalloutText)
     }
 
     private func prepareForDisplay(stop: Stop, delegate: StopAnnotationDelegate) {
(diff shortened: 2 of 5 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. `formatters.map_routes_overflow_fmt` is added as an `OBALoc` key but never added to any `Localizable.strings` file, so no translator can actually substitute anything — every locale falls through to the `value:` default `"%@…"`. That matters here because translator substitutability is the stated reason for wrapping a bare glyph in `OBALoc` at all (the doc comment, the test comment, and `docs/map-route-labels.md` all claim "translators can substitute locale-conventional marks (zh-Hans `……`)"), so as merged those three comments describe a capability the code does not have. Worth noting that `OBAKitCore` is currently at full key coverage — all 175 `OBALoc` keys in `OBAKitCore/*.swift` have an entry in `OBAKitCore/Strings/en.lproj/Localizable.strings` — so this would be the first gap in that module. Also note `LocalizationTests.Every locale has same keys as english` flags any key present in a locale but missing from `en`, so a translator adding `……` to `zh-Hans` later would fail that test until `en` is regenerated. Running `scripts/extract_strings` (and adding the 12 translations) closes it.

https://github.com/OneBusAway/onebusaway-ios/blob/116cca81f2024919dd79f6a820a6beec98f06f29/OBAKitCore/Utilities/Formatters.swift#L624-L632

aaronbrethorst (approved):
This is the right call for #514 — the inconsistency the issue is about is that `StopAnnotationView.titleLabel` truncates with UIKit's own `…` while the route list appended three ASCII periods, and making the pin label and its callout agree is what fixes it.

I checked the safety questions and they're all fine: `prefix(limit)` operates on `[String]`, not `String` indices, so there's no grapheme-cluster splitting possible; the boundary is right (`count > limit` with `prefix(limit)`, exactly 3 routes gets no ellipsis); and U+2026 is bidi-neutral in trailing position, so Arabic rendering is unchanged from the old `"..."`.

Filed #1342 for the follow-up. The short version: `formatters.map_routes_overflow_fmt` isn't in any of the 13 `.strings` files, so every locale falls through to the `value:` default. Runtime is correct everywhere, so this isn't a blocker — but the doc comment, the test comment, and `docs/map-route-labels.md` all say translators can substitute locale-conventional marks, and that isn't true until the key exists. `OBAKitCore` currently has zero key drift, so it's worth keeping it that way with a `scripts/extract_strings` pass.

That issue also notes a product question worth your judgment: bookmark pins show `bookmark.name` in their title, not the route list, so for those the callout was the only place routes appeared and it now truncates to 3 with no consistency benefit.

PR #1322, opened 2026-08-27
Show direction chevrons on trip and stop route lines

### LINKED ISSUE #445: Indicate direction flow on route lines
A possible UI enhancement for the map, suggested by a user on Reddit. Here's the comment:

> I don't know who else gets confused by some bus routes that have their route loops on the same street — there've been times I meant to take one direction when I took the one across the street instead. I know I should look for stop #s and pay attention to the direction (e.g., "SW") in OneBusAway, but if there's some clearer visual way to indicate that flow (maybe animating the black lines with a subtle gradient?) it could be pretty useful.

Borisserz:
## Summary
- Same-street inbound/outbound loops are hard to tell apart from the colored polyline alone (#445). Small `chevron.up` annotations now sit on the remaining trip shape and on each stop-focus route line, every 500m, capped at 24, short remainders omitted.
- Heading is compass degrees; UIKit rotates counter-clockwise, so the view transform is `-heading.radians`. `CLLocationDirection.affineTransform` as-is would point eastbound chevrons west.
- Chevrons tint and dim with the line they annotate. Focus restyle uses the `$focusedRouteID` emission, not a read-back of `focus.focusedRouteID` — `@Published` fires in `willSet`. This is not Transit App's vehicle-dots design.

Closes #445.

## Test plan
- [x] `PolylineDirectionArrowsTests` — spacing, east heading, 24-cap, compass→UIKit transform
- [x] `Direction arrows are placed only on the part of the trip still ahead`
- [x] `Direction arrows dim with their unfocused route`
- [x] `PolylineDirectionArrowsTests` + `TripFocusMapLayerTests` + `StopRouteFocusMapLayerTests` (52)
- [ ] On device: open a trip — chevrons only on the untraveled half, pointing along the line
- [ ] Stop with two routes: chip-tap dims the other route's chevrons with the line

## Summary by CodeRabbit

* **New Features**
  * Added directional chevrons along route and trip lines to indicate travel direction.
  * Arrows appear at regular intervals, follow the route’s geometry, and use route-specific styling.
  * Trip views show arrows only on the remaining portion of the journey.
  * Focused routes remain prominent while other route arrows are dimmed.

* **Documentation**
  * Added guidance on arrow spacing, styling, placement, and rotation behavior.

### DIFF
--- a/OBAKit/Mapping/Layers/PolylineDirectionArrows.swift
+++ b/OBAKit/Mapping/Layers/PolylineDirectionArrows.swift
@@ -147,6 +147,10 @@ final class PolylineArrowAnnotationView: MKAnnotationView {
         super.init(annotation: annotation, reuseIdentifier: reuseIdentifier)
         canShowCallout = false
         isEnabled = false
+        // Decorative only — no title/subtitle. Same opt-out as
+        // BackgroundDotAnnotationView: unlabeled chevrons would flood the
+        // VoiceOver rotor (up to ~144 on a busy stop).
+        isAccessibilityElement = false
         displayPriority = .defaultLow
         collisionMode = .none
         applyAnnotation()
--- a/OBAKit/Mapping/Layers/PolylineDirectionArrows.swift
+++ b/OBAKit/Mapping/Layers/PolylineDirectionArrows.swift
@@ -1,0 +1,168 @@
+//
+//  PolylineDirectionArrows.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import CoreLocation
+import MapKit
+import OBAKitCore
+import UIKit
+
+nonisolated struct PolylineArrowPlacement: Equatable {
+    let coordinate: CLLocationCoordinate2D
+    let headingDegrees: CLLocationDirection
+
+    static func == (lhs: PolylineArrowPlacement, rhs: PolylineArrowPlacement) -> Bool {
+        lhs.coordinate.latitude == rhs.coordinate.latitude
+            && lhs.coordinate.longitude == rhs.coordinate.longitude
+            && lhs.headingDegrees == rhs.headingDegrees
+    }
+}
+
+nonisolated enum PolylineDirectionArrows {
+    private static let maximumPlacementCount = 24
+
+    private struct Segment {
+        let start: MKMapPoint
+        let end: MKMapPoint
+        let startDistance: CLLocationDistance
+        let length: CLLocationDistance
+
+        var endDistance: CLLocationDistance { startDistance + length }
+
+        func mapPoint(at distance: CLLocationDistance) -> MKMapPoint {
+            let fraction = (distance - startDistance) / length
+            return MKMapPoint(
+                x: start.x + (end.x - start.x) * fraction,
+                y: start.y + (end.y - start.y) * fraction
+            )
+        }
+    }
+
+    /// Places arrows along `coordinates` every `spacing` meters, beginning one
+    /// full spacing from the first vertex.
+    static func placements(
+        along coordinates: [CLLocationCoordinate2D],
+        spacing: CLLocationDistance = 500
+    ) -> [PolylineArrowPlacement] {
+        guard coordinates.count >= 2, spacing.isFinite, spacing > 0 else { return [] }
+
+        var segments: [Segment] = []
+        var totalDistance: CLLocationDistance = 0
+        for (startCoordinate, endCoordinate) in zip(coordinates, coordinates.dropFirst()) {
+            let length = startCoordinate.distance(from: endCoordinate)
+            guard length.isFinite, length > 0 else { continue }
+
+            segments.append(Segment(
+                start: MKMapPoint(startCoordinate),
+                end: MKMapPoint(endCoordinate),
+                startDistance: totalDistance,
+                length: length
+            ))
+            totalDistance += length
+        }
+
+        guard !segments.isEmpty else { return [] }
+
+        // Leave half a spacing after the final arrow. This avoids a chevron
+        // crowding the terminal stop or vehicle at the end of a short remainder.
+        let finalPlacementDistance = totalDistance - spacing / 2
+        guard finalPlacementDistance >= spacing else { return [] }
+
+        var placements: [PolylineArrowPlacement] = []
+        var distance = spacing
+        while distance <= finalPlacementDistance, placements.count < maximumPlacementCount {
+            guard let point = mapPoint(at: distance, in: segments) else { break }
+
+            // A short sample on each side follows the local tangent through a
+            // vertex instead of inheriting only the segment before or after it.
+            let previous = mapPoint(at: max(0, distance - 1), in: segments) ?? point
+            let next = mapPoint(at: min(totalDistance, distance + 1), in: segments) ?? point
+            placements.append(PolylineArrowPlacement(
+                coordinate: point.coordinate,
+                headingDegrees: heading(from: previous, to: next)
+            ))
+            distance += spacing
+        }
+
+        return placements
+    }
+
+    /// `chevron.up` points north at identity. Compass heading is clockwise from
+    /// north; `CGAffineTransform` rotates counter-clockwise, so the angle is negated.
+    static func viewTransform(headingDegrees: CLLocationDirection) -> CGAffineTransform {
+        CGAffineTransform(rotationAngle: -CGFloat(headingDegrees.radians))
+    }
+
+    private static func mapPoint(at distance: CLLocationDistance, in segments: [Segment]) -> MKMapPoint? {
+        guard let segment = segments.first(where: { distance <= $0.endDistance }) ?? segments.last else {
+            return nil
+        }
+        return segment.mapPoint(at: min(max(distance, segment.startDistance), segment.endDistance))
+    }
+
+    /// MapKit's x axis points east and y axis points south. Swapping the usual
+    /// atan2 arguments and negating y yields compass degrees: north = 0, east = 90.
+    private static func heading(from start: MKMapPoint, to end: MKMapPoint) -> CLLocationDirection {
+        let degrees = atan2(end.x - start.x, -(end.y - start.y)) * 180 / .pi
+        return degrees >= 0 ? degrees : degrees + 360
+    }
+}
+
+final class PolylineArrowAnnotation: NSObject, MKAnnotation {
+    let coordinate: CLLocationCoordinate2D
+    let headingDegrees: CLLocationDirection
+    let tintColor: UIColor
+    let routeID: RouteID?
+    var alpha: CGFloat
+
+    init(
+        placement: PolylineArrowPlacement,
+        tintColor: UIColor,
+        routeID: RouteID? = nil,
+        alpha: CGFloat = 1
+    ) {
+        coordinate = placement.coordinate
+        headingDegrees = placement.headingDegrees
+        self.tintColor = tintColor
+        self.routeID = routeID
+        self.alpha = alpha
+        super.init()
+    }
+}
+
+final class PolylineArrowAnnotationView: MKAnnotationView {
+    static let reuseIdentifier = "PolylineArrowAnnotationView"
+
+    override var annotation: MKAnnotation? {
+        didSet { applyAnnotation() }
+    }
+
+    override init(annotation: MKAnnotation?, reuseIdentifier: String?) {
+        super.init(annotation: annotation, reuseIdentifier: reuseIdentifier)
+        canShowCallout = false
+        isEnabled = false
+        displayPriority = .defaultLow
+        collisionMode = .none
+        applyAnnotation()
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    private func applyAnnotation() {
+        guard let arrow = annotation as? PolylineArrowAnnotation else { return }
+
+        let configuration = UIImage.SymbolConfiguration(pointSize: 13, weight: .bold)
+        image = UIImage(systemName: "chevron.up", withConfiguration: configuration)?
+            .withTintColor(arrow.tintColor, renderingMode: .alwaysOriginal)
+        transform = PolylineDirectionArrows.viewTransform(headingDegrees: arrow.headingDegrees)
+        alpha = arrow.alpha
+        centerOffset = .zero
+    }
+}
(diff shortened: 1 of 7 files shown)


aaronbrethorst:
### Code review

Found 3 issues:

1. `viewTransform` negates the heading, which mirrors every chevron: eastbound lines point west. In UIKit a *positive* `CGAffineTransform(rotationAngle:)` is clockwise on screen (the view's coordinate space is y-down), which is the same sense as a compass heading. Working with the matrix: `(x', y') = (x·cosθ − y·sinθ, x·sinθ + y·cosθ)`, so at θ = +π/2 the top of the view `(0, −1)` maps to `(1, 0)` — up becomes right, i.e. east. The negation therefore sends 90° (east) to west, 45° (NE) to NW, and so on; only due north and due south survive. This repo already has two shipped precedents for the positive convention: `CLLocationDirection.affineTransform` uses `CGAffineTransform(rotationAngle: CGFloat(radians))` un-negated for `trueHeading` (a compass value), and `PulsingVehicleAnnotationView.updateHeading` negates *because* `TripStatus.orientation` is counter-clockwise-from-east, not compass — its comment spells out that CG-on-iOS runs the other way from the API's convention. The fix is to drop the minus sign. `docs/polyline-direction-arrows.md` and the PR description carry the same inverted rationale and would need updating with it.

https://github.com/OneBusAway/onebusaway-ios/blob/7d1f2427b29964f4fee93cfbb9314282c582aa39/OBAKit/Mapping/Layers/PolylineDirectionArrows.swift#L95-L99

2. The transform test asserts the mirrored expectation, so it can't catch issue 1 — it restates the implementation rather than the desired on-screen result. `east.b ≈ -1` is θ = −π/2, which is the value that renders an eastbound chevron pointing west. A test that pins the rendered direction (e.g. that a heading of 90° maps the view's "up" unit vector `(0, -1)` to `(+1, 0)`) would fail today and pass after the sign fix.

https://github.com/OneBusAway/onebusaway-ios/blob/7d1f2427b29964f4fee93cfbb9314282c582aa39/OBAKitTests/Mapping/PolylineDirectionArrowsTests.swift#L80-L86

3. `apply(style:to:focusedRouteID:)` collapses "explicitly nil" into "not supplied", which undoes this PR's own fix on the un-focus path. The whole point of threading `focusedRouteID` through from the `$focusedRouteID` emission is that `@Published` fires in `willSet`, so `focus?.focusedRouteID` is stale inside the sink — this file's own `apply(arrivals:...)` doc comment calls out that exact hazard. But when the emission is `nil` (tapping the focused chip/marker again, `clearFocus()`, or `apply(routes:)` dropping a dangling focus), `focusedRouteID ?? focus?.focusedRouteID` falls back to the stale, still-non-nil previous route, so `restyleOverlays` re-applies the old focused/dimmed styling to on-screen renderers. `restyleDirectionArrows` takes the value directly and gets it right, so after an un-focus the chevrons brighten while the lines they annotate stay dimmed. Passing the value explicitly at both call sites (or a two-case enum instead of a defaulted Optional) avoids the conflation.

https://github.com/OneBusAway/onebusaway-ios/blob/7d1f2427b29964f4fee93cfbb9314282c582aa39/OBAKit/Mapping/Layers/StopRouteFocus/StopRouteFocusMapLayer.swift#L476-L480

OBAKit/Mapping/Layers/PolylineDirectionArrows.swift
```diff
@@ -0,0 +1,168 @@
...
+            distance += spacing
+        }
+
+        return placements
+    }
+
+    /// `chevron.up` points north at identity. Compass heading is clockwise from
+    /// north; `CGAffineTransform` rotates counter-clockwise, so the angle is negated.
+    static func viewTransform(headingDegrees: CLLocationDirection) -> CGAffineTransform {
+        CGAffineTransform(rotationAngle: -CGFloat(headingDegrees.radians))
```
aaronbrethorst: This is the blocker. `headingDegrees` is a true compass bearing (north = 0, east = 90) — your `heading(from:to:)` above produces exactly that, and its comment says so. `chevron.up` points north at identity. So rotating by the bearing requires a **positive** angle:

```swift
CGAffineTransform(rotationAngle: CGFloat(headingDegrees.radians))
```

Verified by applying the transform to the up vector `(0, -1)`:

| heading | `-radians` (current) | `+radians` |
|---|---|---|
| 90° (east) | **WEST** | EAST |

Positive `CGAffineTransform(rotationAngle:)` is clockwise *on screen* because UIKit view space has y pointing down — the same sense as a compass heading.

If the negation came from `PulsingVehicleAnnotationView.updateHeading`, note that it's flipping a different convention: `TripStatus.orientation` is documented at `OBAKitCore/Models/REST/TripStatus.swift:82` as "0º is east, 90º is north" — counter-clockwise from east. It needs the flip; a compass bearing doesn't.

OBAKitTests/Mapping/PolylineDirectionArrowsTests.swift
```diff
@@ -0,0 +1,86 @@
...
+        ]
+
+        let placements = PolylineDirectionArrows.placements(along: line, spacing: 100)
+
+        #expect(!placements.isEmpty)
+        #expect(placements.count <= 24)
+    }
+
+    /// `chevron.up` points north at identity. UIKit rotates counter-clockwise,
+    /// compass heading is clockwise, so east (90°) must be `-π/2`.
```
aaronbrethorst: This comment carries the premise that produced the bug: UIKit rotates **clockwise** for a positive angle (y is down in view space), not counter-clockwise. Because compass headings are also clockwise, east should be `+π/2`, not `-π/2`.

The assertion below (`east.b ≈ -1`) restates the implementation's matrix coefficient, so it will keep passing whichever sign is used and can't catch the inversion. Asserting the rendered direction makes it a real regression lock:

```swift
let east = PolylineDirectionArrows.viewTransform(headingDegrees: 90)
let up = CGPoint(x: 0, y: -1)
let rotated = up.applying(east)
#expect(rotated.x > 0.99)          // points right / east
#expect(abs(rotated.y) < 0.001)
```

aaronbrethorst (changes requested):
There's a lot to like in the structure here. `PolylineDirectionArrows` as a `nonisolated enum` doing pure geometry over an `MKMapPoint` walk is the right shape, the degenerate-polyline guards are thorough, and choosing fixed geographic spacing over per-zoom respacing is a genuinely good call — it's why there's no O(n) work on the main thread during pan and zoom. The bearing function is correct too: `atan2(dx, -dy)` in `MKMapPoint` space really does give true compass degrees, and Mercator being conformal means the projected bearing is the right one to use.

The problem is one character. `viewTransform` negates the angle, which mirrors every chevron that isn't due north or south — the feature currently points riders the wrong way on any east/west route, which is the opposite of what #445 asked for.

I checked this empirically rather than by eye. Applying the transform to the chevron's "up" vector `(0, -1)`:

```
heading = 90 deg (due EAST); chevron.up starts pointing NORTH
  PR's  -radians  renders: WEST (left)
  the   +radians  renders: EAST (right)
```

The premise in the test's comment is what led here: UIKit's `CGAffineTransform(rotationAngle:)` rotates *clockwise* on screen for a positive angle, not counter-clockwise, because view space has y pointing down. Clockwise is the same sense as a compass heading, so a true bearing needs the angle un-negated.

I suspect the negation was carried over from `PulsingVehicleAnnotationView.updateHeading`, which does negate — but that's a different input convention, not a precedent. `TripStatus.orientation` is documented in `OBAKitCore/Models/REST/TripStatus.swift:82` as "0º is east, 90º is north, 180º is west" — counter-clockwise from east. That's why it needs flipping. Your `heading(from:to:)` already produces a compass bearing, so it doesn't.

The fix is dropping the minus sign and updating the test to assert the rendered direction rather than the matrix coefficient.

Two more things, neither blocking:

- The 24-chevron cap truncates rather than redistributing. On stop focus you're drawing whole route shapes, so a 20 km route gets chevrons over the first 12 km and a bare tail — and a rider whose stop sits in that remainder gets nothing, which is the #445 use case. Scaling spacing to fit the cap would cover the whole shape.
- `focusedRouteID ?? focus?.focusedRouteID` conflates an explicit nil emission with an omitted one. `@Published` fires in `willSet`, so `focus?.focusedRouteID` is stale inside the sink — the file's own `apply(arrivals:)` comment documents that hazard. On un-focus the fallback resurrects the previous route, so the polylines keep their dimmed styling while `restyleDirectionArrows` correctly brightens the chevrons.

Please have a look at the inline comments — the sign fix is small and I'd be glad to re-review quickly once it and the test are updated.

aaronbrethorst:
Heads up: I merged #1323 (vehicle marker interpolation) just now, which also touches `TripFocusMapLayer.swift`, so this branch has conflicts.

Rebase cost only — please merge `main` in and resolve alongside the rotation-sign fix.

OBAKit/Mapping/Layers/PolylineDirectionArrows.swift
Borisserz: Dropped the negation. `viewTransform(headingDegrees: 90)` now rotates the up-vector to `x > 0` (east). Verified red (`east.x = -1`) then green in `PolylineDirectionArrowsTests`.

OBAKitTests/Mapping/PolylineDirectionArrowsTests.swift
Borisserz: Replaced the matrix-coefficient check with the up-vector assertion you wrote. East now requires `rotated.x > 0.99`.

aaronbrethorst:
### Code review

Thanks for the quick turnaround on the rotation sign — I re-checked `viewTransform` and the new up-vector assertion at 7b47413 and both are right now. `restyleOverlays(focusedRouteID:)` threading the emission through to `apply(style:to:focusedRouteID:)` also closes the live half of the stale-focus problem.

Found 2 issues:

1. `docs/polyline-direction-arrows.md` still carries the pre-fix rationale, and now contradicts the code it documents. It says "UIKit rotation is counter-clockwise, so the view transform is `-heading.radians`", while `PolylineDirectionArrows.viewTransform` is `CGAffineTransform(rotationAngle: CGFloat(headingDegrees.radians))` and its own doc comment says "UIKit view space has y down, so a positive `CGAffineTransform` rotation is clockwise too. Do not negate — that mirrored east/west." The whole purpose of that paragraph is to record the rotation convention, so leaving it inverted prescribes exactly the bug that was just removed to whoever reads it next.

https://github.com/OneBusAway/onebusaway-ios/blob/7b4741356e91a3012b45dc0399dd259d0c29541f/docs/polyline-direction-arrows.md#L14-L19

2. `PolylineArrowAnnotationView` never sets `isAccessibilityElement = false`, so every chevron lands in the VoiceOver rotor as an unlabeled element. `PolylineArrowAnnotation` vends no `title` or `subtitle`, and the view is `isEnabled = false` / `canShowCallout = false` — purely decorative. With up to 24 placements per shape and `routeCap = 6` in `StopRouteFocusMapLayer`, a busy stop can put ~144 of these on the map. `BackgroundDotAnnotationView` is the same shape of thing and opts out explicitly, with a comment that describes this case exactly: "an unlabelled, unselectable dot would only add noise to the rotor."

https://github.com/OneBusAway/onebusaway-ios/blob/7b4741356e91a3012b45dc0399dd259d0c29541f/OBAKit/Mapping/Layers/PolylineDirectionArrows.swift#L145-L154

Two smaller things I looked at and am not counting against the PR: the `= nil` default on `apply(style:to:focusedRouteID:)` is now only reached from the `renderer(for:)` delegate callback, which MapKit invokes after `willSet` has completed, so the stale read is latent rather than live — but tightening it to a required parameter would remove the footgun. And the 24-chevron cap still truncates rather than redistributing, which is documented and was non-blocking before; it remains so.

Borisserz:
Fixed both: docs now describe the clockwise `heading.radians` transform, and `PolylineArrowAnnotationView` sets `isAccessibilityElement = false` like `BackgroundDotAnnotationView`.

aaronbrethorst:
### Code review — correction

Both findings in my previous comment were wrong — it reviewed `7b47413` and you'd already pushed `30963ee`.

On the current head the doc says "The view transform is `heading.radians` — do **not** negate, or eastbound chevrons mirror west", which matches `CGAffineTransform(rotationAngle: CGFloat(headingDegrees.radians))`. And `PolylineArrowAnnotationView.init` sets `isAccessibilityElement = false` with a comment citing the BackgroundDot precedent and the rotor-flood reason. Both of the things I flagged were already done.

Apologies for the noise. Nothing outstanding from me on the current head.

aaronbrethorst (approved):
The blocker from last round is genuinely fixed — the negation is gone, and the test now asserts the up-vector (`east.x > 0.99`) rather than the mirrored matrix coefficient, so it is a real regression lock. The doc and the accessibility opt-out I flagged earlier were both already done on this head; that was my error and I have retracted it above.

On the performance question I raised: there is no custom `MKOverlayRenderer` here — these are annotation views, so nothing runs in `draw(mapRect:zoomScale:in:)` and there is no per-frame or per-zoom cost. `restyleDirectionArrows` runs only on focus change. The bearing math (`atan2(dx, -dy)` in `MKMapPoint` space) is right, Mercator conformality makes the projected bearing the correct one, and the degenerate cases — empty, single point, duplicate consecutive points, all-zero-length — are all guarded and tested. Map rotation is disabled, so the screen-space transform cannot desync from a rotated basemap.

Non-blocking leftovers: the 24-chevron cap truncates rather than redistributing, so a long stop-focus route gets chevrons only on its first stretch; and routes shorter than 1.5x spacing get none. Both are documented and deliberate.

Approved on the merits — no re-review needed unless the conflict resolution changes behavior. Sibling PRs landed on `main` today and touched the same files, so this branch now has merge conflicts; please merge/rebase `main` in and resolve.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
The rotation is right now, and I verified it independently rather than taking the last round's word for it: `CGAffineTransform(rotationAngle:)` in UIKit's y-down view space maps the `chevron.up` vector to (sin θ, −cos θ), so 90° points east — which matches the un-negated precedent in `CoreLocationExtensions.affineTransform`. `atan2(dx, −dy)` is the correct compass conversion for MKMapPoint's axes, and Mercator conformality makes the map-space bearing a valid heading. Lifecycle, per-route dimming, and annotation cleanup all check out.

Two things for a follow-up, neither blocking: the 24-chevron cap truncates rather than redistributing, so a long route line goes bare past ~12km; and the `Direction arrows dim with their unfocused route` test uses `allSatisfy` over a filtered array with no non-empty guard, so it would pass vacuously if that shape fetch returned nothing — the sibling test guards with `#expect(!arrows.isEmpty)`.

Also worth updating the PR description before this lands in history: it still says the view transform is `-heading.radians`, which is the opposite of what shipped.

PR #1323, opened 2026-08-27
Interpolate vehicle markers between arrival polls

Borisserz:
## Summary
- Arrival polls used to `removeAnnotation`/`addAnnotation` the trip-page bus. MapKit treats that as a new pin, so the marker pops and any open callout is dismissed.
- The trip layer now keeps the same `VehicleAnnotation` when the vehicle ID matches, and `StopVehicleAnnotation.update` does the same for stop-focus markers.
- `VehicleCoordinateUpdate` eases hops under 500 m (~one 30s poll at city speed) over 0.8 s, snaps farther than that (or to/from null island), and ignores sub-2 m GPS jitter. No new settings.

Addresses item 1 of #1109. Item 2 has an open PR (#1115). Item 3 landed in #1305.

## Test plan
- [x] `A refresh moves the existing vehicle annotation instead of replacing it`
- [x] `VehicleCoordinateUpdateTests` (5) — jitter / city-block / kilometre / null island / invalid
- [x] `TripFocusMapLayerTests` (13) and `StopVehicleAnnotationTests` (6)
- [ ] Device: open a trip page, wait for an arrivals poll — the bus should slide, not jump. A region change or a different vehicle ID should still snap.

## Summary by CodeRabbit

* **Improvements**
  * Vehicle markers now move smoothly between nearby location updates instead of teleporting.
  * Large or invalid location changes snap directly to the correct position.
  * Vehicle markers remain stable when trip information refreshes.
  * Marker status and route colors update without unnecessary replacement.
  * Missing vehicle information now removes the marker appropriately.

* **Bug Fixes**
  * Prevented incorrect animations caused by GPS jitter, invalid coordinates, or Null Island locations.

### DIFF
--- a/OBAKit/Mapping/Layers/StopRouteFocus/StopVehicleAnnotation.swift
+++ b/OBAKit/Mapping/Layers/StopRouteFocus/StopVehicleAnnotation.swift
@@ -94,8 +94,14 @@ final class StopVehicleAnnotation: VehicleAnnotation {
         // same as the initializer's `super.init(tripStatus:)` — so the
         // position-preferred coordinate must be assigned AFTER, or it gets
         // clobbered exactly as the initializer's comment describes.
+        //
+        // Restore `from` before `VehicleCoordinateUpdate.apply` so a city-block
+        // hop interpolates instead of that didSet teleporting the pin. A
+        // kilometre-scale jump still snaps; see VehicleCoordinateUpdate.
+        let from = self.coordinate
         self.tripStatus = tripStatus
-        self.coordinate = coordinate
+        self.coordinate = from
         self.routeColor = routeColor
+        VehicleCoordinateUpdate.apply(from: from, to: coordinate, on: self)
     }
 }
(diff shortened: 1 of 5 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Nice work. Mirroring the diff-by-id pattern you'd already shipped in `StopRouteFocusMapLayer.syncVehicleAnnotations` rather than inventing a second approach is the right instinct, and splitting teardown into `removeShapeAndStops()` / `removeVehicle()` is what makes the whole thing possible.

The part I checked hardest is the restore-then-apply dance, because it's the delicate bit:

```swift
let from = existing.coordinate
existing.tripStatus = status      // didSet writes lastKnownLocation onto coordinate
existing.coordinate = from        // put it back
VehicleCoordinateUpdate.apply(from: from, to: coord, on: existing)
```

That's correct, and `TripStatus.coordinate` really is `lastKnownLocation?.coordinate ?? (0,0)`, so `from` and `to` are consistent. Every edge case degrades to `.snap`, which is today's behavior — invalid coordinates, null island, jumps over the threshold, and a vanished vehicle all route safely. `isSameVehicle` returning true on missing IDs is well-reasoned given this layer only ever draws one marker.

Filed #1341. The main item there: the type doc justifies the threshold as "one or two polls at motorway speed" (830–1670 m) but the constant is 500 m, documented elsewhere as ~50 km/h — which means interpolation effectively won't engage above ~60 km/h. That's a defensible tuning choice, it just isn't what the prose promises. Also noted that `apply` itself has no tests, and that `TripViewController`'s legacy screen still teleports.

PR #1324, opened 2026-08-27
Document FixedRegionName for white-label region pinning

### LINKED ISSUE #608: Add a way to automatically pick region at app launch
Request from @wkulesza:

He'd suggested automatically selecting the region if there's only one available. I wonder if we could do that, or perhaps specify a hardcoded region in the OBAKitConfig data in the Info.plist. That way, it would remove any ambiguity in the case of a data issue.

Borisserz:
## Summary
- `FixedRegionName` / `FixedRegionOBABaseURL` already drive `RegionsService` (name match, then OBA base URL fallback; never overrides a region the user already chose). White-label agencies had no docs for the keys #608 asked for.
- WhiteLabel.md now documents those keys, plus the existing single-active-region auto-select when both are omitted.
- Bundle tests cover the plist read: present, absent, and an unparseable URL degrading to nil. Behavior tests already live in `RegionsServiceAutoSelectTests`.

Closes #608.

## Test plan
- [x] `Fixed region name reads from OBA kit config` / `is nil when absent`
- [x] `Fixed region OBA base URL reads from OBA kit config` / `is nil when absent or unparseable`
- [x] `RegionsServiceAutoSelectTests` (10)
- [ ] White-label: set `FixedRegionName` to a bundled region with no current region in defaults — picker skipped. OneBusAway itself should leave both keys unset.

## Summary by CodeRabbit

- **Documentation**
  - Expanded white-label region configuration guidance to include fixed region name and base URL settings.
  - Documented the order used to resolve a default region, including location-based selection, name and URL matching, and fallback behavior.

- **Tests**
  - Added coverage for reading fixed region settings from configuration.
  - Added validation for missing or invalid fixed region values.

### DIFF
--- a/OBAKit/OBAKit.docc/Articles/WhiteLabel.md
+++ b/OBAKit/OBAKit.docc/Articles/WhiteLabel.md
@@ -164,15 +164,15 @@ If you choose to leave out `RegionsServerBaseAddress` and `RegionsServerAPIPath`
 
 ### Default region from OBAKitConfig
 
-These keys are **not a pin**. `RegionsService.init` picks a region in this order, and later steps never run if an earlier one already set `currentRegion`:
+These keys are **not a pin**. `RegionsService.init` resolves a region in this order:
 
-1. Previously stored region (UserDefaults)
-2. Location-based auto-select — if auto-select is on (the default) *or* no region is stored yet, and `LocationService` already has a cached fix inside a known region
-3. `FixedRegionName`, then `FixedRegionOBABaseURL` if the name matches nothing
+1. Previously stored region (UserDefaults), **unless** location auto-select is on (the default) — then a cached fix inside a known region can overwrite it
+2. Location-based auto-select when auto-select is on *or* no region is stored yet (`|| currentRegion == nil`), and `LocationService` already has a cached fix inside a known region
+3. `FixedRegionName`, then `FixedRegionOBABaseURL` if the name matches nothing — only when `currentRegion` is still nil after steps 1–2
 4. The only *active* region in the list, when there is exactly one
 5. The region picker
 
-A rider whose device already has a location fix inside another region's bounds will get **that** region, not the configured one. `OBAKitTests` covers this as `Location based selection takes priority`. An agency that needs every install on one region must ship a list with only that region (or turn location auto-select off themselves); these keys will not override a location match.
+A rider whose device already has a location fix inside another region's bounds will get **that** region, not the configured one. `OBAKitTests` covers this as `Location based selection takes priority`. Because location runs whenever `currentRegion` is nil (including first launch), registering `OBAAutomaticallySelectRegionUserDefaultsKey` as `false` does **not** skip that match on the launch where `FixedRegionName` could apply. An agency that needs every install on one region must ship a list with only that region (or avoid prompting for location so `currentLocation` stays nil); these keys will not override a location match.
 
 * `FixedRegionName` - Optional. The `name` of a region in the bundled (or downloaded) list. Used only when steps 1–2 left `currentRegion` nil. A successful match writes that region and turns auto-select off so a *later* location fix cannot steal it — an *earlier* one already can.
 * `FixedRegionOBABaseURL` - Optional. Used only when `FixedRegionName` is set and does not match any known region. The first region whose OBA base URL equals this value is selected instead. A missing or unparseable URL is ignored.
--- a/OBAKit/OBAKit.docc/Articles/WhiteLabel.md
+++ b/OBAKit/OBAKit.docc/Articles/WhiteLabel.md
@@ -150,15 +150,28 @@ These are parameters specified in your app's `project.yml` file that customizes 
 
 ### Regions
 
-There are three parameters that are associated with regions:
+There are five parameters that are associated with regions:
 
 * `BundledRegionsFileName` - Required.
 * `RegionsServerBaseAddress` - Optional.
 * `RegionsServerAPIPath` - Optional.
+* `FixedRegionName` - Optional.
+* `FixedRegionOBABaseURL` - Optional.
 
 The regions JSON file that you will specify with the `BundledRegionsFileName` parameter is required so that the app can show the user a list of available regions as soon as the application launches, without having to worry about having an internet connection. However, that list will eventually/inevitably become out of date, and so the app will also regularly download the remote copy of the regions file to make sure that changes (names, features, new regions, deleted regions) are handled.
 
 If you choose to leave out `RegionsServerBaseAddress` and `RegionsServerAPIPath`, then your app will rely solely on the bundled regions file for all regions that your application will know about. It is _highly_ recommended that you provide a regions server URL and path, but it is not strictly required.
+
+### Pinning a region at launch
+
+A white-label app that only serves one agency can skip the region picker:
+
+* `FixedRegionName` - Optional. The `name` of a region in the bundled (or downloaded) list. When the user has no current region, the app selects this one and turns auto-select off so a later location fix cannot steal it.
+* `FixedRegionOBABaseURL` - Optional. Used only when `FixedRegionName` is set and does not match any known region. The first region whose OBA base URL equals this value is selected instead. A missing or unparseable URL is ignored.
+
+Name match wins over URL match. Neither key overrides a region the user (or a previous launch) already chose. If both are omitted, the picker still appears — unless the list contains exactly one *active* region, in which case that region is selected automatically.
+
+Invalid optional URLs degrade to nil; a name that matches nothing and has no URL fallback leaves the current region unset so the picker can still run.
 
 ### Arrival and Departure Display
 
--- a/OBAKit/OBAKit.docc/Articles/WhiteLabel.md
+++ b/OBAKit/OBAKit.docc/Articles/WhiteLabel.md
@@ -162,16 +162,24 @@ The regions JSON file that you will specify with the `BundledRegionsFileName` pa
 
 If you choose to leave out `RegionsServerBaseAddress` and `RegionsServerAPIPath`, then your app will rely solely on the bundled regions file for all regions that your application will know about. It is _highly_ recommended that you provide a regions server URL and path, but it is not strictly required.
 
-### Pinning a region at launch
+### Default region from OBAKitConfig
 
-A white-label app that only serves one agency can skip the region picker:
+These keys are **not a pin**. `RegionsService.init` picks a region in this order, and later steps never run if an earlier one already set `currentRegion`:
 
-* `FixedRegionName` - Optional. The `name` of a region in the bundled (or downloaded) list. When the user has no current region, the app selects this one and turns auto-select off so a later location fix cannot steal it.
+1. Previously stored region (UserDefaults)
+2. Location-based auto-select — if auto-select is on (the default) *or* no region is stored yet, and `LocationService` already has a cached fix inside a known region
+3. `FixedRegionName`, then `FixedRegionOBABaseURL` if the name matches nothing
+4. The only *active* region in the list, when there is exactly one
+5. The region picker
+
+A rider whose device already has a location fix inside another region's bounds will get **that** region, not the configured one. `OBAKitTests` covers this as `Location based selection takes priority`. An agency that needs every install on one region must ship a list with only that region (or turn location auto-select off themselves); these keys will not override a location match.
+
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The `FixedRegionName` description omits that location-based selection outranks it, and the wording implies the opposite. `RegionsService.init` runs the location match *before* the fixed-region block (`if autoSelectRegion, let location = locationService.currentLocation { currentRegion = ... }` at `RegionsService.swift:93-98`, then `if currentRegion == nil, let fixedName = fixedRegionName` at :101). So on a first launch where a location fix is already available — `LocationService.init` seeds `currentLocation = locationManager.location` — the physically-located region wins and `FixedRegionName` never applies, leaving auto-select *on*. `RegionsServiceAutoSelectTests.Location based selection takes priority` asserts exactly this ("Location-based selection should take priority over fixed region config"), and PR #1095's own precedence table ranks location above the fixed region. As written, "When the user has no current region, the app selects this one ... so a later location fix cannot steal it" tells a white-label integrator the key deterministically pins the region, which is the one thing an agency app for a single region cares about. Worth stating the full order: stored region → location → fixed name → fixed URL → single active region → picker.

https://github.com/OneBusAway/onebusaway-ios/blob/11b512b18f1ffbfef2f42eeb4f6e2d1b704726a5/OBAKit/OBAKit.docc/Articles/WhiteLabel.md#L168-L170

OBAKit/OBAKit.docc/Articles/WhiteLabel.md
```diff
@@ -150,16 +150,29 @@ These are parameters specified in your app's `project.yml` file that customizes
...
 
 The regions JSON file that you will specify with the `BundledRegionsFileName` parameter is required so that the app can show the user a list of available regions as soon as the application launches, without having to worry about having an internet connection. However, that list will eventually/inevitably become out of date, and so the app will also regularly download the remote copy of the regions file to make sure that changes (names, features, new regions, deleted regions) are handled.
 
 If you choose to leave out `RegionsServerBaseAddress` and `RegionsServerAPIPath`, then your app will rely solely on the bundled regions file for all regions that your application will know about. It is _highly_ recommended that you provide a regions server URL and path, but it is not strictly required.
 
+### Pinning a region at launch
+
+A white-label app that only serves one agency can skip the region picker:
+
+* `FixedRegionName` - Optional. The `name` of a region in the bundled (or downloaded) list. When the user has no current region, the app selects this one and turns auto-select off so a later location fix cannot steal it.
```
aaronbrethorst: This is the one blocking item. "When the user has no current region" reads as "no region stored from a previous launch", but the actual gate in `OBAKitCore/Location/Regions/RegionsService.swift` is `currentRegion == nil` *after* location matching has already run:

```swift
if autoSelectRegion, let location = locationService.currentLocation {
    currentRegion = RegionsService.firstRegion(in: self.regions, containing: location)
}

// Fixed region from OBAKitConfig: match by name, then fall back to URL.
if currentRegion == nil, let fixedName = fixedRegionName {
```

`LocationService.init` seeds `currentLocation` from `locationManager.location`, so a first launch with a cached fix inside a known region selects *that* region and never reaches the fixed-region block — auto-select stays on, and the claim that "a later location fix cannot steal it" is inverted: an earlier one already did.

`OBAKitTests/Location/RegionsServiceAutoSelectTests.swift` has a test named `Location based selection takes priority` with the comment "Location-based selection should take priority over fixed region config", so this is settled behavior, not an edge case.

Could you state the full order the way #1095's description does — stored → location → fixed name → fixed URL → single active region → picker? An agency pinning one region needs to know that a rider physically inside another region's bounds won't get the pinned one.

OBAKit/OBAKit.docc/Articles/WhiteLabel.md
```diff
@@ -150,16 +150,29 @@ These are parameters specified in your app's `project.yml` file that customizes
...
+* `FixedRegionOBABaseURL` - Optional. Used only when `FixedRegionName` is set and does not match any known region. The first region whose OBA base URL equals this value is selected instead. A missing or unparseable URL is ignored.
+
+Name match wins over URL match. Neither key overrides a region the user (or a previous launch) already chose. If both are omitted, the picker still appears — unless the list contains exactly one *active* region, in which case that region is selected automatically.
```
aaronbrethorst: Minor, and it partly resolves itself once the precedence above is corrected: the next line says a name matching nothing "leaves the current region unset so the picker can still run", but the single-active-region auto-select at `RegionsService.swift:113-119` runs *after* the fixed-region block — so with exactly one active region in the list, a no-match name still ends up with a region and no picker. This sentence and the one above it mildly contradict each other.

aaronbrethorst (changes requested):
Thanks for writing this up — `FixedRegionName` has been undocumented since #1095 landed and it's exactly the kind of white-label key an agency needs spelled out. The characterization tests for the two `Bundle` accessors are a good addition too, and I checked the rest of the section against source: name-over-URL precedence, the "never overrides an existing region" claim, the auto-select-off write, and the single-*active*-region rule are all accurate.

One correction needed before this lands, flagged inline. The section promises pinning, but `FixedRegionName` doesn't actually pin — location-based selection runs first in `RegionsService.init` and the fixed-region block only runs `if currentRegion == nil`. An agency reading this will ship an app that quietly picks a different region for any user with a cached location fix. Our own `Location based selection takes priority` test asserts that ordering, so the doc and the test currently disagree.

Since this file is the authoritative white-label reference, I'd like the precedence stated explicitly rather than implied — something like the stored → location → fixed → single-active → picker chain from #1095's description.

A couple of smaller things while you're in there, not blocking:

- `FixedRegionName` / `FixedRegionOBABaseURL` now appear twice: as bare "Optional." bullets in the five-parameter list above and again with full descriptions in the new section. The other three keys aren't duplicated that way.
- The URL fallback is byte-exact string equality (`$0.OBABaseURL == fixedRegionOBABaseURL`), so a trailing-slash difference silently fails to match. "equals this value" is worth making stricter-sounding.

Please have a look at the inline comments — this is a quick fix and I'm glad to re-review as soon as it's updated.

aaronbrethorst:
Heads up: I merged #1329 just now, which appends a "More Tab" section to the end of `WhiteLabel.md` and adds tests to the same `BundleFeedbackConfigTests` suite — so this branch has conflicts in both files.

Purely textual, both changes are additive. Please merge `main` in and resolve alongside the precedence correction.

OBAKit/OBAKit.docc/Articles/WhiteLabel.md
Borisserz: Rewrote the section to the order from #1095: stored → location → fixed name → fixed URL → single active region → picker. Called out that a cached location fix inside another region wins, matching `Location based selection takes priority`.

OBAKit/OBAKit.docc/Articles/WhiteLabel.md
Borisserz: The no-match sentence now says step 4 (single active region) still runs, so a miss does not always leave the picker up.

aaronbrethorst:
### Code review

Found 1 issue:

1. The escape hatch offered at the end of the new section — "(or turn location auto-select off themselves)" — does not work on the only launch where it matters, and the same mechanism makes the preamble's short-circuit rule wrong for steps 1→2. `RegionsService.init` computes `let autoSelectRegion = userDefaults.bool(forKey: RegionsService.automaticallySelectRegionUserDefaultsKey) || currentRegion == nil` (`OBAKitCore/Location/Regions/RegionsService.swift:93`). Because of the `|| currentRegion == nil`, whenever `currentRegion` is nil — which is the *only* state in which `FixedRegionName` can ever apply, per the `if currentRegion == nil, let fixedName = fixedRegionName` gate at :101 — the location match at :95-97 runs no matter what `OBAAutomaticallySelectRegionUserDefaultsKey` is registered to. An agency that follows this advice and ships the key as `false` still gets the physically-located region on first launch, i.e. exactly the failure the paragraph above it warns about; only the "ship a list with only that region" half of the sentence actually holds. That same `||` also breaks "later steps never run if an earlier one already set `currentRegion`" for step 1 → step 2: with auto-select on (the default), the location match *overwrites* a previously stored region rather than being skipped, which line 170's own "if auto-select is on (the default) *or* no region is stored yet" concedes — so the numbered list contradicts itself one line later. Suggest dropping the parenthetical (or replacing it with "don't ship a location-permission prompt", the only lever that actually nils out `locationService.currentLocation`) and rewording the preamble so step 1 is not presented as short-circuiting step 2.

https://github.com/OneBusAway/onebusaway-ios/blob/cc8ce062f8a08aee241763e2302c04d4c84cedec/OBAKit/OBAKit.docc/Articles/WhiteLabel.md#L174-L176

Borisserz:
Rewrote the FixedRegion section: dropped the auto-select-off escape hatch, and clarified that location can still run when `currentRegion` is nil (including first launch).

aaronbrethorst:
### Code review — correction

Disregard my previous comment: it reviewed `cc8ce06` and you'd already pushed past it. The precedence list now says an earlier location fix can take the region before the fixed-region keys are consulted, and drops the unachievable "turn auto-select off" remedy — which is exactly the correction I was asking for, and it's accurate against `RegionsService.init`.

Nothing outstanding from me on the content. Still needs the rebase.

aaronbrethorst (approved):
The precedence rewrite is accurate — I checked each claim against `RegionsService.init` and they hold: the name-then-URL fallback order, the URL fallback being gated on `FixedRegionName` being set, a successful match turning auto-select off, and neither key overriding a stored region. The "not a pin" framing is the right correction, and the newer commit also fixed the auto-select remedy sentence I flagged; retracted above.

Two leftovers from earlier rounds, both still non-blocking:

- The doc still does not warn that the base-URL match is a byte-exact `URL ==`, so a trailing-slash difference silently misses.
- `FixedRegionName` and `FixedRegionOBABaseURL` appear twice — bare bullets in one place and full descriptions in another — unlike the other three keys.

Also worth fixing while you are in there: the test file's doc comment still says white-label apps "pin a region in `OBAKitConfig` so the picker never appears," which reintroduces the exact framing the doc body now explicitly denies.

Approved on the merits — no re-review needed unless the conflict resolution changes behavior. Sibling PRs landed on `main` today and touched the same files, so this branch now has merge conflicts; please merge/rebase `main` in and resolve.

aaronbrethorst:
### Code review

Found 1 issue:

1. The new test's doc comment asserts exactly the guarantee this PR's own article was rewritten to deny. It says white-label apps "pin a region in `OBAKitConfig` so the picker never appears", while `WhiteLabel.md` four files over now opens the same subject with "These keys are **not a pin**." Both halves of the comment are falsifiable against `RegionsService.init`: the location match at `OBAKitCore/Location/Regions/RegionsService.swift:95-97` runs before the fixed-region gate at :101, so an earlier fix inside another region wins (asserted by `RegionsServiceAutoSelectTests.Location based selection takes priority`); and when the name matches nothing and the URL fallback misses, the block falls through to `Logger.error(...)` leaving `currentRegion` nil — with more than one active region in the list, the picker does appear. Since the deliverable here *is* documentation accuracy, and the last two review rounds were about this precise overstatement, the surviving copy of it in the test file is worth fixing. Rewording along the lines of "these keys are how `RegionsService` is configured at launch; see the precedence rules in `WhiteLabel.md`" would keep the pointer without the false guarantee.

https://github.com/OneBusAway/onebusaway-ios/blob/b970cb29550014b7f327ea3d0a1af16f50ce69d7/OBAKitTests/Extensions/FoundationExtensionsTests.swift#L220-L223

Everything else checked out. The merge of `main` is clean — no conflict markers, #1329's More Tab section and its `BundleFeedbackConfigTests` cases both survive intact, and the only non-additive hunk is one stray blank line dropped before the `// MARK: - String.normalizedSearchQuery` divider. I verified every claim in the new precedence list against `RegionsService.init` (stored region → location → fixed name → fixed URL → sole active region → picker; name-before-URL via `??`; auto-select written to `false` on a successful match so `updateCurrentRegionFromLocation`'s `automaticallySelectRegion` guard blocks later fixes; URL fallback gated on `fixedRegionName` being set) and they all hold.

aaronbrethorst (approved):
This reads well now. I checked every step of the precedence list against `RegionsService.init` and it's accurate — stored region, then location, then `FixedRegionName`, then the base-URL fallback, then a sole active region, then the picker. The "these keys are not a pin" framing is the right correction; that was the whole misconception behind #608.

One small thing to fix on your next pass through here — not worth another round on its own. The doc comment on the new test in `FoundationExtensionsTests.swift` still says white-label apps "pin a region in `OBAKitConfig` so the picker never appears", which is exactly the claim the article you're adding in this same PR refutes. Worth deleting or rewording so the two don't contradict each other.

Merging.

PR #1325, opened 2026-08-27
Move trip planner off MapViewController and lint errors in CI

Borisserz:
## Summary
- `MapViewController`'s type body crossed SwiftLint's error threshold (1000). Local builds failed; CI stayed green because `scripts/swiftlint.sh` no-ops when `$CI=true`.
- Trip planner methods move to `MapViewController+TripPlanner.swift`, same pattern as `+Camera` / `+MapLayers`. Type body is back under the warning line; the `type_body_length` disable on the class is gone.
- CI now runs `swiftlint lint --reporter github-actions-logging` after checkout. Errors fail the job. Warnings annotate and do not fail — `--strict` would collapse the warning/error split in `.swiftlint.yml`. The Xcode build phase still skips in CI so we do not double-lint.

The CLAUDE.md simulator destination is #1307, not this PR.

Addresses the SwiftLint half of #1303.

## Test plan
- [x] `swiftlint lint`: 0 errors; `MapViewController.swift` and `+TripPlanner.swift` clean
- [x] `build-for-testing` (scheme `App`, iPhone 17 Pro)
- [x] `FormattersTests` (14)
- [ ] Device: Map → trip planner → plan a trip. The overlay map still swaps in on `tripStarted`, and dismissing the planner restores the main map.

## Summary by CodeRabbit

- **New Features**
  - Added trip-planning support directly within the map experience.
  - Users can select transportation modes, view trip-planning results, and transition smoothly between the map and planner.
  - Added itinerary updates and trip-start handling, with GraphQL service support and REST fallback.

- **Chores**
  - Added dedicated SwiftLint error checks to the continuous integration workflow while allowing warnings to be reported without failing builds.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,6 +33,23 @@ jobs:
     steps:
     - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
+    # Fail on SwiftLint *errors* (the error-tier rules in .swiftlint.yml), so a
+    # type_body_length breach cannot land on main while staying green on GitHub.
+    # Warnings annotate via the GitHub reporter but do not fail the job —
+    # `--strict` would collapse that split. The Xcode build phase still skips
+    # in CI (scripts/swiftlint.sh); this step is the review-visible gate.
+    # See https://github.com/OneBusAway/onebusaway-ios/issues/1303
+    - name: SwiftLint (errors only)
+      env:
+        HOMEBREW_NO_AUTO_UPDATE: 1
+        HOMEBREW_NO_INSTALL_CLEANUP: 1
+      run: |
+        if ! command -v swiftlint >/dev/null; then
+          brew install swiftlint
+        fi
+        swiftlint lint --reporter github-actions-logging
+
+
     - uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
       with:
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Both halves of this hold up.

On the extraction: I verified mechanically rather than by eye, normalizing the removed lines from `MapViewController.swift` against the added lines in the new file. Every declaration in `MapViewController+TripPlanner.swift` came out of the class — there are no newly-written declarations — and the seven that stayed behind (`longPressGesture`, `semiModalTripPlannerController`, `tripPlanner`, `tripPlannerHostingController`, `tripPlannerMapView`, `floatingPanel`, `createSemiModalPanel`) are still there, just de-privatized. No delegate wiring, lifecycle call, or observer registration was dropped in the move. That's the property that matters for a refactor this size.

On the CI half: errors-only rather than `--strict` is the right choice — it keeps the 7 existing warnings from turning `main` red while still gating the thing #1303 is about. And removing the `type_body_length` disable is legitimate: the body is back under the *warning* threshold, not merely under the error threshold.

Filed #1340 for three follow-ups: `longPressGesture` didn't need de-privatizing (the extension never touches it — its only uses are at `MapViewController.swift:160-163`), the lint step sits inside the test job so one violation costs all test signal for that run, and `brew install swiftlint` is unpinned, which means a SwiftLint release can redden `main` with no code change.

PR #1328, opened 2026-08-28
Tick Live Activity countdown as 8m/NOW without a keepalive push

Borisserz:
## Summary
- Live Activity minutes were a snapshot from `let now = Date()` in `TripLiveActivityCardView`. They only moved when a keepalive push arrived.
- `TripCountdownFormatStyle` is a custom `DiscreteFormatStyle` that still renders `8m` / `NOW` (not `Text.timer` / `0:00`, which #1263 already ruled out). Minute boundaries are anchored on `departure`, not wall-clock minutes. After departure it stays `NOW` instead of counting up.
- Bright Lock Screen and Dynamic Island use `Text(.currentDate, format:)`. Dim Lock Screen (`isLuminanceReduced`) falls back to a static `format(Date())` so a custom style is not redacted to dashes.
- `CountdownView` gained a `departure: Date` init; stop-page rows keep passing `minutes`.
- Chips on the card also tick. `upcomingArrivals(now:)` and the deviation label are still snapshots from the last push — a departed bus still needs a content update to drop off the card. obacloud keepalive relaxations stay out of this PR.

Device confirmation of always-on redaction is still the first item on #1187; this is approach A from that issue.

Addresses #1187.

## Test plan
- [x] `TripCountdownFormatStyleTests` (6/6), including `discreteInput` boundaries and the 1m → NOW step
- [ ] Device: start a Live Activity, leave the phone locked. Bright Lock Screen and Dynamic Island should tick `8m` → `7m` → `NOW` with no keepalive. Dim always-on should show digits, not `—m`.
- [ ] Device: chips next to the headline should tick with it. A bus that has already left should stay on the card until the next push.

## Summary by CodeRabbit

* **New Features**
  * Added live-updating departure countdowns to trip cards and Live Activities.
  * Countdown displays refresh at minute boundaries and show “NOW” at departure time.
  * Countdown text uses localized minute formatting.
  * Lock Screen countdowns remain readable when the display is dimmed.

* **Bug Fixes**
  * Replaced static minute labels with accurate, continuously updating countdowns across supported trip views.

* **Tests**
  * Added coverage for countdown formatting, minute boundaries, localization, and departure transitions.

### DIFF
--- a/OBAKitCore/UI/TripCountdownFormatStyle.swift
+++ b/OBAKitCore/UI/TripCountdownFormatStyle.swift
@@ -1,0 +1,105 @@
+//
+//  TripCountdownFormatStyle.swift
+//  OBAKitCore
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import Foundation
+import SwiftUI
+
+/// Live Activity countdown that keeps the `8m` / `NOW` typography and still
+/// ticks without a keepalive push.
+///
+/// `Text.timer` was tried in #1263 and rejected: `0:00` is not how this app
+/// shows a departure. A custom `DiscreteFormatStyle` is how SwiftUI schedules
+/// the next redraw (minute boundaries *anchored on `departure`*, not on the
+/// wall clock). Past departure it stays `NOW` rather than counting up.
+///
+/// Always-on display redacts unknown format styles. `TickingCountdownText`
+/// falls back to a static `format(Date())` when `isLuminanceReduced` is set,
+/// so dim Lock Screen shows dashes-free digits from the last evaluation.
+///
+/// See: https://github.com/OneBusAway/onebusaway-ios/issues/1187
+public struct TripCountdownFormatStyle: DiscreteFormatStyle, Sendable {
+    public typealias FormatInput = Date
+    public typealias FormatOutput = String
+
+    public var departure: Date
+
+    public init(departure: Date) {
+        self.departure = departure
+    }
+
+    public func format(_ now: Date) -> String {
+        let minutes = Int(departure.timeIntervalSince(now) / 60.0)
+        if minutes <= 0 {
+            return OBALoc(
+                "stop_page.countdown.now",
+                value: "NOW",
+                comment: "Shown in place of the minutes countdown when the vehicle is departing now"
+            )
+        }
+        return "\(minutes)m"
+    }
+
+    public func discreteInput(after date: Date) -> Date? {
+        let minutes = Int(departure.timeIntervalSince(date) / 60.0)
+        guard minutes > 0 else { return nil }
+
+        let boundary = departure.addingTimeInterval(-TimeInterval(minutes) * 60)
+        if boundary > date { return boundary }
+
+        // Sitting on this minute's boundary. Next flip is the next lower
+        // minute, except 1m → NOW which starts the instant remaining < 60s.
+        if minutes == 1 {
+            return date.addingTimeInterval(1)
+        }
+        return departure.addingTimeInterval(-TimeInterval(minutes - 1) * 60)
+    }
+
+    public func discreteInput(before date: Date) -> Date? {
+        let minutes = Int(departure.timeIntervalSince(date) / 60.0)
+        if minutes < 0 { return nil }
+
+        if minutes == 0 {
+            let start = departure.addingTimeInterval(-60)
+            return start < date ? start : nil
+        }
+
+        let startOfCurrent = departure.addingTimeInterval(-TimeInterval(minutes + 1) * 60)
+        return startOfCurrent < date ? startOfCurrent : nil
+    }
+}
+
+/// `8m` / `NOW` that self-updates on a Live Activity. Dim Lock Screen uses a
+/// static snapshot so a custom format style is not redacted to dashes.
+public struct TickingCountdownText: View {
+    public let departure: Date
+    public let font: Font
+    public let color: Color
+
+    @Environment(\.isLuminanceReduced) private var isLuminanceReduced
+
+    public init(departure: Date, font: Font, color: Color) {
+        self.departure = departure
+        self.font = font
+        self.color = color
+    }
+
+    public var body: some View {
+        let style = TripCountdownFormatStyle(departure: departure)
+        Group {
+            if isLuminanceReduced {
+                Text(style.format(Date()))
+            } else {
+                Text(.currentDate, format: style)
+            }
+        }
+        .font(font)
+        .monospacedDigit()
+        .foregroundStyle(color)
+    }
+}
--- a/OBAKitCore/UI/TripCountdownFormatStyle.swift
+++ b/OBAKitCore/UI/TripCountdownFormatStyle.swift
@@ -42,7 +42,15 @@ public struct TripCountdownFormatStyle: DiscreteFormatStyle, Sendable {
                 comment: "Shown in place of the minutes countdown when the vehicle is departing now"
             )
         }
-        return "\(minutes)m"
+
+        // Same key as `Formatters.shortFormattedTime` so the Dynamic Island
+        // keeps ar/fr/ko/ru/vi/zh suffixes instead of a hardcoded English `m`.
+        let formatString = OBALoc(
+            "formatters.short_time_fmt",
+            value: "%dm",
+            comment: "Short formatted time text for arrivals/departures. Example: 7m means that this event happens 7 minutes in the future. -7m means 7 minutes in the past."
+        )
+        return String(format: formatString, minutes)
     }
 
     public func discreteInput(after date: Date) -> Date? {
--- a/OBAKitCore/UI/TripCountdownFormatStyle.swift
+++ b/OBAKitCore/UI/TripCountdownFormatStyle.swift
@@ -57,15 +57,16 @@ public struct TripCountdownFormatStyle: DiscreteFormatStyle, Sendable {
         let minutes = Int(departure.timeIntervalSince(date) / 60.0)
         guard minutes > 0 else { return nil }
 
+        // Truncating `format` flips from N → N-1 the instant remaining drops
+        // *below* N*60, i.e. just after `departure - N*60`. Returning the
+        // floor itself leaves TimeDataSource displaying `format(floor)` (== N)
+        // for a full minute — so 2m shows while only 61s remain (#1187 follow-up).
         let boundary = departure.addingTimeInterval(-TimeInterval(minutes) * 60)
         if boundary > date { return boundary }
 
-        // Sitting on this minute's boundary. Next flip is the next lower
-        // minute, except 1m → NOW which starts the instant remaining < 60s.
-        if minutes == 1 {
-            return date.addingTimeInterval(1)
-        }
-        return departure.addingTimeInterval(-TimeInterval(minutes - 1) * 60)
+        // On the floor: step one second into the lower bucket so `format` flips
+        // (same treatment for every N, including 1m → NOW).
+        return date.addingTimeInterval(1)
     }
 
     public func discreteInput(before date: Date) -> Date? {
--- a/OBAWidget/Widgets/TripLiveActivity.swift
+++ b/OBAWidget/Widgets/TripLiveActivity.swift
@@ -57,12 +57,12 @@ struct TripLiveActivity: Widget {
                 }
                 DynamicIslandExpandedRegion(.trailing) {
                     if let primary {
-                        let primaryMinuteText = presenter.minuteText(for: primary)
-                        Text(primaryMinuteText)
-                            .font(.system(size: 32, weight: .bold, design: .rounded))
-                            .foregroundColor(Color(presenter.color(for: primary)))
-                            .contentTransition(.numericText(value: Double(primaryMinuteText.filter("0123456789".contains)) ?? 0))
-                            .padding(.trailing, 6)
+                        TickingCountdownText(
+                            departure: primary.departureDate,
+                            font: .system(size: 32, weight: .bold, design: .rounded),
+                            color: Color(presenter.color(for: primary))
+                        )
+                        .padding(.trailing, 6)
                     }
                 }
                 DynamicIslandExpandedRegion(.bottom) {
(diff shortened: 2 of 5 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The Dynamic Island loses its localized minute suffix. All four Dynamic Island regions previously rendered `presenter.minuteText(for:)`, which goes through `Formatters.shortFormattedTime(untilMinutes:temporalState:)` and its localized `"formatters.short_time_fmt"` key. This PR replaces every one of them (`TripLiveActivity.swift` L60, L99, L119, L128) with `TripCountdownFormatStyle.format`, which hardcodes `"\(minutes)m"`. Seven of the thirteen shipped locales translate that key away from `%dm` — ar `"%d د"`, zh-Hans `"%d分钟"`, zh-Hant `"%d分"`, ko `"%d분"`, ru `"%d мин"`, fr `"%d min"`, vi `"%dp"` — so those users now get English `8m` in the Dynamic Island. (The lock-screen card already hardcoded the suffix, so nothing regresses there; the Dynamic Island did not.) Routing the format style through `Formatters.shortFormattedTime` would fix it and would also keep `TripActivityPresenter.minuteText` from becoming dead code — after this diff its only remaining reference is `TripActivityPresenterTests`.

https://github.com/OneBusAway/onebusaway-ios/blob/fbeb1df1f129164927113b94fe311e4444559c50/OBAKitCore/UI/TripCountdownFormatStyle.swift#L35-L46

OBAKitCore/UI/TripCountdownFormatStyle.swift
```diff
@@ -0,0 +1,105 @@
...
+    public func format(_ now: Date) -> String {
+        let minutes = Int(departure.timeIntervalSince(now) / 60.0)
+        if minutes <= 0 {
+            return OBALoc(
+                "stop_page.countdown.now",
+                value: "NOW",
+                comment: "Shown in place of the minutes countdown when the vehicle is departing now"
+            )
+        }
+        return "\(minutes)m"
```
aaronbrethorst: This hardcodes the English minute suffix. The path it replaces on the Dynamic Island (`presenter.minuteText(for:)` → `Formatters.shortFormattedTime(untilMinutes:temporalState:)`) resolves it through `OBALoc("formatters.short_time_fmt", value: "%dm")`, which seven shipped locales translate away from `%dm`:

| locale | value |
|---|---|
| ar | `%d د` |
| fr | `%d min` |
| ko | `%d분` |
| ru | `%d мин` |
| vi | `%dp` |
| zh-Hans | `%d分钟` |
| zh-Hant | `%d分` |

Suggest calling `Formatters.shortFormattedTime` here instead of building the string inline. That restores the localized suffix and stops `minuteText` becoming dead code.

(The `NOW` branch above is fine — `stop_page.countdown.now` and `formatters.now` are translated identically in every locale, so that swap is harmless.)

OBAWidget/Widgets/TripLiveActivity.swift
```diff
@@ -57,12 +57,12 @@ struct TripLiveActivity: Widget {
                 }
                 DynamicIslandExpandedRegion(.trailing) {
                     if let primary {
-                        let primaryMinuteText = presenter.minuteText(for: primary)
-                        Text(primaryMinuteText)
-                            .font(.system(size: 32, weight: .bold, design: .rounded))
-                            .foregroundColor(Color(presenter.color(for: primary)))
-                            .contentTransition(.numericText(value: Double(primaryMinuteText.filter("0123456789".contains)) ?? 0))
-                            .padding(.trailing, 6)
+                        TickingCountdownText(
```
aaronbrethorst, line 70: This is one of the four Dynamic Island sites that previously went through `presenter.minuteText(for:)` and so picked up the localized suffix. The lock-screen card already hardcoded `"\(minutes)m"`, so that surface is unchanged — but the Dynamic Island regressions here and at lines 99, 119, and 128 are new. Fixing the format style itself covers all four.

aaronbrethorst (changes requested):
Really nice piece of work — the `DiscreteFormatStyle` approach is the right call, and anchoring the redraw schedule on `departure` rather than the wall clock is exactly the detail that makes this tick correctly. The `isLuminanceReduced` fallback for always-on display is a thoughtful touch too, and the six format/`discreteInput` tests are well chosen.

One blocker before this can land, flagged inline: the four Dynamic Island call sites used to run through `presenter.minuteText(for:)`, which resolves the minute suffix via `OBALoc("formatters.short_time_fmt")`. The new format style hardcodes `"\(minutes)m"`, so seven of our thirteen shipped locales would start seeing English on the app's most glanceable surface — ar `%d د`, fr `%d min`, ko `%d분`, ru `%d мин`, vi `%dp`, zh-Hans `%d分钟`, zh-Hant `%d分`.

It's a small mechanical fix (route the style's non-zero branch through `Formatters.shortFormattedTime`, which also keeps `minuteText` from becoming dead code), and everything else here holds up — I checked the arithmetic and it's pure `TimeInterval` with no `Calendar` or timezone exposure, so there's no DST surface at all.

Two things worth a look while you're in here, neither blocking:

- `discreteInput(before:)` returns `nil` for inputs at or past `departure + 60s`, because `Int(-1.5) == -1` trips the `minutes < 0` guard. Practically inert today, but it does violate the contract.
- `CountdownView.init(departure:)` fabricates `minutes = 0`, leaving a `public let` that lies in one of the type's two states. Worth considering whether that storage should be an enum instead.

Also: your own test plan still has the two device checkboxes unticked, and #1187 names the always-on-display redaction question as the thing that gates this design. Please confirm on hardware before we land it.

Please take a look at the inline comments — happy to re-review as soon as the localization path is restored.

OBAKitCore/UI/TripCountdownFormatStyle.swift
Borisserz: The non-zero branch now uses `formatters.short_time_fmt` the same way `Formatters.shortFormattedTime` does. A static `Formatters` failed Swift 6 Sendable, so this calls `OBALoc` + `String(format:)` instead of allocating a formatter on every tick. `TripCountdownFormatStyleTests` locks that it still matches `shortFormattedTime`.

OBAWidget/Widgets/TripLiveActivity.swift
Borisserz: Covered by the format-style change — all four Dynamic Island sites go through `TickingCountdownText` / `TripCountdownFormatStyle`.

aaronbrethorst:
### Code review

Found 1 issue:

1. `discreteInput(after:)` returns a boundary at which `format` has not yet changed, so the ticking label sits one minute above its own value and `1m` is on screen for one second (bug due to `TripCountdownFormatStyle.discreteInput(after:)`).

   `format` truncates — `Int(remaining / 60.0)` — so the label flips from N to N-1 the instant `remaining` drops *below* `N*60`, i.e. just after `departure - N*60`. But `discreteInput(after:)` returns `departure - N*60` itself (L60–61), where `format` still yields N, and only the *following* call returns `departure - (N-1)*60` (L68). `TimeDataSource` pre-computes the string for each interval (the extension isn't running when a Live Activity updates), so `format(boundary)` is what's displayed until the next boundary:

   | interval | displayed | actual `format` |
   |---|---|---|
   | `[d-480, d-420)` | `8m` | `8m` → `7m` |
   | `[d-120, d-60)` | `2m` | `1m` |
   | `[d-60, d-59)` | `1m` | `1m` |
   | `[d-59, …)` | `NOW` | `NOW` |

   With 61 seconds left the card reads `2m`, and `1m` is visible for a single second before flipping to `NOW` — the state the rider most needs is the one that's skipped.

   Three corroborating signals that this is the bug and not the intended rounding:
   - `discreteInput(before:)` uses the correct truncating bucket start, `departure - (minutes + 1) * 60` (L80). So `before` and `after` disagree about where the boundaries are, which `DiscreteFormatStyle` forbids: "The ordering over `FormatInput` defined by `discreteInput(before:)` / `discreteInput(after:)` must be consistent between the two functions."
   - The `minutes == 1` branch (L65–67) already concedes the point — it advances one second past the boundary *precisely because the boundary itself doesn't flip*. The `minutes > 1` branch skips that step, which is the whole discrepancy.
   - The `isLuminanceReduced` fallback calls `format(Date())` directly, and every in-app surface (`CountdownView(minutes:)`, `arrivalDepartureMinutes`, `TripActivityPresenter.minuteText`) keeps truncating. So the bright Lock Screen would read a minute higher than both the dim Lock Screen and the app, for the same bus at the same instant.

   Applying the `minutes == 1` treatment uniformly (return the boundary plus a small epsilon whenever `boundary <= date`, or equivalently return `departure - minutes*60` only when it's strictly greater than `date` and otherwise step just past it) makes `after` agree with `before` and with `format`.

https://github.com/OneBusAway/onebusaway-ios/blob/124f58ac4a534eb4c379fa6dc6561847df949653/OBAKitCore/UI/TripCountdownFormatStyle.swift#L55-L70

Borisserz:
Fixed `discreteInput(after:)`: on each minute floor we now step +1s into the lower truncating bucket so `format` flips immediately (same treatment as 1m→NOW). Added a regression that at the 2m floor the next tick formats as `1m`.

aaronbrethorst (changes requested):
The localization blocker from last round is genuinely fixed, and fixed well — routing through `OBALoc("formatters.short_time_fmt")`, the same key `Formatters.shortFormattedTime` uses, keeps the ar/fr/ko/ru/vi/zh suffixes at all four Island sites, and there's a regression test pinning the coupling. Your reasoning for inlining `OBALoc` + `String(format:)` rather than holding a `Formatters` (Sendable, no per-tick allocation) is sound.

The mechanism is right too: `Text(.currentDate, format:)` with a `DiscreteFormatStyle` is the supported path for a self-updating Live Activity, so this genuinely ticks without a push, and the `isLuminanceReduced` fallback is justified.

What blocks it is *when* it ticks:

**`discreteInput(after:)` returns a boundary where `format` hasn't changed.** `format` truncates via `Int(remaining/60)`, so the label should flip from N to N−1 just *after* `departure - N*60`. But `discreteInput(after:)` returns `departure - N*60` itself, where `format` still yields N; only the next call returns the following bucket. Since the string is pre-computed per interval, the rendered value sits a minute high for essentially the whole run — 61 seconds out reads `2m` — and `1m` shows for exactly one second before flipping to `NOW`.

Three things corroborate it rather than one: `discreteInput(before:)` uses the *correct* truncating bucket start, so the two functions disagree about where boundaries are, which the `DiscreteFormatStyle` contract explicitly forbids; your own `minutes == 1` branch already advances one second past the boundary for precisely this reason, and the `minutes > 1` branch omits that step; and the `isLuminanceReduced` fallback calls `format(Date())` directly, so the bright Lock Screen would read a minute higher than both the dim one and the app.

Applying the `minutes == 1` epsilon step uniformly should do it.

Still outstanding from last round, both previously non-blocking: `discreteInput(before:)` returns `nil` at/after `departure + 60s` (`Int(-1.5) == -1` trips the `minutes < 0` guard), and `CountdownView.init(departure:)` fabricates `minutes = 0` on a `public let`. And the two device checkboxes in your test plan are still unticked — given #1187 is about exactly this rendering, those are the real gate.

One more: `Future minutes match Formatters.shortFormattedTime` can't fail — `shortFormattedTime` ignores its own `locale` parameter and calls `OBALoc` too, so both sides resolve through the same bundle in any locale.

Sequencing: #1379 rewrites the same four Island regions and the same card file. Worth agreeing an order with it rather than resolving blind.

Borisserz:
Fixed the `discreteInput(after:)` floor bug: when `date` is already on the minute boundary, we now step +1s into the lower bucket so `format` flips immediately (same treatment for every N, including 1m → NOW). Tests cover the 2m→1m and 1m→NOW cases; also fixed a `try #require` compile error in the suite.

Borisserz:
The `discreteInput(after:)` floor bug from this review is fixed on the current head (`e91e404f` / `fb93abd8`): when `date` is already on the minute boundary we step +1s into the lower truncating bucket for every N (not only `minutes == 1`), so `format` flips immediately and `before`/`after` agree. Regression covers 2m→1m and 1m→NOW.

Happy to take `discreteInput(before:)` at/after `departure+60s` and the vacuous `shortFormattedTime` coupling test in a follow-up if you'd rather keep this PR to the tick fix. Device checkboxes still need a physical run — can do that next.

Sequencing vs #1379: I'll land the stale-warning opacity fix on #1379 next, then rebase whichever is second onto the first if they collide on the Island regions.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Approved on the merits — both of my blockers are genuinely fixed.

`format()` now routes through `OBALoc("formatters.short_time_fmt")`, the same key `Formatters.shortFormattedTime` uses, and I confirmed all 13 locales resolve it. And dropping the `minutes == 1` special case so `discreteInput(after:)` steps uniformly is exactly right: my old repro — 61 seconds out reading "2m" — now reads "1m", and the previous code held "2m" across that whole window.

The mechanism is the supported one. `Text(.currentDate, format:)` over a `DiscreteFormatStyle` is what Apple documents for live-updating values in Live Activities, there's no timer anywhere, and the `isLuminanceReduced` fallback is well-founded rather than superstition — an unrecognized format style really does get its digits redacted to dashes on the always-on display.

**Heads up: now that #1379 landed on `main` and touched the same four Dynamic Island regions, this branch has conflicts.** Please merge `main` in and resolve. This is the resolution I'd like, since it's more than textual: #1379 added `.opacity(staleOpacity)` and an accessibility label to exactly the `Text` views you're replacing with `TickingCountdownText`. Rather than re-applying those modifiers around the new type, give `TickingCountdownText` its own opacity and accessibility-label parameters so the staleness chrome composes properly. It's your type, so it's cleaner for you to design that in than for the next person to bolt it on.

No re-review needed unless the resolution changes behavior beyond that.

One follow-up worth its own issue, not this PR: the new Live Activity path localizes the minute suffix, but the stop-page `minutes:` path still hardcodes `"\(minutes)m"`. Pre-existing and unchanged by you — but the two paths now disagree, and that's the repo's most common defect class.

Also FYI, no `lint` check ran here — your branch predates that job. Build is green, so nothing to do.

Borisserz:
Merged `main` and resolved the #1379 overlap by adding `opacity` / `accessibilityLabel` parameters on `TickingCountdownText`, so stale chrome composes on the ticking view rather than re-wrapping modifiers at each Island site.

aaronbrethorst:
### Code review

_Re-review at 2008970, covering commits since the previous review._

aaronbrethorst (approved):
Verified at 3de3635c. Thanks for this.

PR #1329, opened 2026-08-28
Document MoreTab white-label keys

### LINKED ISSUE #666: Customizing the settings tab
It would be great to be able to customize the 'header box' and 'menu items' in the settings tab:
about
settings

menu items could 
- have the links be changed to point to different pages (set in the project.yml)
- be changed completely

header box content to be changed also in project.yml so that it could contain extra information.

Borisserz:
## Summary
- `OBAKitConfig.MoreTab` already drives the More tab header, Help Out section, agency contact rows, and custom Resources links. WhiteLabel.md did not list any of those keys, so agencies still treated the settings tab as a fork.
- Adds the missing `MoreTab` section (with a KiedyBus-style example) and Bundle tests that read the nested dictionary the same way `project.yml` ships it.

Closes #666.

## Test plan
- [x] `BundleFeedbackConfigTests` (9/9), including the two new MoreTab cases
- [x] `MoreTabConfigurationTests` already covered dictionary parsing; this PR only adds the Bundle → `MoreTab` path
- [ ] Open WhiteLabel.md in DocC / GitHub and confirm the YAML example matches `Apps/KiedyBus/project.yml`

### DIFF
--- a/OBAKit/OBAKit.docc/Articles/WhiteLabel.md
+++ b/OBAKit/OBAKit.docc/Articles/WhiteLabel.md
@@ -171,3 +171,28 @@ Stop pages can hide departures that lack real-time data, or hide the real-time o
 * `scheduledOnly` - Show only departures without real-time data.
 
 This is a starting value, not a lock: riders can change it in Settings under Arrival & Departure Display, or from the Departure Type menu on any stop page, and their choice is remembered from then on. An unrecognized value is ignored and treated as `all`.
+
+### More Tab
+
+* `MoreTab` - Optional dictionary. Customizes the More (settings) tab header and menu items without forking the UI. Omit the whole key to keep OneBusAway defaults. KiedyBus is a working example.
+
+Nested keys (all optional):
+
+* `HeaderSupportText` - String shown in the branded header box. Omit or leave empty to keep the default volunteer-support copy.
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Thank you for writing this up — `MoreTab` has been undocumented since #1104 and #1261 shipped the capability, and this is the reference an agency actually needs.

I checked every claim against source rather than taking the doc's word for it: the key names, types, and defaults all match `MoreTabConfiguration.swift`, the UI strings you name are exact against `MoreViewController.swift`, "OneBusAway currently omits `TranslateURL`" matches `Apps/OneBusAway/project.yml`, and the YAML example is a faithful copy of `Apps/KiedyBus/project.yml`. The `URL(string: "")` → nil behavior you rely on for "an empty string hides the row" checks out too.

The two characterization tests are a genuine addition — they cover the Bundle→`MoreTab` nesting that `MoreTabConfigurationTests` didn't reach.

One nuance for a future pass, filed as #1339: an empty `DevelopURL` does hide that row, but if `TranslateURL` is also absent the `guard !contents.isEmpty` in `helpOutSection` drops the entire "Help make the app better" section, not just the row.

PR #1330, opened 2026-08-30
Filter Live Activity arrivals by route and destination

### LINKED ISSUE #1326: Live Activities give incorrect information at transit centers
From a user:

> Setting a lock screen alert/tracking at Transit Centers like Mountlake Terrace causes inaccurate arrival times to be displayed on the lock screen.  The alerts appear to be set on route number alone, rather than route plus destination, meaning if you are at a stop that serves the same route in both directions, you will see the next arrival in either direction rather than the direction that matters to you.

Borisserz:
## Summary
- Tracking from a stop at a transit center built Live Activity content with `filter { \$0.routeID == departure.routeID }`. The card title used the tapped headsign; the countdown and chips used whichever direction left first.
- Content state now uses the same `TripBookmarkKey` as trip bookmarks: stop + route + headsign, soonest first, at most three. If the stop list no longer contains that trip, it falls back to the tapped departure.
- Bookmark Track was already keyed this way. OBACloud registration already sends `trip_headsign`; this PR does not change the push payload shape. If a keepalive still mixes directions after this, that is server-side.

Closes #1326.

## Test plan
- [x] `BookmarkActionsTests` (9/9), including opposite-direction drop and stale-list fallback
- [ ] Device: at a stop that serves the same route both ways (e.g. Mountlake Terrace), Track one destination. Lock Screen headline and chips should only be that direction, even if the other way leaves sooner.

## Summary by CodeRabbit

- **Bug Fixes**
  - Live activity updates now show only arrivals matching the selected departure’s stop, route, and destination.
  - Arrivals are displayed in chronological order.
  - If the tracked departure is no longer listed, the activity continues displaying that departure instead of selecting an unrelated arrival.
- **Tests**
  - Added coverage for filtering opposite-direction arrivals and preserving the selected departure as a fallback.

### DIFF
--- a/OBAKit/Bookmarks/BookmarkActions.swift
+++ b/OBAKit/Bookmarks/BookmarkActions.swift
@@ -94,6 +94,21 @@ final class BookmarkActions {
         return TripAttributes.ContentState(arrivals: Array(arrivals))
     }
 
+    /// Builds content for the trip the rider actually tracked: same stop, route,
+    /// and destination as `departure`. Route-only matching mixes both directions
+    /// at a transit center and shows the opposite bus's countdown (#1326).
+    static func buildContentState(
+        from arrivalDepartures: [ArrivalDeparture],
+        matching departure: ArrivalDeparture
+    ) -> TripAttributes.ContentState? {
+        let key = TripBookmarkKey(arrivalDeparture: departure)
+        let sameTrip = arrivalDepartures
+            .filter { TripBookmarkKey(arrivalDeparture: $0) == key }
+            .sorted { $0.arrivalDepartureDate < $1.arrivalDepartureDate }
+        let source = sameTrip.isEmpty ? [departure] : sameTrip
+        return buildContentState(from: source)
+    }
+
     @discardableResult
     func startLiveActivity(for bookmark: Bookmark, arrivalDepartures: [ArrivalDeparture]) -> TrackResult {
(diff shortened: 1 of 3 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Right fix, keyed to the right thing. Matching on `TripBookmarkKey` rather than route ID means the stop page and trip bookmarks now agree on what "the same trip" is, which is the property that makes this correct rather than just narrower.

Two things I verified before merging. The cap survived the refactor — `prefix(3)` lives inside `buildContentState(from:)`, so moving the filter out didn't quietly uncap the arrival list. And the fallback can't go blank: `sameTrip.isEmpty ? [departure] : sameTrip` guarantees a non-empty source on every path, including `stopArrivals == nil`.

I also checked that the rest of the pipeline agrees with the finer key — `ObacoAPIService.postLiveActivity` already sends `trip_headsign`, `TripAttributes.StaticData.tracksSameTrip(as:)` already compares `routeHeadsign`, and the keepalive path sources arrivals through `BookmarksViewModel.arrivalDepartures(for:)`, which is already `TripBookmarkKey`-filtered. So nothing downstream re-mixes the directions you just separated.

The two tests hit exactly the two interesting branches.

Filed #1337 for the signature: `buildContentState(from:matching:)` can never actually return nil, which makes the `guard let` and both `try #require`s dead branches.

PR #1332, opened 2026-08-31
Silence Bookmarks-tab bulletin when a stop returns JSON null

### LINKED ISSUE #1331: Unable to access bookmarks tab after deletion of bus route
Upon selecting the bookmarks tab I encounter an error message of "Expected to receive json data from the server, but we received nothing instead." Upon tapping the dismiss button nothing happens, and force closing and reopening the app only allows a small window to switch tabs before the error message pops back up.

Borisserz:
## Summary
- HTTP 200 with body `null` on a trip bookmark is silent: no bulletin, no failure haptic, the card settles on "No upcoming departures", and any previous departures for that stop are cleared.
- That is the copy riders see as **"Expected to receive json data from the server, but we received nothing instead."** — `APIService+GetData` throws it as `invalidContentType(..., "json", "nothing")`. Many OBA servers send this instead of a 404 when a stop/route is gone.
- **Empty HTTP 200** still bulletins (San Diego live-stop finding from #1268). Literal HTTP 404 stays silent.

Closes #1331

## Why dismiss looked broken
The Bookmarks tab re-fetches on appear and every 30s. Each failed fetch called `displayError`, so tapping Dismiss was immediately followed by another bulletin. Stopping the false error is the fix; the bookmark itself is left for the rider to delete.

## Test plan
- [x] JSON `null` body — no `displayError`, `hasLoadedArrivalData == true`
- [x] JSON `null` after a successful fetch — stale departures cleared
- [x] Empty HTTP 200 — still `displayError`, `lastRefreshHadError`
- [x] Literal HTTP 404 — still silent
- [x] HTTP 500 — still `displayError`
- [ ] Manual: Bookmarks tab with a bookmark whose stop/route has been deleted — no blocking bulletin; card shows "No upcoming departures"

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of stops removed from the server when responses contain a JSON `null` value.
  * Removed stops no longer trigger an error message or mark refreshes as failed.
  * Previously loaded departures are cleared when a stop is confirmed to be unavailable.
  * Empty responses continue to be treated as temporary errors.

### DIFF
--- a/OBAKitCore/Bookmarks/BookmarkDataLoader.swift
+++ b/OBAKitCore/Bookmarks/BookmarkDataLoader.swift
@@ -53,13 +53,14 @@ public class BookmarkDataLoader: NSObject {
     @MainActor private var batchContinuations: [UInt64: [CheckedContinuation<Void, Never>]] = [:]
 
     /// Stops whose arrival fetch has finished this session — a successful
-    /// payload **or** a literal HTTP 404. Lets consumers distinguish "still
-    /// loading" from "loaded, but no upcoming departures". Empty HTTP 200
-    /// (also thrown as `APIError.requestNotFound`) is not recorded here.
+    /// payload, a literal HTTP 404, or HTTP 200 with body `null`. Lets
+    /// consumers distinguish "still loading" from "loaded, but no upcoming
+    /// departures". Empty HTTP 200 (also thrown as `APIError.requestNotFound`)
+    /// is not recorded here.
     @MainActor private var fetchedStopIDs = Set<StopID>()
 
     /// `true` once an arrival fetch for `stopID` has finished this session
-    /// (success or HTTP 404).
+    /// (success, HTTP 404, or JSON `null`).
     @MainActor public func hasFetchedData(forStopID stopID: StopID) -> Bool {
         fetchedStopIDs.contains(stopID)
     }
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
The diagnosis here is well grounded, and I checked the part that matters most: whether this quietly swallows real errors. It doesn't. `actualContentType == "nothing"` is produced at exactly one site — the literal-`null` branch in `APIService+GetData.swift:90` — so the predicate is a precise discriminator rather than a broad catch. Empty-200, 404-without-404-status, 500, and decode failures all still reach `displayError`.

I also want to note this is *not* a revert of #1268. That PR narrowed the catch because empty-200 was over-matching, and you've explicitly kept empty-200 as a bulletin, San Diego trace comment and all. JSON-`null` is a genuinely distinct shape. Good, surgical widening.

Tests exercise the real path too — `MockDataLoader.buildURLResponse` sets a real `Content-Type` and `Content-Length`, so the mock clears both gates and lands on the `null` branch rather than passing for an unrelated reason.

Filed #1336 for the gap this leaves: `StopViewModel.swift:299` still gates the broken-bookmark UI on `requestNotFound` only, so tapping through from a silenced card lands on a generic error with no delete path. Worth closing so the two surfaces agree.

PR #1351, opened 2026-09-02
Match the Current Trip no-location button to the authorization

### LINKED ISSUE #1343: Current Trip retry comment misstates the 20s auto-refresh
Follow-up from #1317, which added a Try Again button to the `.noLocation` empty state on Current Trip.

The added comment says:

> Without the button the only way to re-run it is to leave the screen and return.

That isn't right. `CurrentTripViewModel.startRefreshTimer()` already re-runs `findVehicle(resetState: false)` every 20 seconds while the screen is visible, in *any* state including `.noLocation`. A future reader will conclude there's no auto-retry here when there is. The PR body repeats the same claim.

The button's strongest real justification is accessibility: `shouldSkipProgrammaticRefresh` disables that 20 s timer when VoiceOver is running, so for VoiceOver users the button genuinely *is* the only in-screen recovery. Worth saying that instead.

Separately, worth considering: when authorization is denied or location services are off, "Try Again" can only ever flash `.loading` and fail. The repo already has the better affordance — `MapViewController.didTapMapStatus` (`OBAKit/Mapping/MapViewController.swift:389`) offers "Turn On in Settings" via `UIApplication.openSettingsURLString`, and `StopPageActionPresenter.swift:547` and `MapPanelRootView.swift:512` do the same. An enhancement, not a regression.

diveshpatil9104:
## Summary

- Fixes the comment you flagged. `CurrentTripViewModel.startRefreshTimer()` does re-run `findVehicle(resetState: false)` every 20 seconds in every state, including `.noLocation`. The old comment could have sent future readers looking for an auto-retry that was already there.
  - The comment now explains the actual reason the button matters: `shouldSkipProgrammaticRefresh` suppresses that timer tick under VoiceOver, making the button the only in-screen recovery path for those riders.

- Also took the enhancement you raised. **"Try Again"** now only appears when a retry can actually do something.

- Returning from Settings fires no view lifecycle callback, and under VoiceOver the 20-second tick that would otherwise catch up is disabled. The controller therefore re-runs the search on `didBecomeActive` while in `.noLocation`.
  - Without this, the Settings button would dead-end for exactly the riders it exists for.

- No new localization keys were added. Both labels reuse keys that already exist in all thirteen locales.

- Added five tests covering every branch of the mapping.

Closes #1343

## Summary by CodeRabbit

- **New Features**
  - Improved the Current Trip experience when location access is unavailable.
  - Users can now request location permission, open system Settings to update access, retry vehicle lookup, or receive clear guidance when recovery isn’t possible.
  - Vehicle information automatically refreshes when returning from Settings after updating location access.
  - Added a Settings icon for controls that open the system Settings app.

- **Bug Fixes**
  - Improved handling of location authorization states to provide the appropriate recovery action.

### DIFF
--- a/OBAKit/Theme/Icons.swift
+++ b/OBAKit/Theme/Icons.swift
@@ -118,6 +118,11 @@ nonisolated class Icons: NSObject {
         configureForButtonIcon(systemImage(named: "line.horizontal.3.decrease"))
     }
 
+    /// A gear icon, for controls that send the user to the system Settings app.
+    public class var settings: UIImage {
+        configureForButtonIcon(systemImage(named: "gearshape"))
+    }
+
     public class var addAlarm: UIImage {
         imageNamed("add_alarm")
     }
--- a/OBAKit/Trip/CurrentTripViewController.swift
+++ b/OBAKit/Trip/CurrentTripViewController.swift
@@ -65,12 +65,30 @@ class CurrentTripViewController: UIViewController,
         super.viewWillAppear(animated)
         disableIdleTimer()
         viewModel.start()
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(applicationDidBecomeActive),
+            name: UIApplication.didBecomeActiveNotification,
+            object: nil
+        )
     }
 
     override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
         enableIdleTimer()
         viewModel.deactivate()
+        NotificationCenter.default.removeObserver(self, name: UIApplication.didBecomeActiveNotification, object: nil)
+    }
+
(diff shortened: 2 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Read the diff. The authorization mapping mirrors `MapViewModel.topPillState` including the `@unknown default`, and every branch lands somewhere real — `.restricted` correctly renders no button rather than a dead one. The `didBecomeActive` hook is the right call, and the comment explaining that VoiceOver's `shouldSkipProgrammaticRefresh` would otherwise strand exactly the riders this button is for is the kind of reasoning I want in the code.

PR #1352, opened 2026-09-02
Label onboarding regions by how they were actually chosen

### LINKED ISSUE #1345: Onboarding still says 'Detected near you' for non-detected regions
Follow-ups from #1316, which made the whole onboarding region row tappable and added a "Your selection" eyebrow.

**1. "Detected near you" can still label a region that was never detected.**

`detectedRegion` returns `regionProvider.currentRegion` first, and `RegionsService` sets that from non-location paths too — `fixedRegionName` (`OBAKitCore/Location/Regions/RegionsService.swift:101-106`) and the sole-active-region fallback (`:114-117`). In those cases the card still reads "Detected near you". #1316 narrows the mislabel but doesn't eliminate it.

**2. The screen's body text now contradicts the new eyebrow.**

`OnboardingRegionView`'s `body` / `body_no_location` still reads "We found the transit network closest to you." after a manual pick, which disagrees with a "YOUR SELECTION" eyebrow directly above it. Previously both were consistently wrong; now they disagree with each other.

**3. Korean phrasing.**

`"내가 선택함"` reads as a verb phrase ("I selected it") rather than a noun label. `"내 선택"` or `"직접 선택"` is more idiomatic for an uppercase card eyebrow. Understandable as-is — worth a native speaker's eye rather than a blocker.

diveshpatil9104:
Fixes #1345.Follow-ups from #1316. Closes #1345.

1. detectedRegion returned currentRegion first, which RegionsService also sets from fixedRegionName and the sole-active-region fallback. A current region now only reads "Detected near you" when the rider's location resolves to it — otherwise the card reads "Current region" (new onboarding.region.current_label). That resolution checks service-area containment before nearest center, matching RegionsService: distanceFrom measures to a region's center, so a rider inside a large region can sit closer to a small neighbour's center.
2. The eyebrow and the body text now derive from one value, so they can't contradict each other. Only the detected state says "We found the transit network closest to you." body_no_location is re-keyed body_choose — same English value, so translations carry over — since it now shows for any non-detected state, not just a missing location.
3. Korean "내가 선택함" → "내 선택".

Tests in OnboardingRegionSelectionTests. "Current region" translated in all 13 catalogs.

## Summary by CodeRabbit

* **New Features**
  * Improved region selection during onboarding by distinguishing detected, chosen, and current regions.
  * Region detection now prioritizes regions containing the rider’s location, then selects the nearest available region.
  * Updated onboarding cards and guidance to clearly identify how each region was selected.

* **Localization**
  * Updated region-selection messaging and added “Current region” labels across supported languages.

* **Tests**
  * Added coverage for region detection, fallback selection, and selection-source behavior.

### DIFF
--- a/OBAKit/Onboarding/Views/OnboardingRegionSelection.swift
+++ b/OBAKit/Onboarding/Views/OnboardingRegionSelection.swift
@@ -1,0 +1,83 @@
+//
+//  OnboardingRegionSelection.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import CoreLocation
+import OBAKitCore
+
+/// The region the onboarding card displays, paired with how the app arrived at it.
+///
+/// Both the card's eyebrow and the screen's body text are derived from `source`, so
+/// they can't drift into describing the region two different ways.
+/// See https://github.com/OneBusAway/onebusaway-ios/issues/1345
+struct OnboardingRegionSelection {
+
+    /// Where the displayed region came from.
+    enum Source {
+        /// The rider's location resolves to this region.
+        case detected
+
+        /// The rider tapped this region in the list on the onboarding screen.
+        case chosen
+
+        /// The app already had this region selected and the rider's location doesn't
+        /// corroborate it — a fixed region from `OBAKitConfig`, the sole-active-region
+        /// fallback, or a pick that predates this run of onboarding.
+        case preselected
+    }
+
+    let region: Region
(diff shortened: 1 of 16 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Read the diff. The provenance trace is sound — `.detected` is unreachable without a real location, so the card can't claim "Detected near you" on a region location never corroborated, which was the point of #1345. Checking containment before nearest-center to match how `RegionsService` actually auto-selects is the right detail, and the localization is complete: all 13 catalogs, the re-key preserves every existing translation, and `current_label` is genuinely translated everywhere rather than English-filled.

Issue #1347, opened 2026-09-02
Proximity alert expiry guard is untested (test can't fail)


aaronbrethorst:
Follow-ups from #1311, which added `ProximityAlertManager`.

**1. The expiry-guard test doesn't reach the guard.**

`OBAKitTests/ProximityAlerts/ProximityAlertManagerTests.swift` — `Entering the geofence of an expired alert delivers nothing` never exercises `guard !alert.isExpired` in `ProximityAlertManager.swift:284`. The second `store.add` posts `.proximityAlertsDidChange`, the observer fires inline, `reconcileMonitoredRegions()` runs (`isReconciling` is false by then) and `deleteExpiredProximityAlerts()` reaps the alert. The crossing then lands on the "no longer stored" branch, which a different test already covers.

Net: delete the expiry guard entirely and this test stays green — while its inline comment claims the opposite. The guard is the last line of defense against waking a rider for a day-old trip, so it deserves real coverage. Setting `store.proximityAlerts = [expired]` directly instead of `store.add` should do it.

**2. Notification carries no region identity.**

`pushService(_:receivedProximityAlertForStopID:)` resolves the region from `regionsService.currentRegion` at tap time; `ProximityAlert` has no `regionID` to carry the originating one. In practice the rider is physically at the geofence and OBA auto-selects by location, so a mismatch needs a manually pinned region — and the failure mode is a stop fetched against the wrong region's API, not corruption. Fixing it means a model change.

**3. Tautological assertion in `ApplicationTests`.**

`#expect(app.pendingStopRegionID == app.regionsService.currentRegion?.regionIdentifier)` holds in both branches — region nil ⇒ both nil; region set ⇒ `queueOrOpenStop` stashes exactly that value. The `pendingStopID` assertion above it is meaningful.

**4. No duplicate-alert-per-stop guard.**

`activeAlert(for:)` exists but `createProximityAlert` never consults it, so two alerts on one stop each burn one of the 20 available region slots. Possibly intended for a later PR in this series.

Borisserz:
Expiry-guard test now asserts the alert is still stored before enter: https://github.com/OneBusAway/onebusaway-ios/pull/1404 (items 2–4 already on main).

PR #1353, opened 2026-09-02
Test the proximity expiry guard and region stash for real

diveshpatil9104:
Items 1 and 3 of #1347, rebased onto current main. Tests only.

Both tests stayed green with the behavior they name deleted.

**`Entering the geofence of an expired alert delivers nothing`** never reached `guard !alert.isExpired`. The second `store.add` posts `.proximityAlertsDidChange`, the observer fires inline, and reconciliation reaps the alert before the crossing arrives — so it landed on the "no longer stored" branch, which another test already covers. Assigning `store.proximityAlerts` directly bypasses the notification and reaches the guard.

**`Push service proximity alert tap is never dropped`** asserted `pendingStopRegionID == currentRegion?.regionIdentifier`, which held in both branches. It stayed inert after the last round too: `pendingStopRegionID` starts nil on a freshly built `Application`, so asserting it is nil proved nothing. Seeding a stale `999` first makes the handler's explicit clear the only reason the assertion passes.

Items 2 and 4 landed separately in ebfca83b, and the region-present branch is covered by the tests added there.

## Summary by CodeRabbit

* **Tests**
  * Expanded proximity-alert coverage for launches without a loaded region, stale region state, explicit or fallback regions, and alerts targeting another region.
  * Added validation for duplicate stops, expired alerts, different stops, lost location authorization, and region recording.
  * Verified that alert notifications include region details when available and omit them when unavailable.

### DIFF
--- a/OBAKitTests/Application/ApplicationTests.swift
+++ b/OBAKitTests/Application/ApplicationTests.swift
@@ -715,18 +715,22 @@ final class ApplicationTests: OBATestCase {
 
     @Test func `Push service proximity alert tap is never dropped`() {
         let (app, pushService) = makeAppAndPushService()
+        // A region left behind by an earlier navigation. Without it the assertion
+        // below is inert: `pendingStopRegionID` starts nil on a freshly built
+        // `Application`, so deleting the handler's explicit clear would leave this
+        // test green.
+        app.pendingStopRegionID = 999
 
         app.pushService(pushService, receivedProximityAlertForStopID: "1_75403", regionID: nil)
 
         // The tap that matters is the one into a terminated app, where neither a
-        // root controller nor a region exists yet. Both branches must stash: with
-        // no region the handler stashes region-less, and with one `queueOrOpenStop`
-        // stashes it alongside the region. Dropping it would strand the rider on
-        // whatever screen the app happened to open to.
+        // root controller nor a region exists yet. Dropping it would strand the
+        // rider on whatever screen the app happened to open to, with no second
+        // chance — the notification is gone once tapped.
         #expect(app.pendingStopID == "1_75403")
-        // Whatever region was stashed must be the one actually current, or the
-        // drain will refuse the stop as belonging somewhere else.
-        #expect(app.pendingStopRegionID == app.regionsService.currentRegion?.regionIdentifier)
+        // Cleared, not merely absent: a stale region would make the drain refuse
+        // this stop as belonging somewhere else.
+        #expect(app.pendingStopRegionID == nil)
     }
 
     @Test func `Push service proximity alert tap stashes the region the alert carries`() {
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The region-stash assertion still can't fail. `pendingStopRegionID` is `var pendingStopRegionID: Int?` on a freshly constructed `Application`, so it is already `nil` before the handler runs — deleting production's explicit `pendingStopRegionID = nil` in `pushService(_:receivedProximityAlertForStopID:)` (`Application.swift`, the line whose own comment reads "Cleared rather than left alone: a region stashed by an earlier navigation would make the drain refuse this stop as belonging to somewhere else") leaves this test green. The new comment here quotes exactly that rationale, but the test never establishes the earlier navigation it guards against. One line — `app.pendingStopRegionID = 999` before the call — makes the assertion load-bearing. Relatedly, the description says this was "split into two tests," but only the region-less one is in the diff, so the region-present branch that reaches `queueOrOpenStop` is still uncovered.

https://github.com/OneBusAway/onebusaway-ios/blob/6f922080b78cdae39b138c9f74951a6b6ecd9b0d/OBAKitTests/Application/ApplicationTests.swift#L716-L721

aaronbrethorst (changes requested):
The `ProximityAlertManagerTests` half is a genuine fix and the diagnosis behind it is correct: `store.add(proximityAlert:)` posts `.proximityAlertsDidChange`, the manager observes it with the selector form, which `NotificationCenter` invokes synchronously — so the second `add` re-entered `reconcileMonitoredRegions()` and emptied the store before the crossing, meaning the old test was exercising the "no longer stored" branch that another test already covers. Assigning `proximityAlerts` directly avoids that, and the test now genuinely reaches the expiry guard: delete it and the delivery assertion fails.

The `ApplicationTests` half doesn't quite deliver the title's "region stash for real", though:

`#expect(app.pendingStopRegionID == nil)` asserts a value that is already nil by default — `pendingStopRegionID` is declared with no initializer and the test builds a fresh `Application`. Delete the `pendingStopRegionID = nil` line in `Application.swift` — the one whose comment says it exists to clear a region stashed by an *earlier* navigation — and this test stays green. Your own new comment invokes that rationale ("stashing a wrong one would make the drain refuse the stop"), but the test never sets up a prior stash. Pre-stashing a wrong region first is a one-liner, and `@testable import OBAKit` is already in scope.

The second test your description promises — the region-present branch, asserting concrete identifiers — isn't in the diff either. That branch has no coverage.

Rebase needed: #1368 landed and changed the proximity-alert handler's signature and its region resolution, touching both of these test files.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is the right fix, and the diagnosis in the body is correct: the old expiry test never reached the guard it named. `add` posts `.proximityAlertsDidChange`, the selector observer runs inline, and reconciliation reaped the alert before the crossing — so it was passing through the "no longer stored" branch that a different test already covers. Assigning `store.proximityAlerts` directly sidesteps the notification and gets to the real guard.

I checked both assertions are now load-bearing by deleting the production code in my head and confirming they'd fail: drop `guard !alert.isExpired` and the delivery happens; drop `pendingStopRegionID = nil` and the seeded 999 survives. That's what was missing before.

All three points from my earlier review are addressed. Merging.

PR #1354, opened 2026-09-02
Make buildContentState(from:matching:) non-optional

### LINKED ISSUE #1337: buildContentState(from:matching:) can never return nil
Follow-up from #1330.

`BookmarkActions.buildContentState(from:matching:)` can never return `nil`:

```swift
let source = sameTrip.isEmpty ? [departure] : sameTrip
return buildContentState(from: source)
```

`source` is non-empty by construction, and the sibling `buildContentState(from:)` only returns `nil` on an empty array. That makes three things dead branches:

- the `Optional` return type itself
- `guard let contentState = …` in `StopPageActionPresenter.startLiveActivity`
- `try #require(...)` in both new tests

Not a bug, but the signature currently claims a failure mode that can't happen, and the `#require`s read as meaningful coverage when they can't fail. Worth either making the return non-optional or giving it a real nil case.

Lower priority, noted while reviewing: if the feed omits `tripHeadsign` for the tracked departure, `TripBookmarkKey` collapses to `""` and matching silently reverts to route-only — i.e. exactly the #1326 symptom — with no log or signal. Rare in OBA feeds and a reasonable degradation, but nothing pins it.

diveshpatil9104:
## Description

`buildContentState(from:matching:)` could never return `nil` — `source` is non-empty by construction. This made the `Optional`, the `guard let` in `startLiveActivity`, and the `try #require` calls in both tests dead code.

### Changes

- Returns `TripAttributes.ContentState` directly; shared mapping logic was moved into a private `contentState(from:)` helper.
- Removed the now-redundant `buildLiveActivityContentState` wrapper from `StopPageActionPresenter`.
- Added a `Logger.warn` for the case also flagged in the issue: a missing trip headsign causes `matching` to silently collapse to stop + route.
  - Behavior is unchanged. Falling back to the tapped departure alone would show only one arrival.
  - The condition is now logged instead of being silent.
- Added two tests:
  - Empty-list fallback.
  - Missing-headsign / matching-collapse case.

Closes #1337

## Summary by CodeRabbit

* **Bug Fixes**
  * Live Activities now continue displaying departure information even when no matching arrivals are available.
  * Departure matching is more resilient when trip headsign information is missing.
  * Live Activity content consistently includes the soonest three arrivals when available.
* **Tests**
  * Added coverage for empty arrival lists and missing trip headsign data.

### DIFF
--- a/OBAKit/Bookmarks/BookmarkActions.swift
+++ b/OBAKit/Bookmarks/BookmarkActions.swift
@@ -79,10 +79,53 @@ final class BookmarkActions {
         (bookmark.routeShortName ?? bookmark.name, bookmark.tripHeadsign ?? "")
     }
 
+    /// Builds content from whatever arrivals a bookmark row has loaded, or `nil`
+    /// when it has none — the bookmark paths have no single departure to fall
+    /// back on, so an empty list really is a failure for them.
     static func buildContentState(from arrivalDepartures: [ArrivalDeparture]) -> TripAttributes.ContentState? {
         guard !arrivalDepartures.isEmpty else {
             return nil
         }
+        return contentState(from: arrivalDepartures)
+    }
+
+    /// Builds content for the trip the rider actually tracked: same stop, route,
+    /// and destination as `departure`. Route-only matching mixes both directions
+    /// at a transit center and shows the opposite bus's countdown (#1326).
+    ///
+    /// Non-optional on purpose. When the stop list no longer contains the tracked
+    /// trip — stale data, or a list that never loaded — this falls back to
+    /// `departure` itself, so there is always at least one arrival to show and
+    /// callers need no failure branch.
+    static func buildContentState(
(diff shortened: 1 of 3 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Read the diff and traced the nil semantics, which is the only thing that mattered here. The Optional really was unreachable — `source` is non-empty by construction — and the `?? []` at the call site is equivalent to the old `?? [departure]` because the pre-existing ternary turns an empty `sameTrip` back into `[departure]`. Nothing is fabricated, and the single-argument overload keeps its Optional and its guard, so the bookmark path's "empty arrivals ⇒ don't start" is intact.

Landing this ahead of #1381, which rewrites the same function to pin on tripID. I've asked that one to keep your non-optional signature and drop the headsign-degradation warn plus its test — under a tripID key a missing headsign can't degrade the match, so that warning would log a permanently false claim.

PR #1355, opened 2026-09-02
Remove unreachable trip-map selection branches

### LINKED ISSUE #1349: Trip-map callouts: dead selection branches and stale docs after #1259
Follow-ups from #1259, which removed MapKit callouts from trip-map stop pins.

The behavior is right — I traced load, list-row tap, pin tap, return-from-stop-page, deselect, and the 30 s refresh, and none of them self-navigate or swallow a tap. These are tidiness items left over from the final round.

**1. Some of the round-3/round-4 machinery is now unreachable.**

`didSelect`'s skip-path `deselectAnnotation` fires `didDeselect`, whose guard (`stopTime == selectedStopTime`) passes and sets `selectedStopTime = nil` immediately after the load-time select. So in production `selectedStopTime` is nil except transiently inside one call stack. That makes two things dead:

- `applyOriginStopSelection`'s `if let current = selectedStopTime, current.stopID != origin.stopID` fall-through (`TripViewController.swift:474`)
- the `oldValue == selectedStopTime` branch in the `didSet` (`:442`)

Behaviorally harmless, but it's the same "a branch that reads like live behavior but can't run" category that got cleaned up in an earlier round.

**2. `docs/trip-map-callouts.md`'s "Refresh" section describes a state machine the app no longer has**, for the same reason.

**3. The PR description's test plan is stale.**

It still claims "the origin stop is selected on the map and highlighted in the list." Neither is true now — `highlightStopInList` sits below the skip guard, and the load-time selection is torn down immediately.

**4. `didSelect` highlights and then navigates away in the same breath.**

The 750 ms throttled `blink()` in `TripFloatingPanelController` fires under the pushed stop page, so the rider never sees it; on return the list is scrolled but unblinked.

**5. No analytics on the new open path.**

`MapViewController` reports `AnalyticsLabels.mapStopAnnotationTapped` for the identical "selection is the open gesture" case; `TripViewController.openStop` reports nothing. Arguably pre-existing.

---

Note for whoever picks this up: a review comment on #1259 claimed the two new deselect tests are inert because `selectAnnotation` is a no-op on a zero-frame unhosted `MKMapView`. I checked that directly — on a zero-frame, never-hosted map with the annotation added, `selectAnnotation` **did** take effect (`selectedAnnotations.count == 1`, `view(for:)` non-nil). So `#expect(selectedAnnotations.isEmpty)` is a meaningful assertion, not a vacuous one, and that comment should be disregarded. (Probe was macOS MapKit; the suite runs on iOS, so if someone wants certainty, revert the `deselectAnnotation` in `openStop` locally and confirm the test goes red.)

diveshpatil9104:
Fixes #1349. Follow-ups from #1259 — no behavior change a rider can see, except the blink.

- `selectedStopTime` is nil except transiently inside one call stack, so `applyOriginStopSelection`'s "rider picked another stop" fall-through and the `oldValue == selectedStopTime` branch in the `didSet` can never run. Both removed. Apply-once is now an explicit `hasAppliedOriginStopSelection`, set after the origin lookup succeeds so an emission that can't supply one doesn't burn the attempt.
- That value-equal branch was clearing a leaked skip arm, so rather than delete the band-aid and leave the wound, `skipNextStopTimeHighlight` is now armed in the `didSet` immediately before `selectAnnotation` — the only statement that can provoke `didSelect`. An assignment that finds no annotation can no longer leave an arm behind to swallow the next pin tap.
- `didSelect` now scrolls the list with `blinksAfterScroll: false`. The 750 ms blink played underneath the stop page it pushes on the next line and was over before the rider came back. The scroll stays, so the row is on screen on their return.
- `openStop` reports `mapStopAnnotationTapped`, matching what `MapViewController` reports for the same callout-less "selection is the open gesture" case.
- `docs/trip-map-callouts.md`'s Refresh section described the old state machine. Rewritten, plus a short section stating that neither the origin pin nor its list row stays highlighted after load — the correction #1259's test plan needs.

Tests: dropped the value-equal test (its branch is gone), reworked the refresh test around the real teardown, and added coverage for the no-matching-stop case and for analytics on both `didSelect` paths.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved trip stop selection to prevent unintended re-selection of the origin stop during refreshes.
  - Ensured trips without a matching origin stop do not trigger an incorrect selection.
  - Refined map and list synchronization when opening a stop.

- **User Experience**
  - Selecting a stop on the map now scrolls to it in the list without an unnecessary blink.
  - Stop openings from map annotations are now tracked for analytics.

- **Documentation**
  - Updated trip map behavior documentation to reflect selection, refresh, and deselection flows.

### DIFF
--- a/OBAKit/Trip/TripFloatingPanelController.swift
+++ b/OBAKit/Trip/TripFloatingPanelController.swift
@@ -165,7 +165,13 @@ class TripFloatingPanelController: UIViewController,
     }
 
     // MARK: - Public Methods
-    public func highlightStopInList(_ matchingStop: Stop) {
+    /// Scrolls the trip stop list to `matchingStop`'s row and, by default, blinks it.
+    /// - Parameters:
+    ///   - matchingStop: The stop whose row should be brought into view.
+    ///   - blinksAfterScroll: Pass `false` when the caller is about to push another view
+    ///     controller. The blink is delayed until scrolling settles, so it would play
+    ///     underneath the pushed page and be finished before the rider returned to see it.
+    public func highlightStopInList(_ matchingStop: Stop, blinksAfterScroll: Bool = true) {
         let data = self.items(for: listView)
 
         var matchingIndexPath: IndexPath?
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Read the diff. The reachability argument holds, and moving the `skipNextStopTimeHighlight` arming down to the `selectAnnotation` call site is the part that makes this more than tidying — it removes the preview-path leak the value-equal branch was quietly healing, rather than just deleting the branch. The `blinksAfterScroll: false` reasoning is right too.

Note for whoever rebases: #1374 adds `stopIndex:` to this same `highlightStopInList` declaration. The parameters are orthogonal and compose fine.

Issue #1341, opened 2026-09-02
Vehicle interpolation: docs contradict the 500m snap threshold


aaronbrethorst:
Follow-ups from #1323, which interpolates the trip-page vehicle marker between arrival polls.

**1. The type doc contradicts the constant it describes.**

`VehicleCoordinateUpdate`'s header says distances "larger than one or two polls at motorway speed" are treated as a new fix — that's 830–1670 m at 100 km/h over 30 s. But `snapBeyondMeters` is **500 m**, documented separately as "a bit over one 30s poll at ~50 km/h". `VehicleCoordinateUpdateTests`'s `A kilometre jump snaps` repeats the motorway framing. The constant is the truth; the prose should match it.

**2. Consequence: interpolation effectively never engages above ~60 km/h.**

A 30 s poll at 60 km/h covers exactly 500 m, right at the cliff — so freeway and express buses snap on most polls, and even a 50 km/h bus at 417 m sits uncomfortably close. That's a defensible tuning choice (it degrades to the old behavior), but it isn't what the docs promise, and it means the feature is a city-traffic improvement specifically.

**3. `VehicleCoordinateUpdate.apply` has no tests.**

The five new tests cover `decision(from:to:)` in isolation. Nothing asserts that `.snap` writes `to`, that `.unchanged` leaves the coordinate alone, or that `StopVehicleAnnotation.update`'s restore-then-apply ordering behaves. #1323's test plan lists "StopVehicleAnnotationTests (6)" but that file is untouched — those are pre-existing tests that pass only because their fixture uses a 0.5°/0.4° jump, which snaps.

**4. `TripViewController` still teleports.**

The legacy trip screen's `currentTripStatus` `didSet` (`OBAKit/Trip/TripViewController.swift:267-290`) still assigns `vehicleAnnotation.tripStatus` directly, so that marker pops rather than slides. Out of scope for #1323, but it's the same issue on the other screen.

PR #1356, opened 2026-09-02
Make the vehicle interpolation docs match the 500 m snap threshold

diveshpatil9104:
## Summary

Part 1 of #1341: `VehicleCoordinateUpdate`'s type doc contradicted the constant it describes.

The header said distances "larger than one or two polls at motorway speed" are treated as a new fix. At 100 km/h over a 30s poll that's 830–1670 m — but `snapBeyondMeters` is **500 m**, documented on its own line as "a bit over one 30s poll at ~50 km/h". `VehicleCoordinateUpdateTests`'s `A kilometre jump snaps` repeated the same motorway framing.

The constant is the truth, so both comments now point at `snapBeyondMeters` rather than restating a threshold that drifts from it.

## Changes

- `OBAKit/Mapping/VehicleCoordinateUpdate.swift` — header doc refers to `snapBeyondMeters` instead of "one or two polls at motorway speed".
- `OBAKitTests/Mapping/VehicleCoordinateUpdateTests.swift` — same fix on the `A kilometre jump snaps` doc comment.

`snapBeyondMeters`' own doc is left as-is; it already describes the 500 m value accurately.

## Scope

Comment-only. No behavior change, no test changes, nothing added or removed from the API.

Parts 2–4 of #1341 (the sub-60 km/h interpolation window, `apply` test coverage, and `TripViewController`'s teleporting marker) are deliberately left out — each is a separate change and will follow in its own PR.

## Testing

- `swiftlint` clean on both touched files.
- No behavioral surface to exercise: the diff is four lines of prose.

### DIFF
--- a/OBAKit/Mapping/VehicleCoordinateUpdate.swift
+++ b/OBAKit/Mapping/VehicleCoordinateUpdate.swift
@@ -15,9 +15,9 @@ import UIKit
 /// Whether a vehicle marker should interpolate to a new coordinate or jump.
 ///
 /// Arrival polls land every 15–30s. Assigning `coordinate` each time teleports
-/// the pin; interpolating every hop looks like motion. Distances larger than
-/// one or two polls at motorway speed are a new fix (or a trip change), not
-/// something to ease across the map.
+/// the pin; interpolating every hop looks like motion. Hops longer than
+/// `snapBeyondMeters` are a new fix (or a trip change), not something to ease
+/// across the map.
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Verified against the source myself — the claim holds. Thanks for the precision fix; these small accuracy corrections are genuinely useful.

PR #1357, opened 2026-09-02
Address weather map button follow-ups from #1344

### LINKED ISSUE #1344: Verify map weather button doesn't truncate at 42pt after Configuration switch
Follow-ups from #1318, which added a condition icon beside the map temperature.

**1. Possible temperature truncation in the hover bar — needs a device check.**

The toolbar is hard-pinned to `widthAnchor == 42.0` with a vertical stack at `alignment = .fill`, so the weather button is exactly 42 pt wide (`OBAKit/Mapping/MapViewController.swift:145`). The PR converts it from `UIButton(type: .system)` (zero content edge insets — the full 42 pt went to the title) to `UIButton.Configuration.plain()`, which brings nonzero default `contentInsets` on all four sides, and simultaneously drops `titleLabel?.adjustsFontSizeToFitWidth`.

That flag was added in `7f2c1b8b` specifically for variable-width temperature text in this bar. It may well have been a no-op — `UILabel.minimumScaleFactor` defaults to `0`, which Apple documents as "the current font size is used as the minimum" — so the residual risk is just the inset narrowing. #1318's own test plan leaves the "On device, map hover bar" box unchecked; a 30-second look settles it. Worth checking a three-digit temperature.

**2. Hardcoded `", "` separator in `buttonAccessibilityValue`.**

`"\(buttonTitle), \(header.conditionSummary)"` in `OBAKit/ViewModels/MapViewModel/WeatherDisplay.swift`. Every other composite user-facing string in that file (`weather.chance_of_rain_format`, `weather.high_low_format`, `weather.percent_format`) goes through an `OBALoc` format string, and the file's own header comment says formatting is delegated to `WeatherFormatter` "so the helpers are testable and reusable". Low impact — VoiceOver only, both components already localized — but inconsistent with the file's stated convention.

**3. `docs/weather-map-button.md` describes only the UIKit layout.**

It says "a sun icon **above** `71°`", but the SwiftUI pill puts the icon **beside** the temperature in an `HStack`.

**4. Redundant config setup.**

`imagePlacement = .top` is set both in the `weatherButton` lazy initializer and again in the `weatherDisplay` `didSet`, which rebuilds the whole configuration on every update.

diveshpatil9104:
## PR Description

Parts 1–4 of #1344.

### Changes

- **Fix the 42pt toolbar button wrapping**
  - Part 1 was a real bug, and worse than expected: the 42pt toolbar button wrapped a two-digit temperature onto three lines — `18°` rendered as `1 / 8 / °`.

    | Before | After |
    |:---:|:---:|
    | Before | After |

  - The cause was `.plain()`, which adds horizontal `contentInsets` that `UIButton(type: .system)` did not have, leaving the title with significantly less than 42pt of usable width.
  - Zeroed the horizontal insets only. The vertical insets remain so the icon and temperature stay separated with `imagePlacement = .top`.
  - Restored `adjustsFontSizeToFitWidth` with a `minimumScaleFactor`. The existing 2019 flag was effectively a no-op without a minimum scale factor, so three-digit temperatures now scale instead of wrapping.

- **Localize `buttonAccessibilityValue`**
  - Routed it through `OBALoc` instead of hardcoding `", "`.
  - `zh-Hans` / `zh-Hant` now use `，`.
  - Arabic now uses `،`.
  - Added the key to all 13 catalogs.

- **Update weather map button documentation**
  - `docs/weather-map-button.md` now documents both surfaces:
    - `imagePlacement = .top` in UIKit.
    - `HStack` in the SwiftUI pill.

- **Remove duplicate configuration**
  - Dropped the duplicate `imagePlacement` assignment from `weatherDisplay`'s `didSet`.

Closes #1344.

### DIFF
--- a/OBAKit/Mapping/MapViewController.swift
+++ b/OBAKit/Mapping/MapViewController.swift
@@ -439,6 +439,15 @@ class MapViewController: UIViewController,
         var config = UIButton.Configuration.plain()
         config.imagePlacement = .top
         config.title = "—"
+        // The toolbar pins this button to 42pt (see the width constraint above),
+        // and `.plain()` ships nonzero horizontal insets that the old
+        // `UIButton(type: .system)` did not have. That left the title too little
+        // room: a two-digit temperature already wrapped onto three lines
+        // ("1", "8", "°"). Zero the horizontal insets only — the vertical ones
+        // still separate the icon from the temperature under `imagePlacement`.
+        // See: https://github.com/OneBusAway/onebusaway-ios/issues/1344
+        config.contentInsets.leading = 0
+        config.contentInsets.trailing = 0
         config.titleTextAttributesTransformer = UIConfigurationTextAttributesTransformer { incoming in
             var outgoing = incoming
             outgoing.font = UIFont.preferredFont(forTextStyle: .body).bold
@@ -446,6 +455,12 @@ class MapViewController: UIViewController,
         }
 
         let button = UIButton(configuration: config)
+        // Restores what `7f2c1b8b` intended before the Configuration switch
+        // dropped it, with the scale factor that version omitted: without a
+        // `minimumScaleFactor`, `adjustsFontSizeToFitWidth` never shrinks
+        // anything. Three-digit temperatures scale down rather than wrap.
+        button.titleLabel?.adjustsFontSizeToFitWidth = true
+        button.titleLabel?.minimumScaleFactor = 0.7
         button.addTarget(self, action: #selector(showWeather), for: .touchUpInside)
         button.accessibilityLabel = OBALoc("map_controller.show_weather_button", value: "Show Weather Forecast", comment: "Accessibility label for a button that provides the current forecast")
         return button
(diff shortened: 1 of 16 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
All four #1344 follow-ups are addressed, and the root-cause diagnosis on the truncation is right — `.plain()` ships nonzero horizontal insets that `UIButton(type: .system)` didn't, which is what left `18°` wrapping to three lines. Localization is clean: all 13 locales, purely additive, and the separator is genuinely localized (Arabic `،`, full-width `，` for zh).

One thing I'd check on device rather than block on: `adjustsFontSizeToFitWidth` may well be inert here. Autoshrink only applies to single-line labels, and this title is demonstrably multi-line — that's what your own before-screenshot shows — with no height constraint and nothing setting `numberOfLines = 1`. So `100°` may still wrap rather than scale. The inset fix stands on its own either way; this is just a no-op addition if I'm right.

PR #1358, opened 2026-09-02
Document that an empty DevelopURL can hide the whole Help Out section

### LINKED ISSUE #1339: Empty DevelopURL hides the whole Help-make-the-app-better section
Follow-up from #1329, which documented the `OBAKitConfig.MoreTab` white-label keys.

The doc says an empty `DevelopURL` "hides the row", which is true. But if `TranslateURL` is also absent, the `guard !contents.isEmpty` in `MoreViewController.helpOutSection` drops the **entire** "Help make the app better" section, not just the one row.

A white-label agency setting `DevelopURL: ""` to hide the GitHub link will find the whole section vanishes. Worth a clause in `WhiteLabel.md`.

diveshpatil9104:
## Description

`helpOutSection` returns `nil` when it has no rows. This means `DevelopURL: ""` doesn't just hide the Develop row — when `TranslateURL` is absent, which is the OneBusAway default, it also removes the **"Help make the app better"** header.

The documentation only promised that the row would be hidden.

The docs now also point to `ShowHelpOutSection: false` as the intended way to deliberately hide the entire section.

Closes #1339.

## Summary by CodeRabbit

* **Documentation**
  * Clarified how an empty `DevelopURL` value affects the More tab.
  * Documented that the “Help make the app better” section may be hidden when no other links are available.
  * Added guidance to disable `ShowHelpOutSection` when the entire section should remain hidden.

### DIFF
--- a/OBAKit/OBAKit.docc/Articles/WhiteLabel.md
+++ b/OBAKit/OBAKit.docc/Articles/WhiteLabel.md
@@ -181,7 +181,7 @@ Nested keys (all optional):
 * `HeaderSupportText` - String shown in the branded header box. Omit or leave empty to keep the default volunteer-support copy.
 * `ShowHelpOutSection` - Boolean. Defaults to `true`. Set `false` to hide the "Help make the app better" section (translate / develop rows).
 * `TranslateURL` - URL for "Help Translate the App". Omit to hide that row. OneBusAway currently omits it; white-label apps can opt in.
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Verified against the source myself — the claim holds. Thanks for the precision fix; these small accuracy corrections are genuinely useful.

Issue #1342, opened 2026-09-02
map_routes_overflow_fmt missing from all locales; comments overclaim


aaronbrethorst:
Follow-ups from #1321, which switched overflowing map route labels to a single U+2026.

**1. The new `OBALoc` key isn't in any `.strings` file.**

`formatters.map_routes_overflow_fmt` appears in zero of the 13 `OBAKitCore/Strings/*.lproj/Localizable.strings` files, so every locale falls through to the `value:` default `"%@…"`. Runtime behavior is correct everywhere — but translator substitutability is the *only* reason to wrap a bare glyph in `OBALoc`, and three places currently claim it works:

- the `formattedMapRoutes` doc comment
- the test comment
- `docs/map-route-labels.md`

All three cite zh-Hans `……` as the motivating example. Until the key exists, that capability is aspirational. `scripts/extract_strings` plus translations closes it.

There's also a latent trap: `OBAKitTests/Strings/LocalizationTests.swift`'s `Every locale has same keys as english` flags keys present in a locale but absent from `en`. So a translator who adds `……` to zh-Hans before `en` is regenerated turns the suite red. Note `OBAKitCore` currently has *no* key drift (175/175), unlike `OBAKit`.

**2. Two vacuous assertions.**

`#expect(!callout.contains("more"))` and `#expect(!subtitle.contains("more"))` can never fail — `formattedMapRoutes` has no "more" branch, and `subtitle` calls `formattedRoutes` with `limit: .max`, which also never emits one.

**3. Redundant `try #require` on a non-optional.**

`try #require(stop.mapCalloutText)` — `mapCalloutText` returns `String`, not `String?`. Swift Testing's redundant-`require` overload emits a deprecation warning.

**4. Product question: bookmark callouts now truncate.**

`buildDetailLabel` sets `numberOfLines = 0`, so the callout wraps freely and previously showed every route via `Stop.subtitle`; it now shows at most 3 + `…`. For stop pins the "pin label and callout must agree" rationale holds. For **bookmark** pins it doesn't — their `titleLabel` shows `bookmark.name`, not the route list, so the callout was the only place routes appeared and is now truncated with no consistency benefit. Worth deciding deliberately.

PR #1359, opened 2026-09-02
Localize the map route overflow marker and drop two vacuous assertions

diveshpatil9104:
## Description

`formatters.map_routes_overflow_fmt` was missing from all 13 OBAKitCore catalogs. As a result, every locale fell back to the `%@…` default, meaning the translator substitutability described in the doc comment was aspirational.

### Changes

- Added `formatters.map_routes_overflow_fmt` to all 13 catalogs.
  - `zh-Hans` / `zh-Hant` use `%@……`.
- Removed two assertions that can never fail.
- Removed the `try #require` on `mapCalloutText`, since it returns a non-optional `String`.
- Kept the `#require` on `subtitle`, since that value is optional.
- Added the localization entries by hand rather than running `scripts/extract_strings`, which would have rewritten unrelated keys in the OBAKit catalog.

Refr #1342.

## Summary by CodeRabbit

* **New Features**
  * Map pins now indicate additional routes with an ellipsis when the full route list cannot be displayed.
  * Added localized support for this display across supported languages.

* **Tests**
  * Updated map callout coverage to reflect ellipsis-based overflow formatting.

### DIFF
--- a/OBAKitCore/Strings/ar.lproj/Localizable.strings
+++ b/OBAKitCore/Strings/ar.lproj/Localizable.strings
@@ -339,6 +339,9 @@
 /* Describes an on-time departure. e.g. departed on time. */
 "formatters.deviation.departure_past_on_time" = "غادرت في الموعد";
 
+/* Overflowing map-pin route list. The argument is the visible names; U+2026 marks that more routes exist. e.g. '10, 12, 49…' */
+"formatters.map_routes_overflow_fmt" = "%@…";
+
 /* Short formatted time text for arrivals/departures occurring now. */
 "formatters.now" = "الآن";
 
--- a/OBAKitCore/Strings/en.lproj/Localizable.strings
+++ b/OBAKitCore/Strings/en.lproj/Localizable.strings
@@ -339,6 +339,9 @@
 /* Describes an on-time departure. e.g. departed on time. */
 "formatters.deviation.departure_past_on_time" = "departed on time";
 
+/* Overflowing map-pin route list. The argument is the visible names; U+2026 marks that more routes exist. e.g. '10, 12, 49…' */
+"formatters.map_routes_overflow_fmt" = "%@…";
+
 /* Short formatted time text for arrivals/departures occurring now. */
 "formatters.now" = "NOW";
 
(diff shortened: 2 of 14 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The PR body says `Closes #1342`, but #1342 has four numbered items and this PR addresses only 1–3. Item 4 — bookmark pin callouts now truncate their route list — is untouched: `prepareForDisplay(bookmark:)` still passes `bookmark.stop.mapCalloutText` (max 3 routes + `…`) into a `numberOfLines = 0` label, where the bookmark's `titleLabel` shows `bookmark.name` rather than routes, so the "pin label and callout must agree" rationale doesn't apply and the callout is the only place routes appear. Merging will auto-close #1342 and drop that product decision. Suggest `Fixes #1342 (items 1–3)` plus a follow-up issue, or reopening #1342 after merge.

https://github.com/OneBusAway/onebusaway-ios/blob/58e1d1d43cce5275b42332ebcb4435cd2138eb07/OBAKit/Mapping/StopAnnotationView.swift#L147-L150

The localization half checks out: `formatters.map_routes_overflow_fmt` lands in all 13 OBAKitCore catalogs, purely additive (+3/−0 each, so no translated→English swap), exactly one `%@` per locale so `Every locale preserves english format specifiers` holds, UTF-8 U+2026 (doubled for zh-Hans/zh-Hant), correct alphabetical slot, and comment/value matching what `genstrings -s OBALoc` would emit. Both dropped assertions are genuinely vacuous: `!callout.contains("more")` is subsumed by the exact-equality assertion on the line above, and `!subtitle.contains("more")` can't fire because `Stop.subtitle` calls `formattedRoutes` with the default `limit: .max`, and any regression that did introduce a limit would already trip `subtitle.contains("62")`.

aaronbrethorst:
### Code review — correction

Retracting the finding in my previous comment. I said `Closes #1342` would auto-close the issue with item 4 (the bookmark callout question) undone. The body actually says `Refr #1342.` — a typo for "Refs", and not a closing keyword, so nothing auto-closes. My error.

The localization work stands on its own and the two dropped assertions really were vacuous. No outstanding issues from me.

(Minor, take it or leave it: `Refr` → `Refs` while you're in there.)

aaronbrethorst (approved):
Localization is correct and complete — all 13 catalogs, one occurrence each, purely additive so nothing was displaced, and the CJK locales correctly use the six-dot `……` rather than a single ellipsis. Nice attention to that.

Both dropped assertions really were vacuous: the preceding exact-equality assertion subsumes the first, and `Stop.subtitle` calls `formattedRoutes` with the default `limit: .max`, so the "+ N more" branch is unreachable from that call site. Good calls, and right to add the key rather than leave the doc comment's translator-substitutability claim aspirational.

PR #1361, opened 2026-09-02
Pluralize rental cluster titles to avoid “1 vehicles here”

### LINKED ISSUE #1350: Rental cluster sheet can render '1 vehicles here'
Follow-ups from #1293, which moved map-layer registration into a shared `MapLayerRegistrar` and brought the Map settings sheet and rental markers to the map panel.

**1. "1 vehicles here" is now reachable.**

`AppSheetViewFactory.rentalClusterView` makes `rental_cluster.title_fmt` ("%d vehicles here") reachable with `rentals.count == 1`, because `rentals(withIDs:)` drops members that have left the feed. The key isn't in any `.stringsdict`, and the `String(format:)` call mirrors the pre-existing one at `RentalDetailViewController.swift:329` — so this is an old string reaching a new state rather than something #1293 introduced. Worth a `.stringsdict` entry now that it's hittable.

**2. `RentalClustering` class doc overstates the marker bound.**

It claims "roughly 90 on an iPhone-sized map", but at the production default `cellSize: 100` on a 390×844 map that's ~36 cells. 90 corresponds to the tests' `cellSize: 60`.

**3. `Marker count is bounded by viewport cells at extreme density` verifies less than it claims.**

The test derives `maxCells` from the viewport, but bucketing happens in global map-point space — the bound holds only because the fixture happens to fit inside one viewport.

**4. New test helpers use `UserDefaults()`.**

Several construct `LocationService(userDefaults: UserDefaults(), ...)`, i.e. standard defaults. CLAUDE.md calls out process-global `UserDefaults.standard` as needing handling since separate suites run concurrently. Test-only, but the pattern spreads.

**5. `RentalMapMarker` fuel label offsets the marker.**

The fuel figure is drawn inside the same `VStack` that `Annotation` centers, so a marker showing one sits slightly above its true coordinate. Cosmetic; unverified without a device.

diveshpatil9104:
## PR Description

`rental_cluster.title_fmt` had no `.stringsdict` entry and wasn't present in any `.strings` catalog, so every locale fell back to `%d vehicles here`. Since `rentals(withIDs:)` drops members that have left the feed, `count == 1` is reachable.

### Changes

- Added the plural entry across all 13 locales, plus the not-found fallback, matching the existing `data_migration_bulletin.*` setup.
- Both call sites — `RentalDetailViewController` and `RentalMapMarker` — use the key, so both are covered.
- French keeps `%d` in one plural category because that category covers both 0 and 1, and 0 is reachable here.
- Arabic's zero/one/two forms carry the number in the localized word, matching the existing entries in that catalog.

closes #1350.

## Summary by CodeRabbit

* **Localization**
  * Added localized rental-cluster titles with accurate vehicle-count pluralization across supported languages.
  * Improved plural handling for languages with complex grammatical rules.
  * Updated region onboarding wording and added current-region labels.
  * Added localized accessibility text for arrival announcements and weather controls.

* **Accessibility**
  * VoiceOver now announces rental-cluster vehicle counts using the correct plural forms.

### DIFF
--- a/OBAKit/Mapping/Layers/RentalDetailViewController.swift
+++ b/OBAKit/Mapping/Layers/RentalDetailViewController.swift
@@ -331,7 +331,7 @@ struct RentalClusterListView: View {
                 .buttonStyle(.plain)
             }
             .listStyle(.plain)
-            .navigationTitle(String(format: OBALoc("rental_cluster.title_fmt", value: "%d vehicles here", comment: "Title of the sheet listing the members of a rental cluster"), rentals.count))
+            .navigationTitle(String(format: OBALoc("rental_cluster.title_fmt", value: "%d vehicles here", comment: "Title of the sheet listing the members of a rental cluster. Plural forms live in Localizable.stringsdict; the value above is only the not-found fallback."), rentals.count))
             .navigationBarTitleDisplayMode(.inline)
             .sheet(item: $selectedRental) { rental in
                 RentalDetailView(
--- a/OBAKit/Mapping/Layers/RentalDetailViewController.swift
+++ b/OBAKit/Mapping/Layers/RentalDetailViewController.swift
@@ -331,7 +331,12 @@ struct RentalClusterListView: View {
                 .buttonStyle(.plain)
             }
             .listStyle(.plain)
-            .navigationTitle(String(format: OBALoc("rental_cluster.title_fmt", value: "%d vehicles here", comment: "Title of the sheet listing the members of a rental cluster. Plural forms live in Localizable.stringsdict; the value above is only the not-found fallback."), rentals.count))
+            // `localizedStringWithFormat`, not `String(format:)`: the latter expands
+            // `%#@count@` but always resolves it against the root plural rule, so the
+            // `few`/`many`/`zero`/`two` forms in the ar, pl, and ru entries could never
+            // be selected. A cluster always holds at least two, so this is the common
+            // case, not an edge one. Same trap as `SearchResultsSheetView`.
+            .navigationTitle(String.localizedStringWithFormat(OBALoc("rental_cluster.title_fmt", value: "%d vehicles here", comment: "Title of the sheet listing the members of a rental cluster. Plural forms live in Localizable.stringsdict; the value above is only the not-found fallback."), rentals.count))
             .navigationBarTitleDisplayMode(.inline)
             .sheet(item: $selectedRental) { rental in
                 RentalDetailView(
--- a/OBAKit/Sheet/Root/Controls/RentalMapMarker.swift
+++ b/OBAKit/Sheet/Root/Controls/RentalMapMarker.swift
@@ -107,8 +107,11 @@ struct RentalClusterMapMarker: View {
                 .foregroundStyle(.white)
         }
         .accessibilityElement(children: .ignore)
-        .accessibilityLabel(Text(String(
-            format: OBALoc(
+        // `localizedStringWithFormat`, not `String(format:)` — see the note on the
+        // sheet's `navigationTitle`; VoiceOver would otherwise read the root plural
+        // form for every count.
+        .accessibilityLabel(Text(String.localizedStringWithFormat(
+            OBALoc(
                 "rental_cluster.title_fmt",
                 value: "%d vehicles here",
                 comment: "Title of the sheet listing the members of a rental cluster. Plural forms live in Localizable.stringsdict; the value above is only the not-found fallback."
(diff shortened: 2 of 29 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. Both call sites still use `String(format:)`, so the new `zero`/`two`/`few`/`many` forms for ar, pl and ru are unreachable and the PR's goal for those locales isn't met. `String(format:)` expands `%#@count@` against the root plural rule — the repo already documents exactly this at `SearchResultsSheetView.swift:109` and `MapTypeButton.swift:74` ("the `few`/`many`/`zero`/`two` forms in the ar, pl, and ru entries could never be selected"), and pins it with `LocalizationTests/Polish layer count reaches its few and many forms`. Loading this PR's own stringsdicts confirms it: Polish n=5 renders `5 pojazdu tutaj` through `String(format:)` versus `5 pojazdów tutaj` through `String.localizedStringWithFormat`; Russian n=5 renders `Здесь 5 транспортного средства` versus `Здесь 5 транспортных средств`; Arabic n=2 renders `2 مركبة هنا` instead of the dual `مركبتان هنا`. Since the `other` category in pl/ru is the fraction form (correctly written that way), every whole count ≥ 2 — and a cluster always has at least two members — is ungrammatical. Both call sites need `String.localizedStringWithFormat`.

https://github.com/OneBusAway/onebusaway-ios/blob/780d49478f1f6cd9d5ad1bc5ce960fec42a96ad8/OBAKit/Mapping/Layers/RentalDetailViewController.swift#L332-L336

https://github.com/OneBusAway/onebusaway-ios/blob/780d49478f1f6cd9d5ad1bc5ce960fec42a96ad8/OBAKit/Sheet/Root/Controls/RentalMapMarker.swift#L102-L111

aaronbrethorst (changes requested):
The stringsdict work here is genuinely good — real `NSStringPluralRuleType` entries rather than a Swift ternary, all 13 locales, correct CLDR category sets (ar with zero/one/two/few/many/other, pl and ru with the fraction `other`, fr keeping `%d` in `one` because French `one` covers 0). That's the hard part and you got it right.

The delivery path undoes it, though:

**Both call sites still use `String(format:)`, so the new `zero`/`two`/`few`/`many` forms are unreachable.** `String(format:)` doesn't consult the plural rules; only `String.localizedStringWithFormat` does. So Polish n=2/5/22 renders "2/5/22 pojazdu tutaj" (the fraction form) instead of pojazdy/pojazdów/pojazdy, Russian n=5 gives "5 транспортного средства" instead of "транспортных средств", and Arabic n=2 misses the dual. Since a cluster always has at least two members, the *common* case is ungrammatical in pl and ru — which means this ships worse grammar than the English fallback it replaces.

The repo already documents this exact trap in `SearchResultsSheetView.swift` and `MapTypeButton.swift`, and pins it with `Polish layer count reaches its few and many forms` in `LocalizationTests`.

So: swap both call sites to `String.localizedStringWithFormat`, and please add `rental_cluster.title_fmt` to `pluralKeys` in `LocalizationTests` — all ten existing plural keys are registered there, and without it the new key isn't covered by the well-formedness test.

One smaller thing while you're in there: pt-BR's `one` hardcodes the literal "1", but CLDR `pt` puts 0 in `one` too, so a count of 0 would render "1 veículo aqui". Unreachable today since both entry points guard on non-empty, but you got this right for French and it's the same shape.

Data's right, wiring's one line. Ping me when it's up.

aaronbrethorst (approved):
This is done properly. I checked the plural machinery myself rather than taking the diff's word for it: all 13 `.stringsdict` catalogs carry the new key, Polish has `one/few/many/other` and Arabic has all six CLDR categories, with `%d` and matching `NSStringFormatValueTypeKey` throughout. A two-form `count == 1` ternary would have been wrong for Polish specifically — which matters here because KiedyBus ships it — and you didn't take that shortcut.

Both call sites moving to `String.localizedStringWithFormat` is the part that makes the non-root categories actually reachable, and registering the key in `pluralKeys` means the well-formedness test covers it going forward. pt-BR reads `%d veículo aqui` now, so that point is closed too.

The `### Code review` comment above this one predates your fixes — disregard it.

Minor, for whenever you're next in the file: the new comment on `navigationTitle` says a cluster "always holds at least two", which holds where clusters are formed but not in `RentalClusterListView`, whose members are re-resolved against `visibleRentals` and can come back as one. That's the very case this PR exists to render correctly, so the comment slightly undercuts its own change. Behavior is right either way.

Merging.

PR #1362, opened 2026-09-02
Shorten the share escape's bar button title so it fits beside Cancel

### LINKED ISSUE #1348: 'Share Without Destination' bar button crowds the nav bar in ru/pl
Follow-up from #1314.

The new "Share Without Destination" escape is a `rightBarButtonItem` on `DestinationStopPickerController`, opposite a left "Cancel" and a "Select Destination" title. In the longer locales that's a lot of bar:

- `ru` — "Поделиться без пункта назначения" (31 chars)
- `pl` — "Udostępnij bez celu podróży" (27 chars)

On a narrow device the title, the button, or both will truncate.

The author flagged the constraint that forced it there — `StandardEmptyDataViewModel` takes a single `buttonConfig`, already used by the empty state's body button — and offered to move it. Worth deciding: either allow a second button in the empty-state view model, or shorten the bar title to something like "Share Anyway" / "Skip".

diveshpatil9104:
## PR Description

The bar button reused the body button's **"Share Without Destination"** title, which reaches 32 characters in Russian and 27 in Polish. That's too long when sharing the bar with **"Select Destination"** and **Cancel**.

It now uses the shorter **"Share Anyway"** key:
- Russian: 20 characters
- Polish: 18 characters

The body button keeps the full phrase since it has a row to itself.

### Why a separate key?

I went with a shorter localized title rather than adding a second `buttonConfig` to `StandardEmptyDataViewModel`. That control is used app-wide, while the constraint is specific to this screen.

The button's accessibility label already carries the full phrase in all 13 locales, so no information is lost for VoiceOver users.

This is intentionally a **distinct localization key**, not a key reuse: it's a shorter phrase for a narrower UI surface.

Also updated the doc comment, which previously stated that using a second key would be a defect.

Closes #1348.

## Summary by CodeRabbit

* **New Features**
  * Added a shorter “Share Anyway” navigation-bar action when sharing a trip without selecting a destination.
  * Added localized labels for this action across supported languages.

* **Accessibility**
  * Preserved the existing accessibility label for the sharing action.

* **Tests**
  * Updated coverage to reflect the new navigation-bar button title.

### DIFF
--- a/OBAKit/Strings/ar.lproj/Localizable.strings
+++ b/OBAKit/Strings/ar.lproj/Localizable.strings
@@ -241,6 +241,9 @@
 /* VoiceOver label for the navigation bar button that shares the trip when no destination could be loaded. */
 "destination_stop_picker.share_without_destination_a11y_label" = "مشاركة الرحلة دون تحديد محطة وصول";
 
+/* Short navigation bar button that shares the trip without a destination stop, used where the full phrase will not fit. */
+"destination_stop_picker.share_without_destination_bar_button" = "مشاركة على أي حال";
+
 /* Button shown when a trip has no remaining stops, letting the user share the trip without picking a destination. */
 "destination_stop_picker.share_without_destination_button" = "مشاركة دون وجهة";
 
--- a/OBAKit/Strings/en.lproj/Localizable.strings
+++ b/OBAKit/Strings/en.lproj/Localizable.strings
@@ -241,6 +241,9 @@
 /* VoiceOver label for the navigation bar button that shares the trip when no destination could be loaded. */
 "destination_stop_picker.share_without_destination_a11y_label" = "Share trip without a destination stop";
 
+/* Short navigation bar button that shares the trip without a destination stop, used where the full phrase will not fit. */
+"destination_stop_picker.share_without_destination_bar_button" = "Share Anyway";
+
 /* Button shown when a trip has no remaining stops, letting the user share the trip without picking a destination. */
 "destination_stop_picker.share_without_destination_button" = "Share Without Destination";
 
--- a/OBAKit/Strings/es.lproj/Localizable.strings
+++ b/OBAKit/Strings/es.lproj/Localizable.strings
@@ -242,7 +242,7 @@
 "destination_stop_picker.share_without_destination_a11y_label" = "Compartir viaje sin parada de destino";
 
 /* Short navigation bar button that shares the trip without a destination stop, used where the full phrase will not fit. */
-"destination_stop_picker.share_without_destination_bar_button" = "Compartir igualmente";
+"destination_stop_picker.share_without_destination_bar_button" = "Aun así";
 
 /* Button shown when a trip has no remaining stops, letting the user share the trip without picking a destination. */
 "destination_stop_picker.share_without_destination_button" = "Compartir sin destino";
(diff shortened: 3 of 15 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The "short" bar-button title is not actually shorter in `pt-BR` (and barely so in `es`), so the crowding this PR exists to fix persists — and pt-BR is a small regression against `main`. `pt-BR` swaps "Compartilhar sem destino" (24 chars) for "Compartilhar assim mesmo" (24 chars); rendered at the 17pt system font that is ~196pt → ~208pt, making it the widest of the 13 new bar titles — wider even than Russian's "Всё равно поделиться" (~178pt). Beside `Cancelar` (~68pt) and the "Selecionar destino" title (~148pt) the bar still needs ~470pt on a ≤440pt device, the exact condition #1348 describes. `es` is the same story with a smaller delta: "Compartir sin destino" (~164pt) → "Compartir igualmente" (~165pt), i.e. no improvement at all. Every other locale shortens substantially (en −91pt, ru −105pt, pl −69pt, vi −134pt, fil −153pt). Suggest something materially shorter for these two, e.g. pt-BR "Compartilhar" or "Mesmo assim", es "Compartir igual" or "Aun así".

https://github.com/OneBusAway/onebusaway-ios/blob/6ecf6fb0bb03a22a222b45cad71d20dc04652dc8/OBAKit/Strings/pt-BR.lproj/Localizable.strings#L244-L246

Otherwise the localization side checks out: the new key is present in all 13 locales, genuinely translated in each (no English fill), every file is additive-only (+3/−0) so no previously-translated string was swapped back to English, all 13 files still parse and remain UTF-8, and the key is inserted in the file's existing sort position. `shareWithoutDestinationTitle` is still used by the empty state's body button, and the tests assert both titles separately.

aaronbrethorst (changes requested):
The approach is right and it's what #1348 asked for, the Swift change is minimal, and the localization hygiene is clean — all 13 locales, purely additive, genuine translations, correct sort position, no translated→English swap.

One thing to fix: **the "short" title isn't shorter in pt-BR, and is a wash in es.** pt-BR goes "Compartilhar sem destino" → "Compartilhar assim mesmo" — same 24 characters, and slightly *wider* at 17pt, making it the widest of the thirteen new bar titles, wider even than Russian. es goes "Compartir sin destino" → "Compartir igualmente", 20 chars → 21. The other nine locales shorten meaningfully (en −91pt, ru −105pt, vi −134pt, fil −153pt), so it's just these two that don't deliver what the PR is named for.

Something like "Compartilhar mesmo assim" doesn't help; you probably want to drop the verb — pt-BR "Mesmo assim" and es "Igualmente" or "Continuar" are the shapes that actually fit.

Not blocking, but worth knowing: the visible title is now "Share Anyway" while `accessibilityLabel` stays "Share trip without a destination stop", so they no longer share any words. VoiceOver is fine, but a Voice Control user saying "Tap Share Anyway" may not match — `accessibilityUserInputLabels` would cover it.

ru still exceeds the bar after this, but UIKit truncates the title before the button, which is the trade #1348 accepted, so I'm not holding on it.

aaronbrethorst (approved):
Both of my points are fixed, and with the wording I'd suggested — pt-BR is now "Mesmo assim" and es "Aun así", which are real reductions rather than the wash the previous round had.

I verified the localization hygiene directly: 13 catalogs touched, zero deletions in any `.strings` file, so this is purely additive with no translated-to-literal swap. Sort position and key/comment coherence are consistent across all of them.

Note for the record: the `### Code review` comment sitting above this one predates your fixes and describes code that no longer exists — ignore it.

One thing to look at whenever you're next in here, not a blocker: in `ar` the "short" title is actually slightly wider than the full phrase it replaces, so that locale gains a translation marginally worse than the one it had. Arabic's bar was never crowded so nothing renders badly; it just makes the "used where the full phrase will not fit" comment untrue for that one locale.

Merging.

PR #1365, opened 2026-09-03
Fix/1336 stop screen missing stop

### LINKED ISSUE #1336: Stop screen doesn't recognize a JSON-null stop as gone
Follow-up from #1332, which taught the Bookmarks tab to treat an HTTP 200 + JSON `null` stop response as "gone" instead of raising a bulletin.

The Stop screen was not taught the same shape. `OBAKit/ViewModels/StopViewModel.swift:299` still gates the broken-bookmark UI on `APIError.requestNotFound` only:

```swift
} catch APIError.requestNotFound {
    operationError = nil
    isBrokenBookmark = bookmarkContext != nil
```

A `null` response falls through to the generic `catch`, which sets `operationError`. So the two surfaces now disagree:

- **Bookmarks tab** — card settles silently on "No upcoming departures"
- **Stop screen** — generic error, and `isBrokenBookmark` never fires

#1332's description says "the bookmark itself is left for the rider to delete", but the path that actually offers deletion doesn't run for this shape. A rider ends up with a card that quietly shows nothing, taps through, and gets a generic error with no way out.

Fix is likely to hoist the `isGoneBookmarkStop(_:)` predicate somewhere both call sites can use it, and widen the `StopViewModel` catch to match.

diveshpatil9104:
## Summary

The Bookmarks tab and the Stop screen disagreed about what a missing stop looks
like, in both directions.

`APIService+GetData` throws three different errors for "the server has no stop at
this ID": a real 404 (`:38`), an empty HTTP 200 on a GET, thrown as
`requestNotFound` carrying a **200** response (`:56`), and HTTP 200 with the body
`null`, thrown as `invalidContentType(_, "json", "nothing")` (`:90`). (A GET with
an empty body and no `Content-Length` header throws `noResponseBody` instead —
that one is a generic error on both surfaces, before and after this PR.)

`BookmarkDataLoader.isGoneBookmarkStop` matched the 404 and the `null`, and
deliberately excluded the empty 200. `StopViewModel` matched `requestNotFound`
with no status filter — so it matched the 404 **and the empty 200**, and missed
the `null`. Both now share `APIError.indicatesMissingStop`.

### What changes for the rider

| Server response | Bookmarks tab | Stop screen, from a bookmark | Stop screen, from a deep link / search / map pin |
|---|---|---|---|
| Real 404 | unchanged | unchanged | spinner → **"stop not found"** |
| Empty HTTP 200 | **bulletin copy fixed** | broken-bookmark UI → **retryable error** | spinner → **retryable error** |
| HTTP 200 + `null` | unchanged | generic error → **broken-bookmark UI** | generic error → **"stop not found"** |

The "spinner" in that last column is a bug on `main`, not something this PR
introduces: a 404 with no bookmark behind it already set `operationError = nil`
with nil arrivals, which is the exact shape `StopPageContent.showsLoadingState`
reads as a fetch still in flight. It renders `StopPageLoadingRow` forever while
the refresh timer re-throws every cycle. Widening the catch would have extended
that to the `null` body, so this PR fixes the underlying state instead.

- **The #1336 fix:** a `null` body on a bookmarked stop now sets
  `isBrokenBookmark` and offers the rider the deletion, instead of a generic
  error with no way out.
- **New terminal state.** A stop the server doesn't have now says so. The view
  model exposes `stopIsMissing`, which `StopPageContent` uses to suppress the
  loading branch and the empty-state row uses to render "This stop isn't in the
  transit agency's data anymore." A `null` body with no bookmark behind it lands
  here rather than on a generic error, matching a real 404 — and the 404 path
  gets the fix too, since it had the same spinner on `main`. The
  review-prompt suppression (`noteStopLoadFailed()`) still fires, so the app
  won't ask for five stars after it. `StopViewController` gets the same state so
  the legacy surface doesn't disagree with the new one.
- **Stale departures are dropped.** A stop that goes missing between refreshes
  now clears `stopArrivals` and `lastUpdated`, so the page can't keep counting
  down to a bus that isn't coming — the same thing `BookmarkDataLoader` already
  does for the card.
- **Deliberate change:** an empty HTTP 200 no longer shows the broken-bookmark UI
  on the Stop screen. That path could previously prompt someone to delete a
  working bookmark because the server returned a blank body once. A live stop
  answers with a full 200, so a blank one is a blip, not a deleted stop.
- **Copy fix, required by the above.** `requestNotFound`'s `errorDescription` is
  hardcoded to `"404 Not found (%@)"` and never reads `statusCode`
  (`NetworkOperation.swift:90`), and `ErrorClassifier` had no case for it — so an
  empty 200 rendered a raw URL and a status code the server never sent. It now
  classifies as `.invalidResponseData`, whose copy already says the right thing
  ("returned data this app can't read… try again shortly") and is already present
  in all 13 catalogs, so this adds no new strings. `Application.displayError`
  runs the same classifier, so this also fixes the identical wrong bulletin on
  the Bookmarks tab, which ships today.

`isGoneBookmarkStop` is deleted; the predicate is now a computed property on
`APIError` so both modules reach it. Renamed to `indicatesMissingStop` because
the Stop screen reaches this path from deep links, search results and map pins
where no bookmark exists — and because the empty-200 exclusion is itself evidence
these servers don't reliably report a stop as *deleted*, only that this response
had no stop in it.

Left alone: the `requestNotFound` catch in `cancelAlarm` (an Obaco DELETE, where
a 404 means the alarm already fired — and the empty-200 heuristic is GET-only, so
it can't reach that path), and `ScheduleForRouteViewModel` /
`OBAEmptyDataSetItem` / `OBAListViewEmptyDataViewModel`, which inspect
`requestNotFound` for their own purposes and aren't stop fetches. The empty-200
error icon does change, as a consequence of the reclassification.

One thing I deliberately did not touch: throwing `requestNotFound` for a blank
200 at all is arguably the root confusion here, but changing that moves every
consumer at once. This PR corrects the presentation without touching the throw.

If you'd rather keep today's Stop-screen behaviour for the empty 200, I'm happy
to put it back — it's a small change, though it means either duplicating the
catch body or writing `where indicatesMissingStop || <any requestNotFound>`,
which puts the divergence back inside the code written to remove it. The copy fix
stands on its own either way.

Closes #1336.

## Testing

- Seven tests on the predicate covering every shape and its negatives, including
  `requestNotFound` carrying a 200 — the boundary the whole design turns on — and
  `invalidContentType(_, "json", nil)`, which `APIService+GetData:83` really does
  throw.
- Six new `StopViewModel` tests: `null` and empty-200, each with and without a
  bookmark context; one asserting a `null` body with no bookmark doesn't leave the
  page in the loading state; and one for a stop that goes missing after a
  successful refresh, checking the arrivals and the freshness label are both
  cleared. All but the first pin behaviour changes, so a future reader doesn't
  "fix" them back.
- Two new `ErrorClassifier` tests for the empty-200 case, alongside the existing
  404 pass-through test that pins its other half. A region-less empty 200 now
  falls back to the neutral "can't read" wording rather than re-emitting the
  "404 Not found" copy the case exists to prevent.
- Three `StopPageContent` tests pinning the loading branch: a missing stop is
  terminal, the same shape without the flag still shows the spinner, and an
  in-flight refresh still wins so a retry isn't swallowed.
- The existing Bookmarks-tab tests from #1331/#1332 pass untouched — including
  `Empty 200 still calls display error`, which is the guard that the hoist didn't
  quietly widen the predicate.

Full suite: 2545 tests, no new failures.

### DIFF
--- a/OBAKit/Stops/StopPage/Shared/StopDeparturesSections.swift
+++ b/OBAKit/Stops/StopPage/Shared/StopDeparturesSections.swift
@@ -177,6 +177,7 @@ struct StopDeparturesSections: View {
                 Section {
                     StopPageEmptyStateRow(
                         isBrokenBookmark: isBrokenBookmark,
+                        stopIsMissing: content.stopIsMissing,
                         errorText: errorText,
                         isFilteredEmpty: content.isFilteredEmpty,
                         isDepartureFilterEmpty: content.isDepartureFilterEmpty,
--- a/OBAKit/Stops/StopPage/Shared/StopPageContent.swift
+++ b/OBAKit/Stops/StopPage/Shared/StopPageContent.swift
@@ -39,6 +39,11 @@ struct StopPageContent {
 
     let hasLoadedArrivals: Bool
     let showsLoadingState: Bool
+
+    /// The server has no stop at this ID. A terminal state: it suppresses the
+    /// loading row, which would otherwise spin forever because a missing stop
+    /// never produces arrivals (#1336).
+    let stopIsMissing: Bool
     /// `true` only when the route filter is what emptied the list.
     let isFilteredEmpty: Bool
     /// `true` only when the Departure Type filter is what emptied the list:
@@ -65,7 +70,8 @@ struct StopPageContent {
         arrivalDepartureFilter: ArrivalDepartureFilter,
         isLoading: Bool,
         hasError: Bool,
-        isBrokenBookmark: Bool
+        isBrokenBookmark: Bool,
+        stopIsMissing: Bool = false
     ) {
         let visible = isListFiltered ? allDepartures.filter(preferences: preferences) : allDepartures
         // Terminal dedup deliberately runs downstream, after the Departure Type
@@ -98,10 +104,14 @@ struct StopPageContent {
         self.listIsEmpty = isGrouped ? routeGroups.isEmpty : departures.isEmpty
 
         self.hasLoadedArrivals = hasLoadedArrivals
+        self.stopIsMissing = stopIsMissing
         // Any in-flight fetch, plus the pre-`.task` first frame (nothing
         // fetched, no error yet) so the page never flashes "No departures"
-        // before the first request has even started.
-        self.showsLoadingState = isLoading || (!hasLoadedArrivals && !hasError && !isBrokenBookmark)
+        // before the first request has even started. A missing stop is excluded
+        // for the opposite reason: it has finished, and it will never have
+        // arrivals, so without it the page spins forever.
+        self.showsLoadingState = isLoading
+            || (!hasLoadedArrivals && !hasError && !isBrokenBookmark && !stopIsMissing)
 
         // The stop has departures but none survive the route preferences.
         // Grouped mode can be empty while `allDepartures` isn't (every departure
@@ -142,7 +152,8 @@ extension StopPageContent {
             arrivalDepartureFilter: viewModel.arrivalDepartureFilter,
             isLoading: viewModel.isLoading,
             hasError: viewModel.operationError != nil,
-            isBrokenBookmark: viewModel.isBrokenBookmark
+            isBrokenBookmark: viewModel.isBrokenBookmark,
+            stopIsMissing: viewModel.stopIsMissing
         )
     }
 }
--- a/OBAKit/Stops/StopPage/StopPageView.swift
+++ b/OBAKit/Stops/StopPage/StopPageView.swift
@@ -542,6 +542,10 @@ struct StopPageEmptyStateRow: View {
     /// leaving them to conclude the bus simply isn't running (parity with
     /// `StopViewController.emptyData(for:)`).
     var isBrokenBookmark: Bool = false
+    /// The server has no stop at this ID and there is no bookmark to repair —
+    /// a deep link, a search result or a map pin pointing at a stop the agency
+    /// has since removed.
+    var stopIsMissing: Bool = false
     let errorText: String?
     let isFilteredEmpty: Bool
     /// The Departure Type filter (real-time only / scheduled only) removed
@@ -580,6 +584,7 @@ struct StopPageEmptyStateRow: View {
 
     private var symbolName: String {
         if isBrokenBookmark { return "bookmark.slash.fill" }
+        if stopIsMissing { return "mappin.slash" }
         if errorText != nil { return "exclamationmark.icloud" }
         if isFilteredEmpty { return "line.3.horizontal.decrease.circle" }
         if isDepartureFilterEmpty { return "antenna.radiowaves.left.and.right" }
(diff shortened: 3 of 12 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. Widening this catch to the JSON-`null` shape leaves the redesigned Stop page stuck on "Loading departures…" forever when there is no bookmark behind it. `FeatureFlags.isNewStopPageEnabled` defaults to `true`, so `StopPageView` is the screen most riders get, and `StopPageContent` computes `showsLoadingState = isLoading || (!hasLoadedArrivals && !hasError && !isBrokenBookmark)`. After this branch runs with `bookmarkContext == nil`, `stopArrivals` is still `nil`, `operationError` is `nil`, `isBrokenBookmark` is `false`, and `isLoading` has gone back to `false` — all three conjuncts hold, so `StopDeparturesSections` renders `StopPageLoadingRow` (spinner + "Loading departures…") instead of `StopPageEmptyStateRow`, and the auto-refresh timer re-throws the same error every cycle without ever changing that. On `main` a `null` body fell to the general `catch`, set `operationError`, and produced an error row with a Retry button, so this is a regression for the deep-link / search-result / map-pin paths the comment itself calls out. The comment's "leaves the page with no arrivals, no error, and nothing to do" accurately describes the legacy `StopViewController` (`items(for:)` returns just `stopHeaderSection`), but not the default screen — there the rider sees an indefinite spinner, which reads as "still working" rather than "this stop is gone". Clearing `stopArrivals`/`lastUpdated` and giving the no-bookmark case a distinguishable terminal state would keep the #1336 fix without collapsing it into a fake loading state.

https://github.com/OneBusAway/onebusaway-ios/blob/130aebe0c1b77aeaa41cabcc3c6e796f5c5702f9/OBAKit/ViewModels/StopViewModel.swift#L298-L318

aaronbrethorst (changes requested):
The shared `APIError.indicatesMissingStop` predicate is the right idea, and the `ErrorClassifier` change is a real improvement — riders stop seeing "404 Not found (url)" for what was actually an empty HTTP 200. The bookmark path now lands on the actionable "delete and recreate" state, which is what #1336 was about.

One thing blocks it:

**The no-bookmark case falls into a permanent "Loading departures…" spinner.** `FeatureFlags.isNewStopPageEnabled` defaults to `true`, so `StopPageView` is the live surface. After the widened catch with `bookmarkContext == nil`, `stopArrivals` is nil, `operationError` is nil, `isBrokenBookmark` is false and `isLoading` is false — so `showsLoadingState` in `StopPageContent` stays true and `StopDeparturesSections` renders `StopPageLoadingRow` forever, with the auto-refresh timer re-throwing every cycle.

On main, a JSON-`null` body reached the general catch, set `operationError`, and produced an error row with a Retry button. So for a rider opening a missing stop *without* a bookmark, this trades a retryable error for a spinner that never resolves — and per #1336 the `null` body is the shape many OBA servers actually send.

Your PR description says this branch "settles with no arrivals and no error card," which is accurate as far as it goes; it's the default screen's rendering of that state that's the problem.

Clearing `stopArrivals` or threading a `stopIsMissing` flag into `StopPageContent` would do it — the point is that it needs a terminal, distinguishable state rather than falling through into the loading branch. Worth a test pinning it.

Non-blocking, but cheap while you're there: the new `ErrorClassifier` case bails to `return apiError` when `regionName` is nil, so a region-less empty-200 still shows the "404 Not found" copy the case exists to prevent — `classifyDecodingError` already has a region-less fallback you could mirror. And `JSON null without bookmark context flags error` asserts `operationError == nil`, so the name says the opposite of the assertion.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
The blocking issue is properly fixed: `stopIsMissing` is now a conjunct in `showsLoadingState`, so the no-bookmark case reaches a real terminal row instead of spinning on "Loading departures…" forever. Clearing stale arrivals and `lastUpdated` alongside it is the right call, and the three points I raised are all addressed.

I also like that you didn't take CodeRabbit's suggestion to short-circuit `shouldRefresh` on `stopIsMissing` — the auto-refresh timer is the only way this page can self-heal if the server recovers, given the terminal state deliberately has no Retry.

The two new `OBALoc` keys aren't in any catalog yet, so every locale renders the English fallback. That matches how this repo batches extraction and there are already ~83 keys in the same state on `main`, so I'm not holding on it — just don't let it drift much further.

The PR title is still the unedited branch default, "Fix/1336 stop screen missing stop". Given how carefully the body is written that's a shame; something like "Give a missing stop a terminal state on the Stop screen" would read better in the history. I'll leave the title to you.

Merging.

PR #1366, opened 2026-09-03
Make the trip-stop id uniqueness test able to fail

diveshpatil9104:
## PR Description

Make the trip-stop ID uniqueness test able to fail.

### Changes

- The **List IDs stay unique across the whole trip** test loaded a fixture whose 53 stops were all distinct.
  - As a result, the assertion held even with the bare `stop.id` identifier it was intended to guard.
  - The test passed for the wrong reason.

- The trip is now doubled out-and-back to represent the shape of a loop route.
  - 106 rows.
  - 106 IDs using `"\(stopIndex)-\(stop.id)"`.
  - Only 53 IDs under the old bare identifier, so the test now fails against the buggy implementation.

- Added a premise guard:
  ```swift
  Set(loop.map(\.stopID)).count < loop.count

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trip stop identification for loop routes, preventing duplicate stops from causing list display crashes.
  * Ensured repeated stops remain uniquely distinguishable based on their position in the trip.

* **Documentation**
  * Clarified identifier behavior and added coverage details for loop-trip scenarios.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/OBAKitTests/ViewModels/TripStopListItemTests.swift
+++ b/OBAKitTests/ViewModels/TripStopListItemTests.swift
@@ -102,10 +102,25 @@ final class TripStopViewModelIdentityTests {
         #expect(first != second)
     }
 
-    @Test func `List ids stay unique across the whole trip`() throws {
+    /// The whole-trip version of the test above. The fixture is an ordinary
+    /// point-to-point trip whose stops are all distinct, so running it as-is
+    /// proves nothing: the ids would come out unique under the bare `stop.id`
+    /// this replaced. Doubling it out-and-back gives every stop a second visit,
+    /// which is the shape a loop route actually has — and which halves the id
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Read the diff and checked the fixture: 53 distinct stops, so the old assertion really was tautological — it would have stayed green under the bare `stop.id` that #538 crashed on. Doubling it out-and-back is the right shape, and pinning the premise with its own `#expect` is what stops it silently reverting. Nice work.

Heads up: #1374 rewrites this same test with a different premise mechanism (injecting one duplicate). Yours is the better of the two, so I'll ask that one to drop its version and rebase.

PR #1367, opened 2026-09-03
Stop a superseded toast task from dismissing the new toast

diveshpatil9104:
## PR Description

Follow-up from #1215, where this was flagged during approval:

> If the previous task's `Task.sleep` has already resumed and its continuation is sitting on the `MainActor` queue, `cancel()` arrives too late and that continuation sets `liveActivityStarted = false` right after the new signal set it to `true`.

- Added `try Task.checkCancellation()` after the sleep to close that race.
- `StopViewModel` is `@MainActor`, so the continuation queues behind the newer signal and now checks cancellation before dismissing the toast.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Live Activity notifications so a newer toast remains visible when another notification arrives before the previous display period ends.

* **Tests**
  * Added coverage confirming that successive Live Activity notifications keep the toast visible as expected.

### DIFF
--- a/OBAKit/ViewModels/StopViewModel.swift
+++ b/OBAKit/ViewModels/StopViewModel.swift
@@ -825,6 +825,13 @@ class StopViewModel: ObservableObject {
         liveActivityToastDismissTask = Task { [weak self] in
             do {
                 try await Task.sleep(for: .seconds(2))
+                // `cancel()` only interrupts the sleep while it is still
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Read the diff myself and the reasoning holds. Nice, tightly-scoped fix.

PR #1368, opened 2026-09-03
 Record a proximity alert's region and refuse duplicates on one stop

diveshpatil9104:
### Summary

Parts 2 and 4 of #1347. Parts 1 and 3 are #1353.

A stop could hold two alerts. `activeAlert(for:)` already existed, but `createProximityAlert` never consulted it. As a result, a second alert on the same stop could arm another geofence at the same coordinate, consume another of the 20 slots guarded by #1292, and fire alongside the first.

It now returns `.alreadyActive(existing)`.

The duplicate check runs **after** the authorization guards, not before. A downgrade from Always neither disarms nor deletes an existing alert, so a stored alert can appear healthy while no longer being able to deliver. That rider needs `.needsLocationAuthorization`, which is actionable and explains the problem, rather than "you already have one". Expired alerts still do not block a new alert.

### Region tracking

The proximity notification previously carried no region.

- `ProximityAlert` now gains `regionID: Int?`.
- The region is stamped at creation using an injected `regionIDProvider` closure.
- The region is carried in the notification's `userInfo`.
- When the rider taps the notification, the carried region is preferred over `currentRegion`.
- `nil` means "fall back to the current region", preserving the behavior of existing alerts and alerts persisted before this field existed.

This also makes `queueOrOpenStop`'s mismatch guard live from this call site for the first time. Previously, `destination.regionID` was derived from `currentRegion` three lines earlier, so it could never disagree with itself.

The comment claiming the current region was correct "by construction" has therefore been removed — that isn't true for a manually pinned region.

### Notification handling

`notificationReceivedHandler` now matches the proximity payload before its single-custom-key check.

A proximity notification contains both a stop and a region, so it has two keys. The previous ordering would have rejected it and fallen through to the fallback that re-presents the notification's own text.

The stop key retains its bare `StopID` shape, so notifications delivered by an earlier build continue to route correctly.

## Testing

14 tests covering:

- A duplicate returns the existing alert by ID.
- Neither the store nor monitored-region count grows when a duplicate is rejected.
- An expired alert does not block a new alert.
- An alert on a different stop is unaffected.
- A stop with an existing alert still reports lost location authorization.
- The region ID round-trips through persistence.
- Missing region IDs decode as `nil`.
- The region reaches the notification payload.
- Both the carried-region and `currentRegion` fallback paths are asserted.
- Cold-launch notification taps preserve the carried region, which previously lost it entirely.

This overlaps with #1353 in two test files. Happy to rebase onto #1353 once it lands.

## Summary by CodeRabbit

- **New Features**
  - Proximity alerts now retain the transit region where they were created.
  - Notifications carry region information so alerts open in the correct regional context.
  - Alerts without region information continue using the currently selected region.

- **Bug Fixes**
  - Prevented duplicate active proximity alerts for the same stop.
  - Improved handling when region information is unavailable or does not match the current region.
  - Preserved location-authorization warnings when creating alerts.

### DIFF
--- a/OBAKit/Orchestration/Application.swift
+++ b/OBAKit/Orchestration/Application.swift
@@ -114,7 +114,8 @@ public class Application: CoreApplication, PushServiceDelegate {
     @MainActor
     public private(set) lazy var proximityAlertManager = ProximityAlertManager(
         locationService: locationService,
-        userDataStore: userDataStore
+        userDataStore: userDataStore,
+        regionIDProvider: { [weak self] in self?.regionsService.currentRegion?.regionIdentifier }
     )
 
     @objc lazy var userActivityBuilder = UserActivityBuilder(application: self)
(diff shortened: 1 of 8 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Read the diff. Preferring the alert's own region over `currentRegion` is the right call and the comment explains the manually-pinned-region case well. `.alreadyActive(existing)` handing the alert back rather than a bare failure is the correct shape for whoever builds the UI on top.

Two things for the follow-up UI PR rather than blockers here: nothing renders `.alreadyActive`, `.regionLimitReached`, or the authorization cases yet, so a refused duplicate is currently invisible to the rider; and the region-mismatch branch in `queueOrOpenStop` drops the tap with no log at all, which is out of step with the deliberate `Logger.info` right next to it.

Note: this will conflict with #1353 in two test files — you'd offered to rebase, and #1353 hasn't landed, so it'll be that one rebasing onto this.

Issue #1340, opened 2026-09-02
SwiftLint CI gate placement, pinning, and a stray de-privatization


aaronbrethorst:
Follow-ups from #1325, which moved the trip planner out of `MapViewController` and added a SwiftLint gate to CI.

**1. `longPressGesture` was de-privatized unnecessarily.**

Every other `private` → internal change in that PR is required because the extension lives in a different file. This one isn't — the extension never touches `longPressGesture`; its only uses are at `MapViewController.swift:160-163`, in the same file. It can go back to `private`.

**2. Lint runs inside the test job rather than beside it.**

The new `SwiftLint (errors only)` step sits before `build-for-testing` in the same job, so a single lint error fails the job and costs all test signal for that run. A separate job would parallelize it and preserve the test results.

**3. `brew install swiftlint` is unpinned.**

A SwiftLint release that adds or promotes an error-severity rule can turn `main` red with no code change, and CI can disagree with whatever version a developer has locally. Matches the existing unpinned `xcodegen` step, so this is a house-style question rather than a defect — but the new step is a gate, which raises the stakes.

Also minor: the new file's doc comment says "the class it hangs off is already at SwiftLint's body-length ceiling", which stopped being true as of that very change — the body is now under the *warning* threshold.

PR #1369, opened 2026-09-03
Run SwiftLint beside the build instead of ahead of it

diveshpatil9104:
- **Run SwiftLint as a separate job**
  - Lint is now its own job instead of a step that runs ahead of `build-for-testing` in the same job.
  - Previously, a single lint error could prevent the build and tests from running, costing all test signal for that run.
  - The jobs now run in parallel instead of serializing lint and build/test work.
  - Workflow permissions remain minimal: `contents: read` is used for checkout, while `checks: write` is required by `xcresulttool`.

- **Make `longPressGesture` private again**
  - Restored `longPressGesture` to `private`.
  - All five references are inside `MapViewController.swift`.
  - Unlike the other de-privatizations in #1325, no extension in another file needs access to it.

- **Correct the SwiftLint body-length justification**
  - The previous "already at SwiftLint's body-length ceiling" justification was incorrect, not merely stale.
  - SwiftLint reports **zero violations** for `MapViewController.swift`, against a `900/1000` threshold.
  - Only `file_length` is suppressed in that file.
  - The same incorrect sentence appears in `+Camera.swift`, so both comments are corrected.

### Left out

- **Pinning `brew install swiftlint`**
  - This was left unchanged as a house-style question.
  - `xcodegen` is also unpinned in the same workflow, so it would be better to decide on pinning for both together.
  - Happy to add it either way.

Refs #1340.

## Summary by CodeRabbit

* **CI Improvements**
  * Linting now runs as a separate parallel job from the build, allowing test results to remain available even when lint checks fail.

* **Code Quality**
  * Updated developer-facing documentation around map camera and trip-planning behavior.
  * Restricted long-press gesture implementation details to the map view controller.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,10 +24,16 @@ on:
 # tests", check the step timings first — it is almost never the tests.
 
 jobs:
-  build:
+  # Its own job, not a step inside `build`. As a step it ran before
+  # build-for-testing in the same job, so one lint error failed the job and took
+  # every test result for that run with it — the two answer different questions
+  # and a PR wants both. Parallel jobs also overlap lint with the build instead
+  # of serialising them. No `needs:`: neither gates the other, and GitHub still
+  # requires both to pass.
+  # See https://github.com/OneBusAway/onebusaway-ios/issues/1340
+  lint:
     runs-on: xcode-27
     permissions:
-      checks: write
       contents: read
 
     steps:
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Read the diff. Good change — and I appreciate that the comment explains *why* it's a separate job rather than just doing it.

PR #1370, opened 2026-09-03
Correct what the region-monitoring radius comment claims

diveshpatil9104:
## Description

Two of the three follow-ups left on #1292 were based on incorrect statements in the same comment, so they are corrected together here.

### Changes

- **Corrected the `CLCircularRegion` radius explanation**
  - The comment previously said that `CLCircularRegion` clamps an oversized radius without reporting it.
  - Current documentation says an oversized region results in `CLError.regionMonitoringFailure`.
  - The existing clamping behavior remains correct; only the stated reason was wrong.

- **Corrected the device maximum explanation**
  - The comment previously said that a non-positive device maximum means the value is unavailable rather than zero.
  - Apple documents `-1` as indicating that region monitoring is unavailable or unsupported on the device, which is more specific than simply "unknown".

- **Documented the remaining gap**
  - The comment now explicitly records that when the device maximum is `-1`, we still arm the alert and return `.started` even though the alert can never fire.

### Scope

**Comment-only on purpose.**

The `isMonitoringAvailable(for:)` guard needs a new case on both `ProximityMonitoringResult` and `ProximityAlertActivationResult`.

#1368 already adds `case alreadyActive` to `ProximityAlertActivationResult`, so this follow-up should land after that change.

Follow-up from #1292.

## Summary by CodeRabbit

* **Documentation**
  * Clarified proximity-monitoring behavior when the requested radius is unavailable or unsupported on a device.
  * Documented that such requests remain honored and that any failures are reported asynchronously.

### DIFF
--- a/OBAKitCore/Location/Location/LocationService.swift
+++ b/OBAKitCore/Location/Location/LocationService.swift
@@ -593,10 +593,19 @@ public protocol LocationServiceDelegate: NSObjectProtocol {
             return .regionLimitReached(limit: Self.maximumMonitoredRegions)
         }
 
-        // `CLCircularRegion` clamps an oversize radius without reporting it, so
-        // the alert would fire at a distance the user never chose. Clamp
-        // deliberately and say so. A non-positive device maximum means the value
-        // is unavailable rather than zero, so honor the request in that case.
+        // Clamp deliberately and report it. Core Location answers an oversize
+        // radius with `CLError.regionMonitoringFailure`, which arrives
+        // asynchronously through `monitoringDidFailFor` carrying no radius — too
+        // late, and too vague, to tell the rider their alert would have fired
+        // somewhere other than where they asked.
+        //
+        // `-1` is not an unknown limit. Apple documents it as region monitoring
+        // being unavailable or unsupported on this device, so honouring the
+        // request in that case hands the caller `.started` for an alert that can
+        // never fire. Left that way for now — the failure does still surface
+        // through `monitoringDidFailFor` — but refusing up front wants an
+        // `isMonitoringAvailable(for:)` check and a result case to carry it.
+        // Both points flagged in review of #1292.
         let requestedRadius = alert.radiusMeters
         let deviceMaximum = locationManager.maximumRegionMonitoringDistance
         let radius = deviceMaximum > 0 ? min(requestedRadius, deviceMaximum) : requestedRadius


aaronbrethorst:
### Code review

Found 1 issue:

1. The corrected claim is fixed in one place but repeated verbatim in three others, so the codebase now contradicts itself about the same API. I verified both new statements against Apple's docs and they're right — `maximumRegionMonitoringDistance` documents that an oversize radius "causes the location manager to send a `CLError.Code.regionMonitoringFailure` error to the delegate," and that `-1` means "region monitoring is unavailable or not supported"; `CLCircularRegion.init(center:radius:identifier:)` documents no clamping at all. But that means the three remaining "silently clamps" comments are now the wrong half of a disagreement, and two of them are in the same directory as the line this PR fixes. The `ProximityMonitoring.swift` one is the worst of the three: it is the public doc comment on `startedWithClampedRadius` (so it ships in DocC) and it additionally asserts the case "is the only signal," which the corrected reading contradicts directly.

https://github.com/OneBusAway/onebusaway-ios/blob/241366845dfb600b20169886ac5a8c30ec53593b/OBAKitCore/Location/Location/ProximityMonitoring.swift#L23-L28

https://github.com/OneBusAway/onebusaway-ios/blob/241366845dfb600b20169886ac5a8c30ec53593b/OBAKitCore/Location/Location/LocationManagerProtocol.swift#L64-L69

https://github.com/OneBusAway/onebusaway-ios/blob/241366845dfb600b20169886ac5a8c30ec53593b/OBAKitCore/Models/UserData/ProximityAlert.swift#L49-L54

Everything else checks out: the `-1` path does return `.started` as the comment now says, `#1292`'s review did raise both points, and neither #1368 nor #1353 changes the behavior described here.

aaronbrethorst (approved):
Read the diff and checked it against the code. The correction is right, and I like that it records the `-1` gap honestly instead of quietly papering over it.

Not holding the merge on this, but worth a follow-up: three other comments still assert the old "Core Location clamps silently" claim this one now contradicts — `ProximityMonitoring.swift` (the `startedWithClampedRadius` doc, which also calls it "the only signal"), `LocationManagerProtocol.swift` on `maximumRegionMonitoringDistance`, and `ProximityAlert.swift`. Same correction, three more places, whenever you're next in there.

PR #1372, opened 2026-09-03
Finish consolidating the schedule-only deviation gate

diveshpatil9104:
## Description

Follow-up from #1270.

You closed #1270 noting that the consolidation was **2-of-3**. `Formatters.deviationLabel` now has the component overload that `StopArrivalItem` needed, and `deviationLabel(for:)` delegates to it, giving us a single gate instead of two.

- Gating on `scheduleStatus == .unknown` is equivalent to gating on `predicted`.
  - `ArrivalDeparture.scheduleStatus` starts with `guard predicted else { return .unknown }`.

It turned out to be **2-of-4**: `fullAttributedArrivalDepartureExplanation` had a fourth inline copy in `Formatters` itself. That has now been folded into the same consolidated path.

### Left unchanged

- The two copies in `TripActivityPresenter`, which `deviationLabel`'s documentation explicitly identifies as the mirror.
- The switch discriminant at `Formatters.swift:190`.

Follow-up from #1270.

### DIFF
--- a/OBAKit/Stops/Sections/StopArrival/StopArrivalItem.swift
+++ b/OBAKit/Stops/Sections/StopArrival/StopArrivalItem.swift
@@ -198,14 +198,11 @@ nonisolated struct ArrivalDepartureContentConfiguration: OBAContentConfiguration
 
     var accessibilityScheduleDeviationText: String? {
         guard let formatters = formatters else { return nil }
-        if viewModel.scheduleStatus == .unknown {
-            return Strings.scheduledNotRealTime
-        } else {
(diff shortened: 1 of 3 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Read the whole diff. The equivalence argument is the right one and it's airtight: `ArrivalDeparture.scheduleStatus` opens with `guard predicted else { return .unknown }`, so gating on `.unknown` is the same condition rather than a looser one — and the doc comment says exactly that, which is what keeps the next person from re-splitting it. Good consolidation.

Issue #1346, opened 2026-09-02
#538 auto-closed by a partial fix; loop-route highlight still wrong in UIKit


aaronbrethorst:
Follow-ups from #1312, which gave loop-route trip stops position-qualified list identifiers.

**1. "Closes #538" overstates the scope — consider reopening.**

#538 is the general tracking issue ("OBAListView throwing duplicate identifier exceptions tops the crash list") and asks for a rethink of item identification across the list framework. #1312 fixes `TripStopViewModel` specifically. Real fix, correct fix, but it auto-closed a broader issue.

**2. The second new test is vacuous as a regression guard.**

`List ids stay unique across the whole trip` loads `trip_details_1_18196913_no_status.json`, whose 53 stop times have 53 distinct stop IDs — so `Set(ids).count == ids.count` passes against the *old* `id: String { stop.id }` implementation too. The first test (`A stop visited twice gets two distinct list ids`) does genuinely cover the fix, so coverage isn't absent; but a fixture with an actually repeated stop would make the second test mean something.

**3. Same loop-route bug remains in the UIKit selection paths (pre-existing).**

`highlightStopInList` and `isUserDestination` both match on bare `stop.id`, so on a loop route they resolve/highlight the wrong (first) occurrence. The SwiftUI side already solved this with an anchored `vehicleIndex` search; the UIKit side hasn't. Not touched by #1312, but it's the same class of bug on the same screen.

Minor: `docs/duplicate-trip-stop-ids.md` ends with a dangling "Tests do not load `TripViewController.view`." with no surrounding context.

Borisserz:
Fixed on `main` by #1374 (merged 2026-09-05): UIKit loop-route highlight keys off `stopSequence` / visit index (`TripStopListModel.userStopIndex`, `isUserDestination`), with regression test `User destination uses stop sequence on a loop`. The PR didn’t auto-close this issue.

@aaronbrethorst okay to close?

Borisserz:
Follow-up residual (UIKit vehicle “here” still used `firstIndex` on loop routes): filed #1425 and opened a fix that routes `TripFloatingPanelController` through `TripStopListModel.vehicleStopIndex` — the same rules as the SwiftUI list.

PR #1373, opened 2026-09-03
Follow-ups: gone stops, loop highlight, proximity test, lint CI

Borisserz:
## Summary
- **#1336** — Hoist `APIError.isGoneStopResponse` and treat JSON-`null` stop fetches as broken bookmarks on the stop page (same predicate as Bookmarks).
- **#1347** — Proximity expiry test arms a live geofence, then swaps in an expired alert without posting `proximityAlertsDidChange`, so it exercises `guard !alert.isExpired`.
- **#1337** — `buildContentState(from:matching:)` is non-optional; the matching overload always has at least the selected departure.
- **#1340** — `longPressGesture` is `private` again; SwiftLint runs in its own CI job so lint failures do not suppress test signal.
- **#1346** — UIKit `highlightStopInList` / `isUserDestination` resolve the rider's stop by `stopSequence`; list-id regression test injects a repeated stop.

## Test plan
- [x] `StopViewModelTests` (35 tests)
- [x] `ProximityAlertManagerTests`, `BookmarkActionsTests`, `TripStopViewModelIdentityTests` (44 tests)

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected stop selection and highlighting for loop routes and trips with repeated stops.
  - Improved handling of unavailable stops, including missing-stop responses and broken bookmarks.
  - Improved Live Activity startup reliability when trip content is available.
  - Refined trip departure information so a valid state is shown even when matching trips are unavailable.

- **Documentation**
  - Clarified how repeated trip stops are identified and selected.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,10 +24,9 @@ on:
 # tests", check the step timings first — it is almost never the tests.
 
 jobs:
-  build:
+  swiftlint:
     runs-on: xcode-27
     permissions:
-      checks: write
       contents: read
 
     steps:
@@ -38,7 +37,7 @@ jobs:
     # Warnings annotate via the GitHub reporter but do not fail the job —
     # `--strict` would collapse that split. The Xcode build phase still skips
     # in CI (scripts/swiftlint.sh); this step is the review-visible gate.
-    # See https://github.com/OneBusAway/onebusaway-ios/issues/1303
+    # Runs in its own job so a lint failure does not suppress test signal (#1340).
     - name: SwiftLint (errors only)
       env:
         HOMEBREW_NO_AUTO_UPDATE: 1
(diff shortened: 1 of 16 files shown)


diveshpatil9104:
there's a lot of overlap here. I have PRs open for four of these five: #1365 (#1336), #1354 (#1337), #1369 (#1340), #1353/#1368 (#1347), and #1366 for #1346 part 2. The loop-highlight half of #1346 is all yours — I hadn't touched it.

Just flagging a couple of things I found:

- **Empty 200:** `requestNotFound` treats it as gone even though `isGoneStopResponse` intentionally treats it as transient. I kept it transient in #1365 to avoid offering to delete a valid bookmark.
- **`buildContentState`:** The `!` can still crash. I moved the shared logic to a non-optional builder in #1354.

Also, `requestNotFound` hardcodes `404`, so an empty 200 can show the wrong status code.

Happy to close mine if Aaron wants to handle these as one batch — just trying to avoid duplicate work :-)

Borisserz:
Thanks for the heads-up — I didn't see those open. You're right on the empty-200 path in `StopViewModel` (comment says transient, code treats it as gone). Closing this batch so #1336 / #1337 / #1340 / #1347 stay with your PRs.

I'll open a smaller PR with only the loop-highlight half of #1346, since you said that part wasn't covered.

Borisserz:
Closing in favor of Divesh's open follow-ups for the overlapping issues; loop-highlight (#1346) will land in a focused PR.

PR #1374, opened 2026-09-03
Fix UIKit loop-route stop highlight and destination (#1346)

Borisserz:
## Summary
- Resolve the rider's stop by `stopSequence` for UIKit `highlightStopInList` and `isUserDestination`, matching `TripStopListModel` (loop routes no longer highlight the first visit).
- Make the list-id uniqueness regression able to fail by injecting a repeated stop.
- Doc cleanup for `docs/duplicate-trip-stop-ids.md`.

Scoped to the half of #1346 Divesh noted was uncovered; his #1366 covers the vacuous-test half.

## Test plan
- [x] `TripStopViewModelIdentityTests` (including `User destination uses stop sequence on a loop`)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stop selection on loop routes by distinguishing repeated visits to the same stop.
  - Map selections now highlight the corresponding stop occurrence in the trip list.
  - Boarding and destination indicators now apply to the correct stop visit.

- **Tests**
  - Added coverage for repeated stop IDs and loop-route destination selection.

- **Documentation**
  - Updated guidance for handling duplicate stop IDs.

### DIFF
--- a/OBAKit/Trip/TripFloatingPanelController.swift
+++ b/OBAKit/Trip/TripFloatingPanelController.swift
@@ -168,10 +168,11 @@ class TripFloatingPanelController: UIViewController,
     /// Scrolls the trip stop list to `matchingStop`'s row and, by default, blinks it.
     /// - Parameters:
     ///   - matchingStop: The stop whose row should be brought into view.
+    ///   - stopIndex: When the trip revisits a stop, disambiguates which visit to scroll to.
     ///   - blinksAfterScroll: Pass `false` when the caller is about to push another view
     ///     controller. The blink is delayed until scrolling settles, so it would play
     ///     underneath the pushed page and be finished before the rider returned to see it.
-    public func highlightStopInList(_ matchingStop: Stop, blinksAfterScroll: Bool = true) {
+    public func highlightStopInList(_ matchingStop: Stop, stopIndex: Int? = nil, blinksAfterScroll: Bool = true) {
         let data = self.items(for: listView)
 
         var matchingIndexPath: IndexPath?
@@ -179,6 +180,7 @@ class TripFloatingPanelController: UIViewController,
             for (itemIndex, item) in section.contents.enumerated() {
                 guard let tripStop = item.as(TripStopViewModel.self),
                       tripStop.stop.id == matchingStop.id else { continue }
+                if let stopIndex, tripStop.stopIndex != stopIndex { continue }
                 matchingIndexPath = IndexPath(item: itemIndex, section: sectionIndex)
                 break
             }
@@ -369,8 +371,8 @@ class TripFloatingPanelController: UIViewController,
         parentTripViewController?.showStopOnMap(tripStop)
     }
 
-    private func showOnList(_ tripStop: TripStopTime) {
-        highlightStopInList(tripStop.stop)
+    private func showOnList(_ tripStop: TripStopViewModel) {
+        highlightStopInList(tripStop.stop, stopIndex: tripStop.stopIndex)
     }
 
     private func serviceAlertsListSection(_ alerts: [ServiceAlert]) -> OBAListViewSection {
@@ -401,6 +403,10 @@ class TripFloatingPanelController: UIViewController,
         }
 
         let closestStopIdx = closestStopIndex(in: tripDetails)
+        let userStopIndex = TripStopListModel.userStopIndex(
+            in: tripDetails.stopTimes,
+            arrivalDeparture: arrivalDeparture
+        )
 
         // Stop times
         let selectTripStopAction: OBAListViewAction<TripStopViewModel> = { [unowned self] item in self.onSelectTripStop(item) }
@@ -410,6 +416,7 @@ class TripFloatingPanelController: UIViewController,
                 arrivalDeparture: arrivalDeparture,
                 stopIndex: index,
                 closestStopIndex: closestStopIdx,
+                userStopIndex: userStopIndex,
                 onSelectAction: selectTripStopAction
             ).typeErased
         }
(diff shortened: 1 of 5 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The map → list disambiguation matches stop times by pointer identity (`$0 === stopTime`), but the map's annotation objects go stale on the very first refresh, so the new `stopIndex` is `nil` from then on and the highlight silently falls back to the first-occurrence behavior this PR is fixing. `TripViewModel` re-decodes `TripDetails` on every 30s timer tick and on every `viewWillAppear` (including returning from the stop page you just pushed), assigning brand-new `TripStopTime` instances to `tripDetailsController.tripDetails`. `mapView.updateAnnotations(with:)` diffs by `Hashable`, and `TripStopTime.isEqual`/`hash` cover `arrival`/`departure`/`regionIdentifier`/`stopID` — all stable across refreshes — so `removed` and `added` are both empty and MKMapView keeps the *old* instances. `view.annotation` in `didSelect` is therefore an object that is no longer in `tripDetails.stopTimes`, and `firstIndex(where: { $0 === stopTime })` returns nil. Note the existing comment ~110 lines below in the same file: "Fixes #220: Find matching trip stop using stop ID instead of using pointers." Value equality (`==`) would both survive refreshes and still distinguish the two visits of a loop route, since their arrival/departure times differ.

https://github.com/OneBusAway/onebusaway-ios/blob/91126ee3883adfdbe2d3d83d62fdc27edb8965f5/OBAKit/Trip/TripViewController.swift#L347-L354

aaronbrethorst (changes requested):
The list-side half of this is correct and valuable — `userStopIndex` plus `isUserDestination = stopIndex == userStopIndex` is the right fix for #1346, the `stopSequence` resolver is properly guarded so a bad feed degrades to the old behavior rather than mis-resolving, and the new loop-destination test is non-vacuous.

One thing blocks it, and it's a one-token fix:

**The map→list highlight uses pointer identity.** `TripViewController.didSelect` does `stopTimes.firstIndex(where: { $0 === stopTime })`. `TripViewModel` re-decodes `TripDetails` on the 30s timer and on every `viewWillAppear`, so `tripDetailsController.tripDetails` holds fresh `TripStopTime` objects — while `mapView.updateAnnotations(with:)` diffs by `Hashable`, and `TripStopTime`'s `isEqual`/`hash` are arrival/departure/regionIdentifier/stopID, all refresh-stable. Nothing is added or removed, so MKMapView keeps the *old* instances. From the first refresh onward `===` never matches, `stopIndex` comes back nil, and the map path silently reverts to the first-occurrence highlight this PR exists to fix.

This file already carries the warning, at line 456: *"Fixes #220: Find matching trip stop using stop ID instead of using pointers."*

`==` works and still separates the two loop visits, since their times differ.

Two more, not blocking but worth folding in while you're here:

- `TripFloatingPanelController.updateProgressView` still resolves the rider's stop with a bare `firstIndex { $0.stopID == ad.stopID }`, 40 lines below the list that now uses `userStopIndex`. On a loop where the rider boards the second visit, the header can say "Passed your stop" while the correctly-highlighted row is still ahead.
- Please drop this PR's edit to `List ids stay unique…` and to the trailing line of `docs/duplicate-trip-stop-ids.md`. #1366 has landed and rewrote the same test with an explicit premise guard, which is the better mechanism.

Rebase note: #1355 landed and adds `blinksAfterScroll:` to the same `highlightStopInList` declaration. The parameters are orthogonal — `highlightStopInList(_ matchingStop: Stop, stopIndex: Int? = nil, blinksAfterScroll: Bool = true)` composes fine.

Borisserz:
Addressed review in `56ef29c4`:

**Blocking**
- `TripViewController.mapView(_:didSelect:)` now uses `==` instead of `===` when resolving the tapped annotation against `stopTimes`, so map refresh/reuse no longer breaks loop-route highlight.

**Also folded in**
- `updateProgressView()` now uses `TripStopListModel.userStopIndex(in:arrivalDeparture:)` (same helper as the list rows).
- Rebased onto `main`; kept #1366's whole-trip id test and dropped this PR's weaker duplicate + the extra `duplicate-trip-stop-ids.md` paragraph.
- Composed `highlightStopInList(_:stopIndex:blinksAfterScroll:)` with #1355's `blinksAfterScroll` default.

**Tests** (iPhone sim `FA364C8B…`, `xcodebuild test-without-building`):
- `TripStopViewModelIdentityTests` — pass
- `TripStopTemporalStateTests` — pass
- `TripProgressViewModelTests` — pass
- `TripStopListModelTests` — pass
- `TripFocusMapLayerTests` — pass

52 tests in 5 suites, all green.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Good fix, and reusing `TripStopListModel`'s `stopSequence` resolver rather than duplicating it is the right instinct — the SwiftUI trip page has resolved loops that way since #1312, so now both surfaces agree by construction.

The `===` problem from the last round is properly gone. `firstIndex(where: { $0 == stopTime })` works here because `TripStopTime`'s `isEqual` compares arrival/departure/region/stopID, all of which survive a re-decode, so a stale annotation from before a 30s refresh still matches — and because the two visits of a loop stop have different arrival times, `==` still tells them apart rather than collapsing back to first-occurrence.

I walked the degenerate cases against `index(in:stopID:stopSequence:)`: stop visited twice or three times resolves by sequence; first and last stop are both in bounds with the stopID guard holding; a stop not on the trip, or a feed whose sequence numbers aren't array indices, falls through to `firstIndex` — i.e. exactly the old behavior, no crash and no regression. The new test is non-vacuous: under the old `stopTime.stopID == ad.stopID` both visits would report `true`, so it would fail.

Two follow-ups, neither blocking. `arrivalDeparture` is now unused in `TripStopViewModel.init` — nothing flags it, but every call site still has to pass it, which will mislead the next reader. And `closestStopIndex(in:)` is still un-anchored, so you've imported the `userStopIndex` half of the SwiftUI model without the `vehicleIndex(in:closestStopID:userIndex:)` half; net better on loops, but the vehicle marker and the destination marker can still disagree when the vehicle's closest stop is the rider's repeated stop.

PR #1377, opened 2026-09-04
Route all three Live Activity start paths through a shared prominence…

### LINKED ISSUE #1375: Trip page's Track path skips the Dynamic Island prominence handshake
Follow-up from #1243, which added `TripLiveActivityRelevance` and the `demoteLivePeers` handshake for #1189 Problem 2.

Three call sites start a Live Activity. Two apply the handshake; the trip page doesn't.

`OBAKit/Bookmarks/BookmarkActions.swift:155` and `OBAKit/Stops/StopPage/StopPageActionPresenter.swift:600` both build content through `TripLiveActivityRelevance.content(state:staleDate:relevanceScore:)` with `prominenceScore()`, then demote the peers.

`OBAKit/Trip/TripPage/TripPageViewController.swift:456` does neither:

```swift
let activity = try Activity.request(
    attributes: TripAttributes(staticData: staticData),
    content: .init(state: contentState, staleDate: nil),
    pushType: .token
)
```

Net: a trip tracked from the trip page gets `ActivityContent`'s default score of `0` and demotes nobody, so when another trip is already live the Dynamic Island keeps showing the older one. That is the #1189 Problem 2 behaviour the other two paths exist to fix.

Worth noting this isn't an oversight in either PR — #1243 and the commit that added this call site were in flight on parallel branches within a day of each other (2026-07-31 and 2026-08-01), so neither could see the other. The merge just didn't reconcile them.

The path is live, not dead code: `StopPageActionPresenter.swift:243` instantiates it ungated, so it's reachable from the current stop page.

`OBAKitTests/Models/TripLiveActivityRelevanceTests.swift:75` already names the hazard — "If the call site ever drops back to `.init(state:staleDate:)`, this is what Island gets" — but it asserts against a locally constructed `ActivityContent`, so nothing observes the real call site. Coverage that reaches the call site would have caught this.

diveshpatil9104:
Fixes #1375.

The trip page's Track path never picked up the Dynamic Island prominence handshake from #1243. It built content with a bare `.init(state:staleDate:)` — which scores `0` — and demoted no peers, so tracking a trip from that page left the Island on whichever trip was already running. That's the #1189 Problem 2 behaviour the bookmark and stop paths exist to prevent.

Neither PR is at fault: #1243 and the commit that added this call site were in flight on parallel branches a day apart, so neither could see the other.

## Why a helper instead of a third copy

The obvious fix is to paste the eight lines the other two sites use. I didn't, because the invariant here can't be tested.

One prominence score has to reach two places that must agree — the new activity's `ActivityContent`, and the demotion its peers are measured against. `Activity.request` doesn't run under test, and no test in `OBAKitTests` reaches one; the existing coverage stops at content-state building and the no-arrivals early return. So a third hand-assembled copy would be a third chance to get an unverifiable pairing wrong, and #1298 is already adding a fourth start path.

`Activity.requestProminent` now owns both halves. Callers keep their own tracking and UI feedback, which is the part that legitimately differs between the three paths. There is exactly one `Activity.request` in the codebase now.

It lives in OBAKit rather than beside its siblings in `TripLiveActivityRelevance`, because `Activity.request` is unavailable to app extensions and OBAKitCore is `APPLICATION_EXTENSION_API_ONLY`. I noted that in the enum's doc comment so the next person doesn't try to move it.

## Tests

Added one that pins the mechanism rather than restating the score math: a bare `ActivityContent` scores `0`, which is *exactly* `demotedScore`. A trip tracked from the trip page was born indistinguishable from a peer that had been deliberately demoted — it couldn't merely lose the Island, it could never win it.

Also corrected two comments that stopped being true: the enum doc and the test-suite header both still described two start paths, and the "Contrast" comment at `TripLiveActivityRelevanceTests.swift` predicted this exact regression ("if the call site ever drops back to `.init(state:staleDate:)`") while asserting against a locally built value instead of a call site.

## One thing I left alone

The trip page also hand-rolls its `ContentState` with a single arrival, where the bookmark and stop paths go through tested builders that carry up to three. That looks like a separate divergence rather than part of this bug, so I haven't touched it — happy to file it if you agree it's one.

## Summary by CodeRabbit

* **New Features**
  * Live Activities started from bookmarks, stops, and trip pages now use consistent prominence handling.
  * Newly started trip activities are prioritized in the Dynamic Island, while other active trip activities are demoted as appropriate.
  * Live Activity updates continue preserving their established relevance and tracking behavior.

* **Tests**
  * Added coverage for Dynamic Island prominence and demotion behavior.

### DIFF
--- a/OBAKit/Bookmarks/BookmarkActions.swift
+++ b/OBAKit/Bookmarks/BookmarkActions.swift
@@ -147,28 +147,12 @@ final class BookmarkActions {
         }
 
         let attributes = TripAttributes(staticData: staticData)
-        // Prominence so the Dynamic Island switches to this Track when another
-        // trip is already live (#1189 Problem 2). Default score is 0 and equal
-        // scores keep the first-started activity.
-        let prominence = TripLiveActivityRelevance.prominenceScore()
         do {
-            let activity = try Activity.request(
+            let activity = try Activity<TripAttributes>.requestProminent(
                 attributes: attributes,
-                content: TripLiveActivityRelevance.content(
-                    state: contentState,
-                    staleDate: nil,
-                    relevanceScore: prominence
-                ),
-                pushType: .token
+                state: contentState
             )
             trackLiveActivity(activity, arrivalDepartures: arrivalDepartures)
-            let activityID = activity.id
-            Task {
-                await Activity<TripAttributes>.demoteLivePeers(
-                    exceptActivityID: activityID,
-                    relativeTo: prominence
-                )
-            }
             Logger.info("Started Live Activity with ID: \(activity.id)")
             showLiveActivityStartedToast()
             return .started
--- a/OBAKit/LiveActivities/Activity+RequestProminent.swift
+++ b/OBAKit/LiveActivities/Activity+RequestProminent.swift
@@ -1,0 +1,61 @@
+//
+//  Activity+RequestProminent.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import ActivityKit
+import Foundation
+import OBAKitCore
+
+extension Activity where Attributes == TripAttributes {
+
+    /// Starts a trip Live Activity that takes over the Dynamic Island, demoting
+    /// any peers that are still live (#1189 Problem 2).
+    ///
+    /// One prominence score has to reach two places that must agree: the new
+    /// activity's `ActivityContent`, and the demotion the peers are measured
+    /// against. Assembling that pairing by hand is what let the trip page ship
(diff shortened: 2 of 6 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Read the diff. The two existing paths come out semantically identical and the trip page gains both halves it was missing — that's exactly #1375. I particularly like that the doc comment says plainly why the pairing can't be covered by a test and that owning both halves in one place is the substitute; that's the right reason for this helper to exist.

Two follow-ups worth filing rather than holding this on: the trip page's duplicate guard still just logs and returns where the other two call `promoteToDynamicIsland`, so re-Tracking an already-live-but-demoted trip leaves the Island elsewhere; and the trip page still hand-rolls a single-arrival ContentState. You flagged the second one yourself.

PR #1378, opened 2026-09-04
Cover the ignored-jitter and animated branches of StopVehicleAnnotation.update

diveshpatil9104:
Refs #1341 (item 3).

`StopVehicleAnnotation.update` had one test, and it moved the vehicle from (47.6, -122.3) to (48.1, -121.9) — roughly 60 km, or **120× `snapBeyondMeters`**. So it only ever reached `.snap`. The other two branches of `VehicleCoordinateUpdate.decision`, and the restore-then-apply ordering they depend on, had no coverage at all.

| Branch | Condition | Covered before |
|---|---|---|
| `.unchanged` | < 2 m | no |
| `.animate` | 2–500 m | no |
| `.snap` | > 500 m | yes |

## The one that matters

The sub-threshold test is the one that pins the ordering. Below `ignoreBelowMeters`, `apply` is a no-op — so whatever sits in `coordinate` when it runs is what the rider keeps. `tripStatus`'s `didSet` has just overwritten that with the fixture's `lastKnownLocation`, and `update`'s `self.coordinate = from` restore is the only thing putting it back. Delete that line and every poll too small to matter teleports the pin to the fixture's location.

The test asserts the coordinate is unchanged *and* that it isn't `lastKnownLocation`, so it fails for the right reason rather than passing on a coincidence.

Both tests assert their distance falls in the intended band rather than assuming it, so neither can quietly go vacuous if the thresholds are retuned — the same guard `#1366` added after a fixture change had silently defanged a test.

## Scope

Item 3 also names `VehicleCoordinateUpdate.apply`'s own branches. Those tests belong in `VehicleCoordinateUpdateTests.swift`, which **#1356** is already editing, so they're deliberately left out to avoid two open PRs touching one file. They can follow once #1356 lands.

Items 1 (docs) is #1356. Items 2 and 4 remain.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for vehicle annotation coordinate updates when movement is below the ignore threshold.
  * Added coverage for city-block-scale movement, confirming annotations move to the updated coordinate.
  * Tests validate distance assumptions against configured movement thresholds.

### DIFF
--- a/OBAKitTests/Mapping/StopRouteFocus/StopVehicleAnnotationTests.swift
+++ b/OBAKitTests/Mapping/StopRouteFocus/StopVehicleAnnotationTests.swift
@@ -110,6 +110,71 @@ final class StopVehicleAnnotationTests {
         #expect(annotation.routeColor == .systemGreen)
     }
 
+    /// #1341 item 3. The test above jumps (47.6, -122.3) → (48.1, -121.9) — about
+    /// 60 km, or 120× `snapBeyondMeters` — so it only ever reaches `.snap`. The
+    /// other two branches of `VehicleCoordinateUpdate.decision`, and the
+    /// restore-then-apply ordering they depend on, had no coverage at all.
+    ///
+    /// This is the branch that proves the ordering. Below `ignoreBelowMeters`
+    /// `apply` is a no-op, so whatever sits in `coordinate` when it runs is what
+    /// the rider keeps. `tripStatus`'s `didSet` has just overwritten that with the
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Good catch that the existing ~63km fixture never reached either branch. The jitter test in particular earns its place — it's the only thing pinning `update`'s restore-then-apply ordering, and it goes red if that write disappears.

PR #1379, opened 2026-09-04
Render Live Activity staleness on lock screen and Island

### LINKED ISSUE #1376: Live Activity staleness is never rendered
Follow-up from #1215, which wired `stale-date` through the promote/demote path.

`OBAKitCore/LiveActivities/TripLiveActivityRelevance.swift:72` notes the server always sends `stale-date`, and both that file and `LiveActivityUpdateCoalescer.swift:64` take deliberate care to carry the marker through score-only touches. So ActivityKit does mark these activities stale.

Nothing renders it. `OBAWidget/Widgets/TripLiveActivity.swift` never reads `context.isStale` — not in the lock screen view, not in any Dynamic Island region. All 134 lines draw from `context.attributes` and `context.state` alone.

Net: when updates stop — device offline, push token registration failed, server unreachable — the card keeps presenting its last known arrival as current. A rider sees "3 min" that is twenty minutes old, visually identical to a live one. The stale marker the promote/demote path works to preserve has no user-visible consequence today.

Secondary: all three `Activity.request` sites create with `staleDate: nil`, so content is unmarked until the first push lands. That is the intended handoff, but it does mean a Track whose push never arrives stays unmarked for the life of the activity.

There's already in-repo precedent for the treatment — `OBAKit/Mapping/Layers/RentalDetailViewController.swift:244-258` tints the freshness line orange and appends "This data may be out of date." when a rental feed ages past its window. Something equivalent keyed on `context.isStale` would close this.

Borisserz:
## Summary
- Closes #1376: lock-screen card and Dynamic Island now react to `context.isStale` (dim + orange warning), matching the rental-map freshness pattern.
- Shared `LiveActivityStaleChrome` owns copy and opacity so the two surfaces cannot drift; key added in all 13 OBAKitCore locales.
- Docs: `docs/live-activity-stale-chrome.md`.

Intentionally unchanged: creating with `staleDate: nil` until the first push (handoff noted in the issue).

## Test plan
- [x] `LiveActivityStaleChromeTests` (3) — opacity + non-empty warning
- [ ] On device: Track a trip, wait past `stale-date` (or airplane mode) → card dims and shows “This data may be out of date.” / Island warning glyph

## Summary by CodeRabbit

- **New Features**
  - Live Activities now clearly indicate when trip data may be outdated.
  - Stale activities display localized warnings while dimming route and timing content for readability.
  - Warning indicators remain visible across lock-screen, Dynamic Island, and compact presentations, with improved accessibility labels.

- **Localization**
  - Added stale-data warnings in supported languages.

- **Documentation**
  - Documented stale Live Activity behavior and presentation details.

### DIFF
--- a/OBAKitCore/LiveActivities/LiveActivityStaleChrome.swift
+++ b/OBAKitCore/LiveActivities/LiveActivityStaleChrome.swift
@@ -1,0 +1,32 @@
+//
+//  LiveActivityStaleChrome.swift
+//  OBAKitCore
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import Foundation
+
+/// User-visible treatment when ActivityKit marks a Live Activity stale.
+///
+/// Promote/demote and push paths carefully preserve `staleDate` (#1215), but
+/// until the widget reads `context.isStale` a frozen countdown looks identical
+/// to a live one (#1376). Keep the copy and dimming here so lock-screen and
+/// Dynamic Island stay in sync without duplicating literals.
+public enum LiveActivityStaleChrome {
+    /// Same tone as the rental-map freshness footer: orange warning, plain language.
+    public static var warningText: String {
+        OBALoc(
+            "live_activity.stale_warning",
+            value: "This data may be out of date.",
+            comment: "Shown on a Live Activity when ActivityKit marks it stale (no recent update)."
+        )
+    }
+
+    /// Dim the minutes/route chrome so a stale card cannot be mistaken for live.
+    public static func contentOpacity(isStale: Bool) -> Double {
+        isStale ? 0.55 : 1.0
+    }
+}
--- a/OBAKitCore/Strings/ar.lproj/Localizable.strings
+++ b/OBAKitCore/Strings/ar.lproj/Localizable.strings
@@ -423,6 +423,9 @@
 /* Title for the schedule/timetable feature tip */
 "schedule_tip.title" = "اعرض جدول مواعيد خطك";
 
+/* Shown on a Live Activity when ActivityKit marks it stale (no recent update). */
+"live_activity.stale_warning" = "قد تكون هذه البيانات غير حديثة.";
+
 /* Shown in place of the minutes countdown when the vehicle is departing now */
 "stop_page.countdown.now" = "الآن";
 
--- a/OBAKitCore/Strings/en.lproj/Localizable.strings
+++ b/OBAKitCore/Strings/en.lproj/Localizable.strings
@@ -423,6 +423,9 @@
 /* Title for the schedule/timetable feature tip */
 "schedule_tip.title" = "View Your Route's Schedule";
 
+/* Shown on a Live Activity when ActivityKit marks it stale (no recent update). */
+"live_activity.stale_warning" = "This data may be out of date.";
+
 /* Shown in place of the minutes countdown when the vehicle is departing now */
 "stop_page.countdown.now" = "NOW";
 
--- a/OBAKitCore/Strings/es.lproj/Localizable.strings
+++ b/OBAKitCore/Strings/es.lproj/Localizable.strings
@@ -423,6 +423,9 @@
 /* Title for the schedule/timetable feature tip */
 "schedule_tip.title" = "Vea el Horario de su Ruta";
 
+/* Shown on a Live Activity when ActivityKit marks it stale (no recent update). */
+"live_activity.stale_warning" = "Estos datos pueden no estar actualizados.";
+
 /* Shown in place of the minutes countdown when the vehicle is departing now */
 "stop_page.countdown.now" = "AHORA";
 
--- a/OBAKitCore/Strings/fil.lproj/Localizable.strings
+++ b/OBAKitCore/Strings/fil.lproj/Localizable.strings
@@ -423,6 +423,9 @@
 /* Title for the schedule/timetable feature tip */
 "schedule_tip.title" = "Tingnan ang Iskedyul ng Iyong Ruta";
 
+/* Shown on a Live Activity when ActivityKit marks it stale (no recent update). */
+"live_activity.stale_warning" = "Maaaring luma na ang data na ito.";
+
 /* Shown in place of the minutes countdown when the vehicle is departing now */
 "stop_page.countdown.now" = "NGAYON";
 
--- a/OBAKitCore/Strings/fr.lproj/Localizable.strings
+++ b/OBAKitCore/Strings/fr.lproj/Localizable.strings
@@ -423,6 +423,9 @@
 /* Title for the schedule/timetable feature tip */
 "schedule_tip.title" = "Consultez les horaires de votre ligne";
 
+/* Shown on a Live Activity when ActivityKit marks it stale (no recent update). */
+"live_activity.stale_warning" = "Ces données peuvent être obsolètes.";
+
 /* Shown in place of the minutes countdown when the vehicle is departing now */
 "stop_page.countdown.now" = "MAINT.";
 
--- a/OBAKitCore/Strings/it.lproj/Localizable.strings
+++ b/OBAKitCore/Strings/it.lproj/Localizable.strings
@@ -423,6 +423,9 @@
 /* Title for the schedule/timetable feature tip */
 "schedule_tip.title" = "Visualizza l'orario della tua linea";
 
+/* Shown on a Live Activity when ActivityKit marks it stale (no recent update). */
+"live_activity.stale_warning" = "Questi dati potrebbero non essere aggiornati.";
+
 /* Shown in place of the minutes countdown when the vehicle is departing now */
 "stop_page.countdown.now" = "ADESSO";
 
--- a/OBAKitCore/Strings/ko.lproj/Localizable.strings
+++ b/OBAKitCore/Strings/ko.lproj/Localizable.strings
@@ -423,6 +423,9 @@
 /* Title for the schedule/timetable feature tip */
 "schedule_tip.title" = "노선 시간표 보기";
 
+/* Shown on a Live Activity when ActivityKit marks it stale (no recent update). */
+"live_activity.stale_warning" = "이 데이터가 오래되었을 수 있습니다.";
+
 /* Shown in place of the minutes countdown when the vehicle is departing now */
 "stop_page.countdown.now" = "지금";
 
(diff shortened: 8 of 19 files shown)


Borisserz:
CI failed on \`BookmarksViewModelTests\` / corrected-time a11y — unrelated to the stale chrome itself. The assertion hardcoded the English \`scheduled …, now expected …\` template while production goes through \`OBALoc\`, so a non-English preferred locale on the runner makes it fail. Pushed a fix that compares against \`DepartureTimeDisplay.accessibilityTimeDescription\` (same path as production).

aaronbrethorst:
### Code review

Found 1 issue:

1. The stale warning is dimmed by the very opacity it exists to explain. On the lock-screen card, `.opacity(LiveActivityStaleChrome.contentOpacity(isStale:))` is attached to the `VStack` that already contains the new warning row, so the orange caption renders at 0.55. Against `Color(uiColor: .systemBackground)` in light mode that is roughly 1.6:1 contrast — systemOrange on white is only ~2.2:1 even undimmed — and under `isLuminanceReduced` (always-on display) ActivityKit dims again on top of that. The Dynamic Island has the same shape: the warning `Text` at lines 97-101 sits inside the `HStack` that gets `.opacity(staleOpacity)` at line 116. The compact and minimal glyphs are correctly left at full opacity, so the two surfaces already disagree with each other. Dim the arrival content and keep the warning row outside the faded container.

https://github.com/OneBusAway/onebusaway-ios/blob/108aa17f040323c96ab18efa252420ceaeb639fc/OBAKitCore/UI/TripLiveActivityCardView.swift#L44-L56

https://github.com/OneBusAway/onebusaway-ios/blob/108aa17f040323c96ab18efa252420ceaeb639fc/OBAWidget/Widgets/TripLiveActivity.swift#L96-L117

aaronbrethorst (changes requested):
Nicely scoped, and the localization is clean — all 13 catalogs, real translations, purely additive. CI is green on the current head (the earlier red was the `BookmarksViewModelTests` assertion, since rewritten).

One thing to fix first:

**The stale warning is dimmed by its own opacity.** The `.opacity(staleOpacity)` sits on the container that already holds the warning row, so the one element telling the rider the data is stale gets faded along with the data. Orange caption text at 0.55 over `.systemBackground` is roughly 1.6:1 — undimmed systemOrange on white is only ~2.2:1 to begin with — so in light mode the warning is close to invisible, which is the opposite of what it's for. It's internally inconsistent too: the compact and minimal glyphs aren't dimmed.

Dim the content and leave the warning at full opacity.

While you're in there, two small ones: the compact/minimal warning glyphs have no `accessibilityLabel`, so VoiceOver reads the SF Symbol's default description rather than the warning copy; and the `minimal` region currently drops the countdown entirely when stale, replacing the only information it carries with a bare triangle — worth a deliberate decision rather than a side effect.

One note on the drive-by `BookmarksViewModelTests` fix: the new expected value is now character-identical to the production expression it's testing, so the clause content isn't really verified any more. More to the point, the stated root cause — that the English template breaks when the host's preferred language isn't English — doesn't hold, because `DepartureTimeDisplayTests` hardcodes the same English template and passed in the same run. Whatever actually broke there is still unidentified and could come back.

Sequencing: #1328 rewrites the same four Dynamic Island regions and the same card file, so it and this one will conflict hard — `minimal` worst of all, where it substitutes a ticking view and you substitute a glyph. Worth agreeing an order rather than resolving blind. For what it's worth I think the combination is coherent: #1328's countdown ticks against a last-known departure time, so it stays arithmetically honest, and dimming plus an explicit warning is the right treatment for that — provided the warning actually reads, which is the fix above.

Borisserz:
Fixed the stale-warning opacity issue:

- **Lock screen card** / **Island expanded**: dim only the arrival content; warning row sits outside the faded container at full opacity.
- **Compact leading**: triangle keeps `accessibilityLabel` = warning copy.
- **Minimal**: deliberate choice — keep the countdown (orange + dim) instead of replacing it with a bare triangle; compact leading already signals staleness, and minutes are the only ETA minimal can show.

Ready for another look. Happy to rebase after #1328 (or the other way) if the Island regions collide.

aaronbrethorst:
### Code review

Found 2 issues:

1. The `BookmarksViewModelTests` comment states a root cause that is provably false, and the fix it justifies is a no-op. `DepartureTimeDisplayTests.swift#L92` hardcodes the identical English template (`"scheduled \(formatted(0)), now expected \(formatted(3 * 60))"`) against the same `accessibilityTimeDescription` → `OBALoc("departure_time.a11y_rescheduled_fmt")` path, with the same `Locale(identifier: "en_US")` formatters, in the same test target — and it is untouched and passing. If a non-English host language broke this assertion, that test would fail in the same run. Conversely, on an English host `en.lproj` defines that key as exactly the `OBALoc` default, so the new expected string is byte-identical to the old one and commit 108aa17f cannot have changed whether the assertion passes. Whatever actually broke here is still unidentified. (bug due to the comment at `BookmarksViewModelTests.swift#L399-L402` contradicting `OBAKitTests/Models/DepartureTimeDisplayTests.swift#L92`)

https://github.com/OneBusAway/onebusaway-ios/blob/e2854394e84fdf9b7281c47008ed43e30607da14/OBAKitTests/ViewModels/BookmarksViewModelTests.swift#L398-L406

2. The `minimal` region's rationale is wrong, and `docs/live-activity-stale-chrome.md` documents the opposite of what the code does. The comment justifies dropping the warning glyph with "Compact leading already shows the triangle," but the two are never on screen together: per Apple's HIG, "the system uses the compact presentation when only one Live Activity is active" and "when multiple Live Activities are active, the system uses the minimal presentation." So in `minimal` the only stale cue a sighted rider gets is an orange tint at 55% opacity, with nothing naming it. Separately, `docs/live-activity-stale-chrome.md#L8-L9` — added by this PR — says "compact/minimal swap the route chip for a warning glyph when stale," which `minimal` explicitly does not do (and `minimal` never showed a route chip to begin with; it shows minutes). The doc was left behind when commit e2854394 changed `minimal`. (bug due to `OBAWidget/Widgets/TripLiveActivity.swift#L151-L153` vs. `docs/live-activity-stale-chrome.md#L8-L9`)

https://github.com/OneBusAway/onebusaway-ios/blob/e2854394e84fdf9b7281c47008ed43e30607da14/OBAWidget/Widgets/TripLiveActivity.swift#L149-L170

aaronbrethorst (changes requested):
The feature itself is in good shape — `LiveActivityStaleChrome` as the single home for the copy and the opacity is the right call, and the localization work is clean: additive key, all 13 locales, real translations, no literals. The lock-screen card and the expanded bottom region both do exactly what you describe. Three things before this lands.

**1. The expanded-leading warning glyph is still dimmed.**

In `OBAWidget/Widgets/TripLiveActivity.swift`, the leading region swaps `bus.fill` for the orange `exclamationmark.triangle.fill`, and then `.opacity(staleOpacity)` is applied to the enclosing `HStack` — so the warning itself renders at 0.55. That's the opposite of what your own comment in the bottom region states ("Dim headsign/minutes only; warning stays full opacity"), and it's the thing the last round asked for. Moving the opacity onto the `Text` (or applying it to the symbol only when not stale) fixes it.

**2. The `minimal` comment's premise isn't right, and the docs say something else again.**

The comment justifies keeping the countdown with "Compact leading already shows the triangle." Compact and minimal are mutually exclusive presentations — the system uses compact when one Live Activity is active and minimal when several are — so in the minimal case there is no compact leading on screen showing anything. The decision to keep the countdown may well still be the right one (minutes are the only ETA the rider gets there, and you do tint it orange), but the reasoning needs to change.

Meanwhile `docs/live-activity-stale-chrome.md` says "compact/minimal swap the route chip for a warning glyph when stale," which the code doesn't do for minimal — and minimal never showed a route chip. Looks like the doc was left behind when the last commit reversed the minimal behavior.

**3. The bookmark a11y test change doesn't do what its comment claims.**

The new comment says hardcoding the English template "flakes on CI" because it disagrees with `OBALoc` on a non-English host. But `DepartureTimeDisplayTests` hardcodes the identical template through the identical `accessibilityTimeDescription` → `OBALoc("departure_time.a11y_rescheduled_fmt")` path, in the same target, and it's untouched and passing — a host-language effect would hit both. And `OBAKitCore/Strings/en.lproj/Localizable.strings` defines that key as exactly the `OBALoc` default, so the new expected string is byte-identical to the old on an English host. The change is a semantic no-op and whatever the real failure was is still undiagnosed. This is the same point from last round; either drop the change or track down the actual cause.

Smaller, non-blocking: the expanded-leading triangle has no `accessibilityLabel` the way compactLeading and minimal do; and the three `LiveActivityStaleChromeTests` cases can't really fail (an `OBALoc` with a non-empty `value:` default can't return empty, and the other two restate a two-line function).

Ping me when the opacity and the doc are sorted and I'll take another pass.

Borisserz:
Addressed the three items from the latest review:

1. **Expanded leading** — `.opacity(staleOpacity)` now applies only to the route name `Text`; the stale warning triangle stays full opacity and has `accessibilityLabel(LiveActivityStaleChrome.warningText)` like compact leading.

2. **Minimal comment + docs** — corrected the rationale (compact and minimal are mutually exclusive). Updated `docs/live-activity-stale-chrome.md` to match code: minimal keeps the orange/dimmed countdown; VoiceOver gets the warning via `accessibilityLabel`.

3. **Bookmark a11y test** — reverted the no-op `DepartureTimeDisplay` comparison and removed the misleading host-language comment; back to the English-template assertion.

`BookmarksViewModelTests` + `LiveActivityStaleChromeTests` pass locally.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
All three blocking points are fixed. The expanded-leading warning glyph now sits outside the `.opacity(staleOpacity)` so it stays at full opacity while the headsign and minutes dim, the `minimal` comment states the mutual exclusion correctly, and reverting the `BookmarksViewModelTests` change was the right move — that failure wants its own diagnosis rather than riding along here.

I checked the thing I'd most expect to break in a widget: the string resolves through `LiveActivityStaleChrome` in OBAKitCore, which goes via `Bundle(for: CoreLocalization.self)`, so it lands in `OBAKitCore.framework` rather than the appex and actually renders in the extension. All 13 catalogs carry it, purely additive.

Two things for later, neither blocking. In `minimal` the stale cue is itself dimmed — there's nothing undimmed beside it to contrast against, so the 0.55 only weakens the one signal there. And the lock-screen warning is low-contrast even at full opacity: `Color.orange` on `systemBackground` is around 2.2:1 in light mode, under AA. A darker orange or a filled capsule would clear both.

Sequencing note, since it affects you twice: this conflicts hard with your #1328. It replaces the `Text` in the trailing, compactTrailing, minimal and bottom regions with `TickingCountdownText`, which exposes only `font` and `color` — the same four sites this PR gives opacity and an accessibility label. I'm landing this one first deliberately, so that when you rebase #1328 you can design `TickingCountdownText` to carry opacity and the a11y label from the start, rather than bolting them onto a type that already shipped.

Merging.

PR #1380, opened 2026-09-04
Document simulator destinations by UDID in CLAUDE.md

Borisserz:
## Summary
- Refs #1303 (docs half): replace stale `name=iPhone 17 Pro` examples with the CI-style UDID resolution.
- Lint/`MapViewController` body-length half already shipped in #1325.

## Test plan
- [x] `CLAUDE.md` no longer mentions `iPhone 17 Pro` as a destination name
- [ ] Skim that the python one-liner matches the intent of `.github/workflows/tests.yml`

## Summary by CodeRabbit

* **Documentation**
  * Updated build and testing guidance to dynamically select an available iPhone simulator.
  * Replaced hardcoded simulator names with simulator UDIDs for more reliable local and CI test execution.
  * Added guidance explaining why UDIDs are preferred over fixed simulator names.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -79,15 +79,31 @@ the single most expensive failure mode here. Install the CLI instead.
 
 Use these only when `xcodebuildmcp` genuinely cannot do the job.
 
+Prefer `id=` over `name=` for the destination — a fixed name like
+`iPhone 17 Pro` breaks when Xcode ships a new default or when multiple
+matching devices exist. CI in `.github/workflows/tests.yml` picks any
+available iPhone on the newest installed runtime and passes its UDID:
+
 ```bash
+# Resolve a simulator UDID (same idea as CI)
+UDID=$(xcrun simctl list devices available -j | python3 -c "
+import json,sys
+data=json.load(sys.stdin)
+for runtime,devs in data.get('devices',{}).items():
+    if 'iOS' not in runtime: continue
+    for d in devs:
+        if d.get('isAvailable') and d['name'].startswith('iPhone'):
+            print(d['udid']); raise SystemExit
+")
+
 # Build for testing
-xcodebuild clean build-for-testing -scheme 'App' -destination 'platform=iOS Simulator,name=iPhone 17 Pro'
+xcodebuild clean build-for-testing -scheme 'App' -destination "platform=iOS Simulator,id=${UDID}"
 
 # Run unit tests
-xcodebuild test-without-building -only-testing:OBAKitTests -project 'OBAKit.xcodeproj' -scheme 'App' -destination 'platform=iOS Simulator,name=iPhone 17 Pro'
+xcodebuild test-without-building -only-testing:OBAKitTests -project 'OBAKit.xcodeproj' -scheme 'App' -destination "platform=iOS Simulator,id=${UDID}"
 
 # Run specific test class
-xcodebuild test-without-building -only-testing:OBAKitTests/SpecificTestClass -project 'OBAKit.xcodeproj' -scheme 'App' -destination 'platform=iOS Simulator,name=iPhone 17 Pro'
+xcodebuild test-without-building -only-testing:OBAKitTests/SpecificTestClass -project 'OBAKit.xcodeproj' -scheme 'App' -destination "platform=iOS Simulator,id=${UDID}"
 ```
 
 ### Code Quality


aaronbrethorst:
### Code review

Found 3 issues:

1. The snippet does not do what the sentence above it says, and does not match CI. L84-85 claims CI "picks any available iPhone on the newest installed runtime," and L88 labels the code "same idea as CI" — but the loop iterates `data.get('devices',{}).items()` in plain JSON key order with no version sort, so it takes the first iPhone on whichever runtime `simctl` happens to emit first. `simctl` does not emit them in version order. On this machine right now: `iOS-27-0`, `iOS-18-5`, `iOS-26-3`, `watchOS-27-0` — iOS 18 is listed *ahead of* iOS 26. CI's step (`.github/workflows/tests.yml`, "Prepare iOS Simulator") sorts explicitly with `sorted(data, key=version, reverse=True)`, and its comment says why: *"Newest-runtime matters: should the image ever carry a pre-SDK runtime alongside iOS 27, the tests must not land on it — that's the isolated-deinit shim crash."* Fed the same device list with a populated `iOS-18-5` bucket ordered first, the two disagree:

   ```
   PR snippet  -> PICKED: OLD-RUNTIME-DEVICE  com.apple.CoreSimulator.SimRuntime.iOS-18-5
   tests.yml   -> PICKED: E573A8A6-…          com.apple.CoreSimulator.SimRuntime.iOS-27-0
   ```

   So this documents a destination that can land on exactly the pre-SDK runtime the workflow goes out of its way to avoid, while telling the reader it is doing the opposite. Adding the same `iOS-(\d+)-(\d+)` reverse sort fixes it.

https://github.com/OneBusAway/onebusaway-ios/blob/e4ecd985057503617564fc0c04a7e6da10d86faf/CLAUDE.md#L83-L97

2. The empty/no-match case is unguarded, and the CI step this claims parity with does guard it. If no iPhone matches, `$UDID` is the empty string and the destination becomes `platform=iOS Simulator,id=`. Verified against this repo's generated project:

   ```
   $ xcodebuild -project OBAKit.xcodeproj -scheme App -destination "platform=iOS Simulator,id=" -showBuildSettings
   xcodebuild: error: missing value for key 'id' of option 'Destination'
   ```

   `tests.yml` has `[ -n "$UDID" ] || { echo "::error::No available iPhone simulator found"; exit 1; }` for precisely this. There's also no `pipefail`: if `xcrun simctl` itself fails — which it does whenever `xcode-select` points at CommandLineTools rather than a full Xcode — `python3` gets empty stdin, throws a `JSONDecodeError` traceback, and the pipeline still yields an empty `$UDID`, so the reader is pointed at a destination problem instead of the real cause. A `set -o pipefail` plus a non-empty check naming the actual remedy would close both.

https://github.com/OneBusAway/onebusaway-ios/blob/e4ecd985057503617564fc0c04a7e6da10d86faf/CLAUDE.md#L87-L101

3. `$UDID` will usually be unset by the time one of these commands runs. The three `xcodebuild` lines are mutually exclusive alternatives — a clean build-for-testing, a full test run, and a single-class run — so the block cannot be executed top to bottom as written; a reader necessarily picks one line. Any selection that omits L89 loses `$UDID` and lands on the `missing value for key 'id'` error above. That is not a copy/paste nicety here: this file's stated audience is Claude Code, whose Bash tool starts a **fresh shell for every call** — the working directory persists, shell and environment state does not. `tests.yml` sidesteps this only because it writes the UDID to `$GITHUB_ENV`, which survives step boundaries; the doc has no equivalent. Repeating the resolution line inside each command's snippet, or a one-line note that it must run in the same shell invocation as the command that consumes it, resolves it.

https://github.com/OneBusAway/onebusaway-ios/blob/e4ecd985057503617564fc0c04a7e6da10d86faf/CLAUDE.md#L88-L106

aaronbrethorst:
Thanks for this — the direction is right, and the underlying complaint is real: a hardcoded device name goes stale on every Xcode release, and `id=` is the correct fix.

I'm not going to merge it, though, because #1307 has been open since 23 August proposing the same change and covers more ground. Comparing them on the things that actually bite:

- **Runtime ordering.** #1307 sorts by newest installed iOS runtime; this one walks `dict.items()` in emission order. `simctl` on my machine emits `iOS-27-0, iOS-18-5, iOS-26-3` — not sorted — so this can land on a pre-SDK runtime, which is the isolated-deinit-shim hazard `tests.yml` warns about. The prose here says it makes "the same choice as CI", and it doesn't.
- **Empty/no-match guard.** #1307 fails with an actionable message; this falls through to `id=` and the reader gets `xcodebuild: error: missing value for key 'id' of option 'Destination'` with nothing explaining it.
- **Shell persistence.** `$UDID` won't survive between Bash tool calls, and the three `xcodebuild` lines here are alternatives rather than a script, so any single-line copy drops the resolution line. #1307 calls this out explicitly and repeats the line in each block.
- **Coverage.** This only touches the raw `xcodebuild` section; the `xcodebuildmcp` block above it — which CLAUDE.md tells you to reach for first — still has bare `<UDID>` placeholders. #1307 fixes both.
- **Duplication.** #1307 extracts `scripts/resolve_simulator_udid` and has `tests.yml` call it, so there's one implementation. This leaves a second copy that has already diverged from CI in the ways above.

So I'm going to close this in favour of #1307. That's a judgement about the two diffs, not about who filed first — I checked whether the simpler inline approach was the better fit for a docs file, and the simplification is exactly what drops the version sort and the guard.

If you'd like a piece of this to survive: #1307's remaining nit is that it attributes the wrong error string to the empty-variable case, which you'd have hit too. Reviewing that one would be genuinely useful.

Borisserz:
Closing in favor of #1307 — Aaron is right that it already covers newest-runtime sort and empty-UDID guards, which this docs-only PR doesn't. Thanks for the review.

Borisserz:
Superseded by #1307 (fuller UDID resolution).

PR #1381, opened 2026-09-04
Pin Live Activity Track to the selected trip

### LINKED ISSUE #1334: Live Activity tracks stop instead of vehicle
When I bring up a particular vehicle, get its ETA as expected, and tap, "track on lock screen," I expect to be following that vehicle's progress toward the selected station. Instead, I get the same rolling "next train in..." ETA that I'd get if I just tracked the station. This isn't very helpful.

Borisserz:
## Summary
- Closes #1334: `buildContentState(from:matching:)` pins by `tripID` so Track follows the vehicle the rider tapped, not the next same-headsign arrival.
- Opposite-direction mixing (#1326) stays impossible under a tripID key.
- Bookmark Track (no `matching:`) still packs up to three same-key arrivals.

## Test plan
- [x] `BookmarkActionsTests` — pins selected trip; later vehicle does not surface earlier; fallback when trip missing from list
- [ ] On device: open a stop with two same-route same-direction vehicles, Track the later one → Lock Screen shows that ETA, not the earlier “next train”

## Summary by CodeRabbit

* **Bug Fixes**
  * Live Activities now continue tracking the selected vehicle when multiple vehicles share the same route and destination.
  * Prevented updates from displaying the wrong vehicle or mixing information between opposing directions.
  * Bookmark Live Activities now refresh with multiple upcoming arrivals, while explicitly tracked trips remain pinned.
  * Improved reliability when refreshing bookmarks or selecting Track again.
  * Existing Live Activities continue working after the update.

* **Documentation**
  * Added guidance explaining selected-trip tracking and Live Activity refresh behavior.

### DIFF
--- a/OBAKit/Bookmarks/BookmarkActions.swift
+++ b/OBAKit/Bookmarks/BookmarkActions.swift
@@ -79,6 +79,45 @@ final class BookmarkActions {
         (bookmark.routeShortName ?? bookmark.name, bookmark.tripHeadsign ?? "")
     }
 
+    /// The `StaticData` stored when Track is tapped on a bookmark row.
+    ///
+    /// Bookmark identity is stop + route + headsign only. `tripID` is left empty
+    /// so duplicate guards reconcile across refresh rollovers and so refresh uses
+    /// the unpinned multi-arrival builder. Trip pinning belongs on the stop-page
+    /// start path, where the rider picks a specific vehicle.
+    static func liveActivityStaticData(
+        for bookmark: Bookmark,
+        regionID: Int,
+        arrivalDepartures: [ArrivalDeparture]
+    ) -> TripAttributes.StaticData {
+        let (routeShortName, routeHeadsign) = liveActivityKeys(for: bookmark)
+        return TripAttributes.StaticData(
+            routeShortName: routeShortName,
+            routeHeadsign: routeHeadsign,
+            stopID: bookmark.stopID,
+            routeColorHex: arrivalDepartures.first?.route.color?.toHex(),
+            regionID: regionID,
+            tripID: ""
+        )
+    }
+
+    /// Builds refreshed content for a running bookmark Live Activity.
+    ///
+    /// Stop-page activities carry a pinned `tripID` in `StaticData` and keep
+    /// following that vehicle. Bookmark activities leave `tripID` empty and show
+    /// up to the soonest three same-bookmark arrivals.
+    static func buildRefreshContentState(
+        for staticData: TripAttributes.StaticData,
+        arrivalDepartures: [ArrivalDeparture]
+    ) -> TripAttributes.ContentState? {
+        guard !arrivalDepartures.isEmpty else { return nil }
+        if !staticData.tripID.isEmpty,
+           let tracked = arrivalDepartures.first(where: { $0.tripID == staticData.tripID }) {
+            return buildContentState(from: arrivalDepartures, matching: tracked)
+        }
+        return buildContentState(from: arrivalDepartures)
+    }
+
     /// Builds content from whatever arrivals a bookmark row has loaded, or `nil`
     /// when it has none — the bookmark paths have no single departure to fall
     /// back on, so an empty list really is a failure for them.
@@ -130,16 +169,10 @@ final class BookmarkActions {
 
     @discardableResult
     func startLiveActivity(for bookmark: Bookmark, arrivalDepartures: [ArrivalDeparture]) -> TrackResult {
-        let (routeShortName, routeHeadsign) = Self.liveActivityKeys(for: bookmark)
-
-        let routeColorHex = arrivalDepartures.first?.route.color?.toHex()
-        let staticData = TripAttributes.StaticData(
-            routeShortName: routeShortName,
-            routeHeadsign: routeHeadsign,
-            stopID: bookmark.stopID,
-            routeColorHex: routeColorHex,
+        let staticData = Self.liveActivityStaticData(
+            for: bookmark,
             regionID: application.currentRegion?.regionIdentifier ?? 0,
-            tripID: arrivalDepartures.first?.tripID ?? ""
+            arrivalDepartures: arrivalDepartures
         )
 
         // Tapping Track again on a bookmark that is already tracked — or tracking
--- a/OBAKit/Bookmarks/BookmarkActions.swift
+++ b/OBAKit/Bookmarks/BookmarkActions.swift
@@ -89,9 +89,11 @@ final class BookmarkActions {
         return contentState(from: arrivalDepartures)
     }
 
-    /// Builds content for the trip the rider actually tracked: same stop, route,
-    /// and destination as `departure`. Route-only matching mixes both directions
-    /// at a transit center and shows the opposite bus's countdown (#1326).
+    /// Builds content for the trip the rider actually tracked.
+    ///
+    /// Pins by `tripID` so Track follows that vehicle's ETA (#1334), not the
+    /// next same-route/headsign arrival. If the stop list no longer contains
+    /// that trip, fall back to the tapped departure alone.
     ///
     /// Non-optional on purpose. When the stop list no longer contains the tracked
     /// trip — stale data, or a list that never loaded — this falls back to
@@ -101,19 +103,8 @@ final class BookmarkActions {
         from arrivalDepartures: [ArrivalDeparture],
         matching departure: ArrivalDeparture
     ) -> TripAttributes.ContentState {
-        let key = TripBookmarkKey(arrivalDeparture: departure)
-
-        if (departure.tripHeadsign ?? "").isEmpty {
-            // `TripBookmarkKey` substitutes "" for a missing headsign, so the
-            // filter below degrades to stop + route and can readmit the opposite
-            // direction — the very symptom this method exists to prevent. Still
-            // better than showing no trip at all, but it must not be silent.
-            // See: https://github.com/OneBusAway/onebusaway-ios/issues/1326
-            Logger.warn("Departure \(departure.tripID) at stop \(departure.stopID) has no trip headsign; Live Activity arrivals match on stop and route only and may mix directions.")
-        }
-
         let sameTrip = arrivalDepartures
-            .filter { TripBookmarkKey(arrivalDeparture: $0) == key }
+            .filter { $0.id == departure.id }
             .sorted { $0.arrivalDepartureDate < $1.arrivalDepartureDate }
         let source = sameTrip.isEmpty ? [departure] : sameTrip
         return contentState(from: source)
@@ -147,7 +138,8 @@ final class BookmarkActions {
             routeHeadsign: routeHeadsign,
             stopID: bookmark.stopID,
             routeColorHex: routeColorHex,
-            regionID: application.currentRegion?.regionIdentifier ?? 0
+            regionID: application.currentRegion?.regionIdentifier ?? 0,
+            tripID: arrivalDepartures.first?.tripID ?? ""
         )
 
         // Tapping Track again on a bookmark that is already tracked — or tracking
--- a/OBAKit/Bookmarks/BookmarkActions.swift
+++ b/OBAKit/Bookmarks/BookmarkActions.swift
@@ -101,11 +101,31 @@ final class BookmarkActions {
         )
     }
 
+    /// The arrival that should drive OBACloud re-registration metadata for a
+    /// running activity on relaunch.
+    ///
+    /// Prefer the pinned `staticData.tripID` when present and still in the list;
+    /// otherwise the soonest bookmark arrival. Kept in lockstep with
+    /// `buildRefreshContentState`'s pin lookup so push registration can't drift
+    /// onto a different vehicle than the Lock Screen card while the pin holds.
+    static func refreshPrimaryArrival(
+        for staticData: TripAttributes.StaticData,
+        arrivalDepartures: [ArrivalDeparture]
+    ) -> ArrivalDeparture? {
+        if !staticData.tripID.isEmpty,
+           let tracked = arrivalDepartures.first(where: { $0.tripID == staticData.tripID }) {
+            return tracked
+        }
+        return arrivalDepartures.first
+    }
+
     /// Builds refreshed content for a running bookmark Live Activity.
     ///
     /// Stop-page activities carry a pinned `tripID` in `StaticData` and keep
     /// following that vehicle. Bookmark activities leave `tripID` empty and show
-    /// up to the soonest three same-bookmark arrivals.
+    /// up to the soonest three same-bookmark arrivals. When a pin is set but
+    /// that trip has left the list, fall back to the unpinned builder
+    /// (degradation) rather than silently re-pinning the soonest arrival.
     static func buildRefreshContentState(
         for staticData: TripAttributes.StaticData,
         arrivalDepartures: [ArrivalDeparture]
--- a/OBAKit/Bookmarks/BookmarksViewController.swift
+++ b/OBAKit/Bookmarks/BookmarksViewController.swift
@@ -215,12 +215,11 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
             let arrivalDepartures = matchingBookmark.map { viewModel.arrivalDepartures(for: $0) } ?? []
 
             let contentState: TripAttributes.ContentState? = {
-                guard matchingBookmark != nil, !arrivalDepartures.isEmpty else { return nil }
-                if !staticData.tripID.isEmpty,
-                   let tracked = arrivalDepartures.first(where: { $0.tripID == staticData.tripID }) {
-                    return BookmarkActions.buildContentState(from: arrivalDepartures, matching: tracked)
-                }
-                return BookmarkActions.buildContentState(from: arrivalDepartures)
+                guard matchingBookmark != nil else { return nil }
+                return BookmarkActions.buildRefreshContentState(
+                    for: staticData,
+                    arrivalDepartures: arrivalDepartures
+                )
             }()
 
             if matchingBookmark != nil, let contentState {
--- a/OBAKit/Bookmarks/BookmarksViewController.swift
+++ b/OBAKit/Bookmarks/BookmarksViewController.swift
@@ -207,13 +207,23 @@ class BookmarksViewController: UIHostingController<BookmarksRootView>,
                 let bookmarkIdentity = TripAttributes.StaticData(
                     routeShortName: keys.routeShortName,
                     routeHeadsign: keys.routeHeadsign,
-                    stopID: bookmark.stopID
+                    stopID: bookmark.stopID,
+                    tripID: staticData.tripID
                 )
                 return bookmarkIdentity.tracksSameTrip(as: staticData)
             })
             let arrivalDepartures = matchingBookmark.map { viewModel.arrivalDepartures(for: $0) } ?? []
 
-            if matchingBookmark != nil, let contentState = BookmarkActions.buildContentState(from: arrivalDepartures) {
+            let contentState: TripAttributes.ContentState? = {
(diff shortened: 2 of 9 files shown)


aaronbrethorst:
### Code review

Found 2 issues:

1. The pin is undone by the Bookmarks tab's refresh. `updateRunningLiveActivities()` matches every running activity to a bookmark via `tracksSameTrip` (stop + routeShortName + headsign — no `tripID`) and then rebuilds its content with the *unpinned* `buildContentState(from:)` over `viewModel.arrivalDepartures(for: bookmark)`, which is the `TripBookmarkKey`-grouped, soonest-first list. So on any stop the rider has bookmarked, tracking a later vehicle holds the pinned ETA only until the bookmarks list next refreshes, after which the headline reverts to the earlier same-headsign vehicle — the exact symptom in #1334. (bug due to `OBAKit/Bookmarks/BookmarksViewController.swift` line 216 still calling the non-matching builder)

https://github.com/OneBusAway/onebusaway-ios/blob/8b004b8b49fe1415871fdf23fdd49a5e6f872b35/OBAKit/Bookmarks/BookmarksViewController.swift#L214-L218

2. Activity identity is still route + headsign + stop, so a second Track can't re-pin. `Activity<TripAttributes>.running(matching:)` / `tracksSameTrip` carry no `tripID`, so tapping Track on a *different* vehicle of the same route/headsign at the same stop hits the duplicate guard: it promotes the activity pinned to the first vehicle and shows the "Tracking on Lock Screen" confirmation, silently leaving the rider tracking the wrong trip. Before this change both vehicles produced identical content, so promoting was harmless; with `tripID` pinning it no longer is. This is the PR's own manual test scenario ("Track the later one") whenever an activity is already running for that route/headsign/stop.

https://github.com/OneBusAway/onebusaway-ios/blob/8b004b8b49fe1415871fdf23fdd49a5e6f872b35/OBAKit/Stops/StopPage/StopPageActionPresenter.swift#L577-L584

aaronbrethorst (changes requested):
The direction is right and it matches the architecture — the OBACloud push registration is already keyed on `tripID`/`serviceDate`/`vehicleID`/`stopSequence`, and the trip page already builds single-arrival content, so this brings the stop page in line with both. The `[departure]` fallback is preserved, so the non-optional return still holds.

But the pin doesn't survive contact with the two identity paths, and #1334's symptom comes back:

**1. The Bookmarks tab refresh clobbers it.** `BookmarksViewController.updateRunningLiveActivities()` matches running activities by `StaticData.tracksSameTrip` — stop + routeShortName + headsign, no tripID — then rebuilds content with the *unpinned* `buildContentState(from:)` over the grouped soonest-first list. For any rider who has a bookmark on that stop/route/headsign, the card reverts to the earlier vehicle. Before this PR both paths produced identical content, so this change is what creates the divergence.

**2. A second Track can't re-pin.** `Activity.running(matching: staticData)` has no tripID either, so tracking vehicle B while an activity for vehicle A on the same route/headsign/stop is live just promotes A and shows the success toast. That guard was fine while content was trip-agnostic; once content is pinned, promoting a differently-pinned activity is a silent wrong answer — and it defeats your own manual test plan.

Threading `tripID` into `TripAttributes.StaticData` and `tracksSameTrip` fixes both. I'd rather that happened here than in a follow-up, since #1334 isn't really closed until it does — but if you'd prefer to split it, say so and I'll merge this with #1334 staying open.

One thing worth knowing, not blocking: `tripID` alone isn't a unique arrival identity at a stop. `ArrivalDeparture.id`'s own doc says serviceDate and stopSequence exist to "disambiguate the same trip serving this stop more than once: loop routes visit a stop at multiple sequences, and trips can repeat across consecutive service dates" — and `docs/duplicate-trip-stop-ids.md` records the shipped crash from exactly that. `$0.id == departure.id` would be exact. The old key grouped those too, so it's not a regression, but the doc comment's "impossible once the key is the trip itself" is stronger than what the code guarantees.

Rebase note: #1354 and #1377 have landed. Keep #1354's non-optional signature and its `contentState(from:)` helper; drop its headsign-degradation `Logger.warn` and the `Content state matching degrades to route only without a headsign` test, since a missing headsign can't degrade a tripID match — leaving that warn in would log a permanently false claim. Where you both rewrote `Content state matching a departure drops the opposite direction`, yours is the one to keep.

Borisserz:
Addressed review feedback in `bf5a4e55` (rebased onto current `main`):

**Root cause:** Track pinned content by `tripID` in `buildContentState(from:matching:)`, but bookmark refresh and `Activity.running(matching:)` still keyed identity on stop+route+headsign only — so foreground reloads clobbered the pin and tracking vehicle B could promote vehicle A.

**Fix:**
- Added `tripID` to `TripAttributes.StaticData` (decodes as `""` for pre-existing activities).
- `tracksSameTrip` now requires matching `tripID`; legacy cards with empty IDs still reconcile with each other.
- All start paths (stop page, trip page, bookmark Track) store `departure.tripID` in `StaticData`.
- `BookmarksViewController.updateRunningLiveActivities()` rebuilds via `buildContentState(from:matching:)` using the arrival whose `tripID` matches the activity.
- Kept #1354's non-optional `buildContentState(from:matching:)` + `contentState(from:)` helper; matching filters on `$0.id == departure.id` (tripID + stop sequence + service date).
- Dropped the headsign-degradation `Logger.warn` and its test — tripID match can't degrade that way.

**Tests:** `BookmarkActionsTests`, `TripAttributesIdentityTests`, `TripLiveActivityRelevanceTests` — 29/29 passed on iPhone 17 Pro simulator (`FA364C8B-D8FF-4900-AE73-FA9FC3A537F4`).

aaronbrethorst:
### Code review

The two blockers from the last round are genuinely fixed: `tripID` now rides in `StaticData`, `tracksSameTrip` compares it, so `running(matching:)` no longer promotes a differently-pinned activity, and the bookmarks refresh rebuilds pinned activities through `buildContentState(from:matching:)`. `$0.id == departure.id` is the exact key I asked for, and the headsign `Logger.warn` and its test are gone. What's left is new, and it's all downstream of one line: bookmark Track now stores a `tripID` it has no business storing.

Found 2 issues:

1. Bookmark Track keys activity identity on `arrivalDepartures.first?.tripID`, which rolls over. That value is whichever vehicle happens to be soonest at Track time, and it changes every time the earliest arrival departs. Since `tracksSameTrip` now compares `tripID`, the duplicate guard three lines below stops matching: tap Track on a bookmark at 10:00 (first arrival = trip A), let A depart, tap Track again at 10:08 (first arrival = trip B) — `running(matching:)` finds nothing and mints a *second* Live Activity for the same bookmark, with a second OBACloud push registration. Same for bookmark Track after a stop-page Track of any vehicle that isn't the bookmark's first arrival. Both are exactly what the comment on lines 145-150 says must not happen, and what `89a47ff` ("Prevent duplicate Live Activities for the same stop and route") added the guard for. Re-Track also loses its promote-to-Island behavior in that state. `updateRunningLiveActivities()` already copies `staticData.tripID` into the bookmark identity precisely so bookmark reconciliation doesn't depend on it — the start path needs the same treatment, i.e. leave bookmark `StaticData.tripID` empty.

https://github.com/OneBusAway/onebusaway-ios/blob/bf5a4e55e30dd0705bccd01f887e552b23971842/OBAKit/Bookmarks/BookmarkActions.swift#L141-L151

2. Bookmark Live Activities silently collapse from three arrivals to one on the first refresh. Because of #1, a bookmark-started activity always has a non-empty `tripID` (an empty arrivals list fails at the content guard before it can start), and that trip is by construction present in the list, so this branch always takes the `matching:` path. `buildContentState(from:matching:)` filters to `$0.id == departure.id`, which yields exactly one arrival — so the card the rider started with a headline plus two chips loses both chips the next time the Bookmarks tab refreshes (`TripLiveActivityCardView` builds chips from `upcoming.dropFirst()`; `TripLiveActivity` from `upcoming.dropFirst().prefix(2)`). That contradicts the PR description's "Bookmark Track (no `matching:`) still packs up to three same-key arrivals" and the `#expect(state.arrivals.count == 3)` case in `BookmarkActionsTests` — which still passes, because nothing tests the refresh path. Fixing #1 fixes this too; worth a test over `updateRunningLiveActivities`-style input either way.

https://github.com/OneBusAway/onebusaway-ios/blob/bf5a4e55e30dd0705bccd01f887e552b23971842/OBAKit/Bookmarks/BookmarksViewController.swift#L216-L224

Two smaller things, not blocking:

- When the pinned trip drops out of the bookmark's arrival list — it departed, or the feed dropped it — line 223 falls through to the *unpinned* builder and repoints a pinned card at the next three vehicles. It's a defensible degradation (`upcomingArrivals(now:)` would otherwise render the card empty), but it's a silent vehicle swap, and `docs/live-activity-track-selected-trip.md`'s claim that "a foreground reload ... cannot clobber or promote the wrong vehicle" isn't true in that case. Returning `nil` there (leaving the last content in place) or narrowing the doc would both be honest.
- A Live Activity still running from a build before this change decodes `tripID` as `""`, so `Does not match when only one trip ID is empty` means a fresh Track on that same trip mints a duplicate card rather than promoting. Self-clearing within one activity lifetime, and the test says it's deliberate — just noting it's a real upgrade-day symptom.

Note for sequencing: #1298 also touches `OBAKit/Bookmarks/BookmarksViewController.swift`, and its `TrackBookmarkIntent` goes through `BookmarkActions.startLiveActivity`, so it inherits issue 1. #1379 and #1328 are widget-render-side and don't overlap.

aaronbrethorst (changes requested):
The core of this is right, and both blockers from the last round are genuinely fixed — `Activity.running(matching:)` carries `tripID` now, and the bookmarks refresh goes through the pinned builder. Matching on `$0.id == departure.id` rather than `tripID` alone was the right correction too. Two things I hit while reading it, and I think they're both real rather than nitpicks.

**1. Bookmark Track stores a `tripID` that rolls over, which breaks the duplicate guard.**

In `startLiveActivity(for bookmark:arrivalDepartures:)` the static data gets `tripID: arrivalDepartures.first?.tripID ?? ""`. That's whichever bus happens to be next at the moment the rider taps Track — it changes as arrivals depart. But `tracksSameTrip(as:)` now compares `tripID`, so the next time that same bookmark is Tracked, `Activity.running(matching: staticData)` builds a different identity and finds nothing. The rider gets a second Lock Screen card and a second OBACloud push registration for one bookmark — exactly what the comment right below that line says it's preventing, and what commit 89a47ff fixed. Re-Track also stops promoting the existing activity to the Island.

The fix looks like one line: don't set `tripID` on the bookmark path. Bookmark identity is stop + route + headsign; the trip only becomes part of identity when the rider picked a specific vehicle from the stop page.

**2. As a consequence, bookmark cards silently collapse from three arrivals to one.**

Because bookmark activities now always carry a non-empty `tripID` that *is* in the arrivals list, `updateRunningLiveActivities()` always takes the `matching:` branch. That calls `buildContentState(from:matching:)`, which filters to `$0.id == departure.id` and yields exactly one arrival — so `upcoming.dropFirst()` renders no chips and the card drops to a single ETA on the first refresh. The PR body says bookmark Track keeps showing three, and the `count == 3` test still passes because it never exercises the refresh path. Fixing #1 fixes this too.

Worth noting: your own `updateRunningLiveActivities` copies `staticData.tripID` into the bookmark identity before comparing, which neutralizes that field for bookmark matching — so I think you already had the instinct that bookmark identity shouldn't depend on the trip. It just needs to hold at the start path too.

One smaller thing, non-blocking: when the pinned trip leaves the arrival list, the fallback rebuilds a pinned activity from the unpinned list, which can quietly repoint the card at a different vehicle. That makes `docs/live-activity-track-selected-trip.md`'s "cannot clobber the wrong vehicle" line not quite true — worth a sentence acknowledging the degradation.

Happy to re-review as soon as the `tripID` line changes.

Borisserz:
Addressed the review in d0349a0a:

1. **Bookmark start path** — `liveActivityStaticData(for:regionID:arrivalDepartures:)` now always sets `tripID: ""`. Stop-page Track still pins `departure.tripID`.
2. **Refresh collapse** — bookmark cards use `buildRefreshContentState`, which takes the unpinned multi-arrival path when `tripID` is empty (fixes the single-arrival refresh regression).
3. **Docs** — `docs/live-activity-track-selected-trip.md` now distinguishes bookmark vs stop-page identity and documents stop-page degradation when the pinned trip leaves the list.
4. **Tests** — added coverage for empty bookmark `tripID`, `tracksSameTrip` reconciliation, and refresh building 3 arrivals with empty `tripID`.

**Test results** (iPhone 17 Pro sim `FA364C8B…`, after clean `build-for-testing`):
- `OBAKitTests/BookmarkActionsTests` — 14/14 passed
- `OBAKitTests/TripAttributesIdentityTests` — 12/12 passed

aaronbrethorst:
### Code review

Both blockers from the last round are fixed at `d0349a0a`: bookmark `StaticData` no longer carries a rolling `tripID`, `buildRefreshContentState` sends the empty-`tripID` (bookmark) case back through the unpinned three-arrival builder, and `docs/live-activity-track-selected-trip.md` now names the stop-page degradation. What's left is the other half of splitting the two identities.

Found 2 issues:

1. Bookmark and stop-page activities can no longer see each other, so one stop can end up with two Lock Screen cards again. `tracksSameTrip` now compares `tripID`; bookmark Track stores `""` and stop-page/trip-page Track store `departure.tripID`. Track a bookmark (activity with `tripID: ""`), then open that stop and Track the same route/headsign departure: `Activity.running(matching:)` finds nothing, so a second activity is minted with a second OBACloud push registration — the exact thing `89a47ff` ("Prevent duplicate Live Activities for the same stop and route") added the guard for, and which that commit deliberately put on *both* start paths. Three comments still assert it can't happen: `BookmarkActions.swift` line 179 ("or tracking the same trip from the stop page"), `StopPageActionPresenter.swift` line 595 ("can be started from here and from the bookmarks list"), `TripPageViewController.swift` line 437. Leaving bookmark `tripID` empty was the right call for the rollover problem; this is the consequence that needs either an asymmetric match (treat an empty `tripID` on either side as a wildcard) or an explicit decision plus corrected comments and a line in the doc.

https://github.com/OneBusAway/onebusaway-ios/blob/d0349a0a84ec2613afd0c3286e13c2769e027391/OBAKit/Bookmarks/BookmarkActions.swift#L97-L102

2. The bookmarks sweep re-registers a pinned activity with the *wrong* trip's push metadata after a relaunch. `updateRunningLiveActivities()` now honors `staticData.tripID` when it builds local content, but the registration three lines up still passes `metadata: .init(arrivalDepartures.first)` — the bookmark's soonest arrival, not the pinned one. In-session this is unreachable (`isForwardingPushToken` is true for an activity the stop page just started), but after a relaunch `tokenTasks` is empty, so a stop-page activity pinned to trip B whose stop/route/headsign is also bookmarked gets re-POSTed to OBACloud with trip A's `tripID`/`serviceDate`/`vehicleID`/`stopSequence`. The server then pushes A's ETA into B's card, and `store(deleteURL:activityID:)` overwrites the previous session's delete URL, orphaning the old (correctly-keyed) row. That is #1334's symptom surviving a relaunch, through the one path this PR left unpinned. The pinned arrival is already in hand — `buildRefreshContentState` locates it with `first(where: { $0.tripID == staticData.tripID })`.

https://github.com/OneBusAway/onebusaway-ios/blob/d0349a0a84ec2613afd0c3286e13c2769e027391/OBAKit/Bookmarks/BookmarksViewController.swift#L237-L242

Two smaller things, not blocking:

- The comment above the bookmark-identity construction says it "delegates to the shared identity rule so this match can't drift from the start paths' duplicate guard" — but it copies `staticData.tripID` into the candidate identity, which neutralizes exactly the field the start paths now key on. Deliberate and correct; the comment no longer describes it.
- `docs/live-activity-track-selected-trip.md` says the degraded card shows a different vehicle "until the user re-Tracks from the stop page." Re-Tracking the pinned trip hits `running(matching:)` and promotes score-only, so it doesn't repair the content. (#1391 would change that, if it lands.)

Sequencing: I merged this against `#1391`, `#1394` and current `main` locally — `BookmarkActions.swift`, `BookmarkActionsTests.swift`, `TripPageViewController.swift` and `StopPageActionPresenter.swift` all merge clean in every combination, including #1394's test insertion right below yours. One semantic follow-up if #1394 lands after this: two of its new comments describe the headsign warning that this PR deletes. #1379 and #1328 don't overlap.

aaronbrethorst (changes requested):
Both of my standing blockers are fixed, and you followed the rebase notes verbatim — the refresh sweep now honors the pin via `buildRefreshContentState`, a second Track can re-pin, and hard-coding `tripID: ""` for bookmark Track with tests pinning it is exactly what I asked for.

Which is why I want to be clear that the blocker below is the unaddressed second half of *my own* suggestion, not you going off-script.

**Leaving bookmark `tripID` empty breaks the cross-path duplicate guard.**

`tracksSameTrip` now compares `tripID` with exact equality. Bookmark Track stores `""`; stop-page and trip-page Track store `departure.tripID`. So `Activity.running(matching:)` can never reconcile a bookmark-started activity with a stop-page-started one for the same stop, route and headsign — which means two Lock Screen cards and two OBACloud push registrations for one stop. That's the exact failure `89a47ff` added the guard for, and it deliberately put it on *both* start paths. Three comments still assert it's prevented (`BookmarkActions.swift:179`, `StopPageActionPresenter.swift:595`, `TripPageViewController.swift:437`), and `Does not match when only one trip ID is empty` currently encodes the gap as intended behavior.

The fix isn't to reverse the empty-`tripID` decision — that part is right. Make empty a wildcard: if either side's `tripID` is empty, treat the trip as matching and fall back to stop + route + headsign. That keeps bookmark identity unpinned, keeps stop-page identity pinned, and restores the guard across paths.

**Second one, only reachable after a relaunch.** `updateRunningLiveActivities()` honors `staticData.tripID` for local content but still calls `track(activity:metadata: .init(arrivalDepartures.first))`. In-session that's unreachable because `isForwardingPushToken` is true, but after a relaunch `tokenTasks` is empty, so a pinned stop-page activity whose stop/route/headsign is also bookmarked gets re-POSTed with the *soonest* arrival's `tripID`/`serviceDate`/`vehicleID`. The server then pushes the wrong vehicle into the pinned card — #1334's symptom returning through the push path — and `store(deleteURL:activityID:)` orphans the previous session's row. The pinned arrival is already resolved two lines up inside `buildRefreshContentState`; pass that as the metadata.

Non-blocking:

`buildRefreshContentState` finds the pinned trip with a bare `$0.tripID == staticData.tripID`, so a loop route serving the stop twice can match the wrong visit. `StaticData` only carries `tripID`, so a real fix needs more identity fields — probably its own issue rather than scope creep here.

The `BookmarksViewController` comment claims it "delegates to the shared identity rule so this match can't drift from the start paths' duplicate guard", while copying `staticData.tripID` into the candidate and thereby neutralizing the field the start paths key on. Behavior is right, comment isn't.

`docs/live-activity-track-selected-trip.md` says re-Tracking from the stop page repairs a degraded card. It hits `running(matching:)` and promotes score-only — though now that #1391 has landed, passing `state:` would make that true.

`LiveActivityLookup.running(matching:)`'s doc still says "stop and route".

Good news on sequencing: I checked this merges clean against #1394 and #1391 — the `BookmarkActionsTests` hunks are adjacent but non-overlapping. No rebase pain expected beyond the two fixes above.

Borisserz:
Addressed the latest review:

1. Empty `tripID` is now a wildcard in `tracksSameTrip` (bookmark ↔ stop-page duplicate guard restored without pinning bookmark identity).
2. `updateRunningLiveActivities` re-POSTs OBACloud metadata from the pinned arrival via `refreshPrimaryArrival`, not `arrivalDepartures.first`.

Tests updated for the wildcard match.

aaronbrethorst:
### Code review

_Re-review at de5a214, covering commits since the previous review._

Borisserz:

aaronbrethorst (approved):
Verified at de5a2147. Thanks for this.

PR #1382, opened 2026-09-04
Give route-grouped departures the same long-press menu as chronological

### LINKED ISSUE #1383: Route-grouped mode has no long-press menu
Follow-up from #1120.

`ChronologicalListView` renders departures through `DepartureRowView`, which carries `.departureRowActions(_:)` — the long-press menu. `GroupedListView` builds its own rows inline and was never given the provider, so long-pressing a departure in route-grouped mode does nothing at all.

Two of the five menu items are reachable another way in grouped mode: the row tap opens the trip, and the bell icon toggles the alarm. Three are not:

- **Schedules**
- **Add Bookmark**
- **Share Trip**

Sort mode is a persisted per-stop preference (`StopViewModel.updateSortType`, `StopViewModel.swift:511`), so a rider who switches to grouped loses those three for that stop indefinitely — and there is nothing on screen to suggest they exist, since the mode never displays them.

The omission is visible in the type's own signature. `GroupedListView` takes nine inputs — `groups`, `expandedRouteID`, `statusProvider`, `alarmLookup`, `alarmLeadTime`, `canAlarm`, `onToggleRoute`, `onSelectDeparture`, `onAlarmToggle` — and none of them is the actions provider. `GroupedListView.swift` contains zero `contextMenu`.

Worth noting the provider was never missing, only unpassed: `StopDeparturesSections.swift:221` hands `actionsProvider` to `ChronologicalListView` twelve lines above the `GroupedListView` construction that omits it. So this is one argument and two modifier applications, reusing the existing `departureRowActions(_:)` rather than rebuilding the menu — and no new strings, since all five labels are already localized.

One detail for whoever picks it up: the menu needs to reach the collapsed card header too, not only the expanded rows. The header is the route's next departure and is the only thing on screen until a card is expanded, so covering expanded rows alone would still leave a rider a tap short of sharing a trip.

Opened as #1382.

diveshpatil9104:
Follow-up from #1120.

Long-pressing a departure in route-grouped mode did nothing. `ChronologicalListView` renders through `DepartureRowView`, which carries `.departureRowActions(_:)`; `GroupedListView` builds its own rows and never received the provider, so the menu wasn't there.

Three actions were unreachable in that mode:

| Action | Chronological | Grouped (before) |
|---|---|---|
| Show Trip Details | long-press | tap already works |
| Alarm | long-press | bell icon already works |
| **Schedules** | long-press | **unreachable** |
| **Add Bookmark** | long-press | **unreachable** |
| **Share Trip** | long-press | **unreachable** |

Sort mode is a persisted per-stop preference, so a rider who prefers grouped mode lost those three permanently — and had no way to discover they existed.

## The change

`actionsProvider` was already in scope at the call site; `StopDeparturesSections.swift:221` passes it to `ChronologicalListView` twelve lines above and simply didn't pass it to `GroupedListView`. Threading it through and applying the existing `departureRowActions(_:)` modifier is the whole fix — no new menu, no duplicated action list, no new strings. All five labels are already localized.

Applied to **both** the card header and the expanded rows. The header is the route's next departure and is the only thing on screen while a card is collapsed, so a menu on expanded rows alone would still leave a rider one tap short of sharing a trip.

Placement matches `ChronologicalListView`: outside the accessibility element, after the custom actions, so the existing VoiceOver composition is untouched.

## Tests

None. `GroupedListView` and `DepartureRowActions` have no existing coverage, and there's no snapshot infrastructure in this target — a context menu's presence isn't reachable from a unit test here. The change is a parameter and two modifier applications, so it's reviewable on inspection, but I'd rather say that outright than imply coverage that doesn't exist.

Verified by reading instead: one construction site (`StopDeparturesSections.swift:221`, patched), no previews to update, SwiftLint clean on both files.

closes #1383

## Summary by CodeRabbit

* **New Features**
  * Added row actions to grouped departure views, including collapsed route headers and expanded departure rows.
  * Grouped and chronological departure lists now provide consistent row actions.

### DIFF
--- a/OBAKit/Stops/StopPage/Departures/GroupedListView.swift
+++ b/OBAKit/Stops/StopPage/Departures/GroupedListView.swift
@@ -28,6 +28,10 @@ struct GroupedListView: View {
     /// hide the alarm affordance when a new alarm can't be created — while still
     /// showing it for a departure that already has one, so it can be cancelled.
     let canAlarm: (ArrivalDeparture) -> Bool
+    /// Supplies the same long-press menu the chronological rows get. Grouped mode
+    /// renders its own rows rather than reusing `DepartureRowView`, so without
+    /// this the menu simply wasn't there — see `departureRowActions(_:)`.
+    let actionsProvider: (ArrivalDeparture) -> DepartureRowActions
     let onToggleRoute: (RouteID) -> Void
     let onSelectDeparture: (ArrivalDeparture) -> Void
     let onAlarmToggle: (ArrivalDeparture) -> Void
@@ -114,6 +118,11 @@ struct GroupedListView: View {
                 }
             }
         }
+        // Applied outside the accessibility element, matching how
+        // `ChronologicalListView` composes it onto `DepartureRowView`. The header
+        // is the route's next departure and is all a rider sees while the card is
+        // collapsed, so the menu has to reach it too — not just expanded rows.
+        .departureRowActions(actionsProvider(next))
     }
 
     private func alarmActionName(for alarm: Alarm?) -> String {
@@ -319,6 +328,7 @@ struct GroupedListView: View {
                     }
                 }
             }
+            .departureRowActions(actionsProvider(departure))
         }
     }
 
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is the right shape for the fix. I went through the new grouped menu against the chronological one action by action — Show Trip Details, add/remove alarm, Schedules, Add Bookmark, Share Trip — and because you reused `departureRowActions(_:)` rather than rebuilding the menu, they're identical by construction and there's no string to drift. Binding the header to `group.next` and the expanded rows to the loop element is correct, and putting the modifier outside the accessibility element matches how `ChronologicalListView` composes it.

Putting the menu on the collapsed header too was the right call — that header is the only thing a rider sees before expanding.

PR #1384, opened 2026-09-04
Highlight transfer trip on new stop page (#1277)

### LINKED ISSUE #1277: Add Toggle for “Arriving at XX:XX pm via ### - Destination” banner on stop

Borisserz:
## Summary
- Carry optional `fromTripID` on `TransferContext` (populated from `ArrivalDeparture.tripID`) so transfer navigation can identify the inbound trip.
- When the new stop page flag is on, `Router.makeStopController` always uses `StopPageViewController` and passes the effective transfer context (still nil when the Settings toggle is off).
- Highlight matching departure rows on the redesigned stop page (tint + leading accent); do not reintroduce relative NOW/−5m banner timing there. Legacy banner remains only when the new-stop-page flag is off.

Closes #1277

## Test plan
- [x] `TransferContextTests` / `TransferTripHighlightTests` (tripID factory + highlight selection)
- [x] `StopViewModelTests` router cases (new page with transfer; toggle off drops context)
- [x] `ArrivalDepartureItemTransferTests` / `StopPagePresentationTests`
- [ ] Manual: trip → another stop with toggle ON → matching trip row highlighted on new stop page (absolute times)
- [ ] Manual: toggle OFF → ordinary stop page, no highlight
- [ ] Manual: disable new-stop-page flag → legacy banner behavior still works with toggle ON

### DIFF
--- a/OBAKit/Stops/StopPage/Departures/ChronologicalListView.swift
+++ b/OBAKit/Stops/StopPage/Departures/ChronologicalListView.swift
@@ -17,6 +17,7 @@ struct ChronologicalListView: View {
     let partition: StopPageListBuilder.ChronologicalPartition<ArrivalDeparture>
     let walkMinutes: Int?
     let showPast: Bool
+    let highlightedTripID: TripIdentifier?
     let statusProvider: (ArrivalDeparture) -> DepartureStatus
     let alarmLookup: (ArrivalDeparture) -> Alarm?
     let actionsProvider: (ArrivalDeparture) -> DepartureRowActions
@@ -56,6 +57,7 @@ struct ChronologicalListView: View {
         // Identity: ArrivalDeparture.id (stable across prediction refreshes).
         ForEach(departures, id: \.id) { departure in
             let actions = actionsProvider(departure)
+            let isTransferTrip = highlightedTripID == departure.tripID
             DepartureRowView(
                 departure: departure,
                 status: statusProvider(departure),
@@ -63,9 +65,11 @@ struct ChronologicalListView: View {
                 canAlarm: actions.canAlarm,
                 onAlarmToggle: actions.onAlarmToggle,
                 style: style,
+                isTransferHighlight: isTransferTrip,
                 onTap: { onSelectDeparture(departure) }
             )
             .departureRowActions(actions)
+            .listRowBackground(DepartureTransferHighlight.rowBackground(isHighlighted: isTransferTrip))
         }
     }
 }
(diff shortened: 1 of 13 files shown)


Borisserz:
Rebased onto main. Kept `StopPageAccessibilityCopy` identity helpers from main and still append the transfer-trip VoiceOver clause.

aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is the right read of #1277 — the issue asked to replace the banner with highlighting the trip on the stop page, and that's what this does. Threading `fromTripID` through the context rather than trying to re-derive it at the row is the clean way to do it, and the routing change is correct: the new page now owns transfer UX whenever the flag is on, with the legacy banner surviving only behind the flag.

I paid particular attention to the merge, given the conflict markers that went out and came back. It's clean. No markers anywhere in the tree, and the `GroupedListView` resolution correctly kept main's `StopPageAccessibilityCopy.upcomingIdentity` / `groupedCardIdentity` + `DepartureAccessibility.label(extraClauses:)` path and re-applied the transfer clause on top, rather than reverting to the older hand-rolled `String(format:)` versions — which matters, because the resolution could easily have traded a localized path for an unlocalized one and didn't.

Two things to pick up, neither blocking:

The Settings footer is now inaccurate for the default configuration. `settings_controller.arrival_display_section.transfer_banner.footer` still promises "opening a stop from a trip shows times relative to when you arrive," but with the new page on (the default) that toggle produces a highlight, not relative times. You updated `docs/transfer-arrival-banner.md` but not the in-app string. Heads up that #1308 is also rewriting that exact footer right now, so whoever lands second should reconcile the copy across all 13 locales in one pass rather than orphaning translations twice.

And the highlight matches on `tripID` alone, so a loop route that serves the same stop twice will tint both rows. That's the same class of bug #1346 just fixed on the trip page by resolving through `stopSequence` — and `TransferContext.arrivalTime` is right there to disambiguate with. Cosmetic, since both rows genuinely are the rider's trip, but worth closing while the pattern is fresh.

PR #1385, opened 2026-09-05
Fix rental fuel labels vanishing in dark mode (#1364)

### LINKED ISSUE #1364: Rental fuel label's  disappears in dark mode on the SwiftUI map
Image

Image

Borisserz:
## Summary

Closes #1364.

On dark / satellite basemap tiles the under-pin rental fuel figure (range or battery `%`) was easy to lose:

- **MapKit** (`RentalAnnotationView`): `rentalPurple` fill + **white** layer shadow
- **SwiftUI** (`RentalMapMarker`): purple fill + adaptive `systemBackground` shadow (near-black in dark mode)

Both now use a fixed **white figure + black halo**, the same map-label convention stop labels already use, independent of color scheme and operative tint.

## Before / after

Dark mode · satellite · scooters layer · downtown Seattle.

| Before | After |
| --- | --- |
| Before — purple text + white glow | After — white text + black halo |

Close-up of the glyph under the pin:

Side-by-side crop

## Test plan

- [x] `OBAKitTests/RentalAnnotationViewTests` — fuel label is white with black shadow (operative and not)
- [x] Simulator: dark appearance, Map layers → Scooters on, Satellite, zoom until fuel/range labels show
- [ ] Confirm labels stay readable on Standard / Hybrid as well as Satellite
- [ ] Confirm non-operative (gray pin) fuel figure still reads as white+halo
- [ ] Optional follow-up: drop `docs/screenshots-1364/` before merge if you prefer screenshots only in the PR body

## Summary by CodeRabbit

- **Improvements**
  - Rental map fuel labels now use white text with a black outline for improved readability across light and dark map backgrounds.
  - Fuel label styling remains consistent between map annotations and map markers.
  - Labels now respond appropriately when Dynamic Type text-size settings change.

- **Documentation**
  - Added guidance on fuel-label appearance and dark-mode visibility.


aaronbrethorst:
### Code review

Found 1 issue:

1. The fuel figure is now hardcoded `.white` in both renderers, so the legibility problem is moved rather than removed: it is now weakest in the app's *default* configuration — light appearance on the `.mutedStandard` basemap (`MapRegionManager.swift:90`, and the registered default at `:237`; `MapBaseType` defaults to `.standard`). On a near-white tile the glyph body has essentially no contrast of its own, and everything depends on the halo. On the UIKit side that halo is a `CALayer` shadow with `shadowRadius = 2` / `shadowOffset = .zero` — a Gaussian blur, not a stroke — so blurring a ~1.5–2pt bold caption glyph over a 2pt radius leaves a peak alpha well below 1, i.e. a white figure sitting on a light-gray smear. Two supporting signals: the SwiftUI side uses a noticeably tighter, denser halo (`radius: 1.25 @ 0.95` plus `radius: 0.5 @ 0.8`), so the two renderers do not agree on how much halo the same figure needs; and #1364 is titled "…on the SwiftUI map", so the weaker treatment is being applied to the renderer that did not exhibit the reported bug (`rentalPurple` + a *white* halo was the one combination that stayed legible on dark tiles). The PR's own test plan leaves "Confirm labels stay readable on Standard / Hybrid as well as Satellite" unchecked. Worth either making the treatment appearance/basemap-aware, or using a real outline (stroke via `NSAttributedString.Key.strokeColor`/`strokeWidth` on the UIKit label) rather than a blur, and verifying light + Standard before merge.

https://github.com/OneBusAway/onebusaway-ios/blob/d893b7a65dd9ce5b1ccb142f61b149b83dd71e72/OBAKit/Mapping/Layers/RentalAnnotationView.swift#L119-L121

https://github.com/OneBusAway/onebusaway-ios/blob/d893b7a65dd9ce5b1ccb142f61b149b83dd71e72/OBAKit/Sheet/Root/Controls/RentalMapMarker.swift#L53-L57

aaronbrethorst (changes requested):
The bug is real and your diagnosis of it is right — purple text with a light halo genuinely does disappear against dark basemap tiles, and I appreciate the screenshots.

My concern is that the fix moves the problem rather than removing it.

Hardcoding `.white` with a black halo makes the label a satellite/dark treatment applied unconditionally. But this app's default basemap is `.mutedStandard` — `MapRegionManager.swift:90`, and it's the registered default at `:237` — so the most common case by far is a near-white tile with a white glyph on it. The only thing holding that up is the halo, and the halo is a `CALayer` shadow (`shadowRadius = 2`, `shadowOffset = .zero`), which is a soft blur rather than an outline. Behind a thin bold-caption glyph its peak alpha is well under 1.

Two things make me think this wasn't verified rather than considered-and-accepted: the test plan's "Confirm labels stay readable on Standard / Hybrid as well as Satellite" is unchecked, and #1364 is specifically about the SwiftUI map, yet the same treatment is applied to the UIKit renderer where purple-plus-white-halo was the pairing that already worked on dark tiles.

Either of these would resolve it:

- Make the treatment appearance- or basemap-aware, the way Apple's own map labels flip polarity.
- Keep a fixed white figure but give it a real outline — `NSAttributedString.Key.strokeColor` / `strokeWidth` — rather than a blurred shadow. That's what makes the white-on-anything idiom actually work.

Either way I'd want a light-mode screenshot on the Standard basemap before this lands, since that's what most riders see.

Two other things:

The two renderers now disagree about how much halo the same figure needs — UIKit uses one shadow at `radius 2, opacity 1`, SwiftUI uses `1.25 @ 0.95` plus `0.5 @ 0.8` — while `RentalMapMarker`'s own doc comment says it "Approximates `RentalAnnotationView`". Worth converging once you pick a treatment.

The ~677 KB of JPEGs in `docs/screenshots-1364/` — `docs/` is all Markdown today with no binary precedent. You offered to drop them; I'd take that. Attaching them to the PR or the issue keeps them visible without putting them in the repo forever. (I realize the hotlinks to your fork die when the branch goes; the issue is the better home.)

The rest is clean — no localization surface, the tests assert something real now, and you covered both render sites rather than just the reported one.

diveshpatil9104 (commented):
The overall result looks good🚀 One small correction to the close-up Before/After: the current screenshots don't clearly show the actual fuel/distance label under the pin that this PR is fixing.

The close-up should show the label underneath the pin in both states, so the comparison clearly demonstrates the actual issue being addressed.

| Before | After |
|:---:|:---:|
| Before — purple fuel distance label under the pin | After — white fuel distance label with black outline under the pin |


The result looks good!!  this is just to make sure the close-up comparison accurately shows the fuel/distance label being changed.

Borisserz:
Addressed the review by switching to Aaron's second option: white figure + real outline (`NSAttributedString` stroke / `MapLabelOutline`), not a soft shadow. That stays legible on the default `.mutedStandard` light basemap and on dark tiles without hardcoding a satellite-only treatment.

Also converged UIKit + SwiftUI on that model and removed the ~677 KB JPEGs from `docs/screenshots-1364/` (screenshots stay on the PR/issue instead).

aaronbrethorst:
### Code review

_Re-review at 906b54b, covering commits since the previous review._

Borisserz:

aaronbrethorst (approved):
Verified at 906b54b6. Thanks for this.

PR #1386, opened 2026-09-05
Keep every route on bookmark map callouts (#1342)

### LINKED ISSUE #1342: map_routes_overflow_fmt missing from all locales; comments overclaim

Borisserz:
## Summary

Closes #1342 (remaining after #1359).

#1359 already localized `formatters.map_routes_overflow_fmt` and dropped the callout/subtitle vacuous asserts. What was still open:

1. **Product decision:** bookmark pin titles are `bookmark.name`, not the route list — truncating the callout hid routes with no pin/callout consistency benefit. Bookmarks now use `Stop.bookmarkCalloutText` (`formattedMapRoutes` with `limit: .max`). Stop pins keep the 3 + `…` callout.
2. Removed the remaining vacuous `#expect(!label.contains("more"))` on the map-routes formatter test.
3. Updated `docs/map-route-labels.md` so it matches the code.

## Test plan

- [x] `FormattersTests` — bookmark callout lists every route; stop callout still truncates
- [ ] Simulator: bookmark a multi-route stop, tap the pin, confirm callout shows all routes
- [ ] Same stop (unbookmarked) still shows `10, 12, 49…` under the pin and in the callout

## Summary by CodeRabbit

- **New Features**
  - Bookmark stop callouts now list every route served by the stop without truncation.
  - Bookmark callouts display the rider-visible stop code alongside the complete route list.

- **Documentation**
  - Updated map and bookmark route-label documentation to clarify overflow markers and full route visibility in bookmark callouts.

### DIFF
--- a/OBAKit/Extensions/ModelExtensions.swift
+++ b/OBAKit/Extensions/ModelExtensions.swift
@@ -66,6 +66,15 @@ nonisolated extension Stop: @retroactive MKAnnotation {
     public var mapCalloutText: String {
         ["#\(code)", mapTitle].compactMap { $0 }.joined(separator: "\n")
     }
+
+    /// Callout text for a **bookmark** pin. The pin title is `bookmark.name`, not
+    /// the route list, so truncating routes in the callout would hide information
+    /// with no pin/callout consistency benefit (#1342). Same compact style as
+    /// `mapCalloutText`, but every route (`limit: .max`).
+    public var bookmarkCalloutText: String {
+        let routes = Formatters.formattedMapRoutes(self.routes, limit: .max)
+        return ["#\(code)", routes].compactMap { $0 }.joined(separator: "\n")
+    }
 }
 
 // MARK: - TripStatus/MKAnnotation
--- a/OBAKit/Mapping/StopAnnotationView.swift
+++ b/OBAKit/Mapping/StopAnnotationView.swift
@@ -146,7 +146,7 @@ class StopAnnotationView: MKAnnotationView {
         labelStack.isHidden = delegate.shouldHideExtraStopAnnotationData
         image = delegate.iconFactory.buildIcon(for: bookmark.stop, isBookmarked: true, traits: traitCollection)
         titleLabel.attributedText = strokedText(bookmark.name)
-        detailCalloutAccessoryView = buildDetailLabel(text: bookmark.stop.mapCalloutText)
+        detailCalloutAccessoryView = buildDetailLabel(text: bookmark.stop.bookmarkCalloutText)
     }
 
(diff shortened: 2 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Checked the part that mattered: `mapTitle` is itself `Formatters.formattedMapRoutes(routes)`, so `bookmarkCalloutText` really is the same list at `limit: .max` rather than swapping what the callout shows. And since `MapRegionManager` drops the `Stop` annotation once a `Bookmark` exists for it, the truncation split is by annotation type — one stop can't render two pins with disagreeing callouts.

Reasoning in the doc holds up too. Merging.

PR #1387, opened 2026-09-05
Align vehicle interpolation docs with 500 m snap (#1341)

### LINKED ISSUE #1341: Vehicle interpolation: docs contradict the 500m snap threshold

Borisserz:
## Summary

Closes #1341.

1. **Docs match the constant.** `VehicleCoordinateUpdate` header (and related comments) now describe `snapBeyondMeters` as **500 m** / city-traffic (~50 km/h). Freeway buses near 60 km/h usually snap — that is intentional, not a motorway-speed promise.
2. **`apply(from:to:on:)` tests** for `.snap`, `.unchanged`, and `.animate` (waits past the 0.8 s duration).
3. **Legacy `TripViewController`** no longer teleports: same restore-then-`apply` pattern as `TripFocusMapLayer` / `StopVehicleAnnotation`.
4. Short doc: `docs/vehicle-coordinate-interpolation.md`.

## Test plan

- [x] `VehicleCoordinateUpdateTests` (8) green, including apply cases
- [ ] Trip page: watch a city bus hop smoothly between polls
- [ ] Large jump (or first fix) still snaps

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved vehicle marker movement between location updates. Short-distance movements now animate smoothly, while larger jumps snap to the latest location.
  - Prevented vehicle annotations from jumping unexpectedly when trip information is refreshed.
  - Removed stale vehicle markers when location data is unavailable, preventing pins from appearing at an incorrect default location.

- **Documentation**
  - Added guidance explaining interpolation, noise suppression, the 500-meter snapping threshold, and behavior when vehicle coordinates are missing.

### DIFF
--- a/OBAKit/Mapping/Layers/StopRouteFocus/StopVehicleAnnotation.swift
+++ b/OBAKit/Mapping/Layers/StopRouteFocus/StopVehicleAnnotation.swift
@@ -95,9 +95,7 @@ final class StopVehicleAnnotation: VehicleAnnotation {
         // position-preferred coordinate must be assigned AFTER, or it gets
         // clobbered exactly as the initializer's comment describes.
         //
-        // Restore `from` before `VehicleCoordinateUpdate.apply` so a city-block
-        // hop interpolates instead of that didSet teleporting the pin. A
-        // kilometre-scale jump still snaps; see VehicleCoordinateUpdate.
+        // A kilometre-scale jump still snaps; see VehicleCoordinateUpdate (#1341).
         let from = self.coordinate
         self.tripStatus = tripStatus
         self.coordinate = from
--- a/OBAKit/Mapping/Layers/StopRouteFocus/StopVehicleAnnotation.swift
+++ b/OBAKit/Mapping/Layers/StopRouteFocus/StopVehicleAnnotation.swift
@@ -95,6 +95,8 @@ final class StopVehicleAnnotation: VehicleAnnotation {
         // position-preferred coordinate must be assigned AFTER, or it gets
         // clobbered exactly as the initializer's comment describes.
         //
+        // Restore `from` before `VehicleCoordinateUpdate.apply` so a city-block
+        // hop interpolates instead of that didSet teleporting the pin.
         // A kilometre-scale jump still snaps; see VehicleCoordinateUpdate (#1341).
         let from = self.coordinate
         self.tripStatus = tripStatus
--- a/OBAKit/Mapping/VehicleCoordinateUpdate.swift
+++ b/OBAKit/Mapping/VehicleCoordinateUpdate.swift
@@ -16,10 +16,13 @@ import UIKit
 ///
 /// Arrival polls land every 15–30s. Assigning `coordinate` each time teleports
 /// the pin; interpolating every hop looks like motion. Hops longer than
-/// `snapBeyondMeters` are a new fix (or a trip change), not something to ease
-/// across the map.
+/// `snapBeyondMeters` (**500 m** — a bit over one 30s poll at ~50 km/h) snap
+/// instead. That is a deliberate city-traffic tuning: freeway/express buses
+/// near or above ~60 km/h usually snap and degrade to the pre-interpolation
+/// behavior rather than easing across hundreds of metres.
 ///
 /// See: https://github.com/OneBusAway/onebusaway-ios/issues/1109
+/// Follow-up: https://github.com/OneBusAway/onebusaway-ios/issues/1341
 enum VehicleCoordinateUpdate {
     enum Decision: Equatable {
         case unchanged
--- a/OBAKit/Trip/TripViewController.swift
+++ b/OBAKit/Trip/TripViewController.swift
@@ -272,7 +272,16 @@ class TripViewController: UIViewController,
             }
 
             if let vehicleAnnotation = vehicleAnnotation {
+                // `tripStatus`'s didSet writes lastKnownLocation onto coordinate
+                // immediately. Restore `from` so `VehicleCoordinateUpdate` can
+                // interpolate instead of teleporting (#1341) — same pattern as
+                // `TripFocusMapLayer.drawVehicle`.
+                let from = vehicleAnnotation.coordinate
                 vehicleAnnotation.tripStatus = currentTripStatus
+                vehicleAnnotation.coordinate = from
+                if let to = currentTripStatus.lastKnownLocation?.coordinate {
+                    VehicleCoordinateUpdate.apply(from: from, to: to, on: vehicleAnnotation)
+                }
                 // Update the annotation view's heading and real-time state since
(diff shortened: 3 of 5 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (changes requested):
The core of this is good — I verified `snapBeyondMeters` is defined once as `500` and that "hops longer than 500 m snap" matches `if meters > snapBeyondMeters`, so the number and the condition are both right. The three new `apply(from:to:on:)` tests are worth having.

Two things to sort out first.

**1. There's a behavior change in here that the PR presents as documentation.**

In `TripViewController`, the old code let `tripStatus`'s `didSet` write `lastKnownLocation?.coordinate ?? (0,0)` and land on the coordinate. The new code restores `from` unconditionally and only calls `apply` when `to` is non-nil. So when `lastKnownLocation` is nil the pin now keeps its previous real coordinate instead of going to null island — which means `annotationsToShow.removeAll(where: { $0.coordinate.isNullIsland })` at line 525 no longer filters it, and the stale pin both stays visible and drags the `showAnnotations` zoom fit.

That may well be an improvement — a stale pin does beat a teleport to the Gulf of Guinea — but it's a user-visible change in a PR titled "align docs", it isn't mentioned in the body, and `TripFocusMapLayer` handles the same case differently by calling `removeVehicle()`. I'd like it either handled consistently with the layer, or called out explicitly and justified.

**2. The diff deletes the rationale it simultaneously declares mandatory.**

`StopVehicleAnnotation.swift` loses "Restore `from` before `VehicleCoordinateUpdate.apply` so a city-block hop interpolates instead of that didSet teleporting the pin", leaving a bare `A kilometre-scale jump still snaps` above a `let from` / `self.coordinate = from` pair with nothing explaining why they exist. Meanwhile the new `docs/vehicle-coordinate-interpolation.md` says callers "must restore the previous `coordinate` before `VehicleCoordinateUpdate.apply`" and lists that exact call site.

So the three sites are less consistent after a PR whose purpose is doc alignment. Please keep the rationale at the call site.

Smaller notes, no need to act unless you agree:

The doc says a 30 s poll at 60 km/h "is exactly 500 m, so freeway/express buses usually snap" — at exactly 500 m the comparison is `>`, so it animates. It's hedged enough to be defensible, and it's close to my own wording in #1341, so I'm not fussed.

`#1341` is cited for the snap behavior in a couple of places, but that came from #1109/#1323.

The 900 ms sleep in the animate test doesn't buy anything — `UIView.animate`'s closure runs synchronously and `MKPointAnnotation.coordinate` has no presentation-layer interpolation, so the assertion holds at t=0. The test still usefully proves the `.animate` branch writes at all; just drop the sleep.

The md file describes interpolate-vs-snap but never mentions `ignoreBelowMeters`, which is one of the three branches.

Borisserz:
Addressed the review:

1. **Behavior called out and aligned:** when `lastKnownLocation` is nil, `TripViewController` now removes the vehicle annotation — same policy as `TripFocusMapLayer.removeVehicle()` — instead of keeping a stale real coordinate that skipped the null-island filter and dragged `showAnnotations`.
2. Restored the “restore `from` before apply” rationale on `StopVehicleAnnotation`.
3. Doc now covers all three branches (including `ignoreBelowMeters`); dropped the useless 900 ms sleep in the animate test.

This is still primarily a docs/threshold PR; the nil-location change is the intentional consistency fix for the behavior the previous revision introduced without naming it.

aaronbrethorst:
### Code review

_Re-review at 9483717, covering commits since the previous review._

Borisserz:
The snap citation in `StopVehicleAnnotation` now points at #1109 / #1323. #1341 is the doc follow-up, not the change that defined the 500 m cliff.

aaronbrethorst:
### Code review

_Re-review at 6d8ad38, covering commits since the previous review._

aaronbrethorst (approved):
This came back in good shape — all six items from the last round are addressed, including both blockers, and the issue citation fix in the latest commit was the last one outstanding.

The part I cared most about was the `TripViewController` change, and it's right: saving `from` before assigning `tripStatus`, restoring it after the `didSet` overwrites `coordinate`, then handing both to `VehicleCoordinateUpdate.apply` is the same dance `TripFocusMapLayer.drawVehicle` does. Both early-return paths still call `updateTitleView()`, so the title doesn't go stale when the pin drops. And not minting a null-island pin when the feed has no location yet is a straight improvement over the old `(0,0)` fallback.

The prose checks out against the code, which is the bar for a doc PR: 500 m is the single definition of `snapBeyondMeters`, the strict `>` at exactly 500 m is called out, `ignoreBelowMeters` is described correctly, and "a bit over one 30 s poll at ~50 km/h" is honest arithmetic (417 m).

One residual edge for whenever you're next in here, not worth holding this up: the new guard is `lastKnownLocation?.coordinate`, which is nil only when the location is *absent*. `TripFocusMapLayer.vehicleCoordinate` additionally rejects a literal null-island fix via `!status.coordinate.isNullIsland`, so a feed reporting `(0,0)` still keeps a pin on the legacy screen where the focus layer would drop it. The zoom filter hides it in practice, and the doc's wording is literally true either way.

PR #1388, opened 2026-09-05
Pin SwiftLint in CI to portable 0.65.1 (#1340)

### LINKED ISSUE #1340: SwiftLint CI gate placement, pinning, and a stray de-privatization

Borisserz:
## Summary

Closes #1340.

Most of the issue is already on `main` from Divesh's follow-ups to #1325:

1. ~~`longPressGesture` → `private`~~ done
2. ~~Lint as its own job beside `build`~~ done
3. **This PR:** pin SwiftLint via the official **portable 0.65.1** release (`SWIFTLINT_PIN`), so an unpinned `brew install` cannot promote a new error-severity rule and turn main red with no app change
4. ~~Stale body-length ceiling comments~~ already fixed on the extension files

Doc: `docs/swiftlint-ci-gate.md`.

## Test plan

- [ ] CI `lint` job downloads portable 0.65.1 and passes
- [ ] Bumping `SWIFTLINT_PIN` without reviewing rules is a deliberate follow-up, not silent drift

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,14 +46,15 @@ jobs:
     # in CI (scripts/swiftlint.sh); this job is the review-visible gate.
     # See https://github.com/OneBusAway/onebusaway-ios/issues/1303
     - name: SwiftLint (errors only)
-      env:
-        HOMEBREW_NO_AUTO_UPDATE: 1
-        HOMEBREW_NO_INSTALL_CLEANUP: 1
       run: |
-        if ! command -v swiftlint >/dev/null; then
-          brew install swiftlint
-        fi
-        swiftlint lint --reporter github-actions-logging
+        # Pin via the official portable release so a Homebrew formula bump cannot
+        # promote a new error-severity rule and turn main red with no app change
+        # (#1340). Bump SWIFTLINT_PIN deliberately after reviewing new rules.
+        SWIFTLINT_PIN="0.65.1"
+        curl -sL "https://github.com/realm/SwiftLint/releases/download/${SWIFTLINT_PIN}/portable_swiftlint.zip" -o /tmp/swiftlint.zip
+        unzip -qo /tmp/swiftlint.zip -d /tmp/swiftlint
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Good change — an unpinned `brew install swiftlint` meant a formula bump could promote a new error-tier rule and turn `main` red with no app change, which is exactly what #1340 was about.

Two things I checked rather than assumed: the pin covers both paths that run SwiftLint (this job and the Xcode build phase via `scripts/swiftlint.sh`, which still skips under `$CI`), and `docs/` is the right home here since it holds committed Markdown while generated docs go to `api-docs/`.

One thing worth a follow-up rather than a block: the pin is a mutable tag plus a self-reported version string, so it guards against formula drift but not against a replaced release asset. A recorded `shasum -a 256 -c` would make it a real pin. Likewise `curl --fail` would name the actual cause on a bad bump instead of surfacing "cannot find zipfile directory" — though Actions' `bash -e` does at least make it fail loudly rather than silently.

Merging.

Issue #1338, opened 2026-09-02
Confirm #421 on device before closing it


aaronbrethorst:
Follow-up from #1327, which added `id` to `OBAListViewSection.==` so it matches `hash(into:)`.

The fix is correct on its own merits — the Equatable/Hashable contract really was violated. But the in-code comment asserts a causal link that hasn't been demonstrated:

```swift
// `id` belongs here: `hash(into:)` already combines it, and
// `NSDiffableDataSource` treats Hashable mismatch as "Failed to find
// index of item" on the header that uses this section's id (#421).
```

The failing lookup in #421's log is an *item* (`OBAListViewHeader`); the fix is at the *section* level. It's plausible that `apply(sectionSnapshot, to: section)` resolves the section identifier by hash+equality and that's the mechanism — but it's inference, and #1327's own test plan left the device verification unticked.

Before closing #421, someone should reproduce the log spam on device and confirm it stops. If it doesn't, #421 is still open and the contract fix was just a good thing to do anyway.

Borisserz:
On the queue: I can verify the Equatable/Hashable contract fix from #1327 is still on \`main\`, but I do not have a reliable device repro of the original \"Failed to find index of item\" spam from #421 in Simulator today. Happy to run a focused device pass if you have a short repro sequence, or leave #421 open until someone hits the log on device.

Borisserz:
Follow-up verification + comment cleanup:

1. **Simulator pass (current main, #1327 already present):** More → Alerts, toggled Community Transit collapse several times while streaming device logs for `Failed to find index` / `DiffableDataSource`. **No spam** in that session. I do not have a physical-device repro of the 2021 log, so this is best-effort, not a proof that #421 is gone forever.

2. **PR softens the overclaim:** the `==` / test comments no longer assert that the Hashable mismatch *caused* the DiffableDataSource spam — only that including `id` is required for a correct contract, and that the spam link is unproven (#1338).

I cannot close #421 from here without a confirmed before/after on the original spam. Happy to leave #421 open until someone hits the log again (or Aaron is satisfied with “no repro + contract fixed”).

PR #1327, opened 2026-08-28
Include section id in OBAListViewSection equality

Borisserz:
## Summary
- `OBAListViewSection.hash(into:)` already combined `id`; `==` compared title, contents, and configuration only. Same title + empty contents with different ids hashed differently but compared equal, so `UICollectionViewDiffableDataSource` could not find `OBAListViewHeader` items (agency alerts used ids like `agency_alerts_Washington State Ferries`).
- Equality now includes `id`, matching `Hashable`.

The freeze-on-bookmarks reports in older comments were a separate issue (Alan + Aaron agreed at the time).

Closes #421.

## Test plan
- [x] `OBAListViewSectionTests` (2/2)
- [ ] Device: open a region with multiple agency-alert sections on the map/stop list; scrolling should not crash with "Failed to find index of item OBAListViewHeader"

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved list section comparisons to distinguish sections with different identifiers.
  * Ensured matching sections produce consistent equality and hash results.

PR #1389, opened 2026-09-05
Soften #421 DiffableDataSource causality claim (#1338)

### LINKED ISSUE #421: Agency Alerts: OBAListView DiffableDataSource can't find item index of headers

Borisserz:
## Summary

Addresses #1338 (does not close #421).

#1327 correctly put `id` into `OBAListViewSection.==` so it matches `hash(into:)`. The in-code comment claimed that mismatch caused DiffableDataSource `"Failed to find index of item"` spam on headers — that causal link was never device-verified (#1338).

This PR only softens the comments in production and tests so they state:

- `id` belongs in `==` for a correct Hashable contract
- a link to the historic spam is **plausible but unproven**

## Verification attempted

On Simulator (Puget Sound → More → Alerts), opened Community Transit and toggled the section header repeatedly while streaming logs for `Failed to find index` / `DiffableDataSource`. **No spam observed** on current `main` with the #1327 fix already present. That does **not** prove #421 is fixed historically — only that I could not reproduce the log today.

## Test plan

- [x] Comment-only change; existing `OBAListViewSectionTests` still describe the contract
- [ ] Optional: anyone with a device repro from #421 can confirm spam is gone and then close #421

## Summary by CodeRabbit

- **Documentation**
  - Clarified the requirements for consistent equality and hashing behavior.
  - Updated related test documentation to distinguish confirmed contract behavior from unverified historical logging causes.
  - No user-facing functionality or public APIs changed.

### DIFF
--- a/OBAKit/Controls/ListView/ViewModels/OBAListViewSection.swift
+++ b/OBAKit/Controls/ListView/ViewModels/OBAListViewSection.swift
@@ -89,9 +89,12 @@ nonisolated public struct OBAListViewSection: Hashable, Identifiable, @unchecked
     }
 
     public static func == (lhs: OBAListViewSection, rhs: OBAListViewSection) -> Bool {
-        // `id` belongs here: `hash(into:)` already combines it, and
-        // `NSDiffableDataSource` treats Hashable mismatch as "Failed to find
-        // index of item" on the header that uses this section's id (#421).
+        // `id` belongs in `==` because `hash(into:)` already combines it —
+        // Hashable requires equal values to share a hash. Two agency-alert
+        // sections can share a title and empty contents while differing only
+        // by id (#421 / #1327). Whether that mismatch alone caused the
+        // DiffableDataSource "Failed to find index of item" spam is unproven
+        // (#1338); including `id` is still required for a correct contract.
         return lhs.id == rhs.id &&
             lhs.title == rhs.title &&
             lhs.contents == rhs.contents
--- a/OBAKit/Controls/ListView/ViewModels/OBAListViewSection.swift
+++ b/OBAKit/Controls/ListView/ViewModels/OBAListViewSection.swift
@@ -90,11 +90,10 @@ nonisolated public struct OBAListViewSection: Hashable, Identifiable, @unchecked
 
     public static func == (lhs: OBAListViewSection, rhs: OBAListViewSection) -> Bool {
         // `id` belongs in `==` because `hash(into:)` already combines it —
-        // Hashable requires equal values to share a hash. Two agency-alert
-        // sections can share a title and empty contents while differing only
-        // by id (#421 / #1327). Whether that mismatch alone caused the
-        // DiffableDataSource "Failed to find index of item" spam is unproven
-        // (#1338); including `id` is still required for a correct contract.
+        // Hashable requires equal values to share a hash. Whether omitting
+        // `id` from `==` caused the DiffableDataSource "Failed to find index
+        // of item" spam is unproven (#1338); including `id` is still required
+        // for a correct contract.
         return lhs.id == rhs.id &&
             lhs.title == rhs.title &&
             lhs.contents == rhs.contents
--- a/OBAKitTests/Controls/OBAListViewSectionTests.swift
+++ b/OBAKitTests/Controls/OBAListViewSectionTests.swift
@@ -13,9 +13,11 @@ import Testing
 /// `NSDiffableDataSource` uses `Hashable` as identity. `hash(into:)` already
 /// combines `id`; `==` used not to, so two agency-alert sections with the same
 /// title and an empty row list compared equal while hashing differently —
-/// undefined behavior, and the source of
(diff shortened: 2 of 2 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The new comment softens the causality claim (good, that's #1338's ask) but replaces it with a concrete scenario that the code contradicts: "Two agency-alert sections can share a title and empty contents while differing only by id." Agency-alert sections are built as `OBAListViewSection(id: "agency_alerts_\(group.key)", title: group.key, contents: alerts)` from a `Dictionary(grouping:by: agency name)` — the id is a pure function of the title, so two agency-alert sections sharing a title necessarily share an id, and each group has at least one alert so `contents` is never empty. That has been true since March 2021 (`d6068ed3`), i.e. before #421 was filed, so it isn't a "used to be true" description either. The Hashable-contract argument in the first sentence stands on its own; the motivating example does not (bug due to `OBAKit/Controls/ListView/Helpers/OBA/AgencyAlertListViewConverters.swift:38`).

https://github.com/OneBusAway/onebusaway-ios/blob/b21cb9e0d0a60088b22b9332a3e488327d737372/OBAKit/Controls/ListView/ViewModels/OBAListViewSection.swift#L91-L98

aaronbrethorst (changes requested):
The direction here is right and it's what #1338 asked for — the first sentence about the `Hashable` contract is accurate and is reason enough for `id` to be in `==`, independent of whether it caused the #421 spam. Softening the causality claim was the correct call.

The problem is the concrete example the new comment adds:

> Two agency-alert sections can share a title and empty contents while differing only by id (#421 / #1327).

That can't happen. `AgencyAlertListViewConverters.listSections(agencyAlerts:)` builds them as `OBAListViewSection(id: "agency_alerts_\(group.key)", title: group.key, contents: alerts)` — the id is a pure function of the title, so two agency-alert sections that share a title necessarily share an id and can't "differ only by id". And because the groups come from `Dictionary(grouping:)`, `contents` is never empty.

I checked whether it was ever true historically — `d6068ed3` (2021) and `929fc58d` (2022) both have the same id/title relationship — and whether another path builds them. It's the only construction site.

Simplest fix is to delete that sentence. The contract argument stands on its own without an example, and an accurate one is harder to find than it looks.

Same reason as the wording note: this is a comment-accuracy PR, so trading one unsupported claim for a different unsupported claim doesn't move us forward.

Two smaller things while you're in there:

The test doc comment says the link "has not been re-confirmed on device", which implies it was confirmed once. It never was — #1338 says so. The production comment's "is unproven" is the more honest phrasing.

Lines 14-15 of `OBAListViewSectionTests` still carry the same agency-alerts scenario. Those lines came from #1327 and aren't in your diff, so I'm not asking you to own them — but you're editing that exact comment block, so it's a natural thing to sweep.

If you want a real example to substitute, the service-alert path may be the one: `listSection(serviceAlerts:showSectionTitle:sectionID:)` takes `sectionID` as a parameter while deriving the title from the item count, which is the shape you'd need. I haven't verified two such sections can coexist in one list — worth a look, or just drop the example.

Borisserz:
Addressed the review feedback:

1. Removed the agency-alert “same title / empty contents / differ only by id” example from the production `==` comment — that construction path can’t produce that scenario.
2. Rephrased the test suite doc comment to say the #421 → DiffableDataSource spam link **is unproven** (matching the production wording; no “re-confirmed” implication).
3. Swept the same inaccurate agency-alerts scenario out of the `OBAListViewSectionTests` header comment while editing that block.

No replacement example added — the Hashable-contract argument stands on its own.

aaronbrethorst:
### Code review

_Re-review at e86310e, covering commits since the previous review._

Borisserz:
Follow-up `e86310eb` drops the impossible agency-alert Hashable example and uses the honest “unproven” phrasing you asked for. Ready for another look.

aaronbrethorst (approved):
Verified at e86310eb. Thanks for this.

PR #1391, opened 2026-09-05
Refresh content when a re-Track promotes a running Live Activity

### LINKED ISSUE #1390: Re-Track promotes a running Live Activity without refreshing its content
Follow-up from #1215.

Both start paths short-circuit a duplicate Track by promoting the running activity and returning — `StopPageActionPresenter.swift:598-607` and `BookmarkActions.swift:159-167`.

`promoteToDynamicIsland` only moves the score. It rebuilds content as `state: activity.content.state`, deliberately, so that a score-only touch doesn't disturb the card. So a re-Track re-ranks the activity and re-shows the confirmation toast, but the times on the card are whatever they already were.

Both call sites have fresher data in hand at that moment. Each has just built `staticData` from a current `departure` / `arrivalDepartures`, and each returns before reaching the `buildContentState` call a few lines below. The rider has explicitly asked for this trip and the app is holding current arrivals for it; nothing is done with them.

That gap matters most for a trip with no matching bookmark. `BookmarksViewController.updateRunningLiveActivities()` is the only local content refresh in the app — three references, all in that file — and it runs solely off the Bookmarks tab's `viewModel.didUpdate`. It schedules new content only when `matchingBookmark != nil`; the `else` branch registers lifecycle observers and nothing more. So an activity started from the stop page for a trip the rider hasn't bookmarked is refreshed by the server push alone, and re-Track is the one local opportunity to correct it.

Fix is to build the content state before the duplicate check and apply it with the promotion, rather than returning early. `promoteToDynamicIsland`'s score-only contract should stay as it is — the state belongs at the call site, which is the only place that knows what fresh looks like.

diveshpatil9104:
Closes #1390. Follow-up from #1215.

Both start paths short-circuit a duplicate Track by promoting the running activity and returning. `promoteToDynamicIsland` only moves the score, so the card kept whatever the last push left — while both call sites were already holding current arrivals, a few lines above the `buildContentState` they never reached.

`promoteToDynamicIsland` now takes an optional `state`: pass it to install fresh content with the score bump, omit it for the score-only behaviour every other caller still gets. Both call sites build content before the duplicate check and hand it over. On the bookmark path a nil builder result simply leaves the promotion score-only, exactly as before.

The stale marker needs opposite handling in the two cases, so it's a named rule rather than an inline ternary: score-only carries the push-set date through (dropping it leaves a stale card looking live); fresh content clears it (the old date was set for content that no longer exists). `promotionStaleDate(installing:existing:)` — tested in all four combinations.

This matters most for unbookmarked trips. `BookmarksViewController.updateRunningLiveActivities()` is the only local content refresh in the app, runs solely off the Bookmarks tab, and schedules content only when a bookmark matches — so a stop-page Track on an unbookmarked trip is refreshed by the server push alone. Re-Track was its one local chance.

**Not built.** `promoteToDynamicIsland` still can't be reached from a test — `Activity.activities` needs a real ActivityKit host — which is why the stale-date rule was extracted into something that can.

### DIFF
--- a/OBAKit/Bookmarks/BookmarkActions.swift
+++ b/OBAKit/Bookmarks/BookmarkActions.swift
@@ -156,17 +156,27 @@ final class BookmarkActions {
         // cards and duplicate OBACloud push registrations. Re-Track still needs
         // to promote the existing activity: after A→B the Island is on B with A
         // demoted to 0, so tapping Track on A again must bump A (not just toast).
+        // Built before the duplicate check so a re-Track can install it: the rider
+        // has asked for this trip while the app holds current arrivals, and the
+        // running card would otherwise keep whatever the last push left (#1390).
+        // Still optional here — a nil simply leaves the promotion score-only,
+        // which is what it always was.
+        let freshContentState = Self.buildContentState(from: arrivalDepartures)
+
         if let existing = Activity<TripAttributes>.running(matching: staticData) {
             Logger.info("Live Activity already running for stop \(staticData.stopID) route \(staticData.routeShortName); promoting instead of duplicating.")
             let existingID = existing.id
             Task {
-                await Activity<TripAttributes>.promoteToDynamicIsland(activityID: existingID)
+                await Activity<TripAttributes>.promoteToDynamicIsland(
+                    activityID: existingID,
+                    state: freshContentState
+                )
             }
             showLiveActivityStartedToast()
             return .promotedExisting
         }
 
-        guard let contentState = Self.buildContentState(from: arrivalDepartures) else {
+        guard let contentState = freshContentState else {
             // Shouldn't happen — the context menu only offers Track once arrival
             // data has loaded — but if data was cleared between the menu render
             // and the tap, tell the user rather than silently doing nothing.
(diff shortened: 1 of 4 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Nicely done. Pulling the stale-marker rule out into `promotionStaleDate(installing:existing:)` as a pure function with a test covering all four combinations is much better than leaving it as an inline ternary — that was the one genuinely subtle decision here, and it now can't quietly invert.

Carrying the push-set date through a score-only promote and clearing it when installing fresh content is the right split, and it follows `LiveActivityUpdateCoalescer`'s existing precedent rather than inventing a second rule. Defaulting `state:` to nil keeps both existing callers correct.

One tension worth being aware of rather than fixing here: on the unbookmarked stop-page trip — the case your body argues matters most — clearing `staleDate` means that if pushes stop, the card shows locally-installed times with no staleness treatment until a push re-arms it. Carrying forward a date computed for the previous content would be worse, and the coalescer already made this call, so I'm not holding on it. If it ever needs closing, the shape is a locally-derived `staleDate` on installed content, and it'd want to change the coalescer too.

Heads up on sequencing: #1394 adds a third promote call site on the trip page and currently omits `state:`. Because your parameter is defaulted that omission compiles silently, so I've asked over there to pass the `contentState` it already builds. Merging this first so that ask is concrete.

PR #1392, opened 2026-09-05
Correct the three remaining "Core Location clamps silently" comments

diveshpatil9104:
## Description

Follow-up from #1370, which Aaron asked for in review:

> three other comments still assert the old "Core Location clamps silently"
> claim this one now contradicts — `ProximityMonitoring.swift` (the
> `startedWithClampedRadius` doc, which also calls it "the only signal"),
> `LocationManagerProtocol.swift` on `maximumRegionMonitoringDistance`, and
> `ProximityAlert.swift`. Same correction, three more places.

Core Location does not clamp an oversize region. It answers with `CLError.regionMonitoringFailure`, delivered asynchronously through `monitoringDidFailFor` and carrying no radius.

| File | Was |
|---|---|
| `LocationManagerProtocol.swift:66` | "silently clamps … with no error to tell anyone" |
| `ProximityMonitoring.swift:25` | "clamps … silently, so this is **the only signal**" |
| `ProximityAlert.swift:59` | "clamps it silently, so the alert would fire at a distance nobody asked for" |

`ProximityMonitoring`'s was wrong twice over — not silent, and not the only signal, since `monitoringDidFailFor` reports it too.

The clamping itself was never the problem and none of it changes; the reason each comment gave for it was wrong. Each now says what actually happens, and why clamping up front is still right: the real error is too late and too vague to tell a rider their alert would have fired somewhere other than where they asked.

Wording follows the corrected comment #1370 landed at `LocationService.swift:596`, so all four now describe one mechanism instead of contradicting each other.

Comments only — no code, no behaviour change, reviewable on inspection. SwiftLint clean.

Deliberately no Closes — this is a follow-up from a merged PR's review, not an issue.

## Summary by CodeRabbit

- **Documentation**
  - Clarified how Core Location handles proximity regions that exceed the supported monitoring distance.
  - Documented that oversized regions fail asynchronously with a region-monitoring error rather than being silently adjusted.
  - Explained that the failure does not include the requested radius, while clamped-radius results communicate when monitoring will occur at a different distance than requested.

### DIFF
--- a/OBAKitCore/Location/Location/LocationManagerProtocol.swift
+++ b/OBAKitCore/Location/Location/LocationManagerProtocol.swift
@@ -63,9 +63,11 @@ public protocol LocationManager {
 
     /// The largest radius, in meters, this device will actually monitor.
     ///
-    /// `CLCircularRegion` silently clamps an oversize radius, so a region built
-    /// past this limit fires at a different distance than the one requested with
-    /// no error to tell anyone. Reading the limit is what makes that detectable.
+    /// An oversize region is not silently clamped: Core Location answers it with
+    /// `CLError.regionMonitoringFailure`, delivered asynchronously through
+    /// `monitoringDidFailFor` and carrying no radius. Reading the limit up front
+    /// is what lets a caller clamp deliberately and report it, rather than learn
+    /// later that something failed without learning what.
     var maximumRegionMonitoringDistance: CLLocationDistance { get }
 }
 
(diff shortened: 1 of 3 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Verified all three rewrites against Apple's documented behavior for `maximumRegionMonitoringDistance` — an oversize region really does come back as `CLError.regionMonitoringFailure` through `monitoringDidFailFor` rather than being silently clamped, so the corrected comments are accurate and now match the wording already merged in `LocationService`.

Every changed line is a doc comment; no behavior moves. Merging.

PR #1394, opened 2026-09-05
Make the trip page promote on re-Track and share the content mapping

### LINKED ISSUE #1393: Trip page's Track path still diverges from the other two
Follow-ups from #1377, which routed the three Live Activity start paths through `Activity.requestProminent`.

**1. The trip page's duplicate guard never promotes.**

`TripPageViewController.swift:439-444` logs and returns:

```swift
    Logger.info("Live Activity already running ...; not starting a duplicate.")
    isTrackingLiveActivity = true
    render()
    return
}
```

`StopPageActionPresenter` and `BookmarkActions` both call `promoteToDynamicIsland` at the same point, for the reason their comments give: after tracking A then B, the Island is on B with A demoted to `demotedScore`, so tapping Track on A again has to bump A back. The trip page doesn't, so re-Tracking an already-live trip there sets `isTrackingLiveActivity`, re-renders, and leaves the Island on whatever else is running.

That is the #1189 Problem 2 behaviour, on the one start path that never received the fix — the same way this path was also the one missing `relevanceScore` and `demoteLivePeers` until #1377.

**2. It still hand-rolls its `ContentState`.**

`TripPageViewController.swift:446-453` builds `TripAttributes.ContentState` inline; the other two paths share `BookmarkActions.buildContentState`. That's the third construction site for one shape.

Worth being precise about the ceiling here: the trip page holds a single `departure` and no stop arrivals list, so routing it through the shared builder is consolidation rather than richer content — the card would still carry one arrival, because there is no second one to offer. The value is that a future change to the builder reaches all three paths instead of two.

diveshpatil9104:
Closes #1393. Follow-up from #1377.

**1. Re-Track now promotes.** `TripPageViewController` was the one start path whose duplicate guard logged and returned. The other two call `promoteToDynamicIsland` there, for the reason their comments give: after tracking A then B the Island sits on B with A demoted to `demotedScore`, so tapping Track on A again has to bump A back. Here it only set `isTrackingLiveActivity` and re-rendered, leaving the Island where it was.

Same path that was missing `relevanceScore` and `demoteLivePeers` until #1377 — this was the remaining half.

**2. The content mapping is shared**, rather than a third hand-rolled copy.

It calls `BookmarkActions.contentState(from:)`, not `buildContentState(from:matching:)`, which needed that method to become internal. The distinction matters: `buildContentState` warns that a missing headsign can let the filter readmit the opposite direction. This screen holds one `ArrivalDeparture` and no stop list, so there is no filtering and no direction to mix — that warning would assert something permanently untrue. Same false-claim shape flagged on #1354 when a tripID key made the warning unreachable.

**Item 2 is a no-op for riders.** Still one arrival — there is no second one on this screen. The mapping is field-for-field what was there, and `prefix(3)` over one element is one element. A test pins that equivalence. What changes is that a future edit to the mapping reaches all three paths instead of two.

## Sequencing

Built against `main`, so it calls `promoteToDynamicIsland(activityID:)`. Open PR #1391 adds an optional `state:` there, with a default — so this compiles either way, but once #1391 lands this call site should pass the `contentState` built directly above it. Happy to fold that in on rebase.

#1381 also touches this file, though not this function.

## Summary by CodeRabbit

- **New Features**
  - Existing trip Live Activities can now be promoted to the Dynamic Island when started again for the same trip.
  - Trip Live Activity information is now presented consistently across trip and bookmark entry points.

- **Tests**
  - Added coverage verifying that departure details, including timing, schedule deviation, and arrival status, are accurately reflected in Live Activity updates.

### DIFF
--- a/OBAKit/Bookmarks/BookmarkActions.swift
+++ b/OBAKit/Bookmarks/BookmarkActions.swift
@@ -121,11 +121,17 @@ final class BookmarkActions {
 
     /// Maps the soonest three arrivals into the Live Activity payload.
     ///
-    /// - Precondition: `arrivalDepartures` is non-empty. Both entry points above
-    ///   establish that — one by guarding, the other by falling back to the
-    ///   tracked departure — which is why only the guarding one returns an
-    ///   `Optional`.
-    private static func contentState(from arrivalDepartures: [ArrivalDeparture]) -> TripAttributes.ContentState {
+    /// Internal rather than private because the trip page starts from a single
+    /// `ArrivalDeparture` with no stop list to filter (#1393). It wants this
+    /// mapping and not `buildContentState(from:matching:)`, whose headsign warning
+    /// describes a direction-mixing risk that a one-element list cannot have —
+    /// logging it there would assert something permanently untrue.
+    ///
+    /// - Precondition: `arrivalDepartures` is non-empty. Every entry point
+    ///   establishes that — by guarding, by falling back to the tracked
+    ///   departure, or by passing that departure directly — which is why only the
+    ///   guarding one returns an `Optional`.
+    static func contentState(from arrivalDepartures: [ArrivalDeparture]) -> TripAttributes.ContentState {
         let arrivals = arrivalDepartures.prefix(3).map { arrDep in
             TripAttributes.ContentState.ArrivalInfo(
                 departureTime: Int(arrDep.arrivalDepartureDate.timeIntervalSince1970),
--- a/OBAKit/Bookmarks/BookmarkActions.swift
+++ b/OBAKit/Bookmarks/BookmarkActions.swift
@@ -132,6 +132,12 @@ final class BookmarkActions {
     ///   departure, or by passing that departure directly — which is why only the
     ///   guarding one returns an `Optional`.
     static func contentState(from arrivalDepartures: [ArrivalDeparture]) -> TripAttributes.ContentState {
+        // Documented above, and unenforced until this became internal: the
+        // guarding wrapper used to be the only way in. An empty array yields a
+        // card with no arrivals rather than being refused, so say so in debug
+        // rather than shipping a blank Live Activity.
+        assert(!arrivalDepartures.isEmpty, "contentState(from:) needs at least one arrival")
+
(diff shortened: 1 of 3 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (changes requested):
This is good work and it does exactly what #1393 and my #1377 note asked for. I checked the extraction field by field against the block you removed — `departureTime`, `scheduleStatus`, `scheduleDeviation`, `isArrival` are identical expressions with only the binding renamed, `ArrivalInfo` has no fifth field to lose, and `contentState(from:)`'s body is untouched so the two existing callers can't have shifted. Hoisting `existing.id` out before the `Task` is the right shape too.

One line to add before this merges.

I've just landed #1391, which gives `promoteToDynamicIsland` an optional `state:` for installing fresh arrivals alongside the score bump. Your promote branch calls `promoteToDynamicIsland(activityID: existingID)` without it — while `contentState` sits computed and unused three lines above.

Because `state:` is defaulted to nil that omission compiles silently, which is exactly why I'd rather fix it now than file it: the trip page would be the one promote path of three that still shows whatever the last push left, and nothing would fail to tell us.

Please rebase on `main` and pass `state: contentState`. You already offered to fold this in, which I appreciate — it should be a one-liner now that #1391 is in.

Two smaller things:

The new test's docstring says it pins the mapping "field for field" but doesn't assert `scheduleStatus`. `ArrivalInfo` is `Hashable`, so a single `#expect(arrival == expected)` would pin all four and make the docstring true.

Widening `contentState(from:)` to internal exposes its documented-but-unenforced non-empty precondition — an internal caller passing `[]` now gets `ContentState(arrivals: [])` instead of being stopped by the guarding wrapper. Your caller always passes `[departure]`, so nothing's broken; just noting the edge got sharper.

Ping me when it's rebased and I'll merge it.

aaronbrethorst:
### Code review

_Re-review at 3aa284a, covering commits since the previous review._

aaronbrethorst (approved):
Verified at 3aa284ab. Thanks for this.

PR #1396, opened 2026-09-05
Drop the inert autoshrink on the weather button

diveshpatil9104:
Follow-up from #1357:

> `adjustsFontSizeToFitWidth` may well be inert here. Autoshrink only applies to
> single-line labels, and this title is demonstrably multi-line. So `100°` may
> still wrap rather than scale.

Right, and #1344's own before-screenshot proves it: `18°` rendered as "1"/"8"/"°". A one-line label cannot wrap, so autoshrink was doing nothing.

Two checks: `numberOfLines` is set nowhere in this file, and the repo's one working autoshrink — `VehicleCalloutView.swift:176` — is a bare `UILabel()`, which defaults to one line. A `UIButton.Configuration` title doesn't.

Removed, with a comment saying why it isn't there. **The inset fix is untouched** — that's what actually bought the room, and what the #1357 screenshots show working.

Not doing the other half: `numberOfLines = 1` would switch autoshrink on and give `100°` scaling instead of wrapping, but it trades wrapping for truncation when 0.7 still doesn't fit — a visible change wanting a device and a before/after. This PR changes no rendering, only removes dead code. Happy to follow up.

The "inset fix is untouched" line is the one I'd keep no matter how short it gets — it's what stops a reviewer thinking you're reverting the thing they merged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the weather toolbar button’s title layout to preserve corrected horizontal spacing and wrapping for narrow temperature displays.

### DIFF
--- a/OBAKit/Mapping/MapViewController.swift
+++ b/OBAKit/Mapping/MapViewController.swift
@@ -455,12 +455,16 @@ class MapViewController: UIViewController,
         }
 
         let button = UIButton(configuration: config)
-        // Restores what `7f2c1b8b` intended before the Configuration switch
-        // dropped it, with the scale factor that version omitted: without a
-        // `minimumScaleFactor`, `adjustsFontSizeToFitWidth` never shrinks
-        // anything. Three-digit temperatures scale down rather than wrap.
-        button.titleLabel?.adjustsFontSizeToFitWidth = true
-        button.titleLabel?.minimumScaleFactor = 0.7
+        // No `adjustsFontSizeToFitWidth` here, deliberately. Autoshrink only acts
+        // on a label pinned to one line, and a `UIButton.Configuration` title is
+        // not — which is exactly how `18°` came to render as "1"/"8"/"°" in the
+        // screenshot on #1344. Setting it (as this did until #1357) shrinks
+        // nothing and reads as though the width problem were handled.
+        //
+        // Making it real needs `titleLabel?.numberOfLines = 1` alongside it, which
+        // trades wrapping for truncation when the scale floor still doesn't fit.
+        // That is a visible change to a 42pt toolbar button and wants a device.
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
### Code review

aaronbrethorst (changes requested):
The change itself is right, and you proved the point I only guessed at on #1357 — I checked Apple's docs and `adjustsFontSizeToFitWidth` really is single-line-only, the button's title label is never pinned to one line, and the toolbar constrains width but not height. So the property genuinely was inert. Good catch, and the "why it's absent" comment is more useful than the code it replaces.

One thing to fix before this goes in, and it's one word.

The new comment says the autoshrink "shrinks nothing and reads as though the width problem were handled — **as this did until #1357**". That's backwards. I traced it: `7f2c1b8b` set it in 2019, **#1318** (`6f5dbba2`, the `UIButton.Configuration` switch) dropped it, and **#1357** (`0aa4822f`) is what added it back — which is why the comment you're deleting says "Restores what `7f2c1b8b` intended". So the file has set it *since* #1357, not *until* it.

`since #1357` fixes it, or cite #1318 if you want to name the PR that actually removed it.

I'm holding on a single word only because this is a comment-correction PR — the comment is the whole deliverable, so a new wrong claim about the history defeats the purpose. Everything else here is good; fix that and I'll merge it.

Minor, take it or leave it: "which is exactly how `18°` came to render as 1/8/°" attributes the overflow to multi-line-ness, but the actual cause was `.plain()`'s horizontal insets against the 42pt width. The block two above already says this correctly, so a reader gets there either way.

aaronbrethorst:
### Code review

_Re-review at 2519af4, covering commits since the previous review._

aaronbrethorst (approved):
Verified at 2519af43. Thanks for this.

PR #1397, opened 2026-09-05
  Track the one-time Always prompt for proximity alerts

diveshpatil9104:
## Track the one-time Always prompt for proximity alerts

Makes background location obtainable for proximity alerts, and works out what a rider still has to grant before one can arm.

Part of [#455](https://github.com/OneBusAway/onebusaway-ios/issues/455). No UI — the stop-page entry point is the next PR.

Today nothing calls `requestAlwaysAuthorization()`, so `createProximityAlert` always returns `.needsLocationAuthorization`. Calling it isn't enough on its own: iOS raises that prompt once per install and silently does nothing afterwards, and `.authorizedWhenInUse` looks identical either side of it. Without a record, the UI would ask forever and never learn to send the rider to Settings.

### What changed

**`LocationService` (`OBAKitCore`)**

`canPromptForAlwaysAuthorization`, persisted, mirroring the existing `canPromptUserForPermission`. `requestAlwaysAuthorization()` spends it itself rather than trusting callers — the sibling flag leaves that to its callers and one of the two call sites already forgets. It refuses, and logs, from statuses iOS would ignore, so a rider who re-enables location in Settings keeps their upgrade path.

**New file — `ProximityAlertAuthorization.swift` (`OBAKit`)**

`ProximityAlertAuthorizationStep.resolve(locationStatus:canPromptForAlways:notificationStatus:)`, a pure function returning one next step: `ready`, `promptForLocation`, `promptForNotifications`, `locationBlocked(status)`, `notificationsBlocked`. Location resolves before notifications deliberately — iOS gives the notification prompt one appearance per install, and spending it on a feature that cannot work, before anything has explained what the notification is for, wastes it.

**`ProximityAlertManager`**

`authorizationStep()` wires the live statuses to the resolver, so the UI asks one question instead of re-deriving it from a raw `CLAuthorizationStatus` — which cannot answer it.

---

> **PR Roadmap for [#455](https://github.com/OneBusAway/onebusaway-ios/issues/455)**
>
> | # | PR | Status |
> |---|-----|--------|
> | 1 | `ProximityAlert` model + `UserDataStore` persistence | `Merged` |
> | 2 | `LocationService` geofencing + hardening | `Merged` |
> | 3 | `ProximityAlertManager` + local notifications | `Merged` |
> | 4a | **Always-authorization tracking + what's-still-needed** | `Current` |
> | 4b | Stop page UI to create and cancel alerts | `Next` |

## Summary by CodeRabbit

- **New Features**
  - Added a unified authorization status for proximity alerts, showing whether location and notification access is ready, requires action, or is blocked.
  - Added an authorization check that identifies the next permission step, prioritizing location before notifications.
  - Added safeguards to ensure the one-time “Always Allow” location prompt is requested only when available.

- **Tests**
  - Added coverage for authorization states, permission ordering, blocked access, and one-time location prompt behavior.

### DIFF
--- a/OBAKit/ProximityAlerts/ProximityAlertAuthorization.swift
+++ b/OBAKit/ProximityAlerts/ProximityAlertAuthorization.swift
@@ -1,0 +1,107 @@
+//
+//  ProximityAlertAuthorization.swift
+//  OBAKit
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import CoreLocation
+import Foundation
+import UserNotifications
+import OBAKitCore
+
+/// The next thing that has to happen before a proximity alert can be armed.
+///
+/// A proximity alert needs two permissions that the rest of the app never asks
+/// for: `.authorizedAlways` location, because a geofence started under When In
+/// Use only delivers while the rider is looking at the screen, and notification
+/// permission, because the notification is the entire product. Each is
+/// separately grantable, separately refusable, and — once refused — reachable
+/// only through Settings.
+///
+/// Reducing all of that to a single next step keeps the decision in one place.
+/// The alternative is every caller re-deriving "can I still ask, or do I send
+/// them to Settings?" from a raw `CLAuthorizationStatus`, which is exactly the
+/// question the status alone cannot answer.
+public enum ProximityAlertAuthorizationStep: Equatable {
+
+    /// Nothing is missing.
+    case ready
+
+    /// The app can still raise the system location prompt. Whether that prompt
+    /// is the first one or the Always upgrade is iOS's business, not the
+    /// caller's — both are `requestAlwaysAuthorization()`.
+    case promptForLocation
+
+    /// Location is settled and the app can still raise the notification prompt.
+    case promptForNotifications
+
+    /// Location cannot be obtained by asking, so only Settings can fix it. The
+    /// status is carried because the rider's situation differs sharply between
+    /// them: `.denied` is a choice they can reverse, `.restricted` may be a
+    /// device policy they cannot, and `.authorizedWhenInUse` means they granted
+    /// something — just not the background access this feature needs.
+    case locationBlocked(CLAuthorizationStatus)
+
+    /// Notifications cannot be obtained by asking. Carries no status: every
+    /// route here wants the same thing from the rider, which is a trip to
+    /// Settings, and the resolver logs the distinction for anyone reading a
+    /// device log.
+    case notificationsBlocked
+}
+
+extension ProximityAlertAuthorizationStep {
+
+    /// Works out the next step from the two authorization states.
+    ///
+    /// Location is settled before notifications are asked about, and the order
+    /// matters rather than being arbitrary: a rider who will not grant
+    /// background location cannot use this feature at all, and iOS gives the
+    /// notification prompt one appearance per install. Asking for notifications
+    /// first would spend that one appearance on a feature already dead — and
+    /// spend it in a context the rider cannot make sense of, since nothing has
+    /// yet told them what the notification would be for.
+    ///
+    /// - Parameters:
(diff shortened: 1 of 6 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The new `canPromptForAlwaysAuthorization` property is inserted directly between `requestAlwaysAuthorization()`'s existing doc comment and the method itself, with no blank line, so the two `///` blocks merge. The whole original block — "Prompts the user to upgrade to Always authorization, which region monitoring requires…" plus the `- Important:` note about `NSLocationAlwaysAndWhenInUseUsageDescription` being required in each white-label app's Info.plist — now documents a read-only `Bool`, and `requestAlwaysAuthorization()` is left with only the new, shorter comment that drops the Info.plist requirement entirely. That note is load-bearing for the white-label story (`Apps/Shared/app_shared.yml` / KiedyBus) and it is also the one case the new comment gets wrong: "Returns without asking — and without spending anything — from any status iOS would ignore the request from" is not true when iOS ignores the call for a missing usage-description key, since the flag is written before `locationManager.requestAlwaysAuthorization()` is called. A blank line before line 274 restores the split; the Info.plist `- Important:` should stay on the method. (bug due to OBAKitCore/Location/Location/LocationService.swift lines 269–286, where the `- Important:` block runs straight into "Whether the one-time Always upgrade prompt…" and terminates at `public var canPromptForAlwaysAuthorization`)

https://github.com/OneBusAway/onebusaway-ios/blob/1034ea8c3a052817b8a32a1af616f54003d26a7d/OBAKitCore/Location/Location/LocationService.swift#L268-L286

aaronbrethorst (changes requested):
The state machine is right, and I checked it against Apple's docs case by case rather than by eye: `.notDetermined` → "Allow While Using App" giving Provisional Always resolves `.ready`; "Keep Only While Using" leaves `.authorizedWhenInUse` with the flag spent and correctly resolves `.locationBlocked`; `.denied`/`.restricted` never spend the flag, so a rider who re-enables in Settings keeps their upgrade path. The notification arm matches `allowsNotificationDelivery` exactly. Persisting the flag rather than holding it in memory is the right call since the limit is per install. The UUID-scoped `UserDefaults` suites in the tests are a nice touch.

One thing to fix, and it's a blank line.

`canPromptForAlwaysAuthorization` is inserted directly between `requestAlwaysAuthorization()`'s existing doc block and its own doc comment, with nothing separating them. I read the file at your head SHA to be sure: the `- Important:` note about `NSLocationAlwaysAndWhenInUseUsageDescription` now runs contiguously into the property's comment, so DocC attaches the white-label Info.plist warning to a `Bool`. And the method's replacement comment drops that requirement entirely.

That note matters more here than it would in most codebases — it's the one telling a new white-label app why it will never reach `.authorizedAlways`. We generate DocC from these via `scripts/docs`, so the misattribution ships.

Add the blank line, and keep the `- Important:` block on the method.

While you're there: the new comment says the method spends nothing "from any status iOS would ignore the request from", but the missing-usage-key case is exactly such a status and the flag *is* written before the call. Worth softening, since it's the case the deleted note warned about.

Two things I'd like you to think about but am not asking you to solve:

If a rider answered onboarding's `requestWhenInUseAuthorization()` with "Allow Once", Apple documents that further `requestAlwaysAuthorization()` calls are ignored. The status still reads `.authorizedWhenInUse`, so the guard passes and the flag burns on a prompt iOS never raises. `CLAuthorizationStatus` can't distinguish temporary from permanent, so I don't think you can detect it — the outcome degrades to `.locationBlocked` and the Settings path, which is recoverable. Just worth a comment acknowledging it.

Also: CodeRabbit's migration finding is wrong. `requestAlwaysAuthorization()` has no production call sites on `main`, so no existing install has spent the prompt and defaulting to `true` is correct. No migration needed.

aaronbrethorst:
### Code review

_Re-review at a572f7e, covering commits since the previous review._

aaronbrethorst (approved):
Verified at a572f7ea. Thanks for this.

PR #1398, opened 2026-09-06
Add voice search mic to map search (#1129)

### LINKED ISSUE #1129: Add Optional Voice Search (Mic Button) to Map Search Field
Would it make sense to add an optional mic button in the map search field for voice search? It could speed up stop/route lookup and help accessibility, while keeping keyboard search as the default fallback

Image

Borisserz:
## Summary
- Optional mic in the SwiftUI search sheet capsule (Google Maps pattern): live transcript fills the field; final utterance auto-runs a classified `SearchRequest` (stop / route / vehicle / address).
- `SFSpeechRecognizer` + mic permissions via shared Info.plist purpose strings (localized); mic hides when speech is unavailable/denied. Keyboard remains the fallback.
- Protocol-backed controller + unit tests for classification and view-model wiring.

Closes #1129

## Test plan
- [ ] Open map search → mic appears in the capsule
- [ ] Tap mic → grant speech/mic → speak a place name → field updates live → search runs on final
- [ ] Say “route 40” / a stop number → correct search type
- [ ] Deny permissions → mic hidden (or error) and keyboard still works
- [ ] Clear / Close cancels listening
- [ ] `VoiceSearchQueryClassifierTests` + `SearchSheetViewModelTests` green

## Summary by CodeRabbit

* **New Features**
  * Added voice search with live transcription for stops, routes, vehicles, and places.
  * Added listening controls, visual feedback, accessibility labels, and clear error messages.
* **Bug Fixes**
  * Improved cancellation and microphone cleanup when listening stops, text is edited, or search closes.
* **Localization**
  * Added voice-search labels, errors, and permission descriptions across supported languages.
* **Documentation**
  * Documented voice-search behavior, permissions, accessibility, cancellation, and speech-processing privacy.


aaronbrethorst:
### Code review

aaronbrethorst (changes requested):
First, the part that usually goes wrong on a PR like this and didn't: the privacy and localization work is genuinely complete. `NSSpeechRecognitionUsageDescription` and `NSMicrophoneUsageDescription` are in `Apps/Shared/info_plist.yml`, which both OneBusAway and KiedyBus include, with localized `InfoPlist.strings` for every `.lproj` each app actually has. All four new UI keys are in all 13 catalogs with fallbacks matching `en`. That's the App Store rejection and the launch crash both avoided, and it's the part most easily missed.

Two things I want fixed before this lands. Note the automated comment above says "no issues found" — that's a confidence threshold, and I'm overriding it; these are real.

**1. Recognition can start after `stop()`, and then nobody owns the microphone.**

`start()` does `Task { @MainActor in await self.beginRecognition() }` and keeps no handle on that task. `beginRecognition()` then suspends twice, on `requestSpeechAuthorization()` and `requestMicrophoneAuthorization()`. If `stop()` runs while it's suspended there, nothing cancels it: the continuation finishes, `tearDownEngine()` runs against an engine that hasn't started, and then the task resumes, activates the audio session, installs the tap and calls `audioEngine.start()`. The task captures `self` strongly, so even after the view model goes away there's a live recognizer with a hot mic and no continuation to deliver anything.

Store the task and cancel it in `stop()`, and check `Task.isCancelled` after each authorization await before touching the session. `SearchSheetViewModel.startVoiceSearch` wants the same guard — right now a `stopVoiceSearch()` issued before its task first runs still starts the mic.

You already have this pattern in the file: `72ec2a29` added the `pendingPresentation` cancellation in `close()` for the same reason.

**2. `isVoiceSearchAvailable` never checks microphone permission.**

`isAvailable` looks only at `SFSpeechRecognizer.authorizationStatus()`, never `AVAudioApplication.shared.recordPermission`. Grant speech, deny the mic, and you keep a visible mic button that raises the same error alert on every tap forever — which contradicts your own design doc ("If speech is unavailable or permanently denied, hide the mic").

Worth a decision rather than a change:

`request.requiresOnDeviceRecognition = false` is set explicitly, so audio goes to Apple's servers even on devices that support on-device recognition. That may be the right trade for accuracy, but it's a deliberate privacy posture for a transit nonprofit and neither the purpose strings nor the design doc mention it. I'd rather we chose it consciously and said so.

Smaller, take or leave:

- `.final` calls `performSearchAndWait(request:)` directly, bypassing `searchTask` cancellation, so an in-flight keyboard search can clear the HUD early and present stale results over the voice result.
- `updateQuery` only stops the mic when the text is empty, so typing mid-listen lets the next partial overwrite what was typed.
- `#expect(voice.stopCount >= 1)` is already satisfied by `startVoiceSearch()`'s own leading `stopVoiceSearch()`. The `isListening == false` assertion beside it does carry weight.
- `if #available(iOS 13, *)` is dead at an iOS 18 deployment target.
- `strippingCue` lowercases the remainder, so "Route Rapid Ride D" searches `rapid ride d` while the address path preserves case.

The architecture is sound — protocol seam with a mock, `AsyncStream` events, backgrounding self-heals because there's no `UIBackgroundModes: audio`. Fix the two above and I'll take it.

Borisserz:
Addressed the CHANGES_REQUESTED items from the review:

1. **Startup race** — `SpeechVoiceSearchController` retains `beginRecognitionTask`, cancels it in `stop()`, and checks `Task.isCancelled` after each authorization await (and before touching the audio session). `startVoiceSearch` also returns immediately if the task was cancelled before first run.
2. **Mic availability** — `isAvailable` now also checks `AVAudioApplication.shared.recordPermission` and hides the button when permanently denied.
3. **Smaller fixes** — `.final` goes through `performSearch` (cancels in-flight keyboard `searchTask`); typing mid-listen stops the mic; removed dead `if #available(iOS 13, *)`; `strippingCue` preserves remainder case (and strips leading punctuation); design doc notes cloud recognition (`requiresOnDeviceRecognition = false`) as a conscious accuracy/privacy trade-off.

Tests added/updated for cancel-before-start, typing-while-listening, availability via mock, and case/punctuation in the classifier. `SearchSheetViewModelTests` + `VoiceSearchQueryClassifierTests` are green.

aaronbrethorst:
### Code review

_Re-review at b1c2b44, covering commits since the previous review._

Found 2 issues:

1. Two callbacks in `SpeechVoiceSearchController` will crash the app at runtime on a real device (bug due to Swift 6 isolation inference, not caught by the compiler). The class is `@MainActor`. The closure passed to `installTap` has a non-Sendable block type (`AVAudioNodeTapBlock` is `NS_SWIFT_NONSENDABLE`), and so does the handler for `SFSpeechRecognizer.requestAuthorization`. In Swift 6 mode the compiler therefore treats both closures as main-actor isolated and inserts a runtime executor check (`swift_task_isCurrentExecutor` → `swift_task_reportUnexpectedExecutor`) at the top of each; I confirmed that by emitting SIL for the same code with `-swift-version 6 -default-isolation MainActor`. AVAudioEngine calls tap blocks on its audio I/O thread, and the Speech header says the authorization handler is not guaranteed to run on the main queue, so the first tap on the mic will crash, either on the authorization callback or as soon as audio starts flowing. `AVAudioApplication.requestRecordPermission` is fine because its handler is imported as `@Sendable`. CI stays green because the tests use `MockVoiceSearchController`. The fix is to build both closures in `nonisolated` helpers, or mark them `@Sendable`.

https://github.com/OneBusAway/onebusaway-ios/blob/b1c2b4401b32dce52b999923272c54824269197f/OBAKit/Search/SpeechVoiceSearchController.swift#L123-L129

https://github.com/OneBusAway/onebusaway-ios/blob/b1c2b4401b32dce52b999923272c54824269197f/OBAKit/Search/SpeechVoiceSearchController.swift#L200-L207

2. The mic keeps running when the user leaves search any way other than Close or editing the text (bug due to `startVoiceSearch` only being stopped from `close()` / `updateQuery`). Partial transcripts refresh the suggestion rows while the user is still speaking. Tapping one of those rows (a recent stop or map item, via `leaveSearchAndPresent`, or a quick-search row, via `performSearch` → `present(.single)`) calls `coordinator.pop()` but never `stopVoiceSearch()`. The voice task's `guard let self` keeps the view model alive across the `for await`, so `deinit` never runs. The `.record` audio session stays active and the mic stays hot after the sheet is gone. When a `.final` arrives later, `performSearch` runs from the dismissed search and pops or pushes sheets on top of the one the user just opened. Stop voice search on every path that leaves search.

https://github.com/OneBusAway/onebusaway-ios/blob/b1c2b4401b32dce52b999923272c54824269197f/OBAKit/Sheet/Content/Search/SearchSheetViewModel.swift#L151-L179

Borisserz:
Audio tap and speech-authorization callbacks are now formed in `nonisolated` helpers, so they are not main-actor isolated when AVAudioEngine and Speech call them off the main queue.

Voice search now stops on every path that leaves search: `performSearch`, disambiguation, the single-result pop, and `leaveSearchAndPresent` (before the resolve await). A later `.final` cannot search a sheet the rider already left.

`SearchSheetViewModelTests` covers the stop-tap and quick-search leave paths. `SpeechVoiceSearchIsolationTests` calls the same tap and authorization helpers off the main actor.

aaronbrethorst:
### Code review

_Re-review at 1c10daf, covering commits since the previous review._

aaronbrethorst:
Both things I asked for last round are genuinely fixed, and fixed at the right level rather than patched over.

The isolation work is the right shape: pulling the audio tap block and the Speech authorization handler into `nonisolated static` helpers, and then having `SpeechVoiceSearchIsolationTests` invoke exactly those helpers from `Task.detached`, means the test actually exercises the off-main path instead of asserting something adjacent to it.

On the mic lifecycle — I walked `SearchDelegate`'s full surface plus `close()`, `present(.disambiguate)`, `present(.single)` and `leaveSearchAndPresent`. All four leave paths call `stopVoiceSearch()`, and the two paths that *don't* leave search correctly don't. That's the complete enumeration, which is what this needed.

The privacy and localization surface is also complete: both Info.plist purpose strings are in `Apps/Shared/info_plist.yml`, and all four new user-facing strings are present in every locale with real translations. On a PR this size that's usually where something slips, and it didn't.

Two things before this can land, neither of them a rewrite:

**It needs a rebase.** Sibling PRs have landed on `main` since you pushed and this branch now conflicts. No re-review needed once it's resolved, unless resolving changes behavior.

**One question I'd like you to answer rather than guess at.** The `recognitionTask(with:resultHandler:)` closure is formed inside `beginAudioCapture`, which is `@MainActor`, and neither `recognitionTaskWithRequest:resultHandler:` nor `requestAuthorization:` is marked `NS_SWIFT_SENDABLE` in the SDK — so that closure inherits `@MainActor` the same way the two you just fixed did. What makes me uncertain is that Apple's own SpokenWord sample assigns `textView.text` directly inside this handler, which suggests it really is delivered on the main queue.

But the code is arguing with itself: you wrapped the body in `Task { @MainActor in … }`, which is only needed if the handler can arrive off-main — and if it can, the executor precondition trips before that `Task` is ever created. So either the wrapper is dead weight or there's a third instance of the bug you just fixed twice. Worth one line of explanation either way.

A few smaller notes for the same pass, none blocking:

- `stopVoiceSearch()` on `performSearch` puts `AVAudioSession.setActive(false, options: .notifyOthersOnDeactivation)` on the hot path of *every* search, including for riders who never touch the mic — `tearDownEngine()` calls it unconditionally. An `audioEngine.isRunning` guard around the session teardown keeps that synchronous IPC off the common path.
- A spoken vehicle query isn't gated on `isVehicleSearchAvailable` the way the typed path is, so with Obaco down the rider gets `APIError.noRegionSelected`'s text, which doesn't describe what happened.
- `tearDownEngine()` calls `recognitionRequest?.endAudio()` before `audioEngine.stop()`; between those the tap can still fire and append after `endAudio()`. Stop the engine and remove the tap first.

Also worth flagging: no build or test job has ever run on this head — checks are CodeRabbit and the CLA only. So none of this has been compiled by CI yet. The rebase will fix that as a side effect.

Borisserz:
Ready for re-review.

Your two blockers from 09-06 are already on the branch (`beginRecognitionTask` cancelled in `stop()` with `Task.isCancelled` after each auth await; `isAvailable` gates on `AVAudioApplication.shared.recordPermission`; `startVoiceSearch` cancels on leave; finals go through `performSearch`).

Also added the English `"line"` route cue with a regression test (`line 40` → route / `40`) — VoiceSearchQueryClassifierTests 11/11 green locally.

Borisserz:

Borisserz:
Rebased onto current `main` (resolved Localizable.strings conflicts by keeping main’s catalog and re-adding the four voice-search keys). Ready for another look — prior review items (cancel on leave, record-permission gate, English `line` cue) are still on the tip.

aaronbrethorst:
### Code review

_Re-review at 1639d98, covering commits since the previous review._

aaronbrethorst (approved):
Verified at 1639d98b. Thanks for this.

PR #1399, opened 2026-09-06
Fix trip page Live Activity tracking parity & add ToastManager unit tests

### LINKED ISSUE #1393: Trip page's Track path still diverges from the other two

ravindra-RKB:
This PR bundles two improvements to the codebase:

### 1. Trip Page Live Activity Parity (Closes #1393)
* **Re-Track Promotion:** Follow-up to #1377. The trip page now correctly promotes an existing Live Activity to the Dynamic Island when a rider re-Tracks an already running trip, rather than silently ignoring it.
* **Shared Mapping:** Centralizes the `ContentState` mapping by utilizing `BookmarkActions.contentState(from:)` rather than relying on a third inline hand-rolled copy.
* *Note: This incorporates the review feedback requested by Aaron in PR #1394 (passing the computed `state: contentState` to `promoteToDynamicIsland` and asserting full struct equality in `BookmarkActionsTests`).*

### 2. Missing Test Coverage: `ToastManager`
* Added comprehensive Swift Testing coverage (`ToastManagerTests.swift`) for the previously untested `ToastMessageBar` component.
* Validates `showSuccess`, `showError`, and `dismiss` behaviors, ensuring `toast` properties and `isShowing` UI states mutate correctly with their specific parameters.

## Summary by CodeRabbit

* **New Features**
  * Restarting live trip tracking now promotes an existing activity to the Dynamic Island and refreshes its content.
  * Trip updates consistently display departure details, schedule status, deviations, and arrival information.

* **Bug Fixes**
  * Improved live activity updates for trips with a single departure.

* **Tests**
  * Added coverage for trip content mapping and toast success, error, visibility, and dismissal behavior.


diveshpatil9104:
@ravindra-RKB, please sign the Contributor License Agreement.

aaronbrethorst:
### Code review

Found 1 issue:

1. The Live Activity half of this PR duplicates #1394, another contributor's open PR for the same issue (#1393). The first commit here, c9aef91, has the same headline and body as #1394's f8d3ec8c ("Make the trip page promote on re-Track and share the content mapping" / "It was the one start path that logged and returned instead of bumping the Island."). The `TripPageViewController`, `BookmarkActions`, and test hunks match #1394 line for line. #1394's head (3aa284a) already has the follow-up this PR adds (`state: contentState`, whole-struct `ArrivalInfo` equality), plus an `assert` on the non-empty precondition that this copy lacks. Merging either PR leaves the other conflicting, and merging this one drops the original author's credit. The `ToastManager` tests are unrelated to that fix and could stand alone as their own PR.

https://github.com/OneBusAway/onebusaway-ios/blob/f0e0ecb5451289b5cc9a8b1a140e2bfcbc91f2b5/OBAKit/Trip/TripPage/TripPageViewController.swift#L435-L462

aaronbrethorst (changes requested):
The `ToastManager` tests here are good work — they're new coverage, nothing else tests that type, and the async cleanup case you added after the bot feedback is the one that would actually catch a regression. I want those.

The problem is the Live Activity half. It's the same change as #1394, by another contributor, for the same issue (#1393). Your first commit `c9aef91` carries the identical headline to #1394's `f8d3ec8c` — "Make the trip page promote on re-Track and share the content mapping" — and the two PRs touch the same three files (`TripPageViewController.swift`, `BookmarkActions.swift`, `BookmarkActionsTests.swift`) with matching hunks. #1394 was opened first.

Whichever of these lands, the other one conflicts, and the duplicated work means somebody's effort gets thrown away. That isn't something I want to resolve by merge order.

What I'd like: drop the Live Activity commits from this PR and let it be the `ToastManager` tests alone. Those stand perfectly well on their own, they're unrelated to #1393, and I can merge them quickly. #1394 carries the Live Activity fix — and it's currently a little ahead of this copy, with the fresh content install and an assertion on the non-empty precondition that this version doesn't have.

To be clear, this isn't a knock on your fix; it's a sequencing problem, and it's partly ours for not making it visible that #1393 was already claimed. If you'd rather take the Live Activity work forward instead, say so and I'll coordinate with the other contributor — but I don't want both PRs carrying it.

Split it and I'll take the toast tests.

Borisserz:
Leaving the Live Activity half to #1394 as you asked — not touching that path from our side. Toast-only split is yours whenever you’re ready.

ravindra-RKB:
@aaronbrethorst Makes complete sense! I've dropped the Live Activity commits so this PR is now exclusively the ToastManager tests. Ready for a quick merge whenever you are!

aaronbrethorst:
### Code review

_Re-review at 1abb6c3, covering commits since the previous review._

aaronbrethorst (approved):
Thanks for splitting this — it's exactly what I asked for and it was quick. The branch is now the ToastManager tests alone, and the async cleanup case is the one that would actually catch a regression in `dismiss()`. Verified the expectations against the implementation, including the 0.3s payload teardown. Merging.

Two cosmetic things for next time, neither worth another round: the file starts with a UTF-8 BOM (no other file in OBAKitTests has one), and the third `@Test` is indented four spaces past its siblings.

PR #1400, opened 2026-09-06
Add unit tests for TipKit configurations (MapLayers, Schedule, TripPlanner)

ravindra-RKB:
This PR improves the overall codebase stability by introducing unit tests for the previously untested `TipKit` components located in the `OBAKit/Tips` directory.

### What changed:
* Created `OBAKitTests/Tips/TipsTests.swift` utilizing Swift Testing.
* Validated that `ScheduleTip` and `TripPlannerTip` correctly initialize with the expected text and icon assets.
* Verified that `MapLayersTip` strictly implements `[Tips.MaxDisplayCount(1), Tips.IgnoresDisplayFrequency(true)]` to ensure it bypasses the hourly display budget and appears reliably on the first launch.

Adding these test assertions prevents future regressions if the underlying tooltip configuration logic is ever refactored.

## Summary by CodeRabbit

- **Tests**
  - Added coverage verifying that schedule, trip planner, and map layers tips provide titles, messages, and images.
  - Added validation that the map layers tip includes the configured maximum display count and frequency-related behavior options with the expected values.

### DIFF
--- a/OBAKitTests/Tips/TipsTests.swift
+++ b/OBAKitTests/Tips/TipsTests.swift
@@ -1,13 +1,14 @@
-//
+﻿//
 //  TipsTests.swift
 //  OBAKitTests
 //
-//  Copyright � Open Transit Software Foundation
+//  Copyright © Open Transit Software Foundation
 //  This source code is licensed under the Apache 2.0 license found in the
 //  LICENSE file in the root directory of this source tree.
 //
 
 import Foundation
+import SwiftUI
 import Testing
 import TipKit
 @testable import OBAKit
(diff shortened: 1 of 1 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The `MaxDisplayCount(1)` check can't catch a wrong count, which is the thing the PR says this test pins. `String(describing: maxDisplayCount).contains("1")` is a substring match. It passes for `MaxDisplayCount(10)`, `(11)`, `(21)` and `(100)`. I ran it against the Xcode 27 TipKit: the description is `Optional(TipKit.Tips.MaxDisplayCount(value: 10))`, and `.contains("1")` returns `true`. The comment on line 51 is also wrong. `Tips.MaxDisplayCount` is not `CustomStringConvertible`, and the TipKit `.swiftinterface` exposes only `init(_:)`. The text comes from Swift's default reflection of a private stored property called `value`, so this assertion (and the `"true"` check on line 55) depends on an undocumented layout that can change with any SDK. At minimum, compare the whole description exactly instead of checking for a substring. Better, read the value through `Mirror` and compare it with `== 1`, and add a comment saying it relies on TipKit's private layout. Otherwise, drop the value assertions and keep only the type-presence checks.

https://github.com/OneBusAway/onebusaway-ios/blob/5cb3d39153430a226aefac9eb9c09191d9c92540/OBAKitTests/Tips/TipsTests.swift#L49-L56

aaronbrethorst (changes requested):
Thanks for this — the tips genuinely had no coverage, and the file is in the right place this time (`OBAKitTests/Tips/` lands in the real test target, so these will actually run).

The title/message/image assertions are fine and will do their job. My concern is the three that carry the real weight of the PR:

```swift
#expect(String(describing: maxDisplayCount).contains("1"))
#expect(String(describing: ignoresFrequency).contains("true"))
```

These lean on the reflected description of opaque TipKit types, which isn't a contract Apple documents or promises to keep stable. Two problems follow from that. It can break on an SDK update with no behavior change at all, and — worse for a test — `"1"` is a substring that could plausibly show up in a description for reasons unrelated to `MaxDisplayCount(1)`, so a pass doesn't firmly establish the configured value. The comment in the test says as much ("its internal value is opaque/not easily equatable in some Swift versions"), which tells me you weren't sure either.

For a test-only PR the assertions *are* the deliverable, so I'd rather get these right than take them as-is.

What I'd suggest: assert the shape you actually care about and can check reliably — that `options` contains exactly one `Tips.MaxDisplayCount` and one `Tips.IgnoresDisplayFrequency`, and that there are two options total. That catches the realistic regression (someone drops or duplicates an option) without depending on reflection internals. If you want to pin the *values*, the honest way is a behavioral test through `Tips.configure` with a datastore, which is a bigger lift and I wouldn't ask for it here.

Worth knowing for context, since it's the reason `MapLayersTip` is configured the way it is: `.hourly` display frequency is a single app-wide budget that all tips compete for, which is how tips end up silently never showing. `IgnoresDisplayFrequency(true)` is what exempts this one. So the option being present is the behaviorally important fact — more so than the literal `1`.

Also: the file starts with a UTF-8 BOM, worth stripping.

Tighten those three assertions and I'll take it.

ravindra-RKB:
@aaronbrethorst Thanks for catching that! Good point about the reflection checks being brittle. 

I just pushed an update. I stripped the UTF-8 BOM and updated the MapLayersTip test. It now just asserts the shape directly (exactly one MaxDisplayCount and one IgnoresDisplayFrequency) instead of relying on string matching.

aaronbrethorst:
### Code review

_Re-review at 42a069b, covering commits since the previous review._

aaronbrethorst (approved):
Verified at 42a069b7. Thanks for this.

PR #1401, opened 2026-09-06
Add AgencyAlertsStore delete-vs-read concurrency coverage (#1145)

### LINKED ISSUE #1145: Follow-ups from PR #1138 (Bookmarks ViewModel Layer)

Borisserz:
## Summary
- Fans `deleteAgencyAlerts()` (via region change) against concurrent reads / `markAlertRead`, waiting for each delete so `cancelAllOperations()` does not drop it.
- After each waited delete the store must be empty (live refetch suppressed). Final counts match the fanout test: every seeded alert is present, and mark-all-read leaves none unread.
- Bookmark VM polish from that issue is already on `main`. This PR does not confirm CI runs the suite under TSan, so it does not close #1145.

## Test plan
- [x] `AgencyAlertsStoreConcurrencyTests` (2 tests) — passed on iPhone simulator
- [ ] Confirm CI with TSan still green on this suite

Refs #1145

### DIFF
--- a/OBAKitTests/Networking/AgencyAlertsStoreConcurrencyTests.swift
+++ b/OBAKitTests/Networking/AgencyAlertsStoreConcurrencyTests.swift
@@ -146,4 +146,62 @@ final class AgencyAlertsStoreConcurrencyTests: OBATestCase {
             #expect(!store.isAlertUnread(alert))
         }
     }
+
+    /// #1145: `deleteAgencyAlerts()`'s `removeAll()` is the mutation most likely to
+    /// corrupt state during a concurrent read. Region change drives that path;
+    /// fan it out against reads so a missing lock fails without relying on TSan.
+    @Test @MainActor
+    func `Concurrent delete-via-region-change and reads do not corrupt state`() async throws {
+        let dataLoader = MockDataLoader(testName: name)
+        let app = createApplication(dataLoader: dataLoader)
+        let store = app.alertsStore
+
+        let alerts = try makeRecentHighSeverityAlerts(count: 20)
+        store.insertAlerts(alerts)
+        #expect(store.recentHighSeverityAlerts.count == alerts.count)
+
+        let iterations = 30
+        await withTaskGroup(of: Void.self) { group in
+            group.addTask { @MainActor in
+                for _ in 0..<iterations {
+                    // `regionsService(_:updatedRegion:)` → `deleteAgencyAlerts()` on the
+                    // store's serial queue. Re-seed after each wipe so the next delete
+                    // still has something to clear.
+                    store.insertAlerts(alerts)
+                    store.regionsService(app.regionsService, updatedRegion: Fixtures.tampaRegion)
+                }
+            }
+            group.addTask {
+                for _ in 0..<iterations {
+                    _ = store.recentHighSeverityAlerts
+                    _ = store.agencyAlerts
+                    for alert in alerts {
+                        _ = store.isAlertUnread(alert)
+                    }
+                }
+            }
+            group.addTask {
+                for _ in 0..<iterations {
+                    for alert in alerts {
+                        store.markAlertRead(alert)
+                    }
+                }
+            }
+        }
+
+        // Wait for the serial delete queue + any in-flight `checkForUpdates`
+        // kicked off by `updatedRegion` to settle.
+        try await Task.sleep(nanoseconds: 300_000_000)
+
+        // `updatedRegion` deletes then refetches — do not assert emptiness.
+        // Coherence under the lock: recent/unread subsets stay inside `agencyAlerts`,
+        // and a final mark-all-read leaves no unread high-severity rows.
+        let allIDs = Set(store.agencyAlerts.map(\.id))
+        let recentIDs = Set(store.recentHighSeverityAlerts.map(\.id))
+        #expect(recentIDs.isSubset(of: allIDs))
+        for alert in store.agencyAlerts {
+            store.markAlertRead(alert)
+        }
+        #expect(store.recentUnreadHighSeverityAlerts.isEmpty)
+    }
 }


aaronbrethorst:
### Code review

Found 2 issues:

1. The new test's assertions can't fail, so it still only catches a missing lock if the process happens to crash. That is the same gap #1145 reported ("can pass even with the lock removed unless run under TSan"). The doc comment says "a missing lock fails without relying on TSan," but:
   - `recentHighSeverityAlerts` is defined as `agencyAlerts.filter { ... }`.
   - Every insert path (`insertAlerts`, `storeAgencyAlerts`) is `@MainActor`.
   - This `@MainActor` test reads `allIDs` and then `recentIDs` with no suspension point in between.
   
   So `recentIDs.isSubset(of: allIDs)` holds with or without the lock. The same goes for "mark every alert read, then expect no unread alerts." Also, `regionsService(_:updatedRegion:)` calls `cancelAllOperations()` before it queues each delete. The driver loop runs straight through on the main actor, so each iteration cancels the previous delete if it hasn't started yet. Only a few of the 30 deletes actually run while the reads are in flight. To make this test catch a missing lock, it needs something that can actually fail, like the count checks in the existing fanout test above it. It also needs deletes that actually run during the reads, for example by waiting for each delete to finish before queuing the next.

https://github.com/OneBusAway/onebusaway-ios/blob/3c4af83bdd90a4df9f08d5b929c5b0c4508fe7c2/OBAKitTests/Networking/AgencyAlertsStoreConcurrencyTests.swift#L149-L173

https://github.com/OneBusAway/onebusaway-ios/blob/3c4af83bdd90a4df9f08d5b929c5b0c4508fe7c2/OBAKitTests/Networking/AgencyAlertsStoreConcurrencyTests.swift#L195-L206

https://github.com/OneBusAway/onebusaway-ios/blob/3c4af83bdd90a4df9f08d5b929c5b0c4508fe7c2/OBAKitCore/Models/AgencyAlertsStore.swift#L176-L179

https://github.com/OneBusAway/onebusaway-ios/blob/3c4af83bdd90a4df9f08d5b929c5b0c4508fe7c2/OBAKitCore/Models/AgencyAlertsStore.swift#L350-L355

2. `Closes #1145` would close the issue with one of its requests still open. The AgencyAlertsStore item asks to "Confirm CI runs OBAKitTests with `-enableThreadSanitizer YES`." CI doesn't do that: the `test-without-building` step in `tests.yml` passes no sanitizer flag, and no file in the repo sets one. The PR's own test plan leaves "Confirm CI with TSan still green on this suite" unchecked. Because of issue 1, the new test only catches this race under TSan, and CI doesn't run TSan, so CI won't catch it. Either enable TSan for this suite or change `Closes` to `Refs` so the TSan follow-up stays tracked.

https://github.com/OneBusAway/onebusaway-ios/blob/3c4af83bdd90a4df9f08d5b929c5b0c4508fe7c2/.github/workflows/tests.yml#L218-L228

Borisserz:
The delete-vs-read test now waits for each queued delete before the next region change, so `cancelAllOperations()` cannot drop it. Live refetch is suppressed, and after each wait the store must be empty — that assertion fails if the delete never ran. The final counts match the fanout test: every seeded alert is present, and mark-all-read leaves none unread.

`Closes #1145` is now `Refs #1145`. This PR does not enable TSan in CI, so the issue stays open for that follow-up.

`AgencyAlertsStoreConcurrencyTests` passed (2 tests).

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Test-only, and it does what the previous round asked for. The store's serial queue is drained after each region change, so `cancelAllOperations()` can no longer swallow the delete you just queued, and the emptiness assertions can actually fail now instead of passing vacuously. Suppressing the live refetch is the right call — otherwise the delete gets overwritten before you can observe it.

Also appreciate the honesty in the description: `Refs #1145` rather than `Closes`, with the note that a missing lock still only shows up under TSan. That's the accurate framing.

Issue #1252, opened 2026-07-31
Follow-up: OBACloud Live Activity pushes should preserve relevance-score


Borisserz:
## Context
PR #1243 (Problem 2 of #1189) sets a monotonic on-device `relevanceScore` so the Dynamic Island follows the newest Track. Aaron noted that OBACloud pushes still send a flat `"relevance-score": 100` (see `docs/LiveActivityPushNotifications.md`), which re-ties scores after a push and undoes the Island ordering.

## Ask
On the OBACloud / push side (onebusaway-cloud-services is archived for issues), either:
1. omit `relevance-score` on Live Activity updates so the on-device score is left alone, or
2. send a per-activity score that preserves newest-Track-highest ordering.

Tracking here so #1189 is not closed by the app-only PR alone.

Related: #1189, #1243.

Borisserz:
Docs/contract for omit-or-per-activity relevance-score (app already preserves on local refresh): https://github.com/OneBusAway/onebusaway-ios/pull/1403 — OBACloud still needs to stop sending flat 100.

PR #1403, opened 2026-09-06
Document Live Activity push relevance-score contract (#1252)

Borisserz:
## Summary
- Push samples and the Node helper in `docs/LiveActivityPushNotifications.md` no longer default to `"relevance-score": 100`.
- New note: omit the field (preferred) or send a per-activity score; a flat 100 re-ties Dynamic Island ordering after Track (#1189 / #1243).
- Adds `docs/live-activity-push-relevance-score.md` for the app-vs-push contract. Local refreshes already use `contentPreservingRelevance`; full fix still needs OBACloud payload change.

## Test plan
- [ ] Docs-only — review the sample payload + Node helper
- [ ] Follow up in OBACloud to stop stamping `100` on updates

Refs #1252

### DIFF
--- a/docs/LiveActivityPushNotifications.md
+++ b/docs/LiveActivityPushNotifications.md
@@ -242,11 +242,26 @@ Your server must send HTTP/2 requests to APNs with specific headers:
             "vehicleId": "1234",
             "occupancyStatus": "MANY_SEATS_AVAILABLE"
         },
-        "stale-date": 1706198520,
-        "relevance-score": 100
+        "stale-date": 1706198520
     }
 }
 ```
+
+**Do not send a flat `"relevance-score": 100` on updates (#1252 / #1189).**
+ActivityKit shows the Live Activity with the highest score in the Dynamic
+Island. The app assigns a monotonic on-device score when the rider Tracks a
+trip (`TripLiveActivityRelevance.prominenceScore`) and demotes peers to `0`.
+A push that stamps every card with `100` re-ties those scores and undoes
+Island ordering. Either:
+
+1. **Omit `relevance-score`** on Live Activity updates so the on-device score
+   is left alone (preferred), or
+2. Send a **per-activity** score that preserves newest-Track-highest ordering.
+
+Local arrivals refreshes already go through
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. `Closes #1252` will auto-close the tracking issue before the fix it tracks exists. #1252 asks for a change on the OBACloud push side (omit `relevance-score`, or send a per-activity score), and it was opened specifically so the work would not be closed by app-side changes alone. This PR only edits the sample payloads in this repo's docs. The PR's own text says the full fix "still needs OBACloud payload change" and leaves "Follow up in OBACloud to stop stamping `100` on updates" unchecked. The new doc says the same thing. Use `Refs #1252` (or "Part of") so the issue stays open until OBACloud stops sending the flat `100`.

https://github.com/OneBusAway/onebusaway-ios/blob/62db015087d05de92e3a1a6eca729692ca4cdf18/docs/live-activity-push-relevance-score.md#L18-L22

Borisserz:
Changed `Closes #1252` to `Refs #1252`. This PR only documents the app-side contract; the issue stays open until OBACloud stops sending a flat `relevance-score: 100`.

aaronbrethorst (approved):
Good catch, and the write-up is the useful kind — it names the mechanism rather than just the symptom. A flat `"relevance-score": 100` on every update re-ties scores that the app deliberately made monotonic on Track, so the Island stops following the newest Track. Documenting that as a push-side contract is right: the client genuinely cannot intercept an APNs content update before ActivityKit applies it, so there's no app-side fix to write.

I verified the symbols you're pointing at rather than assuming: `TripLiveActivityRelevance.prominenceScore` and `contentPreservingRelevance` both exist in `OBAKitCore/LiveActivities/TripLiveActivityRelevance.swift`, and `LiveActivityUpdateCoalescer` calls the latter. So the table's claim that local refreshes preserve ordering and only the push path wipes it is accurate.

The Node helper change is the part that'll actually prevent the bug — omitting the key by default rather than defaulting it to 100 means the mistake has to be made deliberately.

Merging on a build that predates current `main`; that's fine here because this is two Markdown files and touches no compiled code.

PR #1408, opened 2026-09-07
Correct the cluster-count claim in the plural-selection comment

diveshpatil9104:
Follow-up from #1361:

> the comment says a cluster "always holds at least two", which holds where
> clusters are formed but not in `RentalClusterListView`, whose members are
> re-resolved against `visibleRentals` and can come back as one.

Traced it: `rentals(withIDs:)` is `visibleRentals.filter { wanted.contains($0.id) }`, and the caller only guards `isEmpty`. So a two-member cluster whose partner left the feed renders with `rentals.count == 1` — exactly the count #1350 was filed about.

## Summary by CodeRabbit

* **Documentation**
  * Clarified comments describing how rental clusters may be represented in the rental detail view.
  * Documented why a cluster containing one remaining rental can still appear in the interface.


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Verified at 0625854b. Thanks for this.

PR #1409, opened 2026-09-07
Add unit tests for DecodingErrorReporter and UnstructuredError

ravindra-RKB:
This PR improves codebase stability by introducing unit tests for `OBAKitCore/Utilities` components that were previously untested.

**What changed:**
* Created `OBAKitCoreTests/Utilities/DecodingErrorReporterTests.swift` utilizing Swift Testing to validate proper decoding error message formatting (`keyNotFound`, `typeMismatch`, `valueNotFound`, `dataCorrupted`) and ensure the custom report handler callback is successfully triggered.
* Created `OBAKitCoreTests/Utilities/UnstructuredErrorTests.swift` to verify proper error descriptions and recovery suggestions initialization.

Adding these assertions prevents future regressions and expands our automated test coverage within `OBAKitCore`.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for initializing unstructured errors with descriptions and optional recovery suggestions.

### DIFF
--- a/OBAKitCoreTests/Utilities/DecodingErrorReporterTests.swift
+++ b/OBAKitCoreTests/Utilities/DecodingErrorReporterTests.swift
@@ -1,0 +1,98 @@
+﻿//
+//  DecodingErrorReporterTests.swift
+//  OBAKitCoreTests
+//
+//  Copyright © Open Transit Software Foundation
+//  This source code is licensed under the Apache 2.0 license found in the
+//  LICENSE file in the root directory of this source tree.
+//
+
+import Foundation
+import Testing
+@testable import OBAKitCore
+
+struct DummyCodingKey: CodingKey {
+    var stringValue: String
+    var intValue: Int?
+    init(stringValue: String) {
+        self.stringValue = stringValue
+    }
+    init?(intValue: Int) {
+        self.stringValue = "\(intValue)"
+        self.intValue = intValue
+    }
+}
+
+@Suite(.serialized)
+final class DecodingErrorReporterTests {
+    
+    @Test func Message formatting for keyNotFound() {
+        let key = DummyCodingKey(stringValue: "missing_key")
+        let context = DecodingError.Context(codingPath: [], debugDescription: "Key was not found.")
+        let error = DecodingError.keyNotFound(key, context)
+        
+        let message = DecodingErrorReporter.message(from: error)
+        #expect(message.contains("Missing key: 'missing_key'"))
+        #expect(message.contains("Path: root"))
+        #expect(message.contains("Context: Key was not found."))
+    }
+    
+    @Test func Message formatting for typeMismatch() {
+        let context = DecodingError.Context(codingPath: [DummyCodingKey(stringValue: "parent"), DummyCodingKey(stringValue: "child")], debugDescription: "Expected String but found Int.")
+        let error = DecodingError.typeMismatch(String.self, context)
+        
+        let message = DecodingErrorReporter.message(from: error)
+        #expect(message.contains("Type mismatch (expected String)"))
+        #expect(message.contains("Path: parent → child"))
+        #expect(message.contains("Context: Expected String but found Int."))
+    }
+    
+    @Test func Message formatting for valueNotFound() {
+        let context = DecodingError.Context(codingPath: [DummyCodingKey(stringValue: "value")], debugDescription: "Null encountered.")
+        let error = DecodingError.valueNotFound(Int.self, context)
+        
+        let message = DecodingErrorReporter.message(from: error)
+        #expect(message.contains("Missing value (expected Int)"))
+        #expect(message.contains("Path: value"))
+        #expect(message.contains("Context: Null encountered."))
+    }
+    
+    @Test func Message formatting for dataCorrupted() {
+        let context = DecodingError.Context(codingPath: [], debugDescription: "Invalid JSON.")
+        let error = DecodingError.dataCorrupted(context)
+        
+        let message = DecodingErrorReporter.message(from: error)
+        #expect(message.contains("Data corrupted"))
+        #expect(message.contains("Path: root"))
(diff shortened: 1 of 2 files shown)


aaronbrethorst:
### Code review

Found 4 issues:

1. Neither test file is built or run. They were added under a new top-level `OBAKitCoreTests/` directory, but there is no such target. The only unit-test target is `OBAKitTests`, whose XcodeGen spec (`OBAKitTests/project.yml`) uses `sources: ["."]`, so it only picks up files under `OBAKitTests/`. The green `build` job on 629939c confirms this: the log compiles `OBAKitTests/Networking/DecodingErrorReporterTests.swift` and never mentions either new file, even though neither file would compile (see 2 and 4). All 128 lines add no coverage.

https://github.com/OneBusAway/onebusaway-ios/blob/629939c976bc0fd8fb651d39cf9fda718e5d2622/OBAKitCoreTests/Utilities/DecodingErrorReporterTests.swift#L1-L13

2. The test function names contain spaces with no backticks (`@Test func Message formatting for keyNotFound()`, `@Test func Initialization sets properties correctly()`). That is a syntax error. Raw identifiers need backticks, like the existing suite's ``@Test func `Key not found`()``. CI only passes because of issue 1, so moving the files into `OBAKitTests/` would break the build.

https://github.com/OneBusAway/onebusaway-ios/blob/629939c976bc0fd8fb651d39cf9fda718e5d2622/OBAKitCoreTests/Utilities/DecodingErrorReporterTests.swift#L28-L30

https://github.com/OneBusAway/onebusaway-ios/blob/629939c976bc0fd8fb651d39cf9fda718e5d2622/OBAKitCoreTests/Utilities/UnstructuredErrorTests.swift#L16-L18

3. `final class DecodingErrorReporterTests` duplicates a suite that already exists in the test target: `OBAKitTests/Networking/DecodingErrorReporterTests.swift` declares the same type name in the same module. Once this file is in the target, that is an invalid redeclaration. That suite also already covers everything this file tests: message formatting for all four `DecodingError` cases, the `Path: root` and `→` path formatting, and the URL, HTTP method and message passed to the report handler. Any cases that are actually new should go into that existing suite.

https://github.com/OneBusAway/onebusaway-ios/blob/629939c976bc0fd8fb651d39cf9fda718e5d2622/OBAKitCoreTests/Utilities/DecodingErrorReporterTests.swift#L25-L28

4. `reportHandler` has type `@Sendable (DecodingError, URL, String, String) -> Void`, but this closure changes captured local `var`s (`handlerCalled`, `reportedURL`, ...). CLAUDE.md says "every target builds in the **Swift 6 language mode**" with concurrency diagnostics as errors, and in that mode changing a captured var from a `@Sendable` closure is a compile error. The existing suite avoids this by capturing into a `SendableBox` (`capturedURL.value = url`). Also, the cleanup is a trailing `reportHandler = nil` statement, but the existing suite resets it in `deinit`.

https://github.com/OneBusAway/onebusaway-ios/blob/629939c976bc0fd8fb651d39cf9fda718e5d2622/OBAKitCoreTests/Utilities/DecodingErrorReporterTests.swift#L75-L97

aaronbrethorst (changes requested):
Thanks for taking a run at this — `DecodingErrorReporter` and `UnstructuredError` are reasonable things to want covered, and the error-path assertions you wrote are the right *shape*: you check the formatted message, the path rendering, and that the report handler actually fires with the URL and method.

But these files can't run, and there are three separate reasons, so I'd rather flag all of them at once than have you fix one and hit the next.

**1. `OBAKitCoreTests/` isn't a target.** It doesn't exist on `main`, and no XcodeGen spec references it. The only test target is `OBAKitTests` (`Apps/Shared/app_shared.yml`). Files placed in `OBAKitCoreTests/` are never compiled and never executed — which is exactly why CI is green on this PR. The green check here means "your code was ignored," not "your tests pass." OBAKitCore's tests live in `OBAKitTests/` today; `OBAKitTests/Networking/` and `OBAKitTests/Modeling/` are full of them.

**2. The test names aren't valid Swift.** For example:

```swift
@Test func Message formatting for keyNotFound() {
```

An identifier can't contain spaces. This needs backticks, which is also the convention throughout the suite:

```swift
@Test func `Message formatting for keyNotFound`() {
```

`NearbyTripMatcherTests.swift` has a dozen examples. This would be a parse error the moment the file was actually compiled — point 1 is the only reason you haven't seen it.

**3. `DecodingErrorReporterTests` already exists.** `OBAKitTests/Networking/DecodingErrorReporterTests.swift` is 390 lines and 14 tests, and it already covers key-not-found (including nested paths), type mismatch, value not found, data corrupted, root and nested path formatting, and four separate report-handler cases — including handler-not-called-when-nil, which yours doesn't have. Your new file would be a same-named class duplicating a subset of it.

Where that leaves things: the `UnstructuredError` tests are genuinely new — nothing covers that type directly today. If you move those into `OBAKitTests/` with backticked names, that's a real contribution and I'll take it. The `DecodingErrorReporter` half I'd drop, unless you've spotted a case the existing file misses — if so, add it there rather than starting a second file.

One more small thing: both files start with a UTF-8 BOM. Worth stripping.

Happy to re-review as soon as it's moved.

ravindra-RKB:
@aaronbrethorst Thanks for catching all of this! You're totally right I completely missed the existing `DecodingErrorReporterTests` in the Networking folder, and didn't realize `OBAKitCoreTests` wasn't included in the XcodeGen target.

I've pushed an update that addresses all your points:
1. Dropped the redundant `DecodingErrorReporterTests`.
2. Moved `UnstructuredErrorTests` to the correct `OBAKitTests/Utilities/` path.
3. Added backticks to the test names so they compile correctly.
4. Stripped the UTF-8 BOM.

aaronbrethorst:
### Code review

_Re-review at f8b12f8, covering commits since the previous review._

aaronbrethorst (approved):
Verified at f8b12f81. Thanks for this.

PR #1410, opened 2026-09-07
Scope the simulator-UDID note to the commands that actually need one

diveshpatil9104:
## Description

Follow-up from #1307, which Aaron raised on merge:

> line 22's "Every command below needs a simulator UDID" still isn't true —
> `scripts/swiftlint.sh`, `scripts/docs`, `scripts/version` and the `--help`
> blocks need nothing. "Every command that touches a simulator" would fix it.
> Worth a one-line follow-up.

Checked each: `scripts/swiftlint.sh` and `scripts/version` have no simulator reference at all, and `scripts/docs` builds for `-destination 'generic/platform=iOS'` — a generic destination, not a device.

Phrased as a list rather than a rule with exceptions. The failure this line invites is an agent resolving a UDID before running `swiftlint`; "every command that touches a simulator" still leaves it to guess which those are, so the line now names both sides.

Docs only.

## Summary by CodeRabbit

- **Documentation**
  - Clarified which development commands require a simulator UDID.
  - Documented that build-related commands require a UDID, while linting, documentation, versioning, and help commands do not.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,7 +19,9 @@ scripts/generate_project                # Defaults to OneBusAway if no app speci
 
 ### Picking a Simulator
 
-Every command below needs a simulator UDID. Resolve one with:
+The `xcodebuild` and `xcodebuildmcp` blocks below need a simulator UDID —
+`scripts/swiftlint.sh`, `scripts/docs` (which builds for `generic/platform=iOS`),
+`scripts/version` and the `--help` blocks need nothing. Resolve one with:
 
 ```bash
 SIMULATOR_UDID=$(scripts/resolve_simulator_udid)


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Correct on every claim, and I checked each one rather than taking the summary's word for it: `scripts/docs` really does build for `generic/platform=iOS` (line 16), and neither `scripts/swiftlint.sh` nor `scripts/version` references a destination or a simulator at all. The `--help` blocks obviously need nothing.

"Every command below needs a simulator UDID" was overreach, and the cost of that kind of overreach is someone resolving a UDID before running SwiftLint and concluding the docs are wrong about something. Scoping it to the `xcodebuild` and `xcodebuildmcp` blocks is the accurate statement.

Merging on a build that predates current `main`, which I'd normally not do — it's fine here because the diff is one paragraph of Markdown and cannot affect compilation.

PR #1413, opened 2026-09-08
Say Map settings for the button that opens the Map sheet

### LINKED ISSUE #1412: VoiceOver says "Map type" for the button that opens the whole Map settings sheet
`map_controller.map_type.accessibility_label` is used as the VoiceOver label for two different live controls that do different things. Both announce **"Map type"**, and for one of them that is wrong.

## The two call sites

**1. `OBAKit/Mapping/MapViewController.swift:535`** — the UIKit hover-bar button. This one *does* toggle the base map type:

```swift
button.addTarget(self, action: #selector(toggleMapType), for: .touchUpInside)
button.accessibilityLabel = OBALoc("map_controller.map_type.accessibility_label", value: "Map type",
    comment: "Voiceover text indicating that this button toggles the base map type.")
```

"Map type" is accurate here.

**2. `OBAKit/Sheet/Root/Controls/MapTypeButton.swift:44-48`** — the SwiftUI control in `MapControlsCluster`, wired to `onOpenMapSettings`. It does not toggle anything; it opens the Map sheet:

```swift
.accessibilityLabel(Text(OBALoc(
    "map_controller.map_type.accessibility_label",
    value: "Map type",
    comment: "Voiceover text for the button that opens the Map settings sheet."
)))
```

## Why "Map type" is wrong for the second one

`MapSheetView` is titled **"Map"** and contains four sections, only the first of which is about the map type:

- `basemapSection` — Standard / Satellite / Hybrid
- `mapDisplaySection` — "Map display" (Points of Interest, etc.)
- `layerSection` — "Transit"
- `otherModesSection` — "Other ways to get around" (rental bikes, scooters)

So a VoiceOver user hears "Map type", activates it expecting a basemap toggle, and lands in a sheet that also controls transit layers, rental layers and POI display. The label describes about a quarter of what the button opens.

There is corroborating evidence in the control itself: it already sets an accessibility **value** of `map_controller.map_type.accessibility_value_with_layers_fmt` = `"%1$@, %2$d layers on"`. VoiceOver announces something like *"Map type, hybrid, 3 layers on"* — the value already talks about layers that the label does not admit exist.

## How it surfaced

`genstrings` warns when one key is used with two different comments, because it silently keeps only one and a translator can end up with context for the call site they are not looking at:

```
Key "map_controller.map_type.accessibility_label" used with multiple comments
  "Voiceover text for the button that opens the Map settings sheet."
& "Voiceover text indicating that this button toggles the base map type."
```

Four other keys had the same warning for benign reasons (one control described twice) and were unified in #1405. This one was deliberately **left un-unified** so the underlying defect stays visible instead of being tidied away behind agreed prose.

## What needs deciding

The fix is a new key for the sheet-opening button, but the copy is a product call, not a mechanical change:

- **"Map settings"** — matches the sheet's actual scope, conventional for a control that opens a settings surface.
- **"Map options"** — same idea, slightly softer.
- **"Map"** — matches the sheet's own `navigationTitle`, but reads thin as a button label.

Whichever is chosen needs a new `OBALoc` key plus translations in the 12 non-English catalogs. Worth deciding alongside whether the accessibility *value* should keep leading with the basemap name once the label no longer says "type".

Not urgent — nothing is broken for sighted users, and the button works. It is a VoiceOver-only correctness gap.

Found while working through review follow-ups on #1405.

diveshpatil9104:
Closes #1412.
`map_controller.map_type.accessibility_label` labelled two controls that do different things. The UIKit hover-bar button really does toggle the basemap, so "Map type" is right there. The SwiftUI control opens `MapSheetView`, whose basemap picker is one of four sections — the others cover points of interest, transit layers and other travel modes — so the label described a quarter of where it leads.
New key `map_controller.map_settings.accessibility_label` = **"Map settings"**, on the sheet-opening button only. The UIKit button keeps the old key, which now has one call site and one comment, so the `genstrings` duplicate-comment warning is resolved rather than suppressed.
All 13 catalogs, `plutil -lint` clean, every locale at 566 keys.
**On the copy:** you listed "Map settings", "Map options" and "Map". Went with "Map settings" — it matches the sheet's scope, and "Map" alone reads thin on a button. Happy to switch; it's your call.
**On the value:** left it leading with the basemap name. VoiceOver now says "Map settings, hybrid, 3 layers on" — the label names the destination, the value summarises state.
VoiceOver-only; nothing changes for sighted users.

## Summary by CodeRabbit

* **Accessibility**
  * Updated the map controls VoiceOver label to clearly identify Map settings.
  * Clarified that Map settings includes base maps, points of interest, transit layers, and travel modes.

* **Localization**
  * Added translated Map settings accessibility labels across supported languages.
  * Added translations for map layers, rental details, reporting, stop and trip screens, and accessibility announcements.
  * Removed obsolete translations for unused location-permission, stop, trip-panel, and timeline labels.

### DIFF
--- a/OBAKit/Sheet/Root/Controls/MapTypeButton.swift
+++ b/OBAKit/Sheet/Root/Controls/MapTypeButton.swift
@@ -41,10 +41,15 @@ struct MapTypeButton: View {
                     .accessibilityHidden(true)
             }
         }
+        // Distinct from the UIKit hover bar's `map_controller.map_type.*` label,
+        // which really does toggle the basemap. This opens `MapSheetView`, whose
+        // basemap picker is one of four sections — the others cover POI display,
+        // transit layers and rental modes — so "Map type" described a quarter of
+        // where it leads (#1412).
         .accessibilityLabel(Text(OBALoc(
-            "map_controller.map_type.accessibility_label",
-            value: "Map type",
-            comment: "Voiceover text for the button that opens the Map settings sheet."
+            "map_controller.map_settings.accessibility_label",
+            value: "Map settings",
+            comment: "Voiceover text for the button that opens the Map settings sheet, which covers the base map, points of interest, transit layers and other travel modes."
         )))
         .accessibilityValue(Text(accessibilityValueText))
     }
--- a/OBAKit/Strings/ar.lproj/Localizable.strings
+++ b/OBAKit/Strings/ar.lproj/Localizable.strings
@@ -456,6 +456,9 @@
 /* Title of the alert that appears when the user is updating their current region manually. */
 "map_controller.change_region_alert.title" = "هل تريد تغيير المنطقة؟";
 
+/* Voiceover text for the button that opens the Map settings sheet, which covers the base map, points of interest, transit layers and other travel modes. */
+"map_controller.map_settings.accessibility_label" = "إعدادات الخريطة";
+
 /* Voiceover text indicating that this button toggles the base map type. */
 "map_controller.map_type.accessibility_label" = "نوع الخريطة";
 
--- a/OBAKit/Strings/en.lproj/Localizable.strings
+++ b/OBAKit/Strings/en.lproj/Localizable.strings
@@ -520,8 +520,10 @@
 /* Title of the alert that appears when the user is updating their current region manually. */
 "map_controller.change_region_alert.title" = "Change Region?";
 
-/* Voiceover text for the button that opens the Map settings sheet.
-   Voiceover text indicating that this button toggles the base map type. */
+/* Voiceover text for the button that opens the Map settings sheet, which covers the base map, points of interest, transit layers and other travel modes. */
+"map_controller.map_settings.accessibility_label" = "Map settings";
+
+/* Voiceover text indicating that this button toggles the base map type. */
 "map_controller.map_type.accessibility_label" = "Map type";
 
 /* Voiceover value combining the base map type with the number of enabled map layers. %1$@ is the base map type, %2$d is the layer count. Plural forms live in Localizable.stringsdict; the value above is only the not-found fallback. */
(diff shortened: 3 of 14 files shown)


aaronbrethorst:
### Code review

Found 1 issue:

1. The Spanish and Portuguese words for "settings" are swapped compared with the rest of each catalog. pt-BR uses "Configurações do mapa", but every other pt-BR string that mentions settings says "Ajustes": `common.settings` = "Ajustes", "Abrir Ajustes", "Ativar nos Ajustes", "Mais > Ajustes > Migrar Dados". This would be the only "Configurações" in the locale. es uses "Ajustes del mapa", but the Spanish catalog's word is mostly "Configuración": `common.settings` = "Configuración", "Abrir Configuración", "elija una región en Configuración". Swapping the two values ("Configuración del mapa" / "Ajustes do mapa") would match what each catalog already uses, the same kind of fix as the one-word-per-concept cleanup in #1405 (bug due to translation choice in es/pt-BR Localizable.strings)

https://github.com/OneBusAway/onebusaway-ios/blob/5a271fc4495929efb9f3a2c08f694ec5a99b3e70/OBAKit/Strings/pt-BR.lproj/Localizable.strings#L459-L461

https://github.com/OneBusAway/onebusaway-ios/blob/5a271fc4495929efb9f3a2c08f694ec5a99b3e70/OBAKit/Strings/es.lproj/Localizable.strings#L459-L461

aaronbrethorst (approved):
This is the right fix and it's done properly. "Map type" really did describe about a quarter of where that button goes — the sheet covers the basemap picker plus POI display, transit layers and rental modes — so a rider on VoiceOver was getting a materially wrong promise.

What makes this mergeable rather than a half-fix: you added `map_controller.map_settings.accessibility_label` to all 13 locales with a real translation in each, and left `map_controller.map_type.accessibility_label` in place for the UIKit hover bar that genuinely does toggle the basemap. Splitting the merged developer comment in `en.lproj` back into two is correct too — those were two different strings sharing one comment.

Localization changes here have a habit of quietly swapping a translated string for an English one; this one doesn't, and I checked.

PR #1415, opened 2026-09-12
Bring the trip planner to the SwiftUI map panel

mosliem:
Brings OTPKit's trip planner to the SwiftUI map panel. `AppSheetRoute.tripPlanner` existed but had no registered view, so both entry points into trip planning — a map item's "Directions" and a rental's "Plan a trip using this bike" — passed `nil` handlers and hid their buttons. Both work on the panel now.

OTPKit's only shipped `OTPMapProvider` mutates an `MKMapView` outright, taking its delegate and installing its own gesture recognizer; the UIKit surface obliges by handing over a whole second, hidden map view and swapping the two by alpha. The panel has no view to hand over. Rather than fork OTPKit, this implements the protocol a second way — `TripPlannerMapDisplayModel` records OTPKit's calls into published state the panel renders declaratively as `MapPolyline` and `Annotation`. Two upstream OTPKit changes made that possible and are already merged there: `@MainActor` on the protocol (OneBusAway/otpkit#160), without which a conforming type in OBAKit's Swift 6 main-actor-isolated module cannot satisfy it, and `TripPlannerChrome.embedded`, so the planner renders inside the panel's navigation instead of bringing a second header and close button.

## Summary

- **`TripPlannerMapDisplayModel`** — an `OTPMapProvider` backed by state instead of a map view. Camera moves route through the one-shot `CameraTarget` mechanism `MapSearchDisplayModel` already uses, so `setVisibleMapRect` can't yank the map out from under the rider on an unrelated body pass. `setMapType` and friends are deliberate no-ops: OTPKit is a guest on the panel's one shared map, and basemap style is the rider's persisted choice.
- **`AppSheetRoute.tripPlanner` carries a `TripPlannerRequest`** — optional destination, via point and transport mode, covering both entry points from one payload. Analytics keys derive from which fields are *present*, never from coordinates.
- **`TripPlannerSheetView`** holds `TripPlanner` in a `@StateObject` wrapper, so the factory rebuilding a view body can't reset the rider's trip mid-flow. It opens at `.medium` and returns there when OTPKit's directions sheet opens over it — a full-height planner hides the very map the route is drawn on.
- **Both entry points push the route**, gated on `Region.supportsOTP`. The rental case plans *through* the vehicle — a via point paired with `.transitBikeRental`, matching `MapViewController.rentalLayer(planTripUsing:)` — because OTP won't route through a via point in a rental-only mode.
- **`TripPlannerTip` gets a map-panel home** on the home sheet's search row, the counterpart of the UIKit panel's search bar its copy names.
- Trip pins are tagged with OTPKit's own opaque identifier and handed back verbatim on tap, through the `MapPinSelection` binding stops and rentals already share.

## Fixes for defects this flow exposed

- **Publishing from within view updates, twice over.** OTPKit prefills as a side effect of *building* its view, so doing that from `body` mutated observed state mid-update; it now happens once, from `.task`. Separately, the display model published synchronously while OTPKit drove it from inside SwiftUI's update pass (`DirectionsSheetView` calls the map coordinator from `onAppear` and three `onChange` handlers), so every interaction with the directions sheet warned. Its notification is now deferred and coalesced — one itinerary is dozens of provider calls and deserves one re-render. Reads stay synchronous, because `isShowingTrip` gates the ambient stop layer during `body`.
- **Two sheets from one view.** The rental cluster list presents its own detail sheet, which is right on the UIKit surface but collides here: `StackedSheetLayer` attaches the next stacked route's `.sheet` to that same content, and SwiftUI allows one — *"Currently, only presenting a single sheet is supported."* The trip planner pushed from inside the rental sheet stayed queued until the rider dismissed the rental by hand. The list now delegates the drill-in to the host, which pushes `.rentalDetail` through the coordinator.
- **An open rental sheet dying when the map moves.** The sheets resolve their vehicle by id every body pass, against a list that is viewport-scoped *and* zoom-gated. Framing a planned trip re-fetches for the itinerary's bounding box, and every vehicle outside it returns as a removal — indistinguishable, in the snapshot, from one a rider just rode away. Resolution now falls back to vehicles pinned when the sheet was opened, then to the last report the feed made. Pinned vehicles never reach the map, and a vehicle nobody opened a sheet for can still be genuinely gone. Pre-existing and reproducible without any trip: pan the map with a rental sheet open.

## Verification

- **2619 tests in 275 suites**, one pre-existing failure (`RentalFormatTests.rangeFallbackUsesAbbreviatedUnits`, a locale artifact on code this branch does not touch)
- SwiftLint clean on every touched file
- Each of the three fixes above landed against a reproduction in the simulator, iPhone 16 / iOS 26, and carries regression tests

## Follow-up, not fixed here

**Stop details → Directions is hidden on the panel.** `StopTripPlannerAction` already ships "Directions to Here" / "Directions from Here" in the stop page's More menu, but `canPresent` requires `viewRouter.rootController`, which is nil in map-panel mode, so `StopPageActionPresenter` passes `nil` for both and the items disappear. `StopPageActionRow` already renders them when non-nil — only the panel's push is missing. Follow-up, since it also needs an `origin` on `TripPlannerRequest`.

## Summary by CodeRabbit

* **New Features**
  * Added trip planning to the SwiftUI map panel, including route previews, map annotations, and destination, via-point, and transport-mode options.
  * Added a trip-planning tip to the home search bar when available.
  * Added “Plan Trip” actions for map items and rental vehicles.
  * Added rental cluster drill-in navigation.

* **Bug Fixes**
  * Rental details remain available when vehicles move outside the current map viewport.
  * Trip-planning actions are hidden in regions without trip-planning support.

### DIFF
--- a/Apps/OneBusAway/Package.resolved
+++ b/Apps/OneBusAway/Package.resolved
@@ -178,7 +178,7 @@
       "location": "https://github.com/OneBusAway/otpkit.git",
       "state": {
         "branch": "main",
-        "revision": "5d2fb5a0e39b53a98050324b6a09b4d6eb240e0a"
+        "revision": "45438b555cb7071ab9ee47336425aaa1709fbdb4"
       }
     },
     {
--- a/OBAKit/Mapping/Layers/RentalDetailViewController.swift
+++ b/OBAKit/Mapping/Layers/RentalDetailViewController.swift
@@ -63,9 +63,8 @@ struct RentalDetailView: View {
     /// The layer's declared trust window; past it, the footer flags the data as stale.
     let staleAfter: Duration?
     let userLocation: CLLocation?
-    /// Nil on the SwiftUI map panel, which has no trip planner to route into —
-    /// `AppSheetRoute.tripPlanner` has no registered view. The button is hidden
-    /// rather than disabled: a dead primary action is worse than none.
+    /// Nil when the current region has no OTP server, on either surface. The button is
+    /// hidden rather than disabled: a dead primary action is worse than none.
     var onPlanTrip: ((VehicleRental) -> Void)?
     var onOpenURL: (URL, URL?, String?) -> Void
 
@@ -316,6 +315,18 @@ struct RentalClusterListView: View {
     let staleAfter: Duration?
     let userLocation: CLLocation?
     var onPlanTrip: ((VehicleRental) -> Void)?
+    /// Where tapping a row goes, when the host wants to own that.
+    ///
+    /// Nil on the UIKit surface, where this view is a modally-presented hosting
+    /// controller and the `selectedRental` sheet below is the only way to drill in.
+    ///
+    /// The map panel passes a closure instead, because that sheet is not the panel's to
+    /// present: `StackedSheetLayer` already attaches a `.sheet` to this same content for
+    /// the next stacked route, and SwiftUI allows one presentation per view — "Currently,
+    /// only presenting a single sheet is supported." A rental sheet opened here would hold
+    /// that slot, and the trip planner pushed from inside it would silently queue until the
+    /// rider dismissed the rental by hand.
+    var onSelectRental: ((VehicleRental) -> Void)?
     var onOpenURL: (URL, URL?, String?) -> Void
 
     @State private var selectedRental: VehicleRental?
@@ -324,7 +335,11 @@ struct RentalClusterListView: View {
         NavigationStack {
             List(rentals) { rental in
                 Button {
-                    selectedRental = rental
+                    if let onSelectRental {
+                        onSelectRental(rental)
+                    } else {
+                        selectedRental = rental
+                    }
                 } label: {
                     row(for: rental)
                 }
--- a/OBAKit/Mapping/Layers/RentalLayerCoordinator.swift
+++ b/OBAKit/Mapping/Layers/RentalLayerCoordinator.swift
@@ -101,6 +101,16 @@ import OTPKit
 
     var hasEnabledLayers: Bool { !enabledLayerFactors.isEmpty }
 
+    /// Whether the layer is currently in a position to say anything about vehicles.
+    ///
+    /// False when nothing is enabled, and false when the zoom gate is closed —
+    /// `MapRegionManager.forwardViewport(to:)` hands this coordinator a nil viewport
+    /// once the map is zoomed out past the layer's `zoomWindow`, which empties
+    /// `visibleRentals` wholesale. An empty list then means "not looking", not "none
+    /// there", and callers resolving a specific vehicle by id must not read the two the
+    /// same way.
+    var isReportingVehicles: Bool { hasEnabledLayers && lastMapRect != nil }
+
     private var combinedFormFactors: Set<VehicleFormFactor> {
         enabledLayerFactors.values.reduce(into: Set<VehicleFormFactor>()) { $0.formUnion($1) }
     }
--- a/OBAKit/Sheet/Content/Home/HomeSheetView.swift
+++ b/OBAKit/Sheet/Content/Home/HomeSheetView.swift
@@ -8,6 +8,7 @@
 //
 
 import SwiftUI
+import TipKit
 import OBAKitCore
 
 struct HomeSheetView: View {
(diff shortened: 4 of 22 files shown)


aaronbrethorst:
### Code review

Found 2 issues:

1. The planner's navigation title is English in all 12 non-English locales. Switching OTPKit to `.embedded` chrome drops its own title, `OTPLoc("trip_planner.title")`, which OTPKit translates in all 13 locales (es "Planificación de viaje", fr "Planification d’itinéraire", ru, ar, zh-Hans, and so on). The replacement is `OBALoc("trip_planner.title", value: "Trip Planner")`, but that key isn't in any `OBAKit/Strings/*.lproj/Localizable.strings`, so every rider sees "Trip Planner" each time the planner opens. That's a regression from the UIKit surface, where the same screen is localized. `LocalizationTests` won't catch it, because the key is missing from `en` as well. The new `trip_planner.unavailable.*` and `trip_planner.destination.default_title` keys are also missing from every locale. `OTPLoc` is public, so reusing OTPKit's translated string would fix the title.

https://github.com/OneBusAway/onebusaway-ios/blob/71737234783849c582a1797945a6100e1940985b/OBAKit/Sheet/Content/TripPlanner/TripPlannerSheetView.swift#L88-L94

2. After this change, a rental sheet can't show that its vehicle is gone. Both push sites pin the vehicle when the sheet opens: the map-pin tap and the cluster drill-in. `rental(withID:)` and `rentals(withIDs:)` then fall back to the pin whenever the live list is missing that vehicle. This happens even while the feed is reporting, so a bike someone rents while the sheet is open keeps showing as available until the region changes. `rentalUnavailableView` becomes unreachable for these routes. The doc comment says freshness "is reported by the sheet's own `fetchedAt` / `staleAfter` footer," but it isn't: `rentalFetchedAt` is the coordinator's `lastSnapshotAt`, which moves forward on every fetch. So the fetch that drops the bike also marks the old pinned copy as just updated. The doc comment kept on `AppSheetViewFactory.rentalDetailView` still says "a vehicle that leaves the feed while the sheet is open reads as gone rather than as a stale row," which is no longer true. Viewport-driven removals and real ones need to be told apart, for example by only trusting pins while the vehicle is outside the fetched bounding box, rather than trusting every pin.

https://github.com/OneBusAway/onebusaway-ios/blob/71737234783849c582a1797945a6100e1940985b/OBAKit/Sheet/Root/MapPanelLayersModel.swift#L226-L283

aaronbrethorst (changes requested):
This is a substantial piece of work and the direction is right — bringing the planner into the SwiftUI map panel is where this should end up, and the `.embedded` chrome makes the panel feel like one surface instead of two stacked ones.

Two things block it, and the first is the kind of regression that's easy to ship and hard to notice.

**1. The planner title becomes untranslated English in all 12 non-English locales.**

Switching OTPKit to `.embedded` drops its own title, `OTPLoc("trip_planner.title")` — which OTPKit ships translated in all 13 locales ("Planificación de viaje", "Planification d'itinéraire", and so on). The replacement is `OBALoc("trip_planner.title", value: "Trip Planner")`, but this PR doesn't touch a single file under `OBAKit/Strings/`, and that key doesn't exist in any `.lproj` on `main` — I checked all 13, including `en`. So the `value:` fallback is what renders, and every rider outside English sees "Trip Planner" on a screen that used to be in their language.

Same applies to the new `trip_planner.unavailable.*` and `trip_planner.destination.default_title` keys — code-only, no `.strings` entries anywhere.

`LocalizationTests` can't catch this, because it diffs the other locales against `en` and the key is missing from `en` too. That's worth internalizing: a missing key is invisible to that test by construction.

`OTPLoc` is public, so the simplest fix for the title is to keep using OTPKit's already-translated string. The other new keys need real entries in all 13 locales.

**2. A rental sheet can no longer show that its vehicle is gone.**

Both push sites pin the vehicle when the sheet opens — the map-pin tap and the cluster drill-in — and `rental(withID:)` / `rentals(withIDs:)` then fall back to that pin whenever the live list doesn't contain it. That happens even while the feed is actively reporting, so a bike that someone else rents while the sheet is open keeps reading as available until the region changes. `rentalUnavailableView` is unreachable on both of those routes.

The doc comment says freshness "is reported by the sheet's own `fetchedAt` / `staleAfter` footer," but that isn't what happens: `rentalFetchedAt` is the coordinator's `lastSnapshotAt`, which advances on every fetch — so the very fetch that dropped the bike stamps the stale pinned copy as just-updated. And the comment left on `AppSheetViewFactory.rentalDetailView` still claims "a vehicle that leaves the feed while the sheet is open reads as gone rather than as a stale row," which is now false.

The distinction you need is between a vehicle leaving the *viewport* and leaving the *feed*. Trusting the pin only while the vehicle is outside the fetched bounding box would preserve what the pin is for without masking real removals.

Neither of these is a design disagreement — the structure of the PR is fine. Fix the strings and the pin-versus-feed distinction and I'll re-review promptly.

aaronbrethorst:
### Code review

PR #1418, opened 2026-09-14
Key the current-trip list by vehicle, not trip

### LINKED ISSUE #1417: Current-trip list drops a coach when two vehicles share a trip
The SwiftUI “multiple vehicles” list keys its rows on `tripID`. That is not a unique visit.

`NearbyTripMatcher.filterAndSort` already keeps one row per vehicle (`vehicleID`, falling back to `tripID`). Feeds occasionally assign two vehicles to one trip — `filteringDuplicateVehicleReports` documents that — and the matcher fetches up to five stops without collapsing those rows back together. Both coaches reach `matchResults`.

`CurrentTripView` then does:

```swift
ForEach(viewModel.matchResults, id: \.arrivalDeparture.tripID)
```

SwiftUI `ForEach` requires unique ids. The second coach is dropped or updated as the first. The row subtitle prints `vehicleID`, so the screen is trying to tell them apart. The UIKit path (`CurrentTripViewController`) builds a fresh row per result, so the same input stays two rows there.

## Expected

Two realtime vehicles on the same trip, both within range, show as two rows. A missing vehicle id still falls back to the trip id — that is the key the matcher already uses, and those rows were already collapsed to one.

Borisserz:
## Summary
- Closes #1417.
- The “multiple vehicles” list identified rows by `tripID`. A feed can put two vehicles on one trip, and `NearbyTripMatcher` already keeps both, so SwiftUI dropped or reused one row.
- Rows now use the same key the matcher uses: `vehicleID`, falling back to `tripID` when the feed omits a vehicle id.

## Test plan
- [x] `NearbyTripMatcherTests` — two vehicles on `trip_1` get list ids `coach_a` and `coach_b`
- [x] A blank vehicle id falls back to the trip id
- [x] The rest of `NearbyTripMatcherTests` passed
- [ ] On a route with two nearby realtime coaches on the same trip, the SwiftUI sheet shows both rows and each opens the coach that was tapped

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved current-trip results so vehicles operating the same trip are displayed as distinct entries.
  * Added a fallback for services that do not provide vehicle identifiers, ensuring results remain grouped and identifiable using trip information.
  * Corrected list tracking to reduce duplicate or incorrectly merged rows in current-trip views.

* **Tests**
  * Added coverage for multiple vehicles on one trip and missing vehicle identifiers.


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Good fix, and the reasoning behind it is sound. Keying the row on `vehicleID ?? tripID` matches the key `filterAndSort` already dedupes on, so the new identity is collision-free by construction rather than by luck — two rows can't share a `listID` because the second one never survives the `seen` insert. That's the property that makes this safe, and it's worth the line of thought.

The two tests cover the real cases: two vehicles on one trip, and the blank-vehicle-id fallback. Nice touch verifying that `vehicleID` is nilified for blank strings upstream rather than assuming it.

One thing for later, not blocking: `filterAndSort` still computes `arrival.vehicleID ?? arrival.tripID` inline instead of going through `listID`. They agree today, and your doc comment says so, but nothing enforces it — routing both through the one property would keep a future edit from quietly desynchronizing them.

PR #1420, opened 2026-09-14
Pass a locale when formatting stringsdict counts

### LINKED ISSUE #1419: Plural forms in stringsdict never reach Polish, Russian, or Arabic
Several rider-facing counts have correct `few` / `many` / `zero` forms in `Localizable.stringsdict`, and the screens still format them with `String(format:)` and no locale.

That overload resolves `%#@count@` against the root plural rule, which only defines `one` and `other`. English still looks right. A Polish count of 5 hits `other` ("Pokaż 5 minionego odjazdu") instead of `many` ("Pokaż 5 minionych odjazdów"). Russian and Arabic lose the same categories. `LocalizationTests` already documents the trap, and `30545c85` fixed the siblings it found. These call sites were left behind.

## Still on `String(format:)`

- `StopPageListHeaderRow` — VoiceOver "Show N past departures", and the announcement when that disclosure opens
- `ServiceAlertsSection` — "N service alerts" and "Show all N alerts"
- `StopPageView` — "No departures in the next N minutes"
- `StopViewController` — "Show N earlier departures" on a transfer
- `DataMigrationReportView` — "N successful" and "N failures"

## Expected

The same key the catalog already has, rendered with the rider's plural category. A Polish 5 must use `many`, a Polish 3 must use `few`.

`stop_page.timeline.skipped_stops_fmt` has stringsdict entries and no production call site. It is out of scope here.

Borisserz:
## Summary
- Closes #1419.
- `String(format:)` with no locale resolves `%#@count@` against the root plural rule, so only `one` and `other` were reachable. English looked fine. Polish, Russian, and Arabic `few` / `many` / `zero` forms sat in `Localizable.stringsdict` and never reached the rider.
- The leftover call sites now go through `CountPlural`, which passes `Locale.current` into `String(format:locale:)`. Same rule `StopPageAccessibilityCopy` already uses.

Call sites: past-departures VoiceOver label and announcement, service-alert summary and “show all”, empty stop page, transfer “earlier departures”, and the data-migration report counts.

`stop_page.timeline.skipped_stops_fmt` has stringsdict entries and no production caller. Left alone.

## Test plan
- [x] `LocalizationTests` — `Count plural call sites reach Polish few and many` (Polish 5 is `many`, Polish 3 is `few` for the past-departures label)
- [x] Rest of `LocalizationTests` passed
- [ ] Polish (or Russian) device: a stop with 5 past departures, VoiceOver on the Past button, should say the `many` form, not the `other` form

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pluralized localization for departure counts, service alerts, onboarding reports, and accessibility messages.
  * Corrected plural forms for languages such as Polish, including “few” and “many” cases.
  * Updated empty-state and earlier-departure messages to display grammatically correct count-based text.

* **Tests**
  * Updated coverage validating localized plural forms across supported count-based messages.


aaronbrethorst:
### Code review

Found 1 issue:

1. Routing these 8 call sites through `CountPlural.format` takes them out of `OBALoc`, so `scripts/extract_strings` (`genstrings -s OBALoc`) no longer sees them. The `comment:` parameter is accepted and then discarded, which makes the call sites look like they still feed the extractor. Verified by running `genstrings -s OBALoc` over a reduced file with one `OBALoc` call and one `CountPlural.format` call: only the `OBALoc` key was emitted. The next `scripts/extract_strings` run therefore drops all 8 keys — `stop_page.past_toggle_show_a11y_fmt`, `stop_page.a11y.past_shown_fmt`, `stop_page.service_alerts.show_all_fmt`, `stop_page.service_alerts.summary_fmt`, `stop_page.empty.no_departures_fmt`, `stop_controller.transfer_show_earlier_departures_fmt`, `data_migration_bulletin.report_summary_number_of_{successes,failures}` — plus their translator comments from `OBAKit/Strings/en.lproj/Localizable.strings`, which then fails `Every locale has same keys as english` with 8 extra keys across all 12 non-en locales. Every other stringsdict call site in the repo keeps `OBALoc(...)` at the call site and only changes the formatter around it (`String.localizedStringWithFormat` in `SearchResultsSheetView`, `RentalMapMarker`, `MapTypeButton`; `String(format:locale:)` over an `OBALoc` result in `StopPageAccessibilityCopy`), which preserves extraction. Having `CountPlural.format` take an already-localized format string (or the `OBALoc` call) instead of the key would keep both the locale fix and the extraction.

https://github.com/OneBusAway/onebusaway-ios/blob/e060984650ca1d0385baddf74cdabca5cd8eba0b/OBAKit/Strings/Localization.swift#L24-L32

aaronbrethorst (changes requested):
The bug you're fixing is real and the diagnosis is exactly right: `String(format:)` expands `%#@count@` against the root plural rule, so the `few`/`many`/`zero`/`two` forms in `ar`, `pl` and `ru` could never be selected. Threading a locale through is the correct fix, and the Polish test is load-bearing rather than decorative — `many` at 5 and `few` at 3 genuinely differ from the `other` form in `pl.lproj`, so dropping the `locale:` argument really would fail it. That's the kind of test I want on a change like this.

The blocker is what the new wrapper does to string extraction.

`scripts/extract_strings` is `genstrings -s OBALoc`. It finds strings by recognizing the *function name* `OBALoc` at the call site. This PR replaces 8 `OBALoc(...)` calls with `CountPlural.format(...)`, which `genstrings` has never heard of. So the next time anyone runs `scripts/extract_strings`, those 8 keys stop being emitted and get dropped from `en.lproj` — along with their translator comments. The `Every locale has same keys as english` test asserts both directions, so it then fails with 8 extra keys across each of the 12 non-English locales.

The `comment:` parameter is what makes this easy to miss: `CountPlural.format` accepts it and silently discards it, so the call sites still *look* extractable.

The fix is small. Keep `OBALoc` at the call site so `genstrings` still sees it, and have `CountPlural.format` take the already-localized format string:

```swift
CountPlural.format(OBALoc("some.key", value: "%d things", comment: "..."), count: n, locale: locale)
```

That keeps the injectable `locale`/`bundle` your tests need and keeps extraction working. It's also the pattern the rest of the codebase already settled on — `SearchResultsSheetView.swift:104`, `RentalMapMarker.swift:113` and `MapTypeButton.swift:69` all wrap `OBALoc` in `String.localizedStringWithFormat`, and `StopPageAccessibilityCopy.swift:47,74` uses `String(format: OBALoc(...), locale: locale, ...)` — which is the precedent your own description cites.

- `stop_page.a11y.past_shown_fmt` gets nothing from this migration. It's the one migrated key with a `.stringsdict` entry in `en.lproj` only; the other 12 locales resolve it from flat `.strings` (`pl`: `"Pokazano minione odjazdy (%d)"`). Passing a locale can't change that. Its call-site comment claiming "plural forms live in Localizable.stringsdict" is only true for English. Pre-existing data gap, but you're touching the line.
- In `LocalizationTests.swift`, the new `lprojBundle` helper landed directly under the existing doc block for `localizedFormat`, so that comment now reads as documenting `lprojBundle`, which doesn't call `localizedString(forKey:)` at all.

Reshape the wrapper and this is good to go — ping me and I'll re-review.

Borisserz:
Addressed the extraction blocker: call sites are back to `CountPlural.format(OBALoc(...), count:)`, so `genstrings -s OBALoc` still sees the keys. `CountPlural` only applies `String(format:locale:)`. Also fixed the `lprojBundle` / `localizedFormat` doc split and noted that `past_shown_fmt` is English-stringsdict-only.

`LocalizationTests` including `Count plural call sites reach Polish few and many` is green locally. Ready for re-review.

Borisserz:

aaronbrethorst:
### Code review

_Re-review at 8f671b7, covering commits since the previous review._

aaronbrethorst (approved):
Verified at 8f671b7e. Thanks for this.

PR #1422, opened 2026-09-15
Stop stacking error bulletins when displayError repeats

### LINKED ISSUE #1421: Spammy 404 error popping up in app
Image

None of my bookmarks use this stop. It’s making the app unusable. As it pops up every few seconds and immediately when you open the app. Eventually hitting “dismiss” fails and you have to force kill the app

Borisserz:
## Summary
- `Application.displayError` used to create a new `ErrorBulletin` (and `BLTNItemManager`) on every call. When errors arrive every ~30s (e.g. a dead bookmarked stop), a second presentation stacks on the shared overlay window and **Dismiss stops working** — matching #1421.
- Skip presenting while an error bulletin is already showing; expose `ErrorBulletin.isShowing` for that gate.
- Bookmark-tab literal 404 spam itself was already silenced in #1268; this PR covers the force-kill / unusable-Dismiss half of the report.

Closes #1421

## Test plan
- [ ] Unit: `Should not present a second error bulletin while one is showing` in `ApplicationTests`
- [ ] Manual: trigger a repeating `displayError` (or temporarily force a failing bookmark fetch) and confirm only one Error card appears and Dismiss clears it
- [ ] Manual: after Dismiss, a later error can still present a new card

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented multiple error messages from appearing simultaneously.
  * Error notifications are now skipped while another error notification is already visible.

* **Tests**
  * Added coverage confirming that duplicate error notifications are blocked and new notifications can appear once none is showing.


aaronbrethorst:
### Code review

Found 1 issue:

1. The new test can't fail if the fix is removed. It only asserts `!x` against `Application.shouldPresentErrorBulletin(alreadyShowing:)`, a pure one-liner added solely to be testable — it never touches `displayError`, `errorBulletin`, or `ErrorBulletin.isShowing`. Delete the `guard Self.shouldPresentErrorBulletin(...)` line in `displayError` and this test stays green, so it gives no regression protection for #1421 despite its name. Consider asserting on the real wiring instead (e.g. inject/stub the `isShowing` state and verify a second `displayError` doesn't create a new bulletin), which would also remove the need for the tautological static helper in production code.

https://github.com/OneBusAway/onebusaway-ios/blob/862db65f706195ac6aa0e99656134f0c7a214fa1/OBAKitTests/Application/ApplicationTests.swift#L815-L823

aaronbrethorst (approved):
Merging this — the fix is correct and it clears a genuinely nasty symptom (a wedged Dismiss that needed a force-kill). I checked the part that mattered: `displayError` does assign `self.errorBulletin = bulletin` after showing, so the new `errorBulletin?.isShowing` guard reads real state rather than a permanently-nil property. And since `ErrorBulletin` is constructed in exactly one place, the gate covers every error card. A failed presentation leaves `isShowing == false`, so it can't wedge shut either.

One thing I'd like changed in a follow-up: the test doesn't test the fix. `shouldPresentErrorBulletin(alreadyShowing: true) == false` is asserting that `!` works — delete the guard in `displayError` and it stays green. The static exists only to give that assertion something to call, which means production API grew to host a tautology. The useful seam is an injectable `isShowing` so you can drive `displayError` twice and assert one bulletin.

Not blocking the fix on it, but the test currently buys you nothing.

PR #1433, opened 2026-09-18
Add unit tests for Formatters and Theme utilities

ravindra-RKB:
This PR improves codebase stability by introducing unit tests for `OBAKitCore` utilities (`Formatters` and `Theme`) that were previously untested.

## What changed:
* **`OBAKitTests/Utilities/FormattersTests.swift`**: Created tests to validate the `Formatters` class. Verifies correct time output, contextual date strings, and specifically tests `accessibilityValueForArrivalDeparture` to ensure VoiceOver strings format accurately across different temporal states (past, present, future).
* **`OBAKitTests/Theme/ThemeTests.swift`**: Added tests for `ThemeMetrics` and `ThemeColors` to ensure padding constants and core semantic UI colors initialize correctly without regressions.

Adding these assertions prevents future formatting or styling regressions and significantly expands our automated test coverage within `OBAKitCore`.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for theme spacing, corner-radius, and color values.
  * Expanded time-formatting coverage, including reliable handling of “today” dates.
  * Added accessibility text coverage for past, current, future, and scheduled arrivals and departures.
* **Accessibility**
  * Improved validation of departure and arrival announcements across different schedule statuses and time periods.


aaronbrethorst:
### Code review

Found 1 issue:

1. This is described as a test-only PR, but it adds permanent mutable public API to a shipped white-label framework purely as a test seam: `public var now: () -> Date` on `Formatters`. `Formatters` already takes every other environment dependency immutably through `init(locale:calendar:themeColors:)`, so the consistent seam here is an init parameter (or a `contextualDateTimeString(_:relativeTo:)` overload), not a settable property. As written, any consumer — and the single shared instance exposed as `OBAFormattersKey.defaultValue`, which is `nonisolated(unsafe) static let` — can have its clock reassigned from any isolation domain with no compiler check. The production change also isn't mentioned in the PR description.

https://github.com/OneBusAway/onebusaway-ios/blob/2bc2c1dbd4f6b994ca453e26db20eb4dafa024e8/OBAKitCore/Utilities/Formatters.swift#L17-L20

aaronbrethorst:
Closing this in favour of #1434, which contains it in full — no reflection on the work itself.

Your `feature/utilities-badge-logger-tests` branch merged `feature/formatters-tests` into it, so this PR's head (`2bc2c1db`) is a git ancestor of #1434's head (`984bbeb6`). I confirmed the three shared files are byte-identical at both heads. That means #1434 already carries every commit here plus the Logger and ImageBadgeRenderer tests, and reviewing both would be reviewing the same diff twice.

The tests themselves check out — I verified the expectations against the real `en.lproj` strings, and the one hard-coded `"12:00 AM"` correctly pins both `en_US` and a GMT `timeFormatter.timeZone` and normalizes the narrow no-break space. That's the kind of care these usually lack.

Two things to carry over into #1434, which I've also noted there.

The `public var now: () -> Date` added to `Formatters` is the one I'd like changed. It's permanent mutable public API on a white-label framework, added to serve a test — and the repo already has a convention for exactly this, `private let dateProvider: () -> Date` injected via init, in `PushRegistrationManager`. More to the point, the test doesn't need it: `contextualDateTimeString` has zero callers anywhere in the repo (its own definition is the only match for a full-text search), so the method under test is dead API, and calling it with `Date()` is unconditionally "today" under the original `isDateInToday` anyway. An init parameter, or just dropping the seam, both work.

Second, heads up that #1308 has since landed and rewrote the body of `contextualDateTimeString`, so #1434 will need a rebase regardless.

Reopen this if you'd rather take the Formatters/Theme work forward separately and drop it from #1434 — either shape is fine, I just don't want the same diff open twice.

PR #1434, opened 2026-09-18
Add unit tests for Logger and ImageBadgeRenderer utilities

ravindra-RKB:
This PR introduces test coverage for two previously untested utilities in `OBAKitCore`. 
*(Note: This branch builds on top of the `Formatters and Theme` PR, so you might see those commits here until the previous PR is merged).*

### What changed:
* **`OBAKitTests/Utilities/LoggerTests.swift`**: Added tests for the shared `Logger` utility to verify that log statements are successfully recorded and retrieved from `OSLogStore`.
* **`OBAKitTests/Utilities/ImageBadgeRendererTests.swift`**: Validated `ImageBadgeRenderer` size configurations and verified that `drawImageOnRoundedRect` outputs a `UIImage` matching the initialized `badgeSize`.

These tests ensure that diagnostic logging and foundational UI icon rendering behave predictably.

## Summary by CodeRabbit

- **Tests**
  - Added coverage for theme metrics and color initialization.
  - Added tests for image badge sizing, default behavior, and rendering output.
  - Added logging verification tests.
  - Expanded formatter tests for time, date, accessibility, and arrival/departure status handling.
  - Added broader validation for formatting consistency across common date, time, and status scenarios.


aaronbrethorst:
### Code review

aaronbrethorst (changes requested):
I've closed #1433 in favour of this one — your branch merged `feature/formatters-tests` in, so this PR is a strict superset of it and reviewing both meant reviewing the same diff twice. Everything from that PR carries over here.

The tests are in good shape. I checked all six `ThemeMetrics` values against `Theme.swift`, the four `ThemeColors` assertions against their literal assignments in `init`, and all eight VoiceOver strings against both the `OBALoc` defaults and the actual `en.lproj` entries — no drift in any of them. The Logger and ImageBadgeRenderer suites cover types that had nothing before.

**It needs a rebase.** #1308 landed while this was open and rewrote the body of `contextualDateTimeString`, so `Formatters.swift` now conflicts. Nothing you did — just sequencing.

**I'd like the `public var now` seam changed.** This is the one substantive point. It adds permanent mutable public API to a white-label framework in a PR titled "add unit tests," and the PR body doesn't mention the production change at all. The codebase already has a convention for injecting a clock — `private let dateProvider: () -> Date` passed through init, in `PushRegistrationManager` — and `Formatters` takes every other environment dependency the same immutable way via `init(locale:calendar:themeColors:)`.

What decides it for me is that the seam isn't buying anything. `contextualDateTimeString` has zero callers anywhere in the repo — a full-text search returns only its own definition — so it's dead public API, and the test that motivated the seam doesn't need it: under the original `isDateInToday`, `contextualDateTimeString(Date())` is unconditionally the "today" branch. Please either take the date through init or drop the seam and let the test use `Date()`.

A few smaller things, none blocking, take or leave them:

- `LoggerTests.swift:32` polls with up to 20 × 0.1s of `Thread.sleep`. The test target defaults to main-actor isolation and this suite isn't opted out, so on the failing path that blocks the main thread for ~2s while other main-actor suites are running concurrently. A polling loop with `Task.sleep` would avoid it.
- `ThemeTests`'s constant assertions mirror each `static let` back at itself, so they can only fail when someone deliberately edits a constant. We pruned similar assertions from this very file in `58e1d1d4`.
- `Initialization sets properties correctly` in `ImageBadgeRendererTests` asserts only `badgeSize`; `fillColor` and `backgroundColor` are private and unverified, so the name promises more than it delivers. Likewise `Rendering produces a UIImage of expected size` would pass against a blank image — the size is just the bounds handed to the renderer.
- Both new files are missing a trailing newline. SwiftLint never sees them (`included:` covers only `OBAKit`, `OBAKitCore`, `OBAWidget`), so nothing catches it automatically.

Rebase and swap the seam and I'll turn this around quickly.

aaronbrethorst:
### Code review

Found 1 issue:

1. `FormattersTests.swift` was re-saved with the wrong encoding during the rebase (commit 3349b861), and the damage is still there at head. Git shows a UTF-8 BOM (`EF BB BF`) at byte 0 of the file, and 11 existing lines are mojibake: UTF-8 text was read as Windows-1252 and then re-encoded. The copyright header now reads `Copyright Â© Open Transit Software Foundation`, every em dash in a doc comment reads `â€”`, and the #514 doc comment now says UIKit truncates with `â€¦` and that zh-Hans uses `â€¦â€¦`. That comment exists to explain which glyph is correct, so it is now wrong. The latest commit fixed only the two `#expect` literals (by switching them to `\u{2026}`). The compiler does not check comments, and SwiftLint does not lint `OBAKitTests`, so nothing automated will catch this. To fix it, restore the original bytes of the 11 lines outside the new `// MARK: - Time and Accessibility Formatting` section (for example, check out the file from `main` and re-apply only the added tests), and save without a BOM. After that, `grep -n 'Â\|â€' OBAKitTests/Application/FormattersTests.swift` should find nothing, and `head -c3` of the file should print `//\n`.

https://github.com/OneBusAway/onebusaway-ios/blob/326eb6000d339fa4bc7e0117a8a8ee0f850088a6/OBAKitTests/Application/FormattersTests.swift#L1-L6

https://github.com/OneBusAway/onebusaway-ios/blob/326eb6000d339fa4bc7e0117a8a8ee0f850088a6/OBAKitTests/Application/FormattersTests.swift#L244-L248

aaronbrethorst:
### Code review

aaronbrethorst (approved):
Both blockers from my earlier review are resolved: the branch is rebased onto main, and the diff no longer touches Formatters.swift, so the mutable `now` seam is gone. The encoding damage in FormattersTests.swift is fixed too, with no BOM and every non-ASCII line matching main. LoggerTests now polls with `Task.sleep` instead of blocking the main thread, and the new pixel checks in ImageBadgeRendererTests assert real compositing behavior rather than just the output size. Thanks for working through all of it.

PR #1435, opened 2026-09-18
Add unit tests for Orchestration and WebView components

ravindra-RKB:
This PR introduces test coverage for two previously untested component directories: Orchestration (in OBAKitCore) and WebView (in OBAKit).

What changed:
- `OBAKitTests/Orchestration/CoreAppConfigTests.swift`: Verified that CoreAppConfig initializes safely and all required parameters are passed and assigned accurately. Added cleanup for test-specific UserDefaults domains.
- `OBAKitTests/WebView/DocumentWebViewTests.swift`: Added validation for DocumentWebView's HTML token replacement logic to ensure page content and action buttons render correctly.

These tests continue to increase our code coverage on older and foundational utility files.

## Summary by CodeRabbit

- **Refactor**
  - Improved document page rendering so generated content is assembled before being displayed, with optional action buttons handled consistently.

- **Tests**
  - Added coverage for application configuration initialization and property assignment.
  - Added tests confirming document pages include the optional action button when provided and omit action controls when no title is supplied.


aaronbrethorst:
### Code review

Found 1 issue:

1. Both tests in `DocumentWebViewTests` assert `#expect(webView != nil)` on a value the compiler already knows is non-optional. `DocumentWebView(frame:configuration:)` inherits `WKWebView`'s non-failable designated initializer, so the expression is a compile-time constant `true` (Swift even warns "comparing non-optional value ... to 'nil' always returns true"). `Initialization succeeds` therefore contains no assertion at all, and the `setPageContent` test would still pass if `setPageContent` were replaced with an empty body — neither test pins any behavior. Every other `!= nil` expectation in OBAKitTests is on a genuinely optional value. Worth either dropping the vacuous expectations and letting the "does it trap?" smoke test stand on the constructor/method call alone, or asserting something real (e.g. round-tripping the rendered HTML via `evaluateJavaScript` once `loadHTMLString` finishes, so the `{{{oba_page_content}}}` / `{{{oba_page_actions}}}` substitutions are actually covered).

https://github.com/OneBusAway/onebusaway-ios/blob/5ff9d310d010cd65eeca58469137e37bbd58df5b/OBAKitTests/WebView/DocumentWebViewTests.swift#L19-L33

aaronbrethorst (changes requested):
The `CoreAppConfig` half of this is fine — thin, but it does pin against a swapped-parameter regression, and I appreciate that you kept the UserDefaults suite name distinct rather than reaching for `.standard`.

What blocks it is `DocumentWebViewTests`. Both tests assert nothing.

`DocumentWebView` is a `WKWebView` subclass that declares no initializers of its own, so it inherits the non-failable `init(frame:configuration:)`. That makes `#expect(webView != nil)` a compile-time constant `true` — the compiler even says so ("comparing non-optional value of type 'DocumentWebView' to 'nil' always returns true"). It's a warning rather than an error, and `OBAKitTests` isn't in SwiftLint's `included:` list, so nothing in CI flags it.

The consequences are worth spelling out:

- `Initialization succeeds` consists of that one vacuous expectation and nothing else, so the test body is effectively just the constructor call.
- `setPageContent with HTML fragment completes without error` would pass unchanged if `setPageContent` were gutted to an empty function. None of the `{{{oba_page_content}}}` / `{{{accent_color}}}` / `{{{oba_page_actions}}}` substitution work is covered.

That second one is the part I'd actually like tested, and it is testable — the substitution is synchronous string work even though `loadHTMLString` isn't. Pull the template rendering out into something that returns the HTML string and assert on it directly, or load the view and read the DOM back via `evaluateJavaScript` after navigation completes. Either gives you a test that fails when the substitution breaks.

Your comment about `loadHTMLString` being async is accurate and I'm glad you wrote it down rather than papering over it — the issue is just that the test as written doesn't work around the limitation, it accepts it.

Two smaller notes while you're in here:

- `CoreAppConfigTests` passes `queue`, `userDefaults`, `locationService`, and `dataLoader` but never asserts on them, so four of the twelve init parameters are untested. Worth finishing the set given the test's name.
- Both new files are missing a trailing newline, and the hardcoded UserDefaults suite has no `removePersistentDomain` cleanup. It's benign today because nothing reads persisted state, but `OBAKitTests/Helpers/OBATestCase.swift` establishes the per-test-unique-domain pattern for a reason and it's worth following.

Also, the PR description mentions an `OBAKitTests/Orchestration/FeatureFlagsTests.swift` that isn't in the diff — worth fixing so reviewers aren't looking for a file that isn't there.

Drop the `!= nil` expectations, give `setPageContent` a real assertion, and I'll re-review promptly.

ravindra-RKB:
Thanks for the detailed feedback, @aaronbrethorst! I've addressed all the points:
- Extracted the HTML rendering logic into an internal `buildPageContent` method to allow for proper synchronous text assertions in `DocumentWebViewTests`.
- Added the missing parameter assertions and `removePersistentDomain` cleanup to `CoreAppConfigTests`.
- Fixed the missing trailing newlines.
- I've also updated the PR description to remove the obsolete mention of `FeatureFlagsTests`.

aaronbrethorst:
### Code review

PR #1436, opened 2026-09-18
Add unit tests for NetworkHelpers, RESTAPIURLBuilder, and APIError utilities

ravindra-RKB:
This PR introduces test coverage for the network helper and URL builder utilities in `OBAKitCore`.

### What changed:
- **`OBAKitTests/Network/NetworkHelpersTests.swift`**: Added tests for `NetworkHelpers` to verify correct query item dictionary mapping, path variable URL escaping, and HTTP body data generation.
- **`OBAKitTests/Network/RESTAPIURLBuilderTests.swift`**: Validated `RESTAPIURLBuilder` endpoints across `getShape`, `getScheduleForRoute`, `getStopProblem`, `getSurveys`, and other REST endpoints, ensuring correct path generation and query parameter injection.
- **`OBAKitTests/Network/APIErrorMissingStopTests.swift`**: Added coverage for `APIError+MissingStop` to verify that `indicatesMissingStop` correctly identifies `404` errors and empty `null` JSON responses, while gracefully ignoring other HTTP failures.

These tests ensure that foundational network request serialization and endpoint generation remain consistent and bug-free as the app evolves.

## Summary by CodeRabbit

- **Tests**
  - Added coverage for converting dictionaries to query parameters, escaping path values, and encoding HTTP request bodies, including special characters and multiple fields.
  - Improved URL builder tests with calendar-based dates and structured validation of URL paths and query parameters.
  - Updated test formatting and copyright header consistency to improve maintainability and test reliability.


aaronbrethorst:
### Code review

aaronbrethorst (approved):
Good, focused test coverage for two types that had almost none. I traced every expected value back to the implementation — `joinBaseURLToPath`'s no-trailing-slash concat, `escapePathVariable`'s manual `/` replacement, the RFC-3986 unreserved set in the body encoder, and `StopProblemCode.nameWrong.APIStringValue` — and they're all correct. The `URLComponents` query-item assertions are the right shape too: they fail on nil rather than passing silently.

Two things worth folding into your next pass, neither blocking:

The schedule-date tests build the date with `Calendar.current` and the builder formats it with a `DateFormatter` that also reads the default time zone, so the round-trip is consistent by construction and can't actually catch a time zone bug. Pinning an explicit `Calendar(identifier: .gregorian)` with a GMT time zone would make those assertions mean something.

Also, the new files land in `OBAKitTests/Network/` while the existing suite lives in `OBAKitTests/Networking/` — and a couple of the `NetworkHelpers` cases restate coverage that's already in `Networking/Helper Tests/NetworkHelperTests.swift`. Worth consolidating so we don't end up with two sibling folders for the same concern.

Merging.

PR #1437, opened 2026-09-19
Search stop numbers across the full region (#1432)

### LINKED ISSUE #1432: No disambiguation step for duplicated stop numbers
The iOS app does not present a disambiguation page when two agencies have the same zone number. 

Actual behavior: https://youtube.com/shorts/FKXXbSAirAs?is=Y0yKo72wGnW7Q3Sa

Expected behavior: https://youtube.com/shorts/HtQwrPSxXbA?is=gvXRmkIqJ0jqidiV

Borisserz:
## Summary
- Closes #1432: iOS skipped the disambiguation sheet when two agencies share a stop code.
- Root cause: stop-number search used `getStops(circularRegion:query:)`, which caps `radius` at **15 km**, so remote same-code stops never appeared (Android / span queries return both).
- Stop-number lookup now uses `getStops(region:query:)` with `latSpan`/`lonSpan` over the region's `serviceRect`. Existing disambiguation UI already handles `results.count > 1`.

## Test plan
- [x] `NetworkHelperTests` — span+query URL has no `radius=`
- [x] `SearchManagerTests` — stop-number search sends `latSpan`/`lonSpan` + `query`
- [ ] Manual (Puget Sound): search stop code `1000` → disambiguation list with both agencies (not auto-open the downtown hit)

## Summary by CodeRabbit

* **Bug Fixes**
  * Stop-number searches now cover the full service area instead of being limited to a 15 km radius.
  * Searches can find stops across wider agency regions, improving results when the requested stop is far from the map’s center.

* **Tests**
  * Added coverage to verify regional stop searches use the complete service span and return the expected search parameters.


aaronbrethorst:
### Code review

aaronbrethorst (approved):
This is a well-diagnosed fix. The root cause is exactly right — the circular overload clamps radius to 15 km, so a same-code stop at a distant agency in the same region could never reach the disambiguation sheet — and switching to `latSpan`/`lonSpan` over the region's service rect is the correct answer rather than just raising the cap.

I checked the parts that matter: both result consumers already branch on `results.count > 1` (`MapRegionManager.searchResponse` and `SearchSheetViewModel.SearchOutcome`), so the fix lands on the UIKit and SwiftUI paths alike, and the new `SearchManagerTests` case asserts `radius == nil` alongside a non-nil `latSpan`, which means it genuinely fails against the old code instead of just documenting the new. The inline comments pointing at #1432 are the right amount of explanation.

One small thing if you're back in this file: `getStops(region:query:)` is a byte-for-byte copy of `getStops(region:)` plus one parameter, and a single method with `query: String? = nil` would collapse the duplication in both the builder and `RESTAPIService+Get`. The new doc comment also drops the `- important:` subset/zoom-in caveat its two siblings carry, which arguably matters more now that the span can cover an entire region.

Merging.

PR #1443, opened 2026-09-21
Badge the stop list's accessibility clock with its timezone

### LINKED ISSUE #1438: Stop list accessibility clock label misses the region timezone badge
Follow-up from #1308, which landed the region-time-zone display behind the default-off "Show times in region time zone" setting.

`StopArrivalView` populates `accessibilityTimeLabel` from two different call sites, and only one of them got converted:

- `OBAKit/Stops/StopArrivalView.swift:288` — the `arrivalDeparture` didSet path (used by `TripFloatingPanelController`) calls `formatters.formattedClockTime(...)`, which appends the timezone badge when the device offset differs. Correct.
- `OBAKit/Stops/StopArrivalView.swift:251` — the content-configuration path (the stop list) assigns `contentConfiguration.accessibilityTimeLabelText`, which is `OBAKit/Stops/Sections/StopArrival/StopArrivalItem.swift:196`:

```swift
var accessibilityTimeLabelText: String? {
    return formatters?.timeFormatter.string(from: viewModel.arrivalDepartureDate)
}
```

`timeFormatter` is now region-zoned globally, so that label renders a region-zone clock time with no badge.

This matters specifically at accessibility content sizes. `StopArrivalView.swift:304-305` hides `normalInfoStack` and shows `accessibilityInfoStack` when `isAccessibility` is true, and `fullExplanationLabel` is in the former while `accessibilityTimeLabel` is in the latter. So on a stop-list row at large text, `accessibilityTimeLabel` is the only visible clock — and with the setting enabled it shows a time in a zone the rider isn't in, with no indicator. That is the case #332's badge exists to prevent.

VoiceOver *speech* is unaffected: `accessibilityValueForArrivalDeparture` was converted in #1308 and does carry the badge. This is about the visible label at large text sizes.

**Fix:** route `StopArrivalItem.accessibilityTimeLabelText` through `formatters?.formattedClockTime(...)` so both paths feeding the same label agree.

Worth a quick audit of the other non-badged clock surfaces while in here — `TripStopListItem.swift:241`, `TripPage/TripStopListView.swift:111`, `DestinationStopPickerController.swift:295`, `TripActivityPresenter.swift:36`, and `ServiceAlertViewModel.swift:71` (via `dateIntervalFormatter`) all now render region-zone times with no badge on screen. Those are arguably fine since they're all region events, but it's worth a deliberate decision rather than an accident.

diveshpatil9104:
## Description

Closes #1438.

`StopArrivalView` populates `accessibilityTimeLabel` through two different paths, but only one of them included the timezone badge:

- `StopArrivalView.swift:288` — `didSet` path uses `formattedClockTime(...)` ✅
- `StopArrivalItem.swift:196` — configuration path used `timeFormatter` directly ❌

At larger accessibility text sizes, this label can be the only visible clock on screen. When the region-timezone setting is enabled, a stop-list row could therefore display a time from a different timezone without indicating which timezone it belonged to. This is the case that #332's timezone badge was intended to address.

This change routes both paths through `formattedClockTime(...)` so they produce consistent accessibility output.

## Audit

The other clock surfaces were reviewed deliberately. **No behavior was changed**; the reasoning is now documented.

`Formatters.timeFormatter` documents when each format is appropriate:

- **Bare time** — for a sequence of times sharing the same context, such as the three trip stop lists, where adding a timezone badge to every row would create unnecessary noise.
- **Badged time** — for a standalone clock where the timezone context may not otherwise be obvious.

`TripActivityPresenter` was the main borderline case. A Live Activity clock does stand alone on the Lock Screen, but it remains unbadged because the rider is actively taking that trip and is therefore expected to be in that region. The available space on the Live Activity card is also limited.

This should be revisited if Track can continue displaying the Live Activity after the rider leaves the trip's region.

## Summary by CodeRabbit

* **Accessibility**
  * Arrival and departure time labels now include a timezone badge when the displayed time zone differs from the rider’s region, helping clarify the time being shown.

* **Documentation**
  * Clarified how clock formatting and Live Activity times handle timezone badges.

* **Tests**
  * Added coverage for timezone badges appearing only when the displayed and regional time zones differ.


aaronbrethorst:
### Code review

Found 1 issue:

1. The new tests depend on the process time zone, but nothing in this suite sets it. The header comment says "The bundle pins the process to GMT (`OBATestCase`)". That pin is a lazy `private let pinnedToGMT` in `OBATestCase.swift`, and it only fires from `OBATestCase.init`. `StopArrivalAccessibilityClockTests` is a plain struct that doesn't inherit `OBATestCase`, so it never triggers the pin (`WeatherFormatterTests` already notes this caveat for non-inheriting suites). `config.accessibilityTimeLabelText` calls `formattedClockTime(_:)` with the default `deviceTimeZone: .current`, so the result depends on the machine's zone and on whether some `OBATestCase` subclass happened to run first. Run the suite alone (`-only-testing:OBAKitTests/StopArrivalAccessibilityClockTests`) on a Mac set to Pacific time and both tests fail: LA vs. LA gives no badge, and GMT vs. LA does. CI passes only because the runners are on UTC. The #1308 tests avoid this by passing `deviceTimeZone:` explicitly (`TimeZoneScheduleBadgeTests`). The review there called that "the right way to test this", because the GMT pin can't mask or break it. To do the same here, either assert against `formatters.formattedClockTime(date, deviceTimeZone: gmt)` and pin the device zone some other way, or add a `deviceTimeZone` seam to the configuration. Don't rely on the pin.

https://github.com/OneBusAway/onebusaway-ios/blob/9c039b145d1ccfbeab6f29af459fc56dfa6905b0/OBAKitTests/Stops/StopArrivalAccessibilityClockTests.swift#L19-L25

https://github.com/OneBusAway/onebusaway-ios/blob/9c039b145d1ccfbeab6f29af459fc56dfa6905b0/OBAKitTests/Stops/StopArrivalAccessibilityClockTests.swift#L74-L98

PR #1444, opened 2026-09-22
Let riders set a destination proximity alert from the Stop page

### LINKED ISSUE #455: Help users track their proximity to their final destination

diveshpatil9104:
## Summary

Closes #455:  this is the UI half, PR 4. The backend merged across #1071, #1087, #1092, #1292, #1311, #1368 and #1397 had **zero production callers**: `createProximityAlert` and `LocationService.requestAlwaysAuthorization()` were both unreachable until this PR.

## What it does

One item at the end of the **Location** group of the Stop page's More menu, in both live presentations: `StopPageToolbar` (sheet) and `locationMenu()` (pushed). Title and glyph follow the state  *Alert Me When Nearby* / *Cancel Nearby Alert*.

Two-phase, which is what `authorizationStep()` (#1397) exists for: resolve what’s missing *before* offering anything, and only call `createProximityAlert` from `.ready`. The item is never hidden and never disabled, a rider who can’t use it yet is told which of four things is in the way, rather than shown a grey row that does nothing.

A tap arriving while the Always prompt is outstanding is dropped: `requestAlwaysAuthorization()` spends the one-time flag *synchronously*, one line before the prompt appears, so a second tap would otherwise resolve `.locationBlocked` and push "Open Settings" at someone still looking at the prompt they haven’t answered.

Two choices worth a look:

- **`ProximityAlertOutcome`** :  a pure reducer over both manager enums (6 cases and 5), every case by name, no `default:`. An enum return rather than the codebase’s usual `StopPageActionRowState` struct-of-`Bool`s: that shape fits continuous UI state, this produces one of N discrete actions.
- **`liveActivityStarted: Bool` → `transientToast: String?`** , set and cancelled are the second and third users of that one capsule, so it’s generalised rather than duplicated. #1367’s cancellation body is verbatim.

## Deferred, deliberately

1. **`StopPageActionRow.moreMenu`**, the third Location menu, behind `useMapPanelExperienceKey`: read with a plain `bool(forKey:)`, so it defaults **off**. One closure when that flag flips.
2. **Legacy UIKit `StopViewController`**:  same call as #1120’s Share Trip.
3. **A screen listing a rider’s alerts** :  which is why the region-limit alert has no "Manage Alerts" button.

Radius is fixed at 200 m; the manager’s `radiusMeters:` leaves a picker open.

## Your call: the Always prompt shows the wrong justification

iOS renders `NSLocationAlwaysAndWhenInUseUsageDescription` in that prompt, and today it reads *"See where you are in relation to transit, and help you navigate more easily."* - While-Using-era copy that never mentions the background. KiedyBus’s is *"Show your location on the map."*

This PR is the first thing in the app that ever raises that prompt. Both strings are white-label config (`app_shared.yml:81`, `KiedyBus/project.yml:44`) and a change lands in every app’s App Store review, so I’ve flagged it rather than changed it unilaterally.

## Testing

**Unit:** the reducer (every case of both enums, including that the three blocked location statuses give three *different* messages), the four Settings alerts, `proximityAlert` tracking `.proximityAlertsDidChange`, and both `Icons` symbol names resolving on the running OS.

**Walked in the simulator:** item placement in both menus, the location prompt, the While Using → Always alert, notification prompt → toast → item flips, cancel → flips back, cross-surface state sync, notifications-denied alert.

**Not reachable without a device:** a geofence firing, and `.locationBlocked(.restricted)` , covered by a presenter test instead.

## Notes

- 13 locales hand-spliced, not regenerated (`extract_strings` rewrites ~115 unrelated keys). `plutil -lint` clean; `region_limit.message_fmt` uses positional `%1$@`/`%2$d`.
- The flow is in `StopPageActionPresenter+ProximityAlerts.swift` rather than inline, the same split `MapViewController+Camera` uses: it keeps the presenter under SwiftLint’s file-length limit. Zero new lint violations.
- #1291 (Bike Mode) also touches `StopViewModel.swift`, but only the `userDefaultsCancellable` sink in `init`. No semantic conflict; whichever lands second may need a trivial rebase.

## Summary by CodeRabbit

* **New Features**
  * Added stop-level proximity alerts to notify riders when approaching a destination.
  * Added options to set or cancel alerts from the stop page, with status-aware icons and labels.
  * Added guidance for location and notification permissions, Settings access, and monitoring limits.
  * Added confirmation messages for alert and Live Activity actions.

* **Localization**
  * Added proximity-alert translations across supported languages.

* **Bug Fixes**
  * Improved transient toast display and updates on stop pages.


aaronbrethorst:
### Code review

